Issue #30, opened 2019-02-19
Use only one Material Icons


kinow:
We have two material icons assets now. The second is much heavier, and was introduced while creating the login page (mainly because I didn't really know much about that part of Vuetify).

Now I found a site with the list of icons that match the first icon set. We just need to find the three replacements for the login page, remove the other library, and profit!

http://code.meta-platform.com/assets/mdi/preview.html

PR #32, opened 2019-02-21
Fix issue #30 use only one material icon dependency

kinow:
Used the icons from the existing package. Found them by searching in the link in #30. Resulting UI is the same.

image

Even though we may not use this login screen anymore (we will use the one from the hub :-). Fixing it just to remove the dependency size.


kinow:
Hummm, @hjoliver any idea how to get Travis working for this new repo? I had Travis enabled in my repository. And it looks like Travis found out that it was transferred some way?

https://travis-ci.org/cylc/cylc-web

But now it didn't start a job for this pull request...

kinow:
Not using the 2 reviewers approach for now as we are in early stages. Merging so that work on #33 can continue without issues for merging later.

Issue #38, opened 2019-02-27
Add Services


kinow:
Vue.js takes care of view/visualization. It means that you are free to architect the rest of your application. That is quite different from other frameworks such as React or Angular.

However, certain patterns from these other frameworks can be useful. In AngularJS, for instance, most developers are familiar with **services**. And when well designed, it is fairly simple to [write tests mocking the services](https://stackoverflow.com/questions/19274274/how-do-you-mock-a-service-in-angularjs-when-unit-testing-with-jasmine).

Without doing this way, we should still be able to create the application, run, get the right results, etc. However, writing tests without being able to mock a service, forces you to _hack_ your API. By inserting methods that replace parts of an object.

And then eventually you end up having to spend a lot of time maintaining your main code, and also your test code. Which may become a blocker when you have to upgrade or change the framework.

This issue is to design how the components & services should play together, and then create the `UserService` for #33 . That service should be able to use Vuex actions to store the state. The only method necessary now is to retrieve the user profile (i.e. one user, for a given URL like `/user/kinow/userprofile`).

Not necessary, but it would be better if we could have unit tests designed too. IOW, have a test for the `UserService`, in a simple way, that can be replicated later for future services. Later we may have to update the Suite component to remove graphql dependency from Vuex.

kinow:
One implementation is now partially done in https://github.com/kinow/cylc-web/tree/add-services. Now need to try to write a unit test.

If we are able to test everything, except the HTTP call (mocked), then I think this is a good starting point for future services used to talk to suite list service, other parts of the hub like proxy API, etc.

kinow:
Writing the unit test was quite easy and simple. The plan was to see the test failing, and use [Sinon.JS](https://sinonjs.org/) to mock the call.

However, faced a different issue as importing the services, apparently webpack imported the classes with GraphQL. And there is a known issue that running the Apollo GraphQL client in the node environment (i.e. without a browser), it fails due to the missing `fetch`.

If you open a browser window, then go to any URL, open the developer tools, and type in the Console: `window.fetch`, you will see that the browser defines it. But when browserless (as when running the unit tests), that variables does not exist.

This was fixed in Apollo https://github.com/apollographql/apollo-client/issues/3578, and there is now a way to pass a `fetch` object.

![image](https://user-images.githubusercontent.com/304786/53525866-a052ab80-3b47-11e9-938a-71a60822d007.png)

kinow:
[...]

PR #39, opened 2019-02-28
Add services

kinow:
For #38 

- Adds Sinon.JS for mocking
- Create a Suite Service refactoring the existing code in Vuex (which is good, Vuex has less one dependency)
- Create a User Service
- Adds a unit test for the User Service, that tests the Service, the Model, and Vuex. Only the layer for Axios (which does the HTTP query) is mocked.

This unit test should be used as example for future unit tests of services, as they will be pretty much the same. Mock the HTTP/GraphQL layer, and test the rest.

We still need to confirm Vuex will work fine once we have more tests. Can't say if they will run in parallel, or mess with the Store state.

Sending a pull request to see if Travis approves, and if - hopefully - the coverage goes up a bit 🎉

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1074,6 +1074,41 @@
       "integrity": "sha1-p3c2C1s5oaLlEG+OhY8v0tBgxXA=",
       "dev": true
     },
+    "@sinonjs/commons": {
+      "version": "1.3.0",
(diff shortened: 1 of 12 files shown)


kinow:
>Merging #39 will **increase** coverage by `25.7%

Hooray! 🎉

Issue #40, opened 2019-02-28
Move loading mutation call to interceptors


kinow:
It is much easier to have a single place with the `setLoading(true)` and once complted `setLoading(false)`, rather than having to implement that in each call.

e.g. https://github.com/greyby/vue-spinner/issues/19

kinow:
No idea if we have something similar for Apollo GraphQL.

kinow:
There's always a way of intercepting calls to Vuex store (https://vuedose.tips/tips/8/). But I think it wouldn't fit the current design, as the call starts in the service.

kinow:
Humm, it looks like for ApolloClient, we have to work in the Network Layer, using what they call Middleware and Afterweare: https://www.apollographql.com/docs/react/advanced/network-layer.html#linkMiddleware

kinow:
Used [Vuetify Loader](https://vuetifyjs.com/en/components/progress) to display the progress. Then set the interceptors for Axios, and instead of searching for a way to intercept requests in GraphQL, replaced their `fetch` which was using `node-fetch` by the [axios-fetch](https://www.npmjs.com/package/axios-fetch) implementation.

Which means that REST and GraphQL requests share the lower level HTTP client that fetches the data. And we can configure interceptors in a single place 🎉 It deserves a screenshot

PR #46, opened 2019-03-05
Add axios interceptors loading

kinow:
Fix for #40 

Used [Vue.js loader](https://vuetifyjs.com/en/components/progress) to display progress, set the interceptors for Axios, then used `axios-fetch` as the fetch implementation for GraphQL 🎉


kinow:
Just waiting for Travis/Codecov now.

PR #64, opened 2019-05-06
Update dependencies and fix npm audit issues

kinow:
Had a bit of waiting time, while Travis was building a branch in `cylc-flow`. Updated vue.js and other dependencies manually after running `vue ui` and inspecting the changes.

Then updated other dependencies automatically via `npm audit fix`.

There is one pending high priority security bug in `tar`, but it does not affect running the HTML and JS files. I believe it affects the build/development environment. There is no fix yet, but the issue is being tracked at https://github.com/sass/node-sass/issues/2625


dwsutherland:
Only need one reviewer at this stage (depending on changes and author discretion)

PR #66, opened 2019-05-08
Wait for set null before populating again

kinow:
Fixing linter error. Waiting on the set user = null before populating it now, to keep order of actions as expected.


kinow:
One approval is enough for now. Merging :+1:

PR #68, opened 2019-05-13
Fix alias warning in IDE

### LINKED ISSUE #20: Fix issue in WebStorm when using the @ symbol
The `@` symbol is used as alias for the root directory, and appears to be working fine. However, the IDE is not recognizing that option, marking it as an error.

Unit tests with `@` are also not working.

kinow:
Close #20 

Solution found [here](https://intellij-support.jetbrains.com/hc/en-us/community/posts/360000903779-vue-scss-file-cannot-resolve-) fixed the issue.

`npm run build` and the IDE automatic build are both happy now.

Screenshot_2019-05-13_14-18-05

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -19,7 +19,7 @@ export default {
 </script>
 
 <style lang="scss">
-  @import '@/styles/index.scss';
+  @import '~@/styles/index.scss';
 
   /* Remove in 1.2 */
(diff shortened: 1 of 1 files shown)


kinow:
Hard to ask others to review this one. Small change, so merging it as it should not break anything :crossed_fingers:

PR #69, opened 2019-05-13
Fix URIError to display favicon

sadielbartholomew:
When running the UI (using the Cylc UI Server), amongst other log messages to the terminal STDOUT I get, repeatedly:

```console
16:37:41.752 [ConfigProxy] error: Error in handler for GET /user/sbarth/%3C%=%20BASE_URL%20%3Efavicon.png: URIError: URI malformed
```

On investigation, there seems to be some placeholder field ``<%= BASE_URL %>`` which I imagine was added & intended to be replaced at a later stage, but the replacement stage was missed. Playing around, the relative path to use so that the favicon is found & shown is simply the current directory, so that the whole placeholder string can simply be removed.

It's obviously not a crucial problem to have resolved at this early stage of the UI development, but the fix is trivial so I thought I would put in a PR!

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -5,7 +5,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link href='https://fonts.googleapis.com/css?family=Roboto:300,400,500,700,300italic,400italic' rel="stylesheet" type="text/css">
-    <link rel="icon" href="<%= BASE_URL %>favicon.png">
+    <link rel="icon" href="favicon.png">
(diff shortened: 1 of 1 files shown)


sadielbartholomew:
Ah I see! Thank you for the explanation @kinow. Spot the Vue novice over here! It's clearly a set-up issue on my side, so I will close this, but comment later FYI when I work out where I went wrong in getting started to have seen this error.

sadielbartholomew:
Ah I see! Thank you for the explanation @kinow. Spot the Vue novice over here! It's clearly a set-up issue on my side, so I will close this, but comment later FYI when I work out where I went wrong in getting started to have seen this error.

Issue #70, opened 2019-05-20
 Query tasks from GraphQL server in Cylc UI Server, instead of Cylc Flow


kinow:
I am reviewing a [pull request to add graphql to cylc/cylc-uiserver](https://github.com/cylc/cylc-uiserver/pull/34). And while testing it, realized the suite listing call worked, but after that the next call to graphql to query about tasks failed with a CORS issue.

Took me a while to realize the CORS response was not coming from `cylc-uiserver` but from Cylc (cylc/cylc-flow). That's because our initial PoC for GraphQL in cylc-ui worked that way.

Now that it [was already decided](https://github.com/cylc/cylc-flow/issues/3023) to have Cylc talking to Cylc UI Server, the Cylc Web UI can query directly from the Cylc UI Server, without having to talk to Cylc Workflow Service.

kinow:
This issue is pretty much being fixed by the same pull request as #71. When #71 gets fixed (there's a pending PR already), that means every client created will be pointing to the UI server. Including the one for tasks.

All we have to do here, after that, is to make sure we are parsing the result correctly (I tested yesterday and it appeared to work with no modification required), and see if we need to include/remove fields from the GraphQL query, and also update the HTML template.

kinow:
Done as the PR's referenced in the comment above have been done, and now the GraphQL queries are going through the Cylc UI :tada:

PR #72, opened 2019-05-21
Use UI Server GraphQL endpoint for suites

### LINKED ISSUE #71: Query workflows/suites from GraphQL instead of REST endpoint
At the moment we have the left-overs of a PoC, where we added a REST endpoint `/suites`, that executes `cylc scan`. While chatting with @dwsutherland today, he showed me the resolver for `workflows`.

If I run `cylc scan` in the terminal, I get:

```bash
dummy kinow@kinow-VirtualBox:43038
```

And if I run the GraphQL query 

```graphql
query {
  workflows {
    id, owner, name, host, port
  }
}
```

it returns:

```javascript
{
  "data": {
    "workflows": [
      {
        "id": "kinow/dummy",
        "owner": "kinow",
        "name": "dummy",
        "host": "kinow-VirtualBox",
        "port": 43038
      }
    ]
  }
}
```

So that means we don't need to keep GraphQL *and* the REST `/suites` endpoint. The latter is bad too for it forks the `cylc` command found in the `$PATH`.

kinow:
Close #71 

In the past during our first experiments with GraphQL, the endpoint was in `cylc/cylc-flow`. Now it resides in `cylc/cylc-uiserver`. The `cylc scan` was executed actually in a REST endpoint, that was executing the process locally in Python.

So this pull request removes the need to maintain that REST endpoint in `cylc-uiserver` (follow up PR after this one), and changes the Vue.js app to send a GraphQL query to the UI Server instead. The query used is:

```graphql
query {
  tasks {
    id, name, meanElapsedTime, namespace, depth
  }
}
```

This change is based on the pull request by @dwsutherland to `cylc-uiserver`, which is [still pending review](https://github.com/cylc/cylc-uiserver/pull/34). And while I believe we may have code changes, the data model proposed by @dwsutherland has received little negative feedback. So I believe it is safe now to start work on the Vue.js app based on that data model - furthermore, once we are pointing GraphQL queries to the right endpoint, changing the query and model is a 5 minutes job.

### DIFF
--- a/src/model/Suite.model.js
+++ b/src/model/Suite.model.js
@@ -1,17 +1,22 @@
 export default class Suite {
-  constructor(name, user, host, port) {
+  constructor(id, name, owner, host, port) {
+    this.id = id;
     this.name = name;
-    this.user = user;
+    this.owner = owner;
     this.host = host;
     this.port = port
+  }
+
+  getId() {
+    return this.id;
   }
 
   getName() {
     return this.name
   }
 
-  getUser() {
-    return this.user
+  getOwner() {
+    return this.owner
   }
 
   getHost() {
--- a/src/services/suite.service.js
+++ b/src/services/suite.service.js
@@ -1,9 +1,15 @@
 import Alert from '@/model/Alert.model'
 import { createApolloClient } from '@/utils/graphql'
-import axios from 'axios';
 import gql from 'graphql-tag'
 import store from '@/store/'
-import Suite from '@/model/Suite.model';
+
+// query to retrieve all suites
+const suitesQuery = gql`query {
+  workflows {
+    id, name, owner, host, port
+  }
+}
+`;
 
 // query to retrieve all suites
 const tasksQuery = gql`query {
@@ -14,26 +20,23 @@ const tasksQuery = gql`query {
 `;
 
 export const SuiteService = {
-  createGraphqlClient(host, port) {
-    return createApolloClient(`http://${host}:${port}/graphql`);
+  createGraphqlClient() {
+    return createApolloClient(`${window.location.pathname}/graphql`);
   },
   getSuites() {
-    return store.dispatch('suites/setSuites', []).then(() => {
-      return axios.get(window.location.pathname + '/suites').then((response) => {
-        const suites = [];
-        for (let i = 0; i < response.data.length; i++) {
-          const entry = response.data[i];
-          suites.push(new Suite(entry.name, entry.user, entry.host, entry.port));
-        }
-        return store.dispatch('suites/setSuites', suites);
-      }).catch((error) => {
-        const alert = new Alert(error.response.statusText, null, 'error');
-        return store.dispatch('addAlert', alert);
-      });
+    const apolloClient =  this.createGraphqlClient();
+    return apolloClient.query({
+      query: suitesQuery
+    }).then((response) => {
+      const suites = response.data.workflows;
+      return store.dispatch('suites/setSuites', suites);
+    }).catch((error) => {
+      const alert = new Alert(error.message, null, 'error');
+      return store.dispatch('addAlert', alert);
     });
   },
-  getSuiteTasks(suite) {
-    const apolloClient =  this.createGraphqlClient(suite.host, suite.port);
+  getSuiteTasks() {
+    const apolloClient =  this.createGraphqlClient();
     return apolloClient.query({
       query: tasksQuery
     }).then((response) => {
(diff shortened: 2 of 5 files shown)


kinow:
@dwsutherland I believe you have not had the chance to write JS code for Vue.js app. But I think you are the best person to review this PR, as you will understand where we are coming from for these changes. I will annotate the pull request with some comments that may be helpful :+1: 

If you would like, you can also start `cylc/cylc-uiserver` with your GraphQL work, and use this branch to confirm the GraphQL query goes through instead of a REST request.

Below my verification that it worked locally.

image

src/model/Suite.model.js
```diff
@@ -1,17 +1,22 @@
 export default class Suite {
-  constructor(name, user, host, port) {
+  constructor(id, name, owner, host, port) {
+    this.id = id;
     this.name = name;
-    this.user = user;
+    this.owner = owner;
```
kinow: Not sure why but I used `user` for a `Suite` (which will be renamed to Workflow in another change #73 ).

src/services/suite.service.js
```diff
@@ -1,9 +1,15 @@
...
 import gql from 'graphql-tag'
 import store from '@/store/'
-import Suite from '@/model/Suite.model';
+
+// query to retrieve all suites
+const suitesQuery = gql`query {
+  workflows {
+    id, name, owner, host, port
+  }
+}
```
kinow: The GraphQL query. `gql` is a special syntax which some day I will distill to understand how that works in JavaScript. But that comes from this [apollo library to parse graphql queries](https://github.com/apollographql/graphql-tag).

src/services/suite.service.js
```diff
@@ -14,26 +20,23 @@ const tasksQuery = gql`query {
 `;
 
 export const SuiteService = {
-  createGraphqlClient(host, port) {
-    return createApolloClient(`http://${host}:${port}/graphql`);
+  createGraphqlClient() {
+    return createApolloClient(`${window.location.pathname}/graphql`);
```
kinow: Here we were sending requests to whatever `cylc scan` returned for `host:port`. That's because we were sending the queries to `cylc/cylc-flow`. Instead now we copy the URL + `/graphql`, which should match the Tornado Handler for GraphQL.

src/services/suite.service.js
```diff
@@ -14,26 +20,23 @@ const tasksQuery = gql`query {
...
-      });
+    const apolloClient =  this.createGraphqlClient();
+    return apolloClient.query({
+      query: suitesQuery
+    }).then((response) => {
+      const suites = response.data.workflows;
+      return store.dispatch('suites/setSuites', suites);
+    }).catch((error) => {
+      const alert = new Alert(error.message, null, 'error');
+      return store.dispatch('addAlert', alert);
```
kinow: Here is the change to query from GraphQL, removing the query to REST (I copied the existing call to retrieve tasks, which is broken, but will be fixed in #70 )

src/services/suite.service.js
```diff
@@ -14,26 +20,23 @@ const tasksQuery = gql`query {
...
     });
   },
-  getSuiteTasks(suite) {
-    const apolloClient =  this.createGraphqlClient(suite.host, suite.port);
+  getSuiteTasks() {
+    const apolloClient =  this.createGraphqlClient();
```
kinow: No more need for information from the suite, we know we will talk to the UI Server used by this Vue.js app, not to a Cylc Workflow Service anymore.

tests/unit/services/suite.service.spec.js
```diff
@@ -42,10 +52,13 @@ describe('SuiteService', () => {
...
-      e.response = {
-        statusText: 'Test Status'
+      const stubClient = {
+        uri: null,
+        query: function() {
+          return Promise.reject(e);
+        }
       };
-      sandbox.stub(axios, 'get').rejects(e);
+      sandbox.stub(SuiteService, 'createGraphqlClient').returns(stubClient);
```
kinow: The rest here are just changes for tests to match what was changed before :+1: hope that helps you reviewing this.

dwsutherland:
Yes, use the `workflows` query as the `cylc scan`... and you can always include tasks like this:
```
query {
  workflows {
    id
    name
    owner
    host
    port
    stateTotals
    tasks {
      id
      name
      proxies {
        cyclePoint
        state
        meanElapsedTime
        namespace
      }
  }
}
```

kinow:
One review ruling for now :) merging it :+1:

Issue #77, opened 2019-05-24
Investigate simulating GraphQL responses for testing


kinow:
We started with the Suite Graph View issue (#74) but then we had a meeting, and after that others joined the discussion and now we have another broader issue for all Suite Views (#75), and it was also mentioned that it would be important to work on the testing part.

One part of the testing required is some simulated data, to be used by many tests to simulate possible states of a workflow, and pass through our code to verify its behaviour.

In #74, I was going to exactly work on the looking if a VCR-like tool could be used to record-persist-replay the requests of real workflows. And then use this persisted data for tests.

Now with the #75, I think it might be easier to work on this new issue, instead of documenting this part of testing in #74.

kinow:
There is a complication, as most of the existing tools were used by traditional web applications, where a user would send a request to Service#1, then get the response, call Service#2, and so on.

In our case, we have one endpoint only (`/graphql`), where we are sending the same request (`?query=QUERY`), but the response is varying with time, as the workflow state changes.

Which means we would have to find a tool that records multiple interactions. Most of the tools I've seen so far overwrite or raise an error when the same requests appears more than once.

kinow:
The best JS options were:

- https://github.com/linkedin/sepia
    * Looks promising, from Linkedin, quite a few stars on GitHub. But last commit from 2015.
- https://github.com/flickr/yakbak
    * From Flickr, but with last commit from Jan 2018 too.

So will probably check the Python alternatives instead. We could use it to record the requests, and import into this `cylc-ui` project as test resources.

hjoliver:
It might be a lot simpler (if less elegant) to hack cylc-7 and do a JSON dump (to file) of the suite data structure, then manually massage that into the new GraphQL schema?  That would be easy for a small test suite at least.  It might be painful for the 1000+ task complex demo suite.  But we can start small!

kinow:
For Python the most famous is [vcrpy](https://github.com/kevin1024/vcrpy). Even the deprecated [uber/cassette](https://github.com/uber/cassette) now recommends users to use it too.

kinow:
>It might be a lot simpler (if less elegant) to hack cylc-7 and do a JSON dump (to file) of the suite data structure, then manually massage that into the new GraphQL schema? That would be easy for a small test suite at least. It might be painful for the 1000+ task complex demo suite. But we can start small!

[...]

Issue #75, opened 2019-05-23
Implement Views


oliver-sanders:
Implement all Vue views:

* [ ] [GScan](https://github.com/cylc/cylc-ui/issues/139)
* [ ] Implement the original "classic" Cylc suite views.
  * [ ] [Tree View](https://github.com/cylc/cylc-ui/issues/78)
  * [ ] [Dot (LED) View](https://github.com/cylc/cylc-ui/issues/79)
  * [ ] [Graph View](https://github.com/cylc/cylc-ui/issues/74)
* [ ] Implement "static" view(s) as a replacement to `cylc graph`:
  * [ ] [Graph View](https://github.com/cylc/cylc-ui/issues/82)
* [ ] Implement desired new views:
  * [ ] [Dashboard](https://github.com/cylc/cylc-ui/issues/94)
  * [ ] [Gantt View](https://github.com/cylc/cylc-ui/issues/181)
  * [ ] [Job analysis](https://github.com/cylc/cylc-ui/issues/262)
  * [ ] [Suite analysis](https://github.com/cylc/cylc-ui/issues/262)
* [ ] Pursue possible "blue sky" ideas for future suite views:
  * [ ] Matrix View? (#52) ([sketch](https://user-images.githubusercontent.com/16705946/45327187-6ddbd480-b54e-11e8-8af1-d126bd490aa0.png))

kinow:
Note to self: if, when implementing these views, we keep in mind to use or create components that are able to search the current view, that could be useful towards fixing #51

oliver-sanders:
Note: the desire to select a task in one view and highlight it in another (e.g. highlight a task in the tree view and have the graph view focus to that point). Views should have a standard representation of the users focus/selection.

kinow:
>Note: the desire to select a task in one view and highlight it in another (e.g. highlight a task in the tree view and have the graph view focus to that point). Views should have a standard representation of the users focus/selection.

I was thinking about sharing the data model between views too, but I think we will have slightly different data structures for each view. Might have to keep only the selection as you suggested, and each view will handle the work of interpreting the selection values and displaying the data correctly.

kinow:
We (southern hemisphere team) had a quick VC yesterday about GraphQL & cylc-ui Views. I thought I could filter queries by pretty much any field we had in the types we had coming from queries (similar to SQL).

[...]

PR #80, opened 2019-05-26
Show application mode in footer

kinow:
Display the application mode in the footer, when it is not equal to `PRODUCTION`. This is useful for troubleshooting and development.

I am planning to use a different response in the services in an mode (e.g. `offline`, for #75 and #77 ). This way we will be able to run the UI completely offline, with some mocked data. But also use other modes like `development` that could still point to development UI servers.

### DIFF
--- a/src/components/core/Footer.vue
+++ b/src/components/core/Footer.vue
@@ -25,6 +25,7 @@
     </div>
     <v-spacer/>
     <span class="font-weight-light copyright">
+      <strong v-if="env !== 'PRODUCTION'">{{ env }}</strong>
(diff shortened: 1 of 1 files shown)


kinow:
To review/test this PR, this [vue.js doc page](https://cli.vuejs.org/guide/mode-and-env.html#modes) has more about modes, and it's possible to runt he `cylc-ui` project alone simulating modes, e.g.

```bash
$ NODE_ENV=staging npm run serve
```

image

kinow:
Thanks for testing it Tim!!!

Issue #78, opened 2019-05-24
Create Suite Tree View


kinow:
Implement something similar to what we had in Cylc 7: 

![gcylc-text-view](https://user-images.githubusercontent.com/304786/58298113-5c153580-7e2e-11e9-92b9-b939902e7da7.png)

EDIT: more recent screenshots, taken with 7.8.1.

![image](https://user-images.githubusercontent.com/304786/58672058-502ef380-8399-11e9-8c3a-163d3d5b38a7.png)

![image](https://user-images.githubusercontent.com/304786/58672079-62109680-8399-11e9-85a3-c3f66369b460.png)

![image](https://user-images.githubusercontent.com/304786/58672117-8f5d4480-8399-11e9-8586-0c69dc9add9c.png)

And a few features in the current Tree View:

- Users can adjust font size
- Users can choose themes (color blind, high contrast, etc)
- Users can collapse / expand all (though it changes once the family/tasks are updated)
- Users can filter by status
- Users can filter by a regex (not sure what is searched, task name?)
- Users are displayed a legend of what each task color means in the tree view
- Users have access to a context menu with actions such as
    * Browse task URL (if URL is defined, otherwise a warning is displayed)
    * Trigger (if possible)
    * View logs, prereqs, outputs, call `cylc show`
    * Reset
    * Force spawn
    * Hold
    * Release
    * Remove
    * Enable/disable family grouping

Family grouping doesn't seem very simple. Unless we alter the underlying data, but still not sure if the Vue.js component will behave OK with that.

sadielbartholomew:
For the basic tree-view code, there are some really nice pre-existing 'Tree' components in the [Awesome Vue listing under the 'Components & Libraries -> UI Components -> Tree' heading](https://github.com/vuejs/awesome-vue#tree). Too many to pick from! But as some good (while not necessarily the best, as I don't have time to look at them all) examples:

| Tree component: name  | homepage  | demo |
|---|---|----|
| Sl-vue-tree | [here](https://github.com/holiber/sl-vue-tree) | [here](https://holiber.github.io/sl-vue-tree/demo/index) |
| Bosket  | [here](https://github.com/elbywan/bosket)  | [here](https://elbywan.github.io/bosket/vue/index.html#TreeView#Demo) |
| vue-draggable-nested-tree  |  [here](https://github.com/phphe/vue-draggable-nested-tree) | [here](https://codepen.io/phphe/pen/KRapQm) |

[...]

PR #86, opened 2019-05-30
Introduce mode offline with mocked data for testing/development

### LINKED ISSUE #77: Investigate simulating GraphQL responses for testing

kinow:
Close #77 

The only important change for now in this PR is `vue.config.js`, which gets an __alias__ `suite-service`, used in JavaScript when importing the `SuiteService`.

When a developer specifies `NODE_ENV=?`, that value will be used, and if the value is equal to `offline`, then the returned value will be `@/services/mock/suite.service`. Otherwise it will use the real service `@/services/suite.service`, that instead of using the mocked data, it will pull from the GraphQL endpoint.

Other changes here include some new classes that will be used in #78 , with some changes to accommodate the required data structure of the Vue.js component used to render the tree view.

### DIFF
--- a/src/model/Family.model.js
+++ b/src/model/Family.model.js
@@ -1,0 +1,13 @@
+import Task from '@/model/Task.model'
+
+export default class Family {
+  constructor (name, cyclePoint, state, depth, childTasks) {
+    this.name = name
+    this.cyclePoint = cyclePoint
+    this.state = state
+    this.depth = depth
+    this.childTasks = childTasks !== undefined ? childTasks.map((childTask) => {
+      return new Task(childTask.id, childTask.state, childTask.latestMessage, childTask.depth, childTask.jobs)
+    }) : []
+  }
+}
--- a/src/model/Task.model.js
+++ b/src/model/Task.model.js
@@ -1,0 +1,16 @@
+export default class Task {
+  constructor (id, state, latestMessage, depth, jobs) {
+    this.id = id
+    this.state = state
+    this.latestMessage = latestMessage
+    this.depth = depth
+    this.jobs = jobs
+  }
+
+  get jobId () {
+    if (this.jobs !== undefined && this.jobs.length > 0) {
+      return this.jobs.slice(-1)[0].batchSysJobId
+    }
+    return null
+  }
+}
--- a/vue.config.js
+++ b/vue.config.js
@@ -18,5 +18,9 @@ module.exports = {
 
       config.devtool('inline-cheap-module-source-map')
     }
+
+    // set up aliases for mock services, used when the offline mode is used
+    const suiteService = process.env.NODE_ENV === 'offline' ? '@/services/mock/suite.service' : '@/services/suite.service';
+    config.resolve.alias.set('suite-service', suiteService)
   }
 }
--- a/vue.config.js
+++ b/vue.config.js
@@ -20,7 +20,9 @@ module.exports = {
     }
 
     // set up aliases for mock services, used when the offline mode is used
-    const suiteService = process.env.NODE_ENV === 'offline' ? '@/services/mock/suite.service' : '@/services/suite.service';
+    const suiteService = process.env.NODE_ENV === 'offline'
+      ? '@/services/mock/suite.service'
+      : '@/services/suite.service'
     config.resolve.alias.set('suite-service', suiteService)
   }
 }
--- a/src/services/mock/suite.service.js
+++ b/src/services/mock/suite.service.js
@@ -1,0 +1,257 @@
+import store from '@/store/'
+
+export const SuiteService = {
+  getSuites() {
+    // https://github.com/kinow/cylc-cassettes/tree/master/cassettes/five/suite.yaml
+    const suites = [
+        {
+          "id": "kinow/five",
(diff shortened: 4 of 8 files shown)


kinow:
Coverage should go down, but that'll be fixed in #77, when I will know what changes are necessary in these objects to display the suite tree view. Added here just to test and confirm the returned data can be used :+1:

kinow:
Just rebased onto master, fixed unused import linter error, and finished testing and fixed the use without the mocked data - the introduced changes for mocked data needed to change to use a class in JS instead of a constant.

Tested with mocked data, and then with a real suite (using David's pull request). Everything appeared to work OK. Ready for review :+1:

matthewrmshin:
Travis CI not happy.

kinow:
Interesting, we have functional tests to verify the database creation commands :grimacing: will fix those later. But once I fix the 1-to-1 relationship, it would be great if you or somebody else could have a look and see if that makes sense, please, @matthewrmshin .

A few more notes about it:

- There is a risk - well pointed yesterday by @oliver-sanders - of issues with DB updates. Though I think adding a constraint won't break anything. But we still need to take that into consideration when review/testing it.
- If it breaks anything, or if it is not a good change to Cycl 8, we can simply get this generated diagram and close this issue. As this issue is not for adding constraints, but for documenting the DB schema (I reckon our DB scheme changes very rarely, and it's not hard to re-generate the diagram with some `git` diff-fu)

kinow:
Unit test failed as I changed from constant to class, and the mock needed to be updated. Unit tests fixed locally, push-forced the edit commits :+1:

src/model/Task.model.js
```diff
@@ -0,0 +1,16 @@
...
+    this.id = id
+    this.state = state
+    this.latestMessage = latestMessage
+    this.depth = depth
+    this.jobs = jobs
+  }
+
+  get jobId () {
+    if (this.jobs !== undefined && this.jobs.length > 0) {
+      return this.jobs[0].batchSysJobId
```
matthewrmshin: Should go for latest `this.jobs[jobs.length -1]`.

src/model/Family.model.js
```diff
@@ -0,0 +1,13 @@
+import Task from '@/model/Task.model'
+
+export default class Family {
+  constructor (name, cyclePoint, state, depth, childTasks) {
+    this.name = name
+    this.cyclePoint = cyclePoint
```
matthewrmshin: Why do we need a cycle point for a family? I guess this is a family of task proxies instead of a family task definitions?

Would it be clearer to call it `FamilyProxy` (and `TaskProxy`)?

vue.config.js
```diff
@@ -18,5 +18,9 @@ module.exports = {
 
       config.devtool('inline-cheap-module-source-map')
     }
+
+    // set up aliases for mock services, used when the offline mode is used
+    const suiteService = process.env.NODE_ENV === 'offline' ? '@/services/mock/suite.service' : '@/services/suite.service';
```
matthewrmshin: I wonder if you can break up long lines with ternary operator this way?

```javascript
const suiteService = process.env.NODE_ENV === 'offline'
    ? '@/services/mock/suite.service'
    : '@/services/suite.service';
```

src/model/Task.model.js
kinow: Found another way. Done with [slice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice) `this.jobs.slice(-1)[0]`

vue.config.js
kinow: Yup! Worked with no issues.

src/model/Family.model.js
kinow: >Why do we need a cycle point for a family? I guess this is a family of task proxies instead of a family task definitions?

Question for https://github.com/cylc/cylc-flow/pull/3122 I think.

>Would it be clearer to call it FamilyProxy (and TaskProxy)?

I think it could be... but a `FamilyProxy` has `childTasks` and `childFamilies`... which I am not sure if they are proxies too or not... also not clear whether we will have `Family` anywhere else, or if the `FamilyProxy` is supposed to represent the `Family` for the UI... in that case I think `Family` would be easier to understand in the UI... but maybe we can leave this one to be discussed later?

kinow:
Rebased master and addressed two of the points raised by @matthewrmshin  (thanks!).

src/model/Family.model.js
matthewrmshin: Yes. We should clarify this after merging @dwsutherland GraphQL changes in the other projects..

PR #89, opened 2019-06-06
Display view title even after refresh, and include suite name as example

### LINKED ISSUE #88: Toolbar title disappearing when a screen is reloaded
The [Vuetify Material Dashboard](https://www.creative-tim.com/product/vuetify-material-dashboard) includes a Toolbar. That's the component that appears near the top of the page, with a search, notification, and the current View Title.

When you navigate to some URL like `/user/kinow/#/suites/five`, then it displays `Suite` as the screen title.

image

But if you refresh the browser tab, then the title is gone.

image

kinow:
Close #88 

Moves the View title from the Toolbar component data section to a global Vuex (`title` property of the `app` module).

The Suite view now works even after refreshing, and as the data is global, the Suite View is able to set the title to "Suite five", for instance.

Cheers
Bruno

### DIFF
--- a/src/components/core/Toolbar.vue
+++ b/src/components/core/Toolbar.vue
@@ -81,11 +81,9 @@
 
 <script>
 
-import {
-  mapMutations
-} from 'vuex'
+  import { mapMutations, mapState } from 'vuex'
 
-export default {
+  export default {
   data: () => ({
     notifications: [
       'Mike, John responded to your email',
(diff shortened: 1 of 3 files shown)


kinow:
And the reason for the bug in #88, was that the initial value of title was `null`. Then we set a _watcher_ on the Vue Router `$route` object and updated the title when the route changed.

The problem was that on reloading the page, the watch received no events, only when the user navigated through pages. So the title remained `null`. Now it's setting the title on `mounted`.

PR #95, opened 2019-06-28
cylc-ui graph screen

MartinRyan:
adds a graph visualisation screen to cylc-ui


kinow:
Will squash the commits and merge.

Few issues with icons when the family is expanded:

image

And also a few issues in the linter. Other than that, looks pretty good! :clap:

package.json
```diff
@@ -40,16 +56,19 @@
     "@vue/cli-plugin-eslint": "^3.8.0",
     "@vue/cli-plugin-unit-mocha": "^3.8.0",
     "@vue/cli-service": "^3.8.0",
-    "@vue/test-utils": "^1.0.0-beta.20",
```
kinow: This causes issues running unit tests.

PR #97, opened 2019-06-28
Mount component once at test, avoiding timeouts

kinow:
Fix unit tests by avoiding timeouts in tests by mounting the component just once.

There are warnings about canvas not available, but the tests are passing locally. I think we will need to review the tests later.

Feel free to merge if the build is passing on Travis :+1:


kinow:
(if the build fails, I will work on fixing the unit tests next week)

PR #101, opened 2019-07-03
Task job icons

### LINKED ISSUE #100: component: task and job status icons
Create Vue components for task and job state icons for incorporation into the suite views.

oliver-sanders:
close #100 

Add components for the task an job status icons.

* One component for tasks and one for jobs, respond to state changes ala Vue.
* `inline-block` level components `1em x 1em` which can sit in text without disrupting the baseline (much).

### Integration

I'm not sure how best to incorporate these components into a Vue view. We want these components to feed off of the data model of the component which uses them, are `props` good enough for this?

For the moment these are standalone components.

### Example

```vue
<template>
  <span>
    <Task />
    foo
    <Job />
  </span>
</template>
```

Fiddling the `data` in the components you will get something like this:

Screenshot from 2019-07-03 16-41-28

### TODO (future work)

* Incorporate into views.
    * Work out how we are going to feed data into these components (e.g. props or whatever).
* Animate state changes.
    * Pure CSS probably wont be enough.
* Add the more eccentric task states (expired, queued, retrying) and the held icon.
    * Expired and queued are easy, the rest will require a dig into the Cylc Flow code to see what separation is possible with the current internals of Cylc.
* Extract the `<style>` from the Job component to make it theme-able.

### DIFF
--- a/src/components/cylc/Job.vue
+++ b/src/components/cylc/Job.vue
@@ -1,0 +1,82 @@
+<!--
+  Job - The job status icon implemented as an inline-block level object
+        which can sit in text.
+-->
+
+<template>
+  <span style="display:inline-block; vertical-align:middle">
+    <!-- the task icon SVG
+           * comments prefixed `let` are instructions for changing style
+           * contain in a 100x100 viewBox so pixels and percent are equal
+           * bind the job status here, respond to styling in the CSS
+    -->
+    <svg 
+      class="job"
+      v-bind:class="[status]"
+      viewBox="0 0 100 100"
+    >
+      <!-- the job status icon
+           * let width = 100 - x - stroke-width
+           * let height = 100 - y - stroke-width
+      -->
+      <rect
+        x="10" y="10"
+        width="80" height="80"
+        rx="20" ry="20"
+        stroke-width="10"
+      />
+    </svg>
+  </span>
+</template>
+
+
+<style lang="scss">
+    svg.job {
+        /* scale the icon to the font-size */
+        width: 1em;
+        height: 1em;
+
+        rect {
+            /* if no job status display nothing */
+            fill: transparent;
+            stroke: transparent;
+        }
+
+        &.submitted rect {
+            fill: rgb(125,207,212);
+            stroke: rgb(125,207,212);
+        }
+
+        &.running rect {
+            fill: rgb(106,164,241);
+            stroke: rgb(106,164,241);
+        }
+
+        &.succeeded rect {
+            fill: rgb(81,175,81);
+            stroke: rgb(81,175,81);
+        }
+
+        &.failed rect {
+            fill: rgb(207,72,72);
+            stroke: rgb(207,72,72);
+        }
+
+        &.submit-failed rect {
+            fill: rgb(190,106,192);
+            stroke: rgb(190,106,192);
+        }
+    }
+</style>
+
+
+<script>
+    export default {
+        name: 'Job',
+        data: function () {
+            return {
+                'status': 'running'
+            }
+        }
+    }
+</script>
--- a/src/components/cylc/Task.vue
+++ b/src/components/cylc/Task.vue
@@ -1,0 +1,176 @@
+<!--
+  Task - The task status icon implemented as an inline-block level object
+         which can sit in text.
+-->
+
+<template>
+  <span style="display:inline-block; vertical-align:middle">
+    <!-- the task icon SVG
+           * comments prefixed `let` are instructions for changing style
+           * contain in a 100x100 viewBox so pixels and percent are equal
+           * bind the task status here, respond to styling in the CSS
+    -->
+    <svg 
+      class="task"
+      v-bind:class="[status]"
+      viewBox="0 0 100 100"
+    >
+      <!-- progress pie chart:
+             * position in the middle
+             * let radius = 25% radius
+             * let stroke-width = 100% - (2*radius) = 50%
+             * let stroke-dasharray ~= 2 * pi * radius
+             * use dashes as the line style to turn this into a pie chart
+      -->
+      <circle
+        id="progress"
+        cx="50" cy="50"
+        r="25"
+        stroke-width="50"
+        stroke-dasharray="157"
+        v-bind:style="progressStyle"
+      ></circle>
+      <!-- circle in the middle
+             * position in the middle
+             * radius can be changed independently
+      -->
+      <circle
+        id="hub"
+        cx="50" cy="50"
+        r="12">
+      </circle>
+      <!-- outer circle
+             * position in the middle
+             * let radius = 50% - (2*stroke-width)
+      -->
+      <circle
+        id="outline"
+        cx="50" cy="50"
+        r="46"
+        stroke-width="8">
+      </circle>
+      <!-- cross (failure)
+             * create a plus and rotate it 45 deg around the origin
+             * let rx ~= max(width, height) * 0.1
+      -->
+      <g
+        id="cross"
+        transform="rotate(45, 50, 50)"
+      >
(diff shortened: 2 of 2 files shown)


src/components/cylc/Task.vue
```diff
@@ -0,0 +1,176 @@
...
+            return {
+                'status': 'running',
+                'progress': 25
+            }
+        },
+        computed: {
+            progressStyle: function () {
+                return {
+                    'stroke-dashoffset':
+                        'calc(157 - (157 * ' + this.progress/100 + '))'
```
matthewrmshin: I haven't done JS for a long time, but is it OK to concatenate strings and numbers like this?

oliver-sanders: It's OK and quite common, results are predictable:

```js
'a' + 1
"a1"
'a' + 0.1
"a0.1"
'a' + 0.10
"a0.1"
'a' + null
"anull"
```

However, it is nicer to use the new format strings which one can apparently embed statements in (because JS and Python are getting kinda similar nowerdays).

matthewrmshin: OK. So JS is closer to Perl than Python. 😉

kinow:
>Extract the <style> from the Job component to make it theme-able.

+1 :clap: 

Also, currently the `.job` class is global, and could be used by accident somewhere else, or even used by a third party component. The Vuetify components use `v-$class-name` to avoid it, and another option is to use [scoped styles](https://vue-loader.vuejs.org/guide/scoped-css.html#mixing-local-and-global-styles).

Though I don't think we need to bother about it now.

kinow:
>Integration
>
>I'm not sure how best to incorporate these components into a Vue view. We want these components to feed off of the data model of the component which uses them, are props good enough for this?
>
>For the moment these are standalone components.

Would you like to change it? Luckily @hjoliver prompted us to fix the suite tree view, and as it is working with live data, I tried using your new component (`Task`, not sure how to add the `Job` there yet).

Here's what the diff for my current working copy looks like:

```diff
diff --git a/src/components/cylc/Task.vue b/src/components/cylc/Task.vue
index 42bf7c9..497e6a6 100644
--- a/src/components/cylc/Task.vue
+++ b/src/components/cylc/Task.vue
@@ -10,7 +10,7 @@
            * contain in a 100x100 viewBox so pixels and percent are equal
            * bind the task status here, respond to styling in the CSS
     -->
-    <svg 
+    <svg
       class="task"
       v-bind:class="[status]"
       viewBox="0 0 100 100"
@@ -158,10 +158,16 @@
 <script>
     export default {
         name: 'Task',
-        data: function () {
-            return {
-                'status': 'running',
-                'progress': 25
+        props: {
+            status: {
+                type: String,
+                required: true,
+                default: 'waiting'
+            },
+            progress: {
+                type: Number,
+                required: true,
+                default: 0
             }
         },
         computed: {
diff --git a/src/views/Suite.vue b/src/views/Suite.vue
index d795b4d..22f9b4f 100644
--- a/src/views/Suite.vue
+++ b/src/views/Suite.vue
@@ -31,6 +31,9 @@
                 <i class="mdi mdi-plus" v-else></i>
               </span>
             </template>
+            <template slot="state" slot-scope="props">
+              <task :status="props.row.state" :progress=0 />
+            </template>
           </vue-ads-table>
         </material-card>
       </v-flex>
@@ -41,10 +44,14 @@
 <script>
   import { SuiteService } from 'suite-service'
   import {mapState} from 'vuex'
+  import Task from '@/components/cylc/Task'
 
   const suiteService = new SuiteService();
 
   export default {
+    components: {
+      'task': Task
+    },
     metaInfo() {
       return {
         title: 'Cylc UI | Suite ' + this.$route.params.name
``` 

(ignore the `<svg line that was just space-trimmed by the IDE`).

The only important part of the change was the `props`. I believe that the answer for your question on whether it would be good enough for this, would be yes for now. We could also use slots, which gives us more flexibility, but we don't have the need for now.

Here's what it will look like with the suite five: https://kinow.github.io/#/suites/five. You will have to wait a few seconds, as it looks like there is no icon for waiting yet?

Cheers

kinow:
And I think the GraphQL endpoint was either not giving the progress, or I simply forgot to include that in the query (@dwsutherland quick question, is there a field for progress already?)

src/components/cylc/Job.vue
```diff
@@ -0,0 +1,82 @@
...
+
+        &.failed rect {
+            fill: rgb(207,72,72);
+            stroke: rgb(207,72,72);
+        }
+
+        &.submit-failed rect {
+            fill: rgb(190,106,192);
+            stroke: rgb(190,106,192);
+        }
```
kinow: Do we need one value for waiting too? I don't know all the states, but this was blank when running a live suite with this change.

hjoliver:
> as it looks like there is no icon for waiting yet?

I see waiting icons (empty circle).

hjoliver:
> quick question, is there a field for progress already?

I'm guessing not.  As I recall progress bars in the old GUI (which I think I implemented) are computed in the UI, based on run-length-so-far vs expected-or-average-run-length.  Could potentially be done server side, but I think @dwsutherland most just replicated the old data elements to start with.

kinow:
>I see waiting icons (empty circle).

Oh, maybe I confused the status values.

src/components/cylc/Job.vue
hjoliver: So, waiting is a task status, but not a job status (if you look at @oliver-sanders design mock-up doc).

hjoliver:
> > I see waiting icons (empty circle).
> 
> Oh, maybe I confused the status values.

I think answered this above about 1 sec ago :-)  - waiting is a task status but not a job status.

kinow:
>I'm guessing not. As I recall progress bars in the old GUI (which I think I implemented) are computed in the UI, based on run-length-so-far vs expected-or-average-run-length. Could potentially be done server side, but I think @dwsutherland most just replicated the old data elements to start with.

Ah, I think it was discussed (either your or @oliver-sanders) explained exactly that to me. Now that you mention I recall something about it being done on client side... and maybe a median value coming from the server I think... or something similar. We can calculate the progress later too, as we should then have the data available.

kinow:
>I think answered this above about 1 sec ago :-) - waiting is a task status but not a job status.

Hmmm... confusing. Do we have task & job well separated in the old GUI?

The component I am using is `Task`... so it should cover the task statuses... I think?

hjoliver:
I think I resolved the question, and 3 approvals now, so will merge :tada:

hjoliver:
> Do we have task & job well separated in the old GUI?

No, this is a new thing, for the new UI.  In the old GUI we only have "task status" but some states correspond to job states.  E.g.  a "waiting" task can't have a job status, because the job doesn't exist yet.  But a "running" task corresponds to a running job.  And so on.

hjoliver:
> The component I am using is Task... so it should cover the task statuses... I think?

I think that is the right thing to do for now, until we add distinct job status (via the square coloured icons) to the UI.  So all good :+1:

hjoliver:
@oliver-sanders and/or @kinow - I guess the "TODO (future work)" bullet points in the description should be moved to new issues.

oliver-sanders:
> quick question, is there a field for progress already?

There is a the `dt` field, we could either calculate progress client or server side. I guess server side would lead to better consistency...

oliver-sanders:
> The Vuetify components use v-$class-name to avoid it

Ahh, I though Vue would magically namespace my CSS! I'll raise a follow-on to do this.

hjoliver:
> @oliver-sanders and/or @kinow - I guess the "TODO (future work)" bullet points in the description should be moved to new issues.

#109

PR #105, opened 2019-07-06
Fix graph when using a mode that is not offline

kinow:
There are two commits in this PR.

The first is to remove files that I had to play with and see whether they were used or not. I suspect some were used at earlier stage of development, and may be used later - the view is a work-in-progress, and I believe Martin had a few more things to work on his mind.

But removing these I think will help others to focus on only what's being used at the moment. Plus these files should still be present in his branch.

The second commit fixes the issue with the Graph View, when running behind the JupyterHub. In Exeter, we saw the Graph View running in offline mode (pure JS, no Python). The URL in that case is `localhost:8080`. 

There is a JSON file with the graph data, that was being fetched in JS via URL `http://localhost:8080/.../*.json`. The second commit does two things to fix this:

1. Rename the file to *.js
2. Change the URL to be relative, so that it goes through the proxy

The first change #1 is necessary to avoid having to touch the current static assets filter in the Tornado app in Cylc UI Server. The 2# is what actually fixes the bug.

Let me know if that works @hjoliver . I tested with mode offline (i.e. `NODE_ENV=offline npm run serve` and checking `localhost:8080`), and also with the normal end-to-end system with `jupyterhub` + `npm run build:watch`.

Cheers
Bruno


src/views/Graph.vue
```diff
@@ -91,7 +91,8 @@ const config = {
       css: {
         'background-image': function(node) {
           let path = node.data('icon')
-          path === undefined || path === '' ? path = require('@/assets/baseline-donut_large-24px.svg') : ''
+          // TODO: not sure if we can provide the path as a normal string, without importing the svg here?
```
kinow: Oh, also added a TODO that I want to investigate later. I normally use `require` when importing things, but I need to check what's its return type... as I would expect the `path` variable here to be a string... just a note to self to clarify later (we may be able to use relative URL's, removing the need for require, which is easier to test)

hjoliver:
Same test failing on another PR, so it's not a bug on this branch.  https://github.com/cylc/cylc-ui/pull/99#issuecomment-508904687  

I'll merge this so I can use it easily in upcoming demo.

PR #106, opened 2019-07-07
Fix unit test after applying transformer

kinow:
I broke the build when fixing #105 as I applied the transformer to the GraphQL data. The unit test then instead of simply getting the data returned as-is, it started getting the data after the transformation.

So just updated the test to match the returned data. This should increase the coverage a bit, but that's not really relevant. As the component will change, as well as layout :+1:


kinow:
Thanks @matthewrmshin !

PR #108, opened 2019-07-08
Display single error message

kinow:
Instead of accumulating the error messages in a list, after this PR we will have just one alert. If, for any reason, the alert changes (e.g. you get a "Network error", and then something causes a "Unknown error"), then the latest message will be displayed.

Users can still dismiss the error. And when moving to a different route, it should be correctly removed, and no more alert/error should be displayed in the new page/view.

### DIFF
--- a/src/components/core/Alert.vue
+++ b/src/components/core/Alert.vue
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <v-container v-for="alert in alerts" :key="alert.text">
+    <v-container v-if="alert !== null" :key="alert.text">
       <v-alert :value="true" :color="alert.getColor()" :icon="alert.getIcon()" dismissible>
         {{ alert.getText() }}
       </v-alert>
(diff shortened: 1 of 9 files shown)


kinow:
Two ways to test it:

1. Check out branch, then once everything is working, just go to some page like the Suites View, and stop `jupyterhub` process. That will cause the next poll HTTP request to fail.
2. Edit `suite.service.js`, and in `getSuites` throw an Error(...) - or in another part of the code

matthewrmshin:
While this solves our immediate problem, is this going to hide information if there are genuinely multiple alerts?

kinow:
Good point. We could log errors to the browser console, or use a sort of `set` with the old approach, adding each message just once. wdyt?

matthewrmshin:
Logging errors to the browser console sound like the most sensible - as long as the user is alerted to it. A `set` may not work because you may still have multiple errors and you will still end up with a stripy flag.

kinow:
Rebased, fixing conflicts, and then added the `console.log` line when the alert is not `null`.

wxtim:
## Review
- [x] Looks Sane
- [x] Reproduced error in master with a suite 
- [x] Checked that this branch didn't do the same thing.

PR #110, opened 2019-07-09
Propize task component

kinow:
This PR adds props to the `Task` component, and also updates the current suite tree view to use this new component :+1:

### DIFF
--- a/src/components/cylc/Task.vue
+++ b/src/components/cylc/Task.vue
@@ -13,7 +13,7 @@
            * contain in a 100x100 viewBox so pixels and percent are equal
            * bind the task status here, respond to styling in the CSS
     -->
-    <svg 
+    <svg
       class="task"
       v-bind:class="[status]"
       viewBox="0 0 100 100"
@@ -164,10 +164,15 @@
 <script>
     export default {
         name: 'Task',
-        data: function () {
-            return {
-                'status': null,
-                'progress': 0
+        props: {
+            status: {
+                type: String,
+                required: true
+            },
+            progress: {
+                type: Number,
+                required: false,
+                default: 0
             }
         },
         computed: {
--- a/src/views/Suite.vue
+++ b/src/views/Suite.vue
@@ -31,6 +31,9 @@
                 <i class="mdi mdi-plus" v-else></i>
               </span>
             </template>
+            <template slot="state" slot-scope="props">
+              <task :status="props.row.state" :progress=0 />
+            </template>
           </vue-ads-table>
         </material-card>
       </v-flex>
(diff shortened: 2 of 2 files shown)


src/components/cylc/Task.vue
```diff
@@ -158,10 +158,16 @@
...
         name: 'Task',
-        data: function () {
-            return {
-                'status': 'running',
-                'progress': 25
+        props: {
+            status: {
+                type: String,
+                required: true,
+                default: 'waiting'
```
oliver-sanders: *Purely out of academic interest* is there a way of using an enum from this, I think we can extract the allowed values using introspection on GraphQL.

kinow: That's in my browser history from yesterday. Something like 'vue enum props component'. Unfortunately there is no clear solution. Some people recommended `type: Function` or `Object` if I recall correctly, then you would have a bit more of control.

A good improvement. Do you think we should create an issue for this, or add a TODO here?

oliver-sanders: We can leave this as is, a nice improvement but not an important one.

kinow:
Great comment as always @sadielbartholomew, and we don't have any hurry to merge this one, so any constructive discussion is very welcome IMHO. 

>So could we add a note to emphasise it is the status of the Task component, not the task state? Not a blocker, but I think it is an easy assumption to make given the similarity in status & state as descriptors.

Sure, would it be just a comment? But the comment would just be something like `<-- This is the task status, not the task state -->`?

I haven't thought much about it, to be honest, as @oliver-sanders did all the hard work to create the component. His initial work had state and progress, so this PR was more to move them from data values, to props that can be passed to the component.

>Then I was confused as to why it was safe & wise to (as on new line 165) set a default of 'waiting'

I did think a little about this one... but I couldn't come up with a better alternative. What about an empty string then? Not too sure, but I think it would render an empty element, without any CSS class, as the status name is [used to match the CSS style](https://github.com/cylc/cylc-ui/blob/e68f91b9410b5508c8e0fb6f780a9c765ed4638e/src/components/cylc/Task.vue#L15). WDYT?

oliver-sanders:
> I did think a little about this one... but I couldn't come up with a better alternative.

I think the default state value should be blank (`''`) as `waiting` isn't a safe default value (FYI we actually do this in the current GUIs [the "blank" state shows up in black] though you would only see the effects if something went really wrong).

src/views/Suite.vue
```diff
@@ -31,6 +31,9 @@
                 <i class="mdi mdi-plus" v-else></i>
               </span>
             </template>
+            <template slot="state" slot-scope="props">
+              <task :status="props.row.state" :progress=0 />
```
oliver-sanders: Cool, I diddn't realise it was this easy to propagate properties downstream.

kinow:
Rebased fixing conflicts, and then removed the default `waiting` after I realized it was marked already as `required`, so no need to have a default value :grimacing:

PR #119, opened 2019-07-11
Create view control toolbar

### LINKED ISSUE #114: Create views control toolbar
Our current toolbar is:

image

And contains the following:

- The name of the current route
- A non-functional search input field
- A link to the dashboard
- A link to the user profile
- The i18n flag-button

According to the new layout, it should change to be a sort of Control Toolbar, used when a suite/workflow has been positioned/selected.

image

It must contain the following:

- workflow ID
- control buttons (stop/pause/etc)
- a link to expand and show more control information (can be done later)
- a text message
- a link to add more views (can be done later too)

kinow:
Closes #114 

Adds initial version of our new toolbar. This toolbar component is actually based on Vuetify's [toolbar](https://vuetifyjs.com/en/components/toolbars).

With the difference that it 

- overrides a few CSS styles
- has extra CSS styles
- has extra control over the extended feature

Icons are using material design icons, and typography and layout & spacing are also material's default, hence a few differences if you look closely and compare with @oliver-sanders ' sketches.

### DIFF
--- a/src/components/cylc/Toolbar.vue
+++ b/src/components/cylc/Toolbar.vue
@@ -24,7 +24,7 @@
     </div>
 
     <!-- control bar elements displayed only when a suite has been positioned -->
-    <template v-if="suite">
+    <template v-if="tree && tree.length > 0">
       <a @click="onClickPause">
         <v-icon color="#5E5E5E">mdi-pause</v-icon>
       </a>
@@ -77,17 +77,10 @@
 
   computed: {
     ...mapState('app', ['title']),
-    ...mapState('suites', ['suite'])
-  },
-
-  watch: {
-    '$route' (val) {
-      this.$store.commit('app/setTitle', val.name);
-    }
+    ...mapState('suites', ['tree'])
   },
 
   mounted () {
-    this.$store.commit('app/setTitle', this.$route.name);
     this.onResponsiveInverted()
     window.addEventListener('resize', this.onResponsiveInverted)
   },
--- a/src/components/cylc/Toolbar.vue
+++ b/src/components/cylc/Toolbar.vue
@@ -1,0 +1,157 @@
+<template>
+  <v-toolbar
+    flat
+    dense
+    class="c-toolbar"
+  >
+    <div class="v-toolbar-title">
+      <v-toolbar-title
+        class="tertiary--text font-weight-light"
+      >
+        <!-- burger button for mobile -->
+        <v-btn
+          v-if="responsive"
+          class="default v-btn--simple"
+          dark
+          icon
+          @click.stop="onClickBtn"
+        >
+          <v-icon>mdi-view-list</v-icon>
+        </v-btn>
+        <!-- title -->
+        <span class="c-toolbar-title">{{ title }}</span>
+      </v-toolbar-title>
+    </div>
+
+    <!-- control bar elements displayed only when a suite has been positioned -->
+    <template v-if="suite">
+      <a @click="onClickPause">
+        <v-icon color="#5E5E5E">mdi-pause</v-icon>
+      </a>
+
+      <a @click="onClickStop">
+        <v-icon color="#5E5E5E">mdi-stop</v-icon>
+      </a>
+
+      <a>
+        <v-chip color="#E7E7E7" @click="toggleExtended">Control</v-chip>
+      </a>
+
+      <span>Running, will stop at 30000101T0000 cycle</span>
+
+      <v-spacer />
+
+      <a class="add-view" @click="onClickAddView">
+        Add View <v-icon color="#5995EB">mdi-plus-circle</v-icon>
+      </a>
+    </template>
+
+    <!-- displayed only when extended===true -->
+    <template v-slot:extension v-if="extended">
+      <span style="margin-left: 260px;">
+        <a @click="onClickPause">
+          <v-icon color="#5E5E5E">mdi-pause</v-icon>
+        </a>
+
+        <a @click="onClickStop">
+          <v-icon color="#5E5E5E">mdi-stop</v-icon>
+        </a>
+
+        <span>Other controls added in the future</span>
+      </span>
+    </template>
+
+  </v-toolbar>
+</template>
+
+<script>
+
+  import { mapMutations, mapState } from 'vuex'
+
+  export default {
+  data: () => ({
+    responsive: false,
+    responsiveInput: false,
+    extended: false
+  }),
+
+  computed: {
+    ...mapState('app', ['title']),
+    ...mapState('suites', ['suite'])
+  },
+
+  watch: {
+    '$route' (val) {
+      this.$store.commit('app/setTitle', val.name);
+    }
+  },
+
+  mounted () {
+    this.$store.commit('app/setTitle', this.$route.name);
+    this.onResponsiveInverted()
+    window.addEventListener('resize', this.onResponsiveInverted)
+  },
+  beforeDestroy () {
+    window.removeEventListener('resize', this.onResponsiveInverted)
+  },
+
+  methods: {
+    ...mapMutations('app', ['setDrawer', 'toggleDrawer']),
+    onClickBtn () {
+      this.setDrawer(!this.$store.state.app.drawer)
+    },
+    onResponsiveInverted () {
+      if (window.innerWidth < 991) {
+        this.responsive = true
+        this.responsiveInput = false
+      } else {
+        this.responsive = false
+        this.responsiveInput = true
+      }
+    },
+    onClickPause () {
+      console.log("Pausing workflows has not been implemented yet")
+    },
+    onClickStop () {
+      console.log("Stopping workflows has not been implemented yet")
+    },
+    toggleExtended () {
+      this.extended = !this.extended
+    },
+    onClickAddView () {
+      console.log("Adding views has not been implemented yet")
+    }
+  }
+}
+</script>
+
+<style lang="scss">
+  .v-toolbar {
+    min-height: inherit !important;
+    .v-toolbar__content {
+      min-height: inherit !important;
+    }
+  }
+
+  .c-toolbar {
+    background-color: #E7E7E7;
+    border-bottom: 3px solid #ccc;
+
+    .c-toolbar-title, .add-view {
+      font-size: 22px;
+      color: #5995EB;
+      font-weight: 500;
+    }
+
+    .c-toolbar-title {
+      padding-right: 13px;
+    }
+
+    span {
+      color: #7D878F;
+    }
+
+    .add-view {
+    }
+  }
+</style>
--- a/src/components/cylc/Toolbar.vue
+++ b/src/components/cylc/Toolbar.vue
@@ -120,34 +120,3 @@
   }
 }
 </script>
-
-<style lang="scss">
-  .v-toolbar {
-    min-height: inherit !important;
-    .v-toolbar__content {
-      min-height: inherit !important;
-    }
-  }
-
-  .c-toolbar {
-    background-color: #E7E7E7;
-    border-bottom: 3px solid #ccc;
-
-    .c-toolbar-title, .add-view {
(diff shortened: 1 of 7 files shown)


kinow:
Some features:

- icons and links (e.g. pause) are enabled, but instead of doing anything, they simple print messages to console.
- clicking on Control extends the toolbar
- integration with the application is not done via props or slots, but instead by the Vuex values

Screenshots:

image

image

image

kinow:
One thing that I noticed we haven't discussed in Exeter is what to do with suites/workflows with very long names. In the sketches the name fits perfectly, but for the future I believe we will want to trim the text to a certain length.

Otherwise, the control buttons and other links could be in different location because of the name length. WDYT @oliver-sanders ? (I don't think we need to implement it right away, but would be simply a `computed` property based on the suite name I guess)

matthewrmshin:
In which case, you almost always want the suite name to live on its own line.

matthewrmshin:
(Because there is no good way to chop a long name. Some people arrange information in little endian and some people in big endian, so chopping either end of a long string is not going to work for everyone. (And then there are some people who arrange information in chaos...))

oliver-sanders:
> One thing that I noticed we haven't discussed in Exeter is what to do with suites/workflows with very long names.

That's on my list to draw up, for the control bar we can probably get away with truncating the name.

kinow:
Good points, and I was thinking about trimming and displaying long name in a tooltip.

Then rememered tooltips are no-no in mobile 😑

src/components/cylc/Toolbar.vue
```diff
@@ -0,0 +1,150 @@
...
+
+      <a class="add-view" @click="onClickAddView">
+        Add View <v-icon color="#5995EB">mdi-plus-circle</v-icon>
+      </a>
+    </template>
+
+    <!-- displayed only when extended===true -->
+    <template v-slot:extension v-if="extended">
+      <span style="margin-left: 260px;">
+        <a @click="onClickPause">
```
oliver-sanders: The `control` button is supposed to replace the functionality of the current `control` menu so will have things like `poll` and the full range of `stop` options, etc.

We can add these options in a follow-up.

src/components/cylc/Toolbar.vue
```diff
@@ -0,0 +1,150 @@
...
+      min-height: inherit !important;
+    }
+  }
+
+  .c-toolbar {
+    background-color: #E7E7E7;
+    border-bottom: 3px solid #ccc;
+
+    .c-toolbar-title, .add-view {
+      font-size: 22px !important;
```
oliver-sanders: mmm, why is `!important` needed here?

src/components/cylc/Toolbar.vue
```diff
@@ -0,0 +1,150 @@
...
+      color: #5995EB;
+      font-weight: 500;
+    }
+
+    .c-toolbar-title {
+      padding-right: 13px;
```
oliver-sanders: Set in `em` so that it adjusts automatically with `font-size`?

src/components/cylc/Toolbar.vue
```diff
@@ -0,0 +1,150 @@
...
+    }
+  }
+
+  .c-toolbar {
+    background-color: #E7E7E7;
+    border-bottom: 3px solid #ccc;
+
+    .c-toolbar-title, .add-view {
+      font-size: 22px !important;
+      color: #5995EB;
```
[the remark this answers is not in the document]
kinow: Crossed my mind too. We have global variables from vuetify that I wanted to use. Might be easy, just add the scss somewhere, and import using alias `~` (same as `@` in js, the src folder). But I didn't try it, and instead left it local only.

We can try to add it here or in #9, either way is fine for me

src/components/cylc/Toolbar.vue
kinow: I normally use em... but the vuetify theme is using px (even for font size?).

I believe the left padding was something like 26px.

So I just used half its size for the right padding.

src/components/cylc/Toolbar.vue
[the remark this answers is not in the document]
oliver-sanders: Happy to push back if you like, we could attempt this soon.

src/components/cylc/Toolbar.vue
kinow: I can investigate more tomorrow, but I believe it was due to conflicts with attributes from other classes of `a` set in vuetify theme 😓

src/components/cylc/Toolbar.vue
```diff
@@ -0,0 +1,150 @@
...
+      if (window.innerWidth < 991) {
+        this.responsive = true
+        this.responsiveInput = false
+      } else {
+        this.responsive = false
+        this.responsiveInput = true
+      }
+    },
+    onClickPause () {
+      console.log("Pausing workflows has not been implemented yet")
```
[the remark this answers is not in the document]
kinow: Will do, good idea

src/components/cylc/Toolbar.vue
```diff
@@ -0,0 +1,150 @@
...
+    ...mapState('app', ['title']),
+    ...mapState('suites', ['tree'])
+  },
+
+  mounted () {
+    this.onResponsiveInverted()
+    window.addEventListener('resize', this.onResponsiveInverted)
+  },
+  beforeDestroy () {
+    window.removeEventListener('resize', this.onResponsiveInverted)
```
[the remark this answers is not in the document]
kinow: Sure, but this came from the vuetify original code, not written by us... I assumed they had a reason... but didn't think too much about it

src/components/cylc/Toolbar.vue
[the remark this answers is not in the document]
kinow: I will have to update it tomorrow to address your other comments (thanks for doing it so promptly!), so worth giving it a try too. If it seems too hard will suggest moving to the other issue

src/components/cylc/Toolbar.vue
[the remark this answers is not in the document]
oliver-sanders: If we need to do anything programmatic then JS is the way to go, otherwise CSS FTW!

tldr; I have a bit of trauma from a previous project:

I once built an unconventional responsive site which needed to respond to viewport aspect ratio (before CSS3 was sufficiently universal) and had a hell of a lot of problems with resize event listeners, some issues I had:

* Browsers trigger events differently on maximise.
* Mobile screen rotation.
* Callback delay can cause incorrect measurements especially in click+drag resizing.
* Fullscreen mode.
* Jerky resizing.

src/components/cylc/Toolbar.vue
oliver-sanders: Dammit, is vuetify trying to dictate style of downstream components!?

src/components/cylc/Toolbar.vue
oliver-sanders: > but the vuetify theme is using px (even for font size?).

What! That's horrible, it's hard-coded everywhere? What about accessibility.

matthewrmshin: I think people are now going back from `em` to `px` for the finest of fine control.

oliver-sanders: Which means universally changing font size becomes practically impossible...

src/components/cylc/Toolbar.vue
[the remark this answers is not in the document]
kinow: Done

src/components/cylc/Toolbar.vue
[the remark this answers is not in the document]
kinow: >tldr; I have a bit of trauma from a previous project:

Gee! Definitely better if we can avoid repeating those mistakes. Doesn't seem simple to implement responsiveness in JS.

>If we need to do anything programmatic then JS is the way to go, otherwise CSS FTW!

Agreed. As this is code from Vuetify, I would prefer to tackle this one later. When doing so, might be worth to look at other parts of the code for similar pattern as well.

src/components/cylc/Toolbar.vue
kinow: >Dammit, is vuetify trying to dictate style of downstream components!?

Not quite. I think the Toolbar is a component to be used, and not really extended. While it supports some level of extension via props, slots; I think styling is a bit limited.

>mmm, why is !important needed here?

Turns out this !important is not really necessary :tada:

src/components/cylc/Toolbar.vue
kinow: >I think people are now going back from em to px for the finest of fine control.

Interesting. When I learned about `em` (some years ago) I remember it was considered good practice to use it instead of `px`. But now that you mention it, I've seen other components in Vue.js that use `px` instead.

I think if you use `@media` in CSS, you can have different height/font size/etc... meaning that you don't really need `em`. So maybe that's the case in Vuetify? They actually have an issue that I think is related to this: [[Feature Request] Use rem, em instead of px](https://github.com/vuetifyjs/vuetify/issues/3401)

kinow: (said that, I think using `px` for now should be OK, as the whole app is using `px` in Vuetify, and if we use `em` here later Vuetify could be using `rem` instead... so easier to follow that discussion and use the same as the underlying theme/framework IMHO)

src/components/cylc/Toolbar.vue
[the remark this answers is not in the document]
kinow: @oliver-sanders 

Done. And we are now also using the Vuetify material design colours and styles (but I chose the closest to the ones you used in the sketch).

Which means we don't need to create new variables, but instead re-use from our theme :+1: 

Later we will have to adjust the current colours, as the default colour for some links is green, as well as table headers, and selected menu. Instead we should try to use this same blue, as in your sketches :+1:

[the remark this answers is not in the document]
kinow: Screenshot for reference

image

New scss files under `src/styles/cylc/`. The blue colour used was `$blue-A200`. For background `$grey-200`, border `$grey-300`, and text colour `$blue-grey-400`.

kinow:
Ready for review again :+1:

src/components/cylc/Toolbar.vue
oliver-sanders: OK.

BTW my standard approach is along the lines of:

```css
@media only screen and (max-width: x px) {
    font-size: 10px;
}
@media only screen and (max-width: y px) {
    font-size: 12px;
}
@media only screen and (max-width: z px) {
    font-size: 15px;
}
h1 {
    font-size: 2em;
}
h2 {
    font-size: 1.6em;
}
/* etc ... */
```

src/components/cylc/Toolbar.vue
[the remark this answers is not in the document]
oliver-sanders: Cool.

src/components/cylc/Toolbar.vue
kinow: I like this one too, we can aim at having something similar to this :+1:

wxtim:
## Review
- [x] Had a play. All seems generally ok, although not entirely thrilled about the way the text squashes up on a small screen: 
Screenshot from 2019-07-12 12-18-41
- [ ] Read code changes

kinow:
> Had a play. All seems generally ok, although not entirely thrilled about the way the text squashes up on a small screen:

Oh, good point about resizing it, I should have tested that too.

There is also the question about what to do about suites with long names. We will probably have to move the title to another line (or find another alternative), and then we can also lock the elements sizes to a minimum to avoid issues like this when resizing.

wxtim:
Probably an issue for ano0ther day

src/components/cylc/Toolbar.vue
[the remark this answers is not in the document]
wxtim: I was very amused by teh console log messages.

oliver-sanders:
And chuck responsive adaptation to long suite names into it!

PR #121, opened 2019-07-11
Add missing dot for class name

kinow:
The Task scss file is missing a dot in the class name (the Job one looks OK). Resulting in a much larger than expected icon.


kinow:
(@oliver-sanders my bad for running the app for [this change](https://github.com/cylc/cylc-ui/pull/117/) while reviewing the PR)

PR #123, opened 2019-07-12
GraphQL Proxy

### LINKED ISSUE #115: central data store
We will be providing multiple "views" on the same workflow[s] (tree, graph, dot, ...) and permitting users to open multiple views simultaneously.

It is quite likely that users might even want duplicate views open, for example I think this will be a common use case:

* two tree views
  * one showing everything with families collapsed 
  * one filtering out only failed tasks with families expanded

To deal with this elegantly whilst minimising server load and browser memory usage it makes sense for the client to maintain one data model which drives the various views (i.e. a standard MTV data model).

This would mean:

* One model
* Each view registers subscriptions with that model
* Subscription queries are "flat" (tasks, jobs, families are kept separate rather than nested)
* When subscriptions are added / removed all active subscriptions are merged
* The client subscribes to this merged query
* An interface for performing one-off queries for any data we don't need to keep up to date

Merging queries is fairly straight forward (especially if the queries are straight forward), see this demo:

https://github.com/oliver-sanders/cylc-ui-graphql-proxy

Implications:

* Consistency between the data model used by the different views
  * Views must work with the flat data model rather than requesting data in the shape they want
* Central management of all subscriptions by the client
  * Giving us the potential to do clever things like suspending subscriptions for hidden views (e.g. the gantt view in [this sketch](https://user-images.githubusercontent.com/16705946/60719622-36487980-9f20-11e9-8d90-69301420bd4d.png))
* More filtering would be done client-side, less server-side (is this good or bad?)

Some pros some cons.

oliver-sanders:
closes #115

#### Work In Progress

Putting this up now to prevent disruption and help unify our efforts.

#### The Proxy

Implements a central "proxy" called the workflow service for routing all GraphQL queries though.

* Views register subscriptions
  * Ideally subscriptions should request data in the same "flat" format as eachother, but there is nothing to prevent deeply nested subscriptions.
* The workflow service merges subscriptions into a single query
  * The merge is a union so overlapping queries don't result in duplication
  * This is a naive merge so `workflows('*') and workflows('foo')` will not be merged.
* The workflow service commits results into the central data store
  * At present this is done via polling

#### The Views

As a proof of concept I've quickly slung together:

* GScan (adapted from Bruno's Suites view)
* Tree (a simple nested list)

To avoid treading on anyone's toes I've written new code named "workflow" alongside the "suite" code:

* `SuiteService` -> `WorkflowService`
* `Suites` view -> `Workflows` view
* `Suite` view -> `Tree` view
* `suite.module` data model -> `workflow.module` data model

#### TODO:

* [x] Ofline data.
* [x] Find some way of signalling back to the view that the data is loading.
* [x] Tidy tidy!


src/services/workflow.service.js
```diff
@@ -0,0 +1,176 @@
...
+    this.apolloClient.query({
+      query: this.query,
+      fetchPolicy: 'no-cache'
+    }).then((response) => {
+      const workflows = response.data.workflows;
+      store.dispatch('workflows/set', workflows);
+      return true;
+    }).catch((error) => {
+      const alert = new Alert(error.message, null, 'error');
+      store.dispatch('addAlert', alert);
```
kinow: s/addAlert/setAlert, as we now have a single alert (it should be logged to browser console, and the latest alert will be displayed in the UI)

src/store/workflows.module.js
```diff
@@ -0,0 +1,34 @@
+// raw GraphQL response data structure
+const state = {
+    workflows: {}
```
kinow: Needs to be an array for `v-table`, `workflows: []`.

kinow:
Work in progress, but it is looking great already!

First, to get it to work, had to:

```diff
diff --git a/src/components/cylc/Toolbar.vue b/src/components/cylc/Toolbar.vue
index c3d0984..082a997 100644
--- a/src/components/cylc/Toolbar.vue
+++ b/src/components/cylc/Toolbar.vue
@@ -24,7 +24,7 @@
     </div>
 
     <!-- control bar elements displayed only when a suite has been positioned -->
-    <template v-if="tree && tree.length > 0">
+    <template v-if="(workflows && workflows.length > 0 ) || (tree && tree.length > 0)">
       <a @click="onClickPause">
         <v-icon color="#5E5E5E">mdi-pause</v-icon>
       </a>
@@ -77,7 +77,8 @@
 
   computed: {
     ...mapState('app', ['title']),
-    ...mapState('suites', ['tree'])
+    ...mapState('suites', ['tree']),
+    ...mapState('workflows', ['workflows'])
   },
 
   mounted () {
diff --git a/src/services/workflow.service.js b/src/services/workflow.service.js
index f820fba..f07c5f4 100644
--- a/src/services/workflow.service.js
+++ b/src/services/workflow.service.js
@@ -143,7 +143,7 @@ class WorkflowService {
       return true;
     }).catch((error) => {
       const alert = new Alert(error.message, null, 'error');
-      store.dispatch('addAlert', alert);
+      store.dispatch('setAlert', alert);
       return false;
     });
   }
diff --git a/src/store/workflows.module.js b/src/store/workflows.module.js
index 57f84ad..8cf9968 100644
--- a/src/store/workflows.module.js
+++ b/src/store/workflows.module.js
@@ -1,6 +1,6 @@
 // raw GraphQL response data structure
 const state = {
-    workflows: {}
+    workflows: []
 };
 
 const mutations = {
diff --git a/src/views/GScan.vue b/src/views/GScan.vue
index 5dfd4c5..b1f16b2 100644
--- a/src/views/GScan.vue
+++ b/src/views/GScan.vue
@@ -94,6 +94,9 @@
     data: () => ({
       viewId: 'GScan: ' + Math.random(),
       subscriptions: {},
+      pagination: {
+             rowsPerPage: 10
+           },
       headers: [
         {
           sortable: true,
@@ -123,7 +126,8 @@
       ],
     }),
     computed: {
-      ...mapState('workflows', ['workflows'])
+      ...mapState('workflows', ['workflows']),
+      ...mapState('app', ['isLoading'])
     },
     created() {
       workflowService.register(this);
```

Current list of workflows at `/#/workflows`:

image

And the Tree

image


:clap:

oliver-sanders:
#### Implemented off-line data

For the WorkflowService I've written a script which runs a Cylc suite to generate dummy data.

If we change the GraphQL structure or start using new data in the UI we just update the query and re-run the suite.

kinow:
>For the WorkflowService I've written a script which runs a Cylc suite to generate dummy data.

Nice!

The other offline data were generated with https://github.com/kinow/cylc-cassettes/, but I liked the approach of capturing the response from within Cylc. This way you don't have to time the execution of the suite and of the script that captures as I am doing.

Perhaps later we should consider moving everything under the same structure (either here or another repo) for consistency. :+1:

src/services/mock/checkpoint.js
```diff
@@ -0,0 +1,154 @@
...
+                    "id": "cylc/one/20000102T0000Z/eventually_succeeded",
+                    "state": "succeeded",
+                    "cyclePoint": "20000102T0000Z",
+                    "task": {
+                        "name": "eventually_succeeded"
+                    },
+                    "jobs": [
+                        {
+                            "id": "cylc/one/20000102T0000Z/eventually_succeeded/01",
+                            "state": "running",
```
oliver-sanders: FYI: Something funny is going on at the Cylc Flow end with failed jobs reporting back as running.

src/services/mock/checkpoint/get_checkpoint.py
```diff
@@ -0,0 +1,32 @@
...
+            }
+            jobs {
+                id
+                state
+                submitNum
+            }
+        }
+    }
+}
+'''
```
oliver-sanders: We could generate this from queries registered on the views to save having to make changes in two places but thats a reasonably involved piece of work. Can raise as follow-on if others think it's sensible.

src/services/mock/generate
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+
+CHECK_FILE='checkpoint.js'
+
+
+cylc reg "one" one
+
+
+if cylc run "one" --fcp=20000102T0000Z --no-detach; then
```
oliver-sanders: Just a simple static suite for now, we could add a simple cycling suite later.

src/views/Tree.vue
```diff
@@ -0,0 +1,159 @@
+<template>
```
oliver-sanders: Just a really simple tree view for now, I needed a second view to test query merging, view registering / unregistering etc. #78 can rip this template out.

src/views/Tree.vue
```diff
@@ -0,0 +1,159 @@
...
+          {{ workflow.name }}
+        </span>
+        <ul>
+          <li v-for="cycle in cycles" :key="cycle">
+            <!-- Cycle -->
+            <span>
+              {{ cycle }}
+            </span>
+            <ul v-for="proxy in workflow.taskProxies" :key="proxy.id">
+              <li v-if="proxy.cyclePoint === cycle">
```
oliver-sanders: Will need to consider the efficiency of this approach, not sure how to test. If Vue does it's job right I think there may be a slight performance hit on initial-render but each following addition should be quick.

oliver-sanders:
@kinow I think this is now ready for review.

I guess the next step would be to merge across anything left in the `SuiteService` tree?

oliver-sanders:
Wow, that's an extremely opinionated linter.

kinow:
>Is two space indentation really the JS standard.

That's from standard JS... Vue, Angular, React, Apollo, etc, all use 2 spaces too. And most of our code base in cylc-ui is using 2 spaces already. But the linter wasn't checking for unnecessary `;`, indentation, etc.

If you run `npm run lint`, it will try to correct your code (though first you will need to rebase). Otherwise, `npm run lint -- --no-fix` will only show you what issues it has found.

src/services/mock/checkpoint/get_checkpoint.py
kinow: It would be better to keep the Python source somewhere else. Either a new repository, or in a directory that is easy to be excluded when producing the binary to be deployed in environments.

src/views/Tree.vue
kinow: Excellent! #78 just got much easy to implement. Thanks!

src/services/gquery.js
```diff
@@ -0,0 +1,246 @@
...
+        });
+    }
+
+    callbackActive() {
+        /**
+         * Run callback functions registered by views.
+         */
+        // activeCallback - called when all subscriptions are active
+        this.views
+            .filter(v => v.activeCallback)
```
kinow: We could have a `View` (or a subtype) that has the callback when active... or maybe use one of the component lifecycle methods (beforeCreate, beforeDestroy, etc)...

kinow:
Checked out code locally, it appeared to work correctly when running a suite... but it has a delay to display. Maybe we need to invoke the method to fetch as soon as the page is loaded, and then create the polling.

I tried to ignore linter errors, but some of the comments I've added are cosmetic like `const`, `===`... etc. So feel free to ignore them if you prefer (we can work on these later, when writing tests, evolving the component, etc).

src/views/Tree.vue
```diff
@@ -0,0 +1,159 @@
...
+      workflowId: '',
+      subscriptions: {},
+      isLoading: true
+    }),
+
+    computed: {
+      ...mapState('workflows', ['workflows']),
+      cycles: function() {
+        var cycles = new Set();
+        for (let workflow of this.workflows) {
```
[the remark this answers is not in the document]
oliver-sanders: Looks like `forEach` and `every` are used pretty heavily these days too.

src/router/paths.js
```diff
@@ -36,9 +36,23 @@ export default [
...
+    path: '/workflows/:name',
+    view: 'Tree',
+    meta: {
+      'layout': 'default'
+    }
+  },
   {
     path: '/suites/:name',
-    view: 'Suite',
+    view: 'suite',
```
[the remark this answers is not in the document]
oliver-sanders: Think this was accidental!

src/services/mock/checkpoint/get_checkpoint.py
oliver-sanders: I though we were only going to distribute the `dist` dir so this wouldn't matter. Otherwise I could migrate the logic to JS?

src/services/workflow.service.mock.js
```diff
@@ -0,0 +1,35 @@
+import { checkpoint } from '@/services/mock/checkpoint.js'
+import { GQuery } from '@/services/gquery'
+import store from '@/store/'
+
+
+class MockWorkflowService extends GQuery {
+    /**
+     * Standin WorkflowService for off-line work.
```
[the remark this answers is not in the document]
oliver-sanders: Stand-In

oliver-sanders:
Thanks for your detailed review, I'm not quite solid with ES6 yet so really good to get feedback.

> feel free to ignore them if you prefer

Nah, let's get it right!

src/services/gquery.js
```diff
@@ -0,0 +1,246 @@
...
+                    view.activeCallback(true);
+                }
+            });
+    }
+
+    print () {
+        /**
+         * Dump a human-readable representation of all subscriptions.
+         */
+        var ret = 'Combined Query:';
```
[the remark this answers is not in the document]
oliver-sanders: I think this is a `var`?

src/services/gquery.js
```diff
@@ -0,0 +1,246 @@
...
+    recompute () {
+        /**
+         * Compute the combined query from all registered subscriptions.
+         */
+        if (this.subscriptions.length < 1) {
+            this.query = null;
+            return;
+        }
+        // merge together active subscriptions
+        var root = gClone(this.subscriptions[0].query);
```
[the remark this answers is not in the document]
oliver-sanders: This is where my knowledge of JS fails me. The `merge` function on the following line modifies the `root` object. So is the `Object` `root` a constant with variable contents or is `root` a variable object?

src/services/gquery.js
oliver-sanders: There may be a few callbacks, we will definitely need one for sharing selections (e.g. double click on a task in the graph view to get the tree view to select it too).

src/services/workflow.service.mock.js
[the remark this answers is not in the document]
kinow: Ah! Makes sense, thanks! :smile:

oliver-sanders:
Need a second reviewer, rather than cruelly nominating someone: @cylc/core, any volunteers?

sadielbartholomew:
> any volunteers

I'll be having a look anyway as I want to get the data provision to my Dot View PR from this, so I'm happy to review :+1:, & will get a review in soon, but that's not to stop anyone else from reviewing too.

src/services/gquery.js
```diff
@@ -0,0 +1,238 @@
...
+}
+
+function getSelections (a) {
+  /**
+   * Return map of selections present on a node.
+   * @param {Object} a - Node of GraphQL query.
+   * @return {Object} All selections present on this node.
+   */
+  var selections = {}
+  if (!a.selectionSet) {
```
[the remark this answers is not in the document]
oliver-sanders: The reason I diddn't do this initially is I wasn't quite sure whether this was reliable behaviour, some fun digging through a rather old ECMA spec later, it is!

http://www.ecma-international.org/ecma-262/5.1/#sec-11.11

src/services/gquery.js
```diff
@@ -0,0 +1,238 @@
...
+      return
+    }
+    // merge together active subscriptions
+    var root = gClone(this.subscriptions[0].query)
+    for (const subscription of this.subscriptions.slice(1)) {
+      merge(root, subscription.query)
+    }
+    this.query = root
+
+    // TODO - remove
```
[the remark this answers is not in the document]
oliver-sanders: Might leave this in for now for debugging and to help the people writing views as it shows the combined query. Could potentially get it to only display based on some node env.

src/views/Tree.vue
```diff
@@ -0,0 +1,158 @@
...
+    isLoading: true
+  }),
+
+  computed: {
+    ...mapState('workflows', ['workflows']),
+    cycles: function () {
+      var cycles = new Set()
+      for (const workflow of this.workflows) {
+        if (workflow.name === this.workflowId) {
+          for (const proxy of workflow.taskProxies) {
```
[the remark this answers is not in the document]
oliver-sanders: Good spot. My guess would be that the top level `workflows` data is already in the store from the `GScan` page but `workflows.taskProxies` hasn't been loaded yet. If my diagnosis is correct we will run into this one again.

For now I think an easy solution is just to wrap the logic with a `v-if="workflow.taskProxies"`, Vue can update the rest when the data becomes available.

Later we could use `v-if="isActive"` with a `v-else` displaying a loading spinner or something of the form.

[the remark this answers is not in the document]
sadielbartholomew: Yes, great, this solves it as I not longer see the initial error. And in a logical & painless way too, which is a bonus.

oliver-sanders:
:fireworks:

PR #125, opened 2019-07-15
Use standard JS and apply rules to project with eslint

### LINKED ISSUE #124: Apply standard JavaScript style guide to the project
The project was created by [`vue-cli`](https://cli.vuejs.org/) and the default values for the linter are not enforcing things like spaces. At the moment most of the project uses 2 spaces to indent (same as Vuejs, Reactjs, Angular, axios, apollo-client, etc.

[standard JS](https://standardjs.com/) provides a series of pre-defined rules for JavaScript code, which includes indent.

kinow:
Closes #124 

Simply adds standard JS coding style development dependency to the project, and fixes the issues found by the linter.

Also changes Travis to run `npm run lint -- --no-fix`, as otherwise the command automatically fixes the code (similarly to `black` for Python) and exists with `0`.


src/components/cylc/Task.vue
```diff
@@ -160,28 +159,27 @@
...
+      type: Number,
+      required: false,
+      default: 0
+    }
+  },
+  computed: {
+    progressStyle: function () {
+      return {
+        'stroke-dashoffset':
+                        `calc(157 - (157 * ${this.progress / 100} + ))`
```
matthewrmshin: Is this supposed to have so much indentation? (Or just bad Github diff?)

src/views/Graph.vue
```diff
@@ -89,17 +89,19 @@ const config = {
...
+          }
           // console.log('ICON PATH --> ', path)
           return path
         },
         'background-fit': 'contain contain',
-        'background-image-opacity': function(node) {
+        'background-image-opacity': function (node) {
           let opacity
-          node.data('running') > 0 ? opacity =  1 : opacity = .6
+          node.data('running') > 0 ? opacity = 1 : opacity = 0.6
```
matthewrmshin: Can we just do this?

```suggestion
          return node.data('running') > 0 ? 1.0 : 0.6
```

kinow: Oh, I think we can adjust this one too :+1:

src/views/Graph.vue
```diff
@@ -101,7 +101,7 @@ const config = {
         'background-fit': 'contain contain',
         'background-image-opacity': function (node) {
           let opacity
-          node.data('running') > 0 ? opacity = 1 : opacity = 0.6
+          return node.data('running') > 0 ? 1.0 : 0.6
           return opacity
```
matthewrmshin: Sorry I forgot to say earlier, but in which case, you can also remove the lines before and after.

kinow: Oh, sure. I was just testing the GitHub "Commit this diff" feature. Quite cool :) Will edit the previous commit.

src/components/cylc/Task.vue
kinow: This change was done automatically by `npm run lint`. Interesting that it seems to try to keep the line under 120 characters I think (there is no limit in the linter I believe). Fixed.

kinow:
Thanks for the review @matthewrmshin ! Feedback addressed.

hjoliver:
I must say `get ()` instead of `get()` strikes me as weird, but if that's considered standard in JS, so be it!

kinow:
>I must say get () instead of get() strikes me as weird

Initially it was weird to me too, in Java we normally don't use space before parentheses. But in JS it is common to have spaces before function definition, and no spacing when calling the function.

kinow:
>With Matt's review this might be enough to merge at the this early stage of development?

I think so. And we can tweak these settings later if necessary, and the linter can act like Python's `black`, re-writing the code to fix whatever is defined in rules... so updating the code after rules were changed later is easy/cheap.

PR #128, opened 2019-07-15
dot/LED view, w/ dismissable pulse on state change

### LINKED ISSUE #79: Create Suite Dot/LED View
Part of #75 , we need to implement a view similar to Cylc 7's Dot (aka LED) view.

58298409-769bde80-7e2f-11e9-8b36-20c362f3f2db

In the screenshot above, the dot view is the right part.

### LINKED ISSUE #127: A USP for Dot View: task state transition monitoring
Some [basic user research has indicated](https://cylc.github.io/cylc-admin/research/MO-gcylc-rc-report.html#results) (along with awareness from user support) that Dot View is seldom-used. This is not too surprising, since it shows a sub-set of the information provided by the Tree View, just rearranged to have tasks (as rows) organised by their cycle point (as columns). Thinking about the two other views & what they uniquely convey:

* Graph View -> task/family dependencies;
* Tree View ->  full information about each task;

Given that Dot View just shows the task state, & the crucial aspect to that is transitions between states (especially from "inactive" to "active" categories of state, or to task failures), I propose we **rebrand it as a light-weight view designed to allow the user to keep track of any changes in task state so they can easily monitor the *progression*, i.e. *dynamics*, of their workflow** (disregarding even dependencies).

I further suggest we **do so via means of a minimal animation that fires any time a task state changes, & that remains until the user dismisses it**. That way, a user is notified whenever a task transitions, but in a way that is flexible, such that any time they are not monitoring the UI (e.g. away from their desk, working in a different tab, etc.), they can return to it & know what transitions have occurred.

I am about to open a PR where I have implemented this as a Proof of Concept, using a nice pulsation animation, so you can try it out live on the current UI, to demonstrate that the notification animation can be done in a subtle but detectable (& useful) way.

My vision is that ultimately we can support configuration to enable & disable the state-transition notification animation for:

* the entire view i.e. all tasks;
* on a per-task or per-family basis (users could set it only for the tasks/families they particularly care about the state of).

Some related questions:

* Tasks may transition through a number of states, e.g. waiting -> ready -> submitted -> running, & we may want to skip certain tasks transition flows such as these if they happen quickly, but not more quickly than the UI detects them & raises the transition alert (e.g. 'ready' doesn't seem relevant to alert on in that example).
* Can we, & should we, incorporate *task-job separation* into this view in some way? It seems like a good idea, since the other views will make that distinction.

sadielbartholomew:
**WIP** (awaiting agreed cross-view data provision approach, see :star:). Close #79 & provide a POC example for #127.

**Cylc 8 UI live implementation of the old Dot/LED ``gcylc`` view, with some (proposed) changes**:

* **actual dots** (circles, not squares)!
* (see #127) a new feature where the **dot icons undergo a pulse animation (like an LED!) when the task's state changes, until the user clicks to dismiss it, or refreshes the whole page to dismiss all pulsating tasks.**

**************

#### Tips for those (eventually) reviewing (or just trying this branch out):

* You can use ``cylc reset`` to manually reset some task states to test the pulse animation.
* For this branch (& PR) I've re-based to ``master`` & systemised my commits, but I developed this all on [another branch](https://github.com/sadielbartholomew/cylc-ui/tree/dot-view), should you be confused as to timestamps implying this took just a few hours!

**************

### This PR will (with progress indicated):

- [x] Add a new URL path for each of a user's suites at ``.../suites/dot/<suite name>`` (renaming the Tree View path correspondingly to ``.../suites/tree/<suite name>``), with both accessible, for now, via separate icons in individual table columns from the Suites Listing page (``.../suites``).
- [x] Incorporate text for internationalisation on the Dot & Tree View tables' headings in both currently-supported languages.
  * *Note*: I've simply Google-translated the headings text from English to Portugese for this, @kinow please let me know how bad it did; I definitely anticipate changes will be needed to this "Portugese"!
- [x] Populate each suite's ``/dot/`` page with a table with information for all tasks for a suite.
- [ ] Recast the table, making it dynamically-sized with cycle points as columns rather than them being the top-level of the nested structure of task data.
  * :star:: **I am waiting for a resolution (#123?) to #115 before attempting this**, because I have struggled to get this aspect working with the current *nested* model for subscription queries (which also goes through a transformer adding a further layer of manipulation); a *flat* model for them  would make this much simpler to implement, & the implementation much cleaner!
- [x] Replace the text-based report of the task state with icons of circles with appropriately-mapped colours taken from the [``status-indicator`` Vue component](https://github.com/coderdiaz/vue-status-indicator) which has been incorporated into the repository.
- [x] Customise those icons so they are larger than the default provided by the component, to better sit with the table & page dimensions.
- [x] Apply the pulse animation: the dot icons (as above) will pulse any time a task changes state & do so until they are clicked.
- [ ] Revise (after discussion) the [colour mappings given to each task state](https://github.com/sadielbartholomew/cylc-ui/blob/bdb5d621da96f4528ee58019066ccd9c9825e29e/src/components/cylc/Dot.vue#L38-L52), where we may want to create further colours on dots, as there are only five currently.
* ~~Create tests (or do so as a follow-on PR)?~~ [agreed that in context this should be future work]
- [ ] Concluding self-review (ensure tests pass).


kinow:
>actual dots (circles, not squares)!

:smile: that makes sense I guess, for the dot view.

> Create tests (or do so as a follow-on PR)?

For now I think the `Graph` component from Martin is the only one that included tests. I think we can move the tests to a follow-on, as we are all more or less changing the whole code and a lot will change (especially regarding the central store).

I have a post-it to start investigating how to write simple unit tests for the small components (Task, Job, new Header). For views it will be a little harder I think.

src/lang/pt-BR/DotView.json
```diff
@@ -0,0 +1,4 @@
+{
+  "tableHeader": "Ponto Vista",
+  "tableSubHeader": "O status das tarefas de um conjunto conforme organizado por ponto de ciclo"
```
kinow: Ponto Vista sounds like Point of View in Portuguese (ponto de vista). But "vista" doesn't really mean the same as in English like a display or visualization... I guess we would need to go with visualization here.

Maybe "Visualização por Pontos" for the `tableHeader` and for `tableSubHeader` "Os status das tarefas de uma suite organizados por pontos do ciclo."

src/lang/pt-BR/Suites.json
```diff
@@ -1,6 +1,6 @@
 {
   "tableHeader": "Lista de Suites",
-  "tableSubHeader": "Esta é a lista de suites que este usuário possui permissoões para visualizar",
+  "tableSubHeader": "Esta é a lista de suites que tenho permissoões para visualizar",
```
kinow: s/tenho/possuo (can't explain why... in this case the verb possuir/possess I guess sounds better)... but if you could fix my old typo as well, please haha. s/permissoões/permissões (extra o)

src/lang/pt-BR/TreeView.json
```diff
@@ -0,0 +1,4 @@
+{
+  "tableHeader": "Vista de árvore",
+  "tableSubHeader": "Dados personalizáveis ​​baseados em colunas para todas as tarefas desta suíte"
```
kinow: Grammatically correct, but sounds like "The view from a tree". I think for `tableHeader` "Visualização em Árvore". And for the `tableSubHeader` maybe "Dados em formato colunar e customizáveis disponíveis para as tarefas desta suite." ("suíte" is the correct translation, but we still use "suite" as in "suite de testes" / "test suite" and "suíte" more for a hotel room name... can't explain why :confused: )

kinow:
I really liked this status indicator! Great idea @sadielbartholomew ! Here's what it looks like after applying a few changes mentioned above, then running `NODE_ENV=offline npm run serve`, and visiting http://localhost:8080.

image

Or what it looks after merging locally the #129 changes:

image

And because it looks much nicer with the animation:

GIFrecord_2019-07-15_204305

Bravissimo Sadie!

src/components/cylc/Dot.vue
```diff
@@ -0,0 +1,88 @@
...
+
+    v-on:click="dismissPulse()"
+    :pulse="isPulseOn"
+  >
+  </status-indicator>
+</template>
+
+<script>
+  // This script customises the default 'dot' icons and animations on them:
+  import 'status-indicator/styles.css'
```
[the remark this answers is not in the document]
sadielbartholomew: Ah, great spot, thank you! For context, there are actually two separate components ``status-indicator`` & ``vue-status-indicator`` (where the [creator specifies on the ``README.md``](https://github.com/coderdiaz/vue-status-indicator?ref=madewithvuejs.com) for the latter 'This is a fork of status-indicator with a few changes for use with Vue.js'), & [on my development branch by mistake I installed both instead of just the Vue version](https://github.com/cylc/cylc-ui/compare/master...sadielbartholomew:dot-view). And when I tidied up for this branch, I removed the non-Vue variation as I got build errors for it without it in my ``package-lock.json``, without realising, as it turns out from your observation, I was importing the wrong one. Then I was very confused as to why Travis threw an error in the same way on the ``status-indicator`` aspect, which I thought had gone completely from the code!

I'll get ``vue-`` prepended in my next commit.

src/components/cylc/Dot.vue
```diff
@@ -0,0 +1,88 @@
...
+      },
+      isIntermediaryMappedState() {
+        return this.testInState(this.dotClass, 'intermediary');
+      },
+      testInState(stateString, requiredState) {
+        // For safety with 'undefined' values, check variety of negation cases:
+        if (stateString == null ||
+            this.stateToDotClassMappings[stateString] == null ||
+            this.stateToDotClassMappings[stateString] != requiredState) {
+          return false;
```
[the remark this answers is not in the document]
sadielbartholomew: > In JS it is safer to use === and !==

Sure, I remember reading about that. But in this case I have just checked & ``null === undefined`` evaluates to ``false``, so I think I might need to change ``null`` -> ``undefined``. I'll take a closer look. I want to make sure all error cases are covered. I guess I am adjusting from my native language, Python, as there isn't a ``KeyError`` equivalent raised in JS for an invalid dict key.

> I think this function can be rewritten

Absolutely, good point (confused as to how I missed that, as it is pure Boolean logic, but I didn't really do a self-review for consolidation etc. at this stage, & think I was just eager to get everything working).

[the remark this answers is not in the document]
kinow: I would probably write it like you did too. My IDE warned me about the === and also suggest the rewrite 😁

src/lang/pt-BR/Suites.json
sadielbartholomew: Oh, BTW, just to clarify in case there is the potential for misinterpretation, I was not changing your own translation here (I do not know Brazilian Portuguese of course so to change your translation would be ridiculous), it is that I changed the text slightly in so that for the ``en-GB`` it reads 'I have' rather than 'the current user', which is more appropriate I think (probably the previous was for information during development). So I tried to change the ``pt-BR`` equivalent using Google translate, awaiting some help from a native speaker :smile:.

Thanks for help with the translations here, I'll add them in a new commit soon. Let me know if I missed anything once that goes up.

src/lang/pt-BR/DotView.json
sadielbartholomew: Excellent, thank you. The subtleties of translation 😁 Google translate has a tough job, for sure!

src/components/cylc/Dot.vue
```diff
@@ -0,0 +1,88 @@
...
+  <status-indicator
+    :active="isActiveMappedState()"
+    :positive="isPositiveMappedState()"
+    :negative="isNegativeMappedState()"
+    :intermediary="isIntermediaryMappedState()"
+
+    v-on:click="dismissPulse()"
+    :pulse="isPulseOn"
+  >
+  </status-indicator>
```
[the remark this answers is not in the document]
sadielbartholomew: I see, thank you for the advice. I'll add this in for the next commit.

src/components/cylc/Dot.vue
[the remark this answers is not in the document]
sadielbartholomew: Good old emacs doesn't have that sort of feature. Maybe it is time for me to transition to using Atom :thinking:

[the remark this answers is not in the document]
sadielbartholomew: Actually, for your suggestion on strict (in)equality being safer than abstract (in)equality for comparison, I have just amended the final ``!=``, because (I forgot I already checked this!) in the first pair of cases, comparing in the abstract to ``null`` catches both the ``null`` & ``undefined``, whereas comparing strictly will [only catch one](https://www.w3schools.com/js/tryit.asp?filename=tryjs_datatypes_undefined_3).

sadielbartholomew:
Tests now passing (after importing ``vue-status-indicator`` rather than ``status-indicator``, as per https://github.com/cylc/cylc-ui/pull/128#discussion_r303731605).

I have a few non-development tasks to address urgently, so won't get back onto this PR until tomorrow. But the next step is to negotiate with everyone how we plan to go ahead with the data provision so I can access the data to formulate it in the Dot View arrangement with cycle point columns.

src/lang/pt-BR/TreeView.json
hjoliver: "The view from a tree" - he heh, I like that.  I love being up a tree, looking out at the landscape. :grin:

hjoliver:
Nice, great progress.

hjoliver:
@kinow - how did you record that gif, BTW?

kinow:
>@kinow - how did you record that gif, BTW?

With `byzanz`. I think I used [this post](https://www.maketecheasier.com/record-screen-as-animated-gif-ubuntu/) to install it. I can't remember if I modified anything further, but I also have the script I run [here](https://github.com/kinow/dork-scripts/blob/55ee3e7926a0205e898f21b3b39f12a4d000b946/gifs/byzanz-gui.sh)

src/components/cylc/Dot.vue
```diff
@@ -0,0 +1,88 @@
...
+@import '~@/styles/status-indicator/resize.css';
+</style>
+
+<template>
+  <!--
+  The first group of classes below set the dot colour based on the task state.
+  The second group start, upon a state (hence colour) change for that task, and
+  stop, upon a user click onto that dot, the pulse animation.
+  -->
+  <status-indicator
```
oliver-sanders: Should use the `<task>` component here?

sadielbartholomew: I went for different components, at least for now, since:

1. (Predominantly) we haven't decided how to incorporate task-job separation into this view. I don't think the approach applied for the Tree View & Graph View mock-ups, with physically-separated icons for the jobs & tasks, is a good approach here, as since state icons are the only information in this view & they are all condensed in a close-packed arrangement, it would become a cluttered mess of icons. I plan to draw up some ideas on the Design Issue for how to indicate both task & job status here.
2. (Also) I wanted to get a POC for state change animations, & we do not have any animated components ready yet (or indeed a decision on how they would look when animated).

oliver-sanders: The dot view is a little awkward here I agree but having separated the task an job icons we really don't want to go back to merging them. I guess would expect the dot view to just show task state with job state available on request (e.g. context menu)?

sadielbartholomew: > but having separated the task an job icons we really don't want to go back to merging them

Agreed, I wasn't suggesting that. We should be definitely be consistent with the segregation. I was thinking either along the line you suggest, where the table could be toggled to display either all tasks *or* all jobs (collecting multiple jobs for one task entry together in some condensed way, like in a self-contained grid rather than linearly in a row, for clarity), or having a new design component that indicates both task & job status in a discrete way (all within one shape). I will try to get up some drawings soon.

sadielbartholomew:
Hi all, I will close this PR since the code is probably too out of date now, such that it will be easier for me to start from scratch, but I would still like to implement the Dot View eventually (I think the natural time to add it would be when @kinow's Tree Component is ready), where I will use the experience in Vue & the ``cylc-ui`` codebase I gained here to guide me in the new attempt. We can also view this closed PR as a POC for task state transition animations.

sadielbartholomew:
Hi all, I will close this PR since the code is probably too out of date now, such that it will be easier for me to start from scratch, but I would still like to implement the Dot View eventually (I think the natural time to add it would be when @kinow's Tree Component is ready), where I will use the experience in Vue & the ``cylc-ui`` codebase I gained here to guide me in the new attempt.

We can also view this closed PR as a POC for task state transition animations.

Issue #109, opened 2019-07-08
follow up 101 (task icons)


hjoliver:
From bottom of description in #101 

(We should either replace this issue with multiple separate issues, or turn the following into a 
check-point list and address with multiple PRs)

TODO (future work)
- [x] Incorporate into views.
   - Work out how we are going to feed data into these components (e.g. props or whatever).
- [ ] Animate state changes (see #127)
   -  Pure CSS probably wont be enough.
   - [edit] Actually CSS is probably the best bet, we are now using CSS animations to set the clock-face position.
   - More below...
- [x] Add the more eccentric task state~~s~~ (expired, ~~queued, retrying~~) and the held badge.
   - Expired and queued are easy, the rest will require a dig into the Cylc Flow code to see what separation is possible with the current internals of Cylc.
- [x] (#134) Extract the <style> from the Job component to make it theme-able.

kinow:
I think I can raise a PR for the first item about props.

kinow:
>I think I can raise a PR for the first item about props.

Done: https://github.com/cylc/cylc-ui/pull/110

oliver-sanders:
Support for held and retrying states is blocked by:

* https://github.com/cylc/cylc-flow/issues/3223

hjoliver:
Suggested icon design for `expired` task state (carried forward from #201):
![image](https://user-images.githubusercontent.com/2362137/68527136-988b0e00-0348-11ea-9f51-1d25fb6ad2db.png)

oliver-sanders:
The last thing on this issue is animation of state changes.

It might be useful to highlight state changes to make them more obvious to the user. IIn particular changes to the `failed` and `submit-failed` states.

Any animations would need to be minimal and subtle to avoid becoming more hindrance than help.

We are using CSS animations to set the clock-face position as the task is running. We have had to limit the number of frames for performance reasons so animations should be short lived and easy to compute too.

Suggest a quick, simple size or colour change.

oliver-sanders:
The last point is covered by #127 so closing this umbrella issue.

Issue #9, opened 2019-02-01
Themes


kinow:
For accessibility, trying to cover everything in one interface might be too hard. There are certain visual elements such as buttons and icons provided by Vuetify components that do not pass the accessibility tests for WAVE tool (which appears to focus more on screen readers).

For colour blindness, there are over three types of colour blindness, and they depend on certain rules for contrast and colors.

Using themes would give us options to have things like a high contrast theme, a theme compatible for protanomaly, another one for protanopia, perhaps even one with higher font size (or that allows the user to alter the font size and store in a cookie), etc.

The Vuetify template used in the current PoC is the [Vuetify Material Dashboard, by Tim Creative](https://www.creative-tim.com/product/vuetify-material-dashboard).

It provides several options and libraries out of the box. But I don't believe it includes themes. It is possible to have themes within a template like this, or the theme coming from a lower level library, that controls which template to use.

Not sure which approach is the best/easier to implement. So will have to start researching in the Vue communities. Will open an issue in the vuetify-material-dashboard GitHub repository too, to ask them about it.

I suspect this is yet another one of the issues that if you leave to implement later, it might be much too late and hard to implement. Causing issues with regressions, code refactoring, etc. So better to start soon-ish if possible.

kinow:
Done https://github.com/creativetimofficial/vuetify-material-dashboard/issues/12

kinow:
And here's a tool to verify a site against some types of colour blindness: https://www.toptal.com/designers/colorfilter/

kinow:
See comment in the linked ticket, quoting:

>It is possible to do that but you must have another scss for this who changes what you need and replace the existent scss.

Good idea to investigate. WebPack is already dealing with scss. So in theory we can have as many as we want. The tricky part is finding a good way to switch between the generated CSS, and also instructing webpack to not combine all files, but instead create separate final CSS style sheets.

kinow:
Dang it, Vuetify also contains docs about it https://vuetifyjs.com/en/framework/theme

Vuetify comes with i18n, validation, lots of other useful code, plus something for themes! hooray :tada:

kinow:
A tool that could help to create and test several palettes https://www.metoffice.gov.uk/hadobs/climviz/climpal/

oliver-sanders:
I think we may end up with two types of theme one determining the appearence of the GUI in general and the other just affecting the colour of the job state icons (in the future potentially the colour of the task state icons as well):

[...]

PR #134, opened 2019-07-17
theme: job icons

oliver-sanders:
Address theme component of #109.

#### Description

Add theme support for Job icons:

* Normal
* Colour-blind
* Greyscale

These respond to the `job_theme--<name>` class of a containing element intended to match the application `theme--<name>` class.

At present we don't have a way to toggle themes or for storing user preferences.

#### Example application to demmonstrate:

i.e. `App.vue`.

```vue
<template>
  <v-app>
    <div v-for="theme in themes" :class="['job_theme--' + theme]">
      <h2>{{theme}}</h2>
      <ul v-for="state in states">
        <li>
          <task :status="state" progress="0" />
          foo
          <job :status="state" />
        </li>
      </ul>
    </div>
  </v-app>
</template>

<script>
const defaultLayout = 'empty'

  import Job from '@/components/cylc/Job'

export default {
  computed: {
    layout () {
      return (this.$route.meta.layout || defaultLayout) + '-layout'
    }
  },
  components: {
    task: Task,
    job: Job
  },
  data: () => ({
    states: [
      'waiting',
      'submitted',
      'running',
      'succeeded',
      'failed',
      'submit-failed'
    ],
    themes: [
      'normal',
      'greyscale',
      'colour_blind'
    ]
  })
}
</script>

<style lang="scss">
  @import '~@/styles/index.scss';

  /* Remove in 1.2 */
  .v-datatable thead th.column.sortable i {
    vertical-align: unset;
  }

  div {
    margin-left: 20px;
    float: left;
  }
</style>
```

#### Example Output

Screenshot from 2019-07-17 16-09-55

### DIFF
--- a/src/styles/cylc/_job.scss
+++ b/src/styles/cylc/_job.scss
@@ -1,127 +1,112 @@
-.job_theme--normal {
+/* css namespace for job element */
+$cjob: ".c-job svg.job rect";
 
+
+#{$cjob} {
+    /* if no job status display nothing */
+    fill: transparent;
+    stroke: transparent;
+}
+
+
+.job_theme--normal {
     $teal: rgb(125,207,212);
     $blue: rgb(106,164,241);
     $green: rgb(81,175,81);
     $red: rgb(207,72,72);
     $pink: rgb(190,106,192);
 
-    .c-job {
-        svg.job {
-            rect {
-                /* if no job status display nothing */
-                fill: transparent;
-                stroke: transparent;
-            }
+    #{$cjob} {
+        &.submitted {
+            fill: $teal;
+            stroke: $teal;
+        }
 
-            &.submitted rect {
-                fill: $teal;
-                stroke: $teal;
-            }
+        &.running {
+            fill: $blue;
+            stroke: $blue;
+        }
 
-            &.running rect {
-                fill: $blue;
-                stroke: $blue;
-            }
+        &.succeeded {
+            fill: $green;
+            stroke: $green;
+        }
 
-            &.succeeded rect {
-                fill: $green;
-                stroke: $green;
-            }
+        &.failed {
+            fill: $red;
+            stroke: $red;
+        }
 
-            &.failed rect {
-                fill: $red;
-                stroke: $red;
-            }
-
-            &.submit-failed rect {
-                fill: $pink;
-                stroke: $pink;
-            }
+        &.submit-failed {
+            fill: $pink;
+            stroke: $pink;
         }
     }
 }
 
-.job_theme--greyscale {
 
+.job_theme--greyscale {
     $light: rgb(208,208,208);
     $medium: rgb(133,133,133);
     $dark: rgb(0,0,0);
 
-    .c-job {
-        svg.job {
-            rect {
-                /* if no job status display nothing */
-                fill: transparent;
-                stroke: transparent;
-            }
+    #{$cjob} {
+        &.submitted {
+            fill: transparent;
+            stroke: $light;
+        }
 
-            &.submitted rect {
-                fill: transparent;
-                stroke: $light;
-            }
+        &.running {
+            fill: $light;
+            stroke: $light;
+        }
 
-            &.running rect {
-                fill: $light;
-                stroke: $light;
-            }
+        &.succeeded {
+            fill: $medium;
+            stroke: $medium;
+        }
 
-            &.succeeded rect {
-                fill: $medium;
-                stroke: $medium;
-            }
+        &.failed {
+            fill: $dark;
+            stroke: $dark;
+        }
 
-            &.failed rect {
-                fill: $dark;
-                stroke: $dark;
-            }
-
-            &.submit-failed rect {
-                fill: transparent;
-                stroke: $dark;
-            }
+        &.submit-failed {
+            fill: transparent;
+            stroke: $dark;
         }
     }
 }
 
 .job_theme--colour_blind {
-
     $grey: rgb(152,152,152);
     $success: rgb(108,218,255);
     $error: rgb(146,0,0);
 
-    .c-job {
-        svg.job {
-            rect {
-                /* if no job status display nothing */
-                fill: transparent;
-                stroke: transparent;
-            }
+    #{$cjob} {
+        &.submitted {
+            fill: transparent;
+            stroke: $grey;
+        }
 
-            &.submitted rect {
-                fill: transparent;
-                stroke: $grey;
-            }
+        &.running {
+            fill: $grey;
+            stroke: $grey;
+        }
 
-            &.running rect {
-                fill: $grey;
-                stroke: $grey;
-            }
+        &.succeeded {
+            fill: $success;
+            stroke: $success;
+        }
 
-            &.succeeded rect {
-                fill: $success;
-                stroke: $success;
-            }
+        &.failed {
+            fill: $error;
+            stroke: $error;
+        }
 
-            &.failed rect {
-                fill: $error;
-                stroke: $error;
-            }
-
-            &.submit-failed rect {
-                fill: transparent;
-                stroke: $error;
-            }
+        &.submit-failed {
+            fill: transparent;
+            stroke: $error;
         }
     }
 }
--- a/src/styles/cylc/_job.scss
+++ b/src/styles/cylc/_job.scss
@@ -1,0 +1,127 @@
+.job_theme--normal {
+
+    $teal: rgb(125,207,212);
+    $blue: rgb(106,164,241);
+    $green: rgb(81,175,81);
+    $red: rgb(207,72,72);
+    $pink: rgb(190,106,192);
+
+    .c-job {
+        svg.job {
+            rect {
+                /* if no job status display nothing */
+                fill: transparent;
+                stroke: transparent;
+            }
+
+            &.submitted rect {
+                fill: $teal;
+                stroke: $teal;
+            }
+
+            &.running rect {
+                fill: $blue;
+                stroke: $blue;
+            }
+
+            &.succeeded rect {
+                fill: $green;
+                stroke: $green;
+            }
+
+            &.failed rect {
+                fill: $red;
+                stroke: $red;
+            }
+
+            &.submit-failed rect {
+                fill: $pink;
+                stroke: $pink;
+            }
+        }
+    }
+}
+
+.job_theme--greyscale {
+
+    $light: rgb(208,208,208);
+    $medium: rgb(133,133,133);
+    $dark: rgb(0,0,0);
+
+    .c-job {
+        svg.job {
+            rect {
+                /* if no job status display nothing */
+                fill: transparent;
+                stroke: transparent;
+            }
+
+            &.submitted rect {
+                fill: transparent;
+                stroke: $light;
+            }
+
+            &.running rect {
+                fill: $light;
+                stroke: $light;
+            }
+
+            &.succeeded rect {
+                fill: $medium;
+                stroke: $medium;
+            }
+
+            &.failed rect {
+                fill: $dark;
+                stroke: $dark;
+            }
+
+            &.submit-failed rect {
+                fill: transparent;
+                stroke: $dark;
+            }
+        }
+    }
+}
+
+.job_theme--colour_blind {
+
+    $grey: rgb(152,152,152);
+    $success: rgb(108,218,255);
+    $error: rgb(146,0,0);
+
+    .c-job {
+        svg.job {
+            rect {
+                /* if no job status display nothing */
+                fill: transparent;
+                stroke: transparent;
+            }
+
+            &.submitted rect {
+                fill: transparent;
+                stroke: $grey;
+            }
+
+            &.running rect {
+                fill: $grey;
+                stroke: $grey;
+            }
+
+            &.succeeded rect {
+                fill: $success;
+                stroke: $success;
+            }
+
+            &.failed rect {
+                fill: $error;
+                stroke: $error;
+            }
+
+            &.submit-failed rect {
+                fill: transparent;
+                stroke: $error;
+            }
+        }
+    }
+}
--- a/src/components/cylc/Job.vue
+++ b/src/components/cylc/Job.vue
@@ -15,7 +15,6 @@
     -->
     <svg
       class="job"
-      v-bind:class="[status]"
       viewBox="0 0 100 100"
     >
       <!-- the job status icon
@@ -23,6 +22,7 @@
            * let height = 100 - y - stroke-width
       -->
       <rect
+        v-bind:class="[status]"
         x="10" y="10"
         width="80" height="80"
         rx="20" ry="20"
--- a/src/components/cylc/Job.vue
+++ b/src/components/cylc/Job.vue
@@ -33,44 +33,13 @@
 </template>
 
 <style lang="scss">
-    .c-job {
-        svg.job {
-            /* scale the icon to the font-size */
-            width: 1em;
-            height: 1em;
-
-            rect {
-                /* if no job status display nothing */
-                fill: transparent;
-                stroke: transparent;
-            }
-
-            &.submitted rect {
-                fill: rgb(125,207,212);
-                stroke: rgb(125,207,212);
-            }
-
-            &.running rect {
-                fill: rgb(106,164,241);
-                stroke: rgb(106,164,241);
-            }
-
-            &.succeeded rect {
-                fill: rgb(81,175,81);
-                stroke: rgb(81,175,81);
-            }
-
-            &.failed rect {
-                fill: rgb(207,72,72);
-                stroke: rgb(207,72,72);
-            }
-
-            &.submit-failed rect {
-                fill: rgb(190,106,192);
-                stroke: rgb(190,106,192);
-            }
-        }
+  .c-job {
+    svg.job {
+      /* scale the icon to the font-size */
+      width: 1em;
+      height: 1em;
     }
+  }
 </style>
 
 <script>
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -5,6 +5,7 @@
 @import "cylc/toolbar";
 @import "cylc/drawer";
 @import "cylc/footer";
+@import "cylc/job";
 
 .theme--light.v-application {
   background-color: transparent;


src/styles/cylc/_job.scss
```diff
@@ -0,0 +1,127 @@
...
+    }
+}
+
+.job_theme--greyscale {
+
+    $light: rgb(208,208,208);
+    $medium: rgb(133,133,133);
+    $dark: rgb(0,0,0);
+
+    .c-job {
```
kinow: With the approach of theming in css/scss, we will end up with a lot of repetition like the `.c-job` being re-defined as child of `.job_theme--themeN`, and may lead to maintenance problems if we end up customizing more than just colors.

Could we hold back theming, and implement in each component after https://github.com/cylc/cylc-ui/issues/9? Vuetify provides a way to have themes in [JS](https://github.com/cylc/cylc-ui/blob/aa6e75244dd4add7537552e4baf7fe9d904ebe27/src/plugins/theme.js), but very simple.

I think the best approach for `cylc-ui` would be to have the theme data in JS, but with Vuex. Similar to [this example](https://codesandbox.io/s/3no663nv6).

src/styles/cylc/_job.scss
```diff
@@ -0,0 +1,127 @@
+.job_theme--normal {
+
+    $teal: rgb(125,207,212);
+    $blue: rgb(106,164,241);
+    $green: rgb(81,175,81);
+    $red: rgb(207,72,72);
+    $pink: rgb(190,106,192);
```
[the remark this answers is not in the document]
oliver-sanders: I purposely didn't do that as the job theme is not related to the material-dashboard theme.

src/styles/cylc/_job.scss
oliver-sanders: The job theme isn't related to the app theme (users will need to be able to change the two separately) so this really is un-related to #9.

I'm not convinced by theming in JS but do try to convince me, I think in-line style is a much bigger maintenance headache then well structured CSS.

kinow: >The job theme isn't related to the app theme (users will need to be able to change the two separately)

Agreed.

>I'm not convinced by theming in JS but do try to convince me, I think in-line style is a much bigger maintenance headache then well structured CSS.

When the user switches the themes, we will need to store the theme chosen in Vuex. Then I believe we would also use that value to dynamically set the value in `.job_theme--$value`.

So up to here both approaches would work and are quite similar.

1. As a user, 

I would like to be able to easily create themes and customize them in my site. With the CSS approach, they would need to edit SCSS and re-build the project.

With the JS approach, I **think** (would need to investigate) we could have dynamically loaded themes. Users woulds copy a file like `$cylc-ui-src/src/plugins/default-theme.js`, edit the values and put the JS somewhere in the server with a name `my-theme-name.js`, then the theme `my-theme-name` could be selected.

2. As a developer,

Either JS or CSS, I would find it easier if we had the definition of colours/variables/etc in one place, with the definition of elements/classes/etc somewhere else and without repetition.

Isn't there a way to achieve the same in CSS without the need to re-define classes under each theme? If the `.c-job` got more attributes or nested types, or if a component needs to define several classes, that will get annoyingly long.

If we are able to organize the current CSS code without this repetition, we can merge it and work around the CSS vs. JS approach in a separate issue. WDYT?

oliver-sanders: Good points!

> Isn't there a way to achieve the same in CSS without the need to re-define classes under each theme

I think so, depends how scss compiles down (I'm used to sylus), will take a crack.

> Users woulds copy a file like `$cylc-ui-src/src/plugins/default-theme.js`

I guess the big advantage is that the user could create a custom theme from within the browser without having to work in the FS.

> we can merge it and work around the CSS vs. JS approach in a separate issue. WDYT?

I'm perfectly happy for this code to get scrapped either now or in the future, whatever's best.
(was only working on this because it's unlikely to get into conflict with other current work).

kinow: >I'm perfectly happy for this code to get scrapped either now or in the future, whatever's best.
(was only working on this because it's unlikely to get into conflict with other current work).

So let's get it merged and discuss the theming elsewhere if you prefer. Besides the comment above :point_up: about `--fix` in the linter (which I think is not necessary), I think this is good to be merged.

oliver-sanders: OK, I'm trying to work out how to to re-write the SCSS, I could put the job theme as a SCSS variable but then the CSS would have to be recompiled to change theme...

oliver-sanders: Bit of hacking, the best I can come up with is:

```scss
/* normal theme */
$n-teal: rgb(125,207,212);
$n-blue: rgb(106,164,241);
$n-green: rgb(81,175,81);
$n-red: rgb(207,72,72);
$n-pink: rgb(190,106,192);

$job-theme-normal: (
    "submitted": (
        "fill": $n-teal,
        "stroke": $n-teal
    ),
    "running": (
        "fill": $n-blue,
        "stroke": $n-blue
    ),
    "suceeded": (
        "fill": $n-green,
        "stroke": $n-green
    ),
    "failed": (
        "fill": $n-red,
        "stroke": $n-red
    ),
    "submit-failed": (
        "fill": $n-pink,
        "stroke": $n-pink
    ),
);


/* greyscale theme */
$g-light: rgb(208,208,208);
$g-medium: rgb(133,133,133);
$g-dark: rgb(0,0,0);

$job-theme-greyscale: (
    "submitted": (
        "stroke": $g-light
    ),
    "running": (
        "fill": $g-light,
        "stroke": $g-light
    ),
    "succeeded": (
        "fill": $g-medium,
        "stroke": $g-medium
    ),
    "failed": (
        "fill": $g-dark,
        "stroke": $g-dark
    ),
    "submit-failed": (
        "stroke" $g-dark
    )
);


/* colour-blind theme */
$b-grey: rgb(152,152,152);
$b-success: rgb(108,218,255);
$b-error: rgb(146,0,0);

$job-theme-colour-blind: (
    "submitted": (
        "stroke": $b-grey
    ),
    "running": (
        "fill": $b-grey,
        "stroke": $b-grey
    ),
    "succeeded": (
        "fill": $b-success,
        "stroke": $b-success
    ),
    "failed": (
        "fill": $b-error,
        "stroke": $b-error
    ),
    "submit-failed": (
        "stroke": $b-error
    ),
);


/* job theme registry */
$job-themes: (
    "normal": $job-theme-normal,
    "greyscale": $job-theme-greyscale,
    "colour-blind": $job-theme-colour-blind
);


/* dump themes */
@each $name, $theme in $job-themes {
    .c-job {
        svg.job {
            rect {
                /* if no job status display nothing */
                fill: transparent;
                stroke: transparent;
            }
            @each $state, $style in $theme {
                &.#{$state} rect {
                    @each $key, $value in $style {
                        #{$key}: $value;
                    }
                }
            }
        }
    }
}
```

oliver-sanders: Let me know which way you'd prefer.

kinow: This one looks better. But we still have the CSS class redefined within each theme. I think we can have the CSS class defined in one place, and then the theme variables defined elsewhere... e.g.

```scss
// we can define the colours of the default job-theme in a
// ~styles/cylc/job-themes/_colours.scss or _variables.scss...
// I used this one just to avoid creating global variables in this file,
// and because the colours are quite similar to the ones you picked...
@import "../material-dashboard/colors";

/* job css class wrapper */
.c-job {
    svg.job {
        rect {
            /* if no job status display nothing */
            fill: var(--job_theme--fill);
            stroke: var(--job_theme--stroke);
        }

        &.submitted rect {
            fill: var(--job_theme--submitted-fill);
            stroke: var(--job_theme--submitted-stroke);
        }

        &.running rect {
            fill: var(--job_theme--running-fill);
            stroke: var(--job_theme--running-stroke);
        }

        &.succeeded rect {
            fill: var(--job_theme--succeeded-fill);
            stroke: var(--job_theme--succeeded-stroke);
        }

        &.failed rect {
            fill: var(--job_theme--failed-fill);
            stroke: var(--job_theme--failed-stroke);
        }

        &.submit-failed rect {
            fill: var(--job_theme--submit-failed-fill);
            stroke: var(--job_theme--submit-failed-stroke);
        }
    }
}

// --- Theming with CSS variables [1], could be moved to theme file(s)?

/**
 The normal theme uses the same variables from the original theme.

 The other classes simply re-define these variables locally, in their own
 scope. Once a component adopts the class, the variables are replaced in
 "runtime". This way SCSS variables are used in "compile time" (dev), and
 CSS variables in "runtime" (live), with the option of being changed with
 CSS and/or JS.
 */

// instead of the putting variables in :root, another alternative is to
// provide a second argument to var($variable, $fallback), moving the
// default values to the fallback. Maybe that's better?

/* job_theme global variables */
:root {
    --job_theme--fill: transparent;
    --job_theme--stroke: transparent;
    --job_theme--submitted-fill: #{$teal}; // see note [2]
    --job_theme--submitted-stroke: #{$teal};
    --job_theme--running-fill: #{$blue};
    --job_theme--running-stroke: #{$blue};
    --job_theme--succeeded-fill: #{$green};
    --job_theme--succeeded-stroke: #{$green};
    --job_theme--failed-fill: #{$red};
    --job_theme--failed-stroke: #{$red};
    --job_theme--submit-failed-fill: #{$pink};
    --job_theme--submit-failed-stroke: #{$pink};
}

/* normal theme variable values */
.job_theme--normal {
    --job_theme--fill: transparent;
    --job_theme--stroke: transparent;
    --job_theme--submitted-fill: #{$teal};
    --job_theme--submitted-stroke: #{$teal};
    --job_theme--running-fill: #{$blue};
    --job_theme--running-stroke: #{$blue};
    --job_theme--succeeded-fill: #{$green};
    --job_theme--succeeded-stroke: #{$green};
    --job_theme--failed-fill: #{$red};
    --job_theme--failed-stroke: #{$red};
    --job_theme--submit-failed-fill: #{$pink};
    --job_theme--submit-failed-stroke: #{$pink};
}

/* greyscale theme variable values */
.job_theme--greyscale {
    $light: rgb(208,208,208);
    $medium: rgb(133,133,133);
    $dark: rgb(0,0,0);
    --job_theme--fill: transparent;
    --job_theme--stroke: transparent;
    --job_theme--submitted-fill: transparent;
    --job_theme--submitted-stroke: #{$light};
    --job_theme--running-fill: #{$light};
    --job_theme--running-stroke: #{$light};
    --job_theme--succeeded-fill: #{$medium};
    --job_theme--succeeded-stroke: #{$medium};
    --job_theme--failed-fill: #{$dark};
    --job_theme--failed-stroke: #{$dark};
    --job_theme--submit-failed-fill: transparent;
    --job_theme--submit-failed-stroke: #{$dark};
}

/* colour blind theme variable values */
.job_theme--colour_blind {
    $grey: rgb(152,152,152);
    $success: rgb(108,218,255);
    $error: rgb(146,0,0);
    --job_theme--fill: transparent;
    --job_theme--stroke: transparent;
    --job_theme--submitted-fill: transparent;
    --job_theme--submitted-stroke: #{$grey};
    --job_theme--running-fill: #{$grey};
    --job_theme--running-stroke: #{$grey};
    --job_theme--succeeded-fill: #{$success};
    --job_theme--succeeded-stroke: #{$success};
    --job_theme--failed-fill: #{$error};
    --job_theme--failed-stroke: #{$error};
    --job_theme--submit-failed-fill: transparent;
    --job_theme--submit-failed-stroke: #{$error};
}

// [1] https://www.bolajiayodeji.com/introducing-css-custom-properties-variables/
// [2] See issue https://github.com/sass/sass/issues/1128, need string interpolation
//     here instead of accessing SCSS variable directly

```

kinow: This way I think users should be able to override the values in their own themes with some CSS... and maybe a JS somewhere that activates/loads the themes (still need to investigate this part, though we may be able to use CSS variables, which seem to be a compromise between our approaches?).

Not sure if we need to define the CSS [`!default`](https://stackoverflow.com/questions/10643107/what-does-default-in-a-css-property-value-mean) marker as well, to avoid users having to use `!important`... but still needs investigation as well.

oliver-sanders: Ooh, pure CSS3 variables. Will have a go.

oliver-sanders: Issues with this approach are:

* Themes don't have freedom over the CSS (e.g. can't add a new property)
* The `:root` theme can leak into the job themes (imagine what will happen to custom themes if we add a new property)

Latest idea:

Define the css namespace `c-job svg.job rect` as a variable at the top of the file, use it to namespace theme CSS that way we aren't re-defining the inheritance only the fact that `.job` must live within `.job_theme` which is a fundamental assumption anyway.

kinow: >Themes don't have freedom over the CSS (e.g. can't add a new property)

I thought that was expected. I mean, we would define the properties to be re-defined by users.

>The :root theme can leak into the job themes (imagine what will happen to custom themes if we add a new property)

Yup, see the TODO in the example above. Instead of pollyting root, we can simply use a default value in the `var(--..., $default-value-instead-of-root)`.

>As we are re-defining the variables, how does theme switching work?

A user would choose a new theme. That theme would be set in Vuex via JavaScript. The components would rely on Vuex to retrieve the application theme name, and concatenate the component base class (e.g. `job_theme--`) with the theme name.

>Define the css namespace c-job svg.job rect as a variable at the top of the file, use it to namespace theme CSS that way we aren't re-defining the inheritance only the fact that .job must live within .job_theme which is a fundamental assumption anyway.

Uh, that sounds interesting. Can you elaborate with a short code example? Would it be something like:

```scss
$theme-namespace: .c-job svg.job react;
```

And then `#{$theme-namespace}` in some parts of the CSS?

oliver-sanders: > Uh, that sounds interesting. Can you elaborate with a short code example?

Commited as https://github.com/cylc/cylc-ui/pull/134/commits/3dfda938c9a03cf43920a6ab552371ddc0fd7b33

kinow: Ah, similar to what I thought you suggested.

It reduces duplication in the code, but the final produced CSS is still the same. If we had to add a new state (e.g. `dead`). We would have to change the contents of `#{$cjob} {` under N themes (i.e. .job_theme--normal, .job_theme--greyscale, job_theme--...).

Whereas with the CSS variables we would change it in one place I think?

oliver-sanders: > If we had to add a new state (e.g. dead). We would have to change the contents of #{$cjob} { under N themes

Yes! To add a new state we would need to modify each theme, that is the correct behaviour to me, we wouldn't want to fallback on a global definition.

kinow: I think we have an impasse here. Perhaps a third person could come up with another solution, or convince one of us :) (but I feel like it's clarified how we both were thinking around theming, and we've made some good progress in the discussion)

hjoliver: I guess this still needs to be resolved?

kinow: Yup, no consensus on which way to implement theming. But I think I digressed too much, so perhaps we should use @oliver-sanders ' approach and move the discussion around how to do theming for another issue.

@oliver-sanders feel free to update the PR with your latest version (if you still want to modify it). I can test it again, and then approve it if there are no errors on console/IDE/etc :+1:

oliver-sanders: I'm happy to wait for your theming PR then adapt this into it.

oliver-sanders: Ah, wait, you've closed #143....

oliver-sanders:
I forget where we are with this.

Looks like this PR is waiting for #134 and #134 was closed pending this PR.

@kinow you wanted to go down a more javascript based approach for ease of configuration if I remember correctly. Where did we get with user-configuration?

kinow:
> 
> 
> I forget where we are with this.
> 
> Looks like this PR is waiting for #134 and #134 was closed pending this PR.
> 
> @kinow you wanted to go down a more javascript based approach for ease of configuration if I remember correctly. Where did we get with user-configuration?

We made no progress with user-configuration yet. There is one conflict on this PR now, but it'd be a shame to block this any longer. Travis passed, and this shouldn't break anything, so will leave the approval pending only fixing the conflict.

We can sort out user-configuration or other approaches later if necessary. LGTM 👍

package.json
```diff
@@ -8,6 +8,7 @@
     "build:report": "vue-cli-service build --report",
     "build:watch": "vue-cli-service build --watch --mode development",
     "lint": "vue-cli-service lint",
+    "lint-fix": "vue-cli-service lint --fix",
```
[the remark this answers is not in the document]
oliver-sanders: (removed)

oliver-sanders:
OK, we can extract variables into the user configuration later.

src/styles/cylc/_job.scss
```diff
@@ -0,0 +1,112 @@
...
+            fill: transparent;
+            stroke: $dark;
+        }
+    }
+}
+
+.job_theme--colour_blind {
+    $grey: rgb(152,152,152);
+    $success: rgb(108,218,255);
+    $error: rgb(146,0,0);
```
sadielbartholomew: For future development I think it would for clarity be better to name these variables to give an indication as to the resulting colours rather than the corresponding job state category, which can be seen from the maps in ``#{$cjob}`` anyway (for the ``greyscale`` theme it is obvious from context they are all a shade of grey of lightness as per the name).

So perhaps you could use ``cb-blue`` & ``cb-red`` instead of ``success`` & ``error``, respectively?

oliver-sanders: This was kinda on purpose on the basis of it being a colour-blind theme.

oliver-sanders:
> most of the SCSS selectors set the same (colour) variable for both fill & stroke

It is not possible to do anything like:

```
fill = color: $colour
```

src/styles/cylc/_job.scss
sadielbartholomew: Are the variable names ever exposed to the users? I assumed that only developers would encounter them, & if so I think my point stands (though it is minor & isn't a blocker in any way). But if the names are somehow processed & displayed to users then definitely, your point makes a lot of sense.

Issue #59, opened 2012-10-25
gcylc - keep view options client-side


benfitzpatrick:
Adjusting some view options in `gcontrol`'s graph view affects other `gcontrol` instances - for example 'Group All Families'. This would be a problem for shared suites, and would probably be best kept on the client side rather than stored remotely.

kinow:
Notes from discussion to move to cylc-web:

>  * (Matt) Browser cookies? Or user session settings on server side? Should probably move to cylc/cylc-web.
>  * (Hilary) Agreed. Not sure how to do it yet, but we don't want one user's family-collapse choices to affect other users viewing the same suite (as happens now)

oliver-sanders:
Superseded by #335

Issue #53, opened 2019-03-10
cylc run --set=NAME=VALUE inputs persistence and gui issues


hjoliver:
The GUI needs somewhere for users to enter Jinja2 `var=value` pairs for use with any spawned command that takes the `--set` option.  As of #745 this can be achieved for `cylc run` and `cylc restart` with a new "other options" entry in the suite start popup, but not for validation etc.

kinow:
Notes from discussion about moving it to cylc-web:

>  * (Matt) Yes. Values are stored in the runtime SQLite file. We need to be able to enable read/write from the web UI.
>  * (Hilary) Agreed. (BUT it might be good to require that such input values are stored in a file on disk - the to-be-migrated rose-suite.conf - and rather than specified dynamically in the UI, and allow that file to be edited via the UI?)

MetRonnie:
The up-to-date solution for this would be to query the workflow's template vars to automatically populate them in the Play form:

![image](https://github.com/cylc/cylc-ui/assets/61982285/c4efbbb8-4aee-4e0a-9480-57d0f2cc71d3)

PR #143, opened 2019-07-23
Local user settings

kinow:
WIP example of storing user settings locally (client-side) using the `LocalStorage`. Not sure if we wouldn't still use it later, or if we would replace it entirely by backend stored settings (see #53, #59, for example).

This allows users (mainly developers for now) to store locally in their browsers the following settings:

- applicationTheme
- jobStatesTheme (#134 #9) 
- language

Right now only language is being used with our i18n Vue plugin. But we could hook up the job state theme name, and try to use it with CSS/SCSS/Stylys/JS/etc (see #134).

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -7,6 +7,8 @@
 </template>
 
 <script>
+import { UserService } from 'user-service'
+
 const defaultLayout = 'empty'
 
 export default {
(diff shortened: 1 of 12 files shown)


kinow:
A few things missing:

- [ ] implement real service (probably same as mocked for now)
- [ ] use the theme with components to see if that works well
- [ ] user input client validation (vee-validate)

kinow:
@oliver-sanders will try to finish it this week to show what I had in mind for theming :+1:

oliver-sanders:
> @oliver-sanders will try to finish it this week to show what I had in mind for theming +1

Thanks, I'll hold off speculating!

Looks good BTW.

kinow:
Closing as I think there will be a different approach to the theming (see #134). Will rebase branch after it gets merged and see which direction to go :+1:

PR #151, opened 2019-07-29
Use same favicon from cylc-doc

kinow:
Was looking at a pull request in Cylc Doc that has a new Sphinx theme, but also a new favicon. I copied the current favicon from Cylc GUI (GTK). Or we can use another icon if anyone has other suggestions.


kinow:
What it looks like: 

image

kinow:
And before the change:

image

One downside is that if you open many tabs with Cylc UI and Cylc Doc, it may be hard to distinguish them?

PR #153, opened 2019-07-29
Fix logo chrome

### LINKED ISSUE #148: Logo not working on Chrome
Synced the code, and started the test server this morning, launching on Firefox and Chrome, but the logo was not displayed on Chrome.

Not sure if due to some update... I am quite sure I used Chrome before since we changed the logo to use SVG.

kinow:
Close #148 

@oliver-sanders apparently just needed to specify the width and height. Also copied extra attribute from the [StackOverflow answer](https://stackoverflow.com/a/37074597) where I found the fix.

Firefox 68.0.1:

image

Chrome 75.0.3770.142 (with adblock on, and cropped a piece of the logo to the left because of my bad skills with screenshots, but it's displaying just like in Ffox):

image

One review should do?

### DIFF
--- a/src/components/cylc/Header.vue
+++ b/src/components/cylc/Header.vue
@@ -2,7 +2,8 @@
   <v-list-tile-action-text class="pt-3 pb-3 c-header">
     <v-layout align-center justify-center row wrap fill-height>
       <div class="mb-2">
-        <svg version="1.1" viewBox="0 0 655 260" xmlns="http://www.w3.org/2000/svg" xmlns:osb="http://www.openswatchbook.org/uri/2009/osb">
+        <svg version="1.1" preserveAspectRatio="xMinYMin meet" width="100%" height="100%" viewBox="0 0 655 260" xmlns="http://www.w3.org/2000/svg"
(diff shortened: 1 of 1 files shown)


src/components/cylc/Header.vue
```diff
@@ -2,7 +2,8 @@
   <v-list-tile-action-text class="pt-3 pb-3 c-header">
     <v-layout align-center justify-center row wrap fill-height>
       <div class="mb-2">
-        <svg version="1.1" viewBox="0 0 655 260" xmlns="http://www.w3.org/2000/svg" xmlns:osb="http://www.openswatchbook.org/uri/2009/osb">
+        <svg version="1.1" preserveAspectRatio="xMinYMin slice" width="100%" height="100%" viewBox="0 0 655 260" xmlns="http://www.w3.org/2000/svg"
```
oliver-sanders: We want `meet` rather than `slice` as `slice` can crop the SVG

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/preserveAspectRatio#Syntax

kinow: Aaahh. I will admit that I blindly copied the code without checking the slice. Done! Thanks!

PR #156, opened 2019-07-31
Remove (unnecessary) vuex getters

### LINKED ISSUE #122: Remove getters from Vuex stores if they are not depending on other data
Vuex has getters, that work similar to `computed` in Vue components, caching and updating values reactively.

I wrote getters in some stores that are redundant, and instead could access the data directly.

kinow:
Remove the [getters](https://vuex.vuejs.org/guide/getters.html) from Vuex stores. The getters are similar to `computed` values in Vuejs.

So you will only need them when you have a value that depends on some `state` data from Vuex, using reactivity. This will eliminate unused code, and make it easier to write the unit tests.

This closes #122

### DIFF
--- a/src/components/core/Footer.vue
+++ b/src/components/core/Footer.vue
@@ -6,21 +6,25 @@
   >
     <v-spacer/>
     <span class="font-weight-light copyright">
-      <strong v-if="env !== 'PRODUCTION'">{{ env }}</strong>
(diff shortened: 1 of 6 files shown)


kinow:
One of the getters was for the Cylc UI version, which was broken. It appears it was accidentally removed (from looking at the git commit history), so am adding it back. Meaning that if merged, the footer will have Cylc 1.0.0, 1.0.0 being the current version in `package.json` (not necessarily final version). This information can be moved elsewhere later if others prefer :+1:

matthewrmshin:
(Conflict.)

kinow:
Fixed, thanks @matthewrmshin

PR #158, opened 2019-07-31
Remove notifications view

kinow:
Removing another View that came with the theme, but we are not using.


hjoliver:
Might we actually want to use a notifications view?  List all notifiable task failures etc. that occured...  I suppose we can always put this back later if needed?

kinow:
Yup, we can definitely add it back. There are other views that we have already removed (e.g. typography) that contain examples on how to use the theme. We can copy them either from the git history, or from the git repo: https://github.com/creativetimofficial/vuetify-material-dashboard/

PR #167, opened 2019-08-07
hide tooltip on navigation and fix undo

MartinRyan:
hide tooltips on navigating away from the graph view and fix undo

### DIFF
--- a/src/views/Graph.vue
+++ b/src/views/Graph.vue
@@ -1069,9 +1069,9 @@ export default {
       document.addEventListener(
         'keydown',
         function (event) {
-          if (event.ctrlKey && event.which === '90') {
+          if (event.ctrlKey && event.which === 90) {
             cy.undoRedo().undo()
-          } else if (event.ctrlKey && event.which === '89') {
+          } else if (event.ctrlKey && event.which === 89) {
             cy.undoRedo().redo()
           }
         },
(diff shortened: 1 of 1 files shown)


src/views/Graph.vue
```diff
@@ -504,7 +505,7 @@ export default {
       }
 
       async function registerExtensions () {
-        // register extensions
+        // // register extensions
```
kinow: accidental double comment?

kinow:
Tested `master` first. Clicking on `Graph`, then clicking on a node, it displays a popup. By clicking on another link like `Dashboard`, now the new view is loaded, but the popup is still being displayed.

Also, while in the `Graph` view, moving nodes around, CTRL+Z does nothing on `master`.

Happy to report this PR indeed fixes both issues :clap: and I had no idea we had an undo feature. Thanks @MartinRyan !

src/views/Graph.vue
MartinRyan: lo yep!

hjoliver:
@MartinRyan - did you accidentally merge this?

kinow:
@hjoliver, maybe when @MartinRyan left for vacation we were still doing the 1 review only? I think I asked in Riot about that around the time he was leaving or had left already...

hjoliver:
@kinow - I was wondering that too, but @MartinRyan also merged this without fixing the double comment you noted above ... so I thought maybe it was an accident!  (Also it isn't "customary" to merge your own PRs!).  (No biggie, just trying to figure out what the situation is).

PR #168, opened 2019-08-07
task: add isHeld attribute

oliver-sanders:
Companion to https://github.com/cylc/cylc-flow/pull/3230

Add a held icon to the task state.

### Further Work

For the moment I've kept the task icon square (which is much simpler). We may want to change this later, perhaps make the task icon a little wider and budge the held icon to the left a bit.

Will leave this as is for now until we have decided what we are doing with the retry state.

### Example Screenshot

Screenshot from 2019-08-07 16-12-22

### Example App To Replicate

```vue
<template>
  <div>
    <ul v-for="state in states">
      <li>
        <task
          :status="state"
          :isHeld="false"
        />
        <span style="padding-left:0.5em"></span>
        <task
          :status="state"
          :isHeld="true"
          progress="0.3"
        />
        foo
        <job :status="state" />
      </li>
    </ul>
  </div>
</template>

<script>
import Job from '@/components/cylc/Job'

export default {
  components: {
    task: Task,
    job: Job
  },
  data: () => ({
    states: [
      'waiting',
      'submitted',
      'running',
      'succeeded',
      'failed',
      'submit-failed',
    ]
  })
}
</script>

<style lang="scss">
  @import '~@/styles/index.scss';
</style>
```

### DIFF
--- a/src/components/cylc/Task.vue
+++ b/src/components/cylc/Task.vue
@@ -15,60 +15,91 @@
     -->
     <svg
       class="task"
-      v-bind:class="[status]"
+      v-bind:class="[classes]"
       viewBox="0 0 100 100"
     >
-      <!-- progress pie chart:
-             * position in the middle
-             * let radius = 25% radius
-             * let stroke-width = 100% - (2*radius) = 50%
-             * let stroke-dasharray ~= 2 * pi * radius
-             * use dashes as the line style to turn this into a pie chart
-      -->
-      <circle
-        id="progress"
-        cx="50" cy="50"
-        r="25"
-        stroke-width="50"
-        stroke-dasharray="157"
-        v-bind:style="progressStyle"
-      ></circle>
-      <!-- circle in the middle
-             * position in the middle
-             * radius can be changed independently
-      -->
-      <circle
-        id="hub"
-        cx="50" cy="50"
-        r="12">
-      </circle>
-      <!-- outer circle
-             * position in the middle
-             * let radius = 50% - (2*stroke-width)
-      -->
-      <circle
-        id="outline"
-        cx="50" cy="50"
-        r="46"
-        stroke-width="8">
-      </circle>
-      <!-- cross (failure)
-             * create a plus and rotate it 45 deg around the origin
-             * let rx ~= max(width, height) * 0.1
+      <!--
+        Status indicator - the status portion of the task state.
       -->
       <g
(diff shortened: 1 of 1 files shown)


oliver-sanders:
Note the isHeld field isn't in the GraphQL schema yet (will need a separate PR to Cylc Flow) so this cann't (yet) be incorporated into any of the views.

hjoliver:
Very nice!

src/components/cylc/Task.vue
```diff
@@ -178,6 +232,13 @@ export default {
       return {
         'stroke-dashoffset': `calc(157 - (157 * ${this.progress / 100} + ))`
       }
+    },
+    classes: function () {
+      var classes = [this.status]
```
kinow: s/var/const

src/components/cylc/Task.vue
```diff
@@ -15,60 +15,91 @@
...
-        r="46"
-        stroke-width="8">
-      </circle>
-      <!-- cross (failure)
-             * create a plus and rotate it 45 deg around the origin
-             * let rx ~= max(width, height) * 0.1
+      <g
+        id="status"
+      >
+        <!-- progress pie chart:
```
kinow: Thanks a lot for the great comments!

src/components/cylc/Task.vue
oliver-sanders: Dammit, I'm really going to have to do some reading!

kinow: I think `var` is considered deprecated (or even a risk, because... I think JS moves the var variable declarations to the top of the code, which may cause some weird bugs). If the variable doesn't change, `const`. If it needs to change (e.g. a counter), then `let`. I acquired the habit to always go with `const`, and change to `let` if necessary (used to do the same with Java's `final`)

oliver-sanders: In lines 237-8 the list `classes` may get extended (`classes.push`) so should this be `let`?

kinow: Sorry, I should have said "if the variable is not re-assigned". `const` creates a mutable variable.

You are correct that `classes` is changed, but only the internal structure, not the variable pointer. So still a `const`. Sorry for the confusion.

oliver-sanders: Aaah, now it makes sense!

PR #172, opened 2019-08-08
Add logo file

### LINKED ISSUE #163: Move SVG logo to disk
JupyterHub, when initialized to run Cylc UI Servers, loads the Cylc logo from the same location defined for the static assets (well, it's hard-coded in `jupyterhub_config.py` so it could be another value).

After we replaced the old Cylc logo (`logo.png`) by embedded `<svg>` tags in the code of Cylc UI, JupyterHub couldn't locate the logo file anymore, and is now displaying just the `alt` text "JupyterHub".

We can either move the SVG logo back to the disk and load here in Cylc UI, or include the logo somewhere in Cylc UI Server. I think the former is easier :+1:

kinow:
Close #163 

@oliver-sanders I tried moving the current SVG to disk, by simply copying the `<svg>...</svg>` contents. It worked, in a way that I could view the SVG. But the browser refused to display it as ``, or `<v-img>`. Tried `<object>` too, but looking at the element inspector, the SVG was always rendered as 0x0, hidden from the user.

I was reading about it, and some people pointed there are browsers that may block SVG's being added due to some security vulnerabilities. I decided I had spent long enough on this simple issue, and went with the approach to just have the logo in two places (hope it's not likely to change?).

Would this change be alright? If so, I will raise another PR against cylc-uiserver, as it currently looks for `logo.png`. I did a quick testing, and it seems to work once we use `logo.svg` in our `jupyterhub_config.py`.

Cheers
Bruno


kinow:
image

oliver-sanders:
> the browser refused to display it as 

SVG should work in an `` tag in all modern browsers https://caniuse.com/#feat=svg-img.

We've tried the following example on Firefox & Edge:

```html
<html>
  <head></head>
  <body>
    <img src="cylc-logo.svg" width="100%" height="100%" />
  </body>
</html>
```

public/img/logo.svg
```diff
@@ -0,0 +1,16 @@
+<svg version="1.1" preserveAspectRatio="xMinYMin meet" width="100%" height="100%" viewBox="0 0 655 260" xmlns="http://www.w3.org/2000/svg">
```
oliver-sanders: We should probably move the `width` and `height` attributes out of the `<svg/>` tag so that they don't cause conflicts with the containing element.

oliver-sanders:
> I decided I had spent long enough on this simple issue, and went with the approach to just have the logo in two places

OK, I might take a crack at this later.

kinow:
Thanks @oliver-sanders! If your approach works, feel free to choose between committing directly to this branch, or raising a new PR and closing this one  :+1:

Issue #170, opened 2019-08-08
e2e tests


kinow:
We have [some automated e2e tests](https://github.com/cylc/cylc-ui/tree/815002a4da295d23e87eb0f0eec5604e3e7d2f69/tests/e2e) in Cylc UI at the moment with [Cypress](https://www.cypress.io/).

As we already have a few components done, that probably won't change much, I think we can start writing tests now as these won't change much either.

We need:

- [x] Make sure it works well in multiple developer environments (normally some Linux flavour, but the browser and Node/etc versions will probably vary a lot)
- [ ] Collect coverage
- [x] Check if others have a way of running it against multiple browsers (mobile/desktop/etc) in a cheap (as free) way

kinow:
From Feb 2019 left-over actions: https://cylc.github.io/cylc-admin/meetings/left-over-actions.html

kinow:
##  Check if others have a way of running it against multiple browsers (mobile/desktop/etc) in a cheap (as free) way

Unfortunately looks like there is no easy way to integrate it with cloud providers. One very popular provider (though not free), SauceLabs, doesn't support Cypress yet - https://support.saucelabs.com/hc/en-us/community/posts/115007199793-Support-for-Cypress.

It supports Selenium and Nightwatch, but we would still have to pay. We could use a normal cloud provider like AWS to run our tests, but that is still quite an enterprise, and maintenance won't be very easy - see https://stuartsandine.com/cypress-on-aws-lambda-part-1/

**Unfortunately I think the only solution for now is running the e2e tests manually before a release or before/after a big change.**

kinow:
# Make sure it works well in multiple developer environments (normally some Linux flavour, but the browser and Node/etc versions will probably vary a lot)

Well, it is supposed to work I think. Cypress - framework we are currently using - runs Google chrome, by default.

So we just need any other developer that could review #177, running `npm run test:e2e -- --headless` (or remove headless if you would like to see the test in the browser).

kinow:
Created #178 to make sure the expectation in this issue is just to get the functional end-to-end tests working.

kinow:
Closing as we already have the e2e tests :-)

PR #177, opened 2019-08-12
Fix e2e tests

kinow:
Part of #170 

Should fix our old e2e tests. We don't have any screen complete and ready for a test. The ones we have right now are going through changes to match with the design sketch.

The only two tests in our e2e battery are one for the Dashboard, and one for the User Profile views. These tests are simply looking for some text in the page once it is loaded.

This used to work, but got broken after some work on `master`. These tests are not executed on Travis CI for now.

Note: removed the test for `Graph.vue` as that was empty and not doing any assertion now (cc @MartinRyan :+1: )


kinow:
De-conflicted.

kinow:
Rebased.

oliver-sanders:
Looks good, could you rebase this to make sure they still pass after recent merges.

kinow:
>Looks good, could you rebase this to make sure they still pass after recent merges.

Good idea. Tested locally.

1. rebased onto `master`
2. `npm install` OK
3. `npm run lint` OK
4. `npm run test:unit` OK
5. `npm run test:e2e -- --headless` OK

Pushing the rebased branch now so Travis can double check this (except e2e, which is not yet rigged in Travis).

PR #180, opened 2019-08-12
Use only cy instance provided by vue-cytoscape

### LINKED ISSUE #112: static graph view: issues when leaving page
In the current ``master`` Static Graph View (which I appreciate is in a very early stage) there are a few issues I have noticed regarding leaving the Graph View page.

1. **Mostly blank page on return:** If you access & then leave it by changing page (via the sidebar menu or specifying some non-root URL), then return to the Graph VIew page, the graph & the top-left navigation bar both disappear (though you can still see the "minimap"). (However, if you specify & go to the root URL ``localhost:8000/user/<user>`` & access the Graph View after that, it is displayed again.) Example screenshot:

    bruno_new_scr

2. **Persistent task node information boxes:** You can click on a task node to view information about it in a box. Whilst still on the Graph View page, you can click away from the graph in the blank surrounding area & it will disappear. But if you change page  whilst a box is still up, it will stay up on the page & you cannot get rid of it (except, as above, by setting the browser to return to the root URL):

    sticky_node_!
    sticky_node_2

Due to other priorities I cannot investigate this yet, but I suspect it is a problem relating to the router paths.

kinow:
Closes #112

@MartinRyan I think I found what was going on. The `cy` instance was being created and re-used, but from looking at the [vue-cytoscape](https://github.com/rcarcasses/vue-cytoscape), I found that `afterCreated` actually takes a `cy` parameter.

That means we don't have to worry about the Cytoscape instance, and can abstract that away and just assume the instance provided is OK.

GIFrecord_2019-08-13_090721

I think this issue happened maybe because you didn't start with `vue-cytoscape` from the beginning, and adding it later probably you didn't need the `cy` instance as the code was working.

Hope that helps
Bruno

### DIFF
--- a/src/views/Graph.vue
+++ b/src/views/Graph.vue
@@ -44,7 +44,6 @@ import VueCytoscape from '@/components/core/Cytoscape.vue'
 import { mixin } from '@/mixins/index'
 
 const DATA_URL = 'simple-cytoscape-dot.7.js'
-let cy = {}
 let ur = {}
 const elements = []
 // eslint-disable-next-line no-unused-vars
@@ -340,7 +339,7 @@ export default {
       cytoscape.use(coseBilkent)
       cytoscape.use(klay)
     },
-    async afterCreated () {
+    async afterCreated (cy) {
       console.log('after created')
       const dagreOptions = {
         name: 'dagre',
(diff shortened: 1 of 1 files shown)


MartinRyan:
I've just created a pull request with my fix for the reload issue Bruno

hjoliver:
@MartinRyan @kinow - does this PR fix the same problem as #182 ?  If so, I guess we need to compare and choose (at first glance, this one is approx 100 times smaller code change).

MartinRyan:
size is not all that matters!

hjoliver:
> size is not all that matters!

True, but it helps!

kinow:
#182 fixes the issue and has a few more improvements (mainly removing code), so I think we should go with that one and close this one then?

Issue #21, opened 2019-02-10
Fix deprecation warning for escape function in vue.config.js


kinow:
The `vue.config.js` created by `vue-cli` some weeks ago appears to be using now a deprecated function `escape`. It suggests to use `encodeURI` or `encodeURIComponent` instead.

kinow:
Not necessary anymore. The new template created by `vue create $project` does not use this file any longer :tada:

PR #182, opened 2019-08-13
graph - debug reload

MartinRyan:
This fixes a problem with displaying the graph when returning to the graph screen.

### DIFF
--- a/src/styles/cytoscape/cytoscape-custom.css
+++ b/src/styles/cytoscape/cytoscape-custom.css
@@ -5,6 +5,13 @@
   z-index: 10;
 }
 
+#cytoscape { 
+  width: 100%;
+  height: 800px;
+  background-color: #222;
+  z-index: 100;
+}
+
 #app {
   font-family: 'Avenir', Helvetica, Arial, sans-serif;
   -webkit-font-smoothing: antialiased;
@@ -39,10 +46,6 @@
   position: fixed;
   top: 50%;
   left: 54%;
-}
-
-.tippy-popper {
-  /* transition: none !important; */
 }
 
 .collapsed-child {
--- a/src/views/Graph.vue
+++ b/src/views/Graph.vue
@@ -18,7 +18,7 @@
       <div class='cytoscape-navigatorOverlay'></div>
     </div>
     <b id='collapseAll' class='collapseAll' style='cursor: pointer; color: white'>collapse all</b> /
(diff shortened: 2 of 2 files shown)


src/styles/cytoscape/cytoscape-custom.css
```diff
@@ -5,6 +5,13 @@
   z-index: 10;
 }
 
+#cytoscape { 
+  width: 100%;
+  height: 800px;
+  background-color: #222;
+  z-index: 100;
```
kinow: Even though you have z-index here, note that #183 happens already with the code on `master`

src/views/Graph.vue
```diff
@@ -46,6 +46,7 @@ import { mixin } from '@/mixins/index'
 const DATA_URL = 'simple-cytoscape-dot.7.js'
 let cy = {}
 let ur = {}
+const config = {}
```
kinow: I think if we remove this `config` now, and the reference in `data`, it should still work? Looks like the change from line #21 was made this variable unnecessary?

src/views/Graph.vue
```diff
@@ -622,7 +562,12 @@ export default {
 
       // load graph data and run layout
       const { data: elements } = await updateData()
-      cy = await this.$cytoscape.instance
+      const stylesheet = await updateStyle()
+      cy = await cytoscape({
+        container: document.getElementById('cytoscape'),
+        elements: elements,
+        style: stylesheet.style
+      })
```
kinow: I think this is the actual fix for the bug? Instead of re-using the `this.$cytoscape` instance, we are now creating a new instance. Is that correct @MartinRyan ?

kinow:
Tested locally and it indeed fixed the issue :tada:

src/views/Graph.vue
MartinRyan: As far as I understand it this is getting a reference to the same cytoscape instance, not creating a new one

src/styles/cytoscape/cytoscape-custom.css
MartinRyan: I have a fix for this, it just involves z-indexes. I can create a pull request once this issue is resolved

src/views/Graph.vue
MartinRyan: yes agreed :-)

PR #188, opened 2019-08-14
Task failed

### LINKED ISSUE #173: task icons: failed & submit-failed too similar?
(Sorry to bring this up now after the icon implementation, but this is the first time I have noticed this, & better to raise it sooner rather than later when it could be more deep-rooted into the overall UI design.)

The task icons for 'failed' & 'submit-failed' as implemented, but still (though less notably) in the design mock-up, are barely distinguishable. Even upon zooming in so they are much larger than would be practicable for the viewing of realistically-sized suites by task, it is difficult to tell them apart:

scr-failed-states

At 100% zoom the perceptible difference becomes negligible.

I appreciate it is intentional that the two look similar, given they are both states of a variant of task failure, but I think they still need to be discernible at standard size. Can we make a change to one or both to accomplish this?

The natural approach would be to increase the size of the inner circle for the 'submit-failed' icon, but even then I am not convinced it would be distinct enough. I would personally like to change the nature of the 'submit-failed' icon to be different in some other way, by not using the dot & instead for example:

* rotating the ``x`` cross 45 degrees to a plus-style cross (``+``); or
* adding a few lines to make an hourglass-like symbol, like [this](https://en.wikipedia.org/wiki/Extinction_symbol#/media/File:Extinction_Symbol.svg); or
* making the ``x`` cross legs much shorter, & to contrast make them fill out all the way to the outer circle for the 'failed' icon.

oliver-sanders:
Proposal to close #173 (because it's faster to code than draw for this one)

* Visually clarify that succeeded and failed are both "completed" states.
* Visually distinguish failed from submit-failed
* Remove hub (circle) from submit-failed icon (because it's just messy)

Idea credit @hjoliver 

Screenshot from 2019-08-14 10-30-09

### DIFF
--- a/src/components/cylc/Task.vue
+++ b/src/components/cylc/Task.vue
@@ -39,14 +39,14 @@
           stroke-dasharray="157"
           v-bind:style="progressStyle"
         ></circle>
-        <!-- circle in the middle
+        <!-- circle in the middle (hub)
                * position in the middle
                * radius can be changed independently
         -->
         <circle
           id="hub"
           cx="50" cy="50"
-          r="12">
+          r="15">
         </circle>
         <!-- outer circle
                * position in the middle
@@ -203,7 +203,6 @@
 
             &.submit-failed {
                 @include outline();
-                @include hub();
                 @include cross($foreground);
             }
         }
--- a/src/components/cylc/Task.vue
+++ b/src/components/cylc/Task.vue
@@ -141,9 +141,9 @@
             }
         }
 
-        @mixin cross() {
+        @mixin cross($colour) {
             #cross rect {
-                fill: $foreground;
+                fill: $colour;
             }
         }
 
@@ -196,14 +196,15 @@
             }
 
             &.failed {
+                @include disk();
                 @include outline();
-                @include cross();
+                @include cross($background);
             }
 
             &.submit-failed {
                 @include outline();
                 @include hub();
-                @include cross();
+                @include cross($foreground);
             }
         }
     }


hjoliver:
A post-approval comment.  Your white-cross-on-black takes my suggestion https://github.com/cylc/cylc-ui/pulls#issuecomment-521120410 to its logical task-completed limit.  That does make it very clear that execution-failed is a "completed" state like succeeded, as you say. 

But maybe it's still worth considering a black cross on top of the executing/running radial progress bar. That would show more clearly that the job failed during execution (particularly if we can put the progress bar in the right place; but maybe that's difficult).  On the downside, that's a busier-looking icon. 

Thoughts?

oliver-sanders:
I did think about that but thought it would be to busy.

Screenshot from 2019-08-15 09-36-43

```diff
--- a/src/components/cylc/Task.vue
+++ b/src/components/cylc/Task.vue
@@ -196,9 +196,10 @@
             }
 
             &.failed {
-                @include disk();
                 @include outline();
-                @include cross($background);
+                @include hub();
+                @include progress();
+                @include cross($foreground);
             }
 
             &.submit-failed {

```

hjoliver:
Fair enough, I think I agree.

sadielbartholomew:
> But maybe it's still worth considering a black cross on top of the executing/running radial progress bar.
>
> > I did think about that but thought it would be to busy.

There are benefits to both alternatives. However I agree that Hilary's suggestion, as shown when implemented in the screenshot above, looks a bit too busy, & awkward with the overlay, if even it does include more information as I described for its main advantage in per https://github.com/cylc/cylc-ui/issues/173#issuecomment-521228997...

> Visually clarify that succeeded and failed are both "completed" states.

... but this is the main benefit for me, achieving parity with the succeeded icon due to the background colour. So I am happy to go with this suggestion.

Thanks for the quick response to the Issue raised, & it's a good sign that the design for that icon could be revised by changing so few lines of code!

PR #195, opened 2019-08-21
Remove border from: Header, Sidebar, and use floating style for NavDrawer

### LINKED ISSUE #193: Remove borders in-between divs like sidebar, toolbar, and main panel
See comment from https://github.com/cylc/cylc-ui/issues/145#issuecomment-522875938

We have a thin border (3px I believe, gray-300 color from Material UI) between certain elements, following the initial sketch from #87 .

This issue is for discussion on removing these borders, and having a more flat layout. That would give us a bit more of space, and make the layout "thinner". But there are probably counterarguments for keep using the borders.

Cheers
Bruno

kinow:
Closes #193 

@hjoliver here's a preview

image

(strange dark thin border at the bottom is the XFCE menu, not part of the UI)

### DIFF
--- a/src/components/core/Drawer.vue
+++ b/src/components/core/Drawer.vue
@@ -4,6 +4,7 @@
     v-model="inputValue"
     app
     persistent
+    floating
     mobile-break-point="991"
     width="260"
     clipped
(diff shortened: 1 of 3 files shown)


kinow:
Not sure whether 1 review is fine or not, but added @hjoliver as he requested this change :+1:

hjoliver:
I think one review will do, because we got some agreement, and it's a trivial change that's easy to revisit later if necessary.

PR #196, opened 2019-08-21
Remove unused functions in Dashboard.vue

kinow:
Had noticed some days ago that the `Dashboard.vue` when displayed always resulted in some console errors.

```
Error in mounted hook: "TypeError: Cannot read property 'children' of null"
```

This was due to some function I left from the initial implementation, when I was working on that graph-like display, with connected dots.

We decided not to add that for now, but I forgot to remote the functions. Sorry.

One review should be enough, as this code is not used.

### DIFF
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -137,10 +137,6 @@
 import { mixin } from '@/mixins/index'
 import { mapState } from 'vuex'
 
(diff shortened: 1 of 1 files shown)


kinow:
These errors while trivial, get in the way when you are working on code and trying not to add more errors. Thanks!!!

PR #197, opened 2019-08-21
Remove duplicate Dashboard route

kinow:
Fix browser console warning (yellow-ish console log, not really an Error and frequently ignored) for "[vue-router] Duplicate named routes definition: { name: "Dashboard", path: "" }".

I probably accidentally added a duplicate route while adding the Dashboard view I guess. Trivial change (click on the blue icon in the diff near the top of the change to see more of the file, and keep doing that until you can see another route that is also mapped to `Dashboard` view), one review should do :+1:

### DIFF
--- a/src/components/core/Drawer.vue
+++ b/src/components/core/Drawer.vue
@@ -26,7 +26,7 @@
       </v-list-tile>
 
       <v-list-tile
-          to="/dashboard"
+          to="/"
           :active-class="color"
           avatar
           class="v-list-item"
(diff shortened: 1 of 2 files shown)


kinow:
Warning as displayed in the browser:

image

kinow:
Oh, actually we had the duplicate routes, and were still using the Dashboard alias... so we had `/` -> Dashboard, `/dashboard` -> Dashboard... removed the alias (though I thought I had done it some time ago... :confused: )

kinow:
Now the menu also correctly displays the Dashboard as active.

Issue #192, opened 2019-08-19
Suite one used by generate script getting stalled


kinow:
It worked once for me this morning, out of near 10 runs. I tried changing the `sleep` times in the suite definition with no luck.

It is running for a long time, until it displays

```
2019-08-19T22:25:03Z INFO - [retrying.20000101T0000Z] -health check settings: execution timeout=None
2019-08-19T22:25:03Z CRITICAL - [retrying.20000101T0000Z] status=running: (received)failed/EXIT at 2019-08-19T22:25:03Z for job(02)
2019-08-19T22:25:03Z CRITICAL - [retrying.20000101T0000Z] -job(02) failed
2019-08-19T22:25:05Z WARNING - suite stalled
2019-08-19T22:25:05Z WARNING - Unmet prerequisites for sleepy.20000101T0000Z:
2019-08-19T22:25:05Z WARNING -  * checkpoint.20000101T0000Z succeeded
2019-08-19T22:25:05Z WARNING - Unmet prerequisites for waiting.20000101T0000Z:
2019-08-19T22:25:05Z WARNING -  * checkpoint.20000101T0000Z succeeded
2019-08-19T22:25:05Z WARNING - Unmet prerequisites for failed.20000102T0000Z:
2019-08-19T22:25:05Z WARNING -  * succeeded.20000102T0000Z succeeded
2019-08-19T22:25:05Z WARNING -  * eventually_succeeded.20000102T0000Z succeeded
```

![Screenshot_2019-08-20_10-25-36](https://user-images.githubusercontent.com/304786/63305000-4d5cf280-c339-11e9-993c-faf5c3ea7748.png)

Then nothing happens for several minutes. The last time I tried, it started at `2019-08-20T10:44:19` and I killed it at `2019-08-20T10:58:29`.

When I killed it, the log had several similar messages:

```
2019-08-19T22:56:47Z INFO - [client-command] pb_entire_workflow kinow@kinow-VirtualBox:cylc-uiserver
2019-08-19T22:56:52Z INFO - [client-command] pb_entire_workflow kinow@kinow-VirtualBox:cylc-uiserver
2019-08-19T22:56:57Z INFO - [client-command] identify kinow@kinow-VirtualBox:cylc-uiserver
2019-08-19T22:56:57Z INFO - [client-command] pb_entire_workflow kinow@kinow-VirtualBox:cylc-uiserver
[...]

Issue #145, opened 2019-07-24
Tree Component


oliver-sanders:
A component / style library for displaying nested trees for use by:

* [GScan](#139)
* [Tree View](#78)
* [Dot View](#79)

#### Requirements:

1. Nested trees
2. Arbitrary expansion / collapsing
3. Selection
   1. Single selection (e.g. the current workflow in GScan)
   2. Multiple selection (click and drag as in the current Tree view)
   3. Multiple non-contiguous selection (e.g. ctrl+click as in the current Tree view)
4. Activatable (e.g. in GScan the current workflow is "activated", that is highlighted in blue)
5. Tabulated data representation
   1. Ability to have columns of data (e.g. full tree view)
   2. Column sorting
   3. Show/hide columns
6. Can operate from the central data store
    * No requirement to reformat data to fit it into the component
7. Responsive Design
    * Desktop mode:  Collapsible nested tree
    * Mobile mode: Paged data with a bread-crumb trail
    * See [design document](https://github.com/cylc/cylc-ui/files/3362486/design.document.pdf)

#### Options:

| Component  | 1 | 2 | 3.1 | 3.2 | 3.3 | 4 | 5.1 | 5.2 | 5.3 | 6 | 7 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| Vuetify Treeview | :ok: | :ok: | :no_entry: | :no_entry: | :no_entry: | :ok: | :no_entry: | :no_entry: | :no_entry: | :no_entry: | :no_entry: |
| Vue Tree View | :ok:  | :ok: | :ok: | :no_entry: | :no_entry: | :no_entry: | :no_entry: | :no_entry: | :no_entry: | :no_entry: | :no_entry: |

#### Likely Conclusion

The tree view for Cylc is quite complicated and there likely isn't an off-the-shelf component which can do the job.

Can we borrow code from existing components to give ourselves a head start?

oliver-sanders:
Taking a quick look around I think we are unlikely to find a tree which will marry into our data structure as our data is kinda unstructured in a way, its about how you combine families, cycle points, etc which forms the tree in Cylc.

Unless anything crops up which ticks all the boxes I'd be inclined to nick the Vuetify CSS and apply it to a `<ul />` tree to give the composer complete control over the tree structure:

```html
<ul v-for="x in y">
  <li>
     <ul v-for="j in j">
        <li>
          ....
```

[...]

PR #198, opened 2019-08-21
Add host and progress information (mean elapsed time, execution time) to GraphQL query (mock and real)

### LINKED ISSUE #191: generate script used for mock data produces invalid JS file
I used the script once today successfully. But after changing the query and re-running it a few times, my `checkpoint.js` script got the header

```js
const checkpoint=const checkpoint=const checkpoint=const checkpoint={
```

and footer

```js
export { checkpoint }

export { checkpoint }

export { checkpoint }

export { checkpoint }

```

I think there must be something missing in the Shell script. I couldn't find anything mentioning to delete the file before re-running `generate`, but that may solve it too?

kinow:
The issue with `generate` and mock data was in my environment (see #192 and thanks @matthewrmshin !). Once fixed, managed to re-run the script using `master` from both Cylc UI Server and Cylc Flow, and update the mock data.

Also updated `Tree.vue`'s GraphQL query, and tested with offline mode, and with real data.

Part of #145 , just submitting before as it will take a bit more to finish that component's PR, and it will make it easier reviewing the PR.

Also closes #191 

Cheers
Bruno

### DIFF
--- a/src/services/mock/generate
+++ b/src/services/mock/generate
@@ -16,5 +16,5 @@ else
     exit 1
 fi
 
-sed -i "1 s/^/const checkpoint=/; s/${USER}/cylc/g" "${CHECK_FILE}"
+sed -i "1 s/^/const checkpoint = /; s/${USER}/cylc/g" "${CHECK_FILE}"
 echo -e '\nexport { checkpoint }' >> "${CHECK_FILE}"
--- a/src/views/Tree.vue
+++ b/src/views/Tree.vue
@@ -53,15 +53,21 @@ const QUERIES = {
           id
           name
           status
+          owner
+          host
+          port
           taskProxies {
             id
             state
             cyclePoint
             task {
+              meanElapsedTime
               name
             }
             jobs {
               id
+              host
+              startedTime
               state
               submitNum
             }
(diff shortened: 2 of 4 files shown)


src/services/mock/generate
```diff
@@ -16,5 +16,5 @@ else
     exit 1
 fi
 
-sed -i "1 s/^/const checkpoint=/; s/${USER}/cylc/g" "${CHECK_FILE}"
+sed -i "1 s/^/const checkpoint = /; s/${USER}/cylc/g" "${CHECK_FILE}"
```
kinow: For the linter.

src/views/Tree.vue
```diff
@@ -53,15 +53,21 @@ const QUERIES = {
...
           status
+          owner
+          host
+          port
           taskProxies {
             id
             state
             cyclePoint
             task {
+              meanElapsedTime
```
oliver-sanders: The Tree View isn't actually using this data yet though...

kinow: That's correct, I separated the pull requests. I thought this would be reviewed much quicker, and then either the Tree View PR could be updated, or it could be done in another PR.

oliver-sanders: OK, probably easier to merge PRs for the future so the subscription can't get out of date with the component?

oliver-sanders: (and sorry it's been hanging around for so long)

kinow:
Thanks y'all! Once the Tree Component is merged, I will submit a new PR to address progress in the icons for the tree view :+1:

Issue #162, opened 2019-08-01
Define where the data will be loaded for views like Dashboard


kinow:
Follow up from #149 

When browsing `/#/dashboard`, the user will initiate a Route to a view. The view is handled by a component, in this case `src/views/Dashboard.vue`.

At the moment, `Dashboard.vue` is initiating a polling to retrieve data from GraphQL for the query. The design sketch document has what is in `GScan.vue` in the place of what we have in the `Drawer.vue` at the moment.

![image](https://user-images.githubusercontent.com/304786/62259703-6c1b4800-b463-11e9-90fc-571c2b76ab22.png)

Once we move `GScan.vue` there, both components (`Dashboard.vue` and `GScan.vue`) will be displayed at the same time. This issue is for discussion around choosing when and how to fetch the data for components in Cylc UI.

kinow:
I thought the application do something like:

1. User requests `/`, so the router understands it needs to display the `Dashboard.vue`
2. Before doing that, the user is displayed with the loading bar (we could add status messages later like "Loading User Data", "Fetching Workflows")
  2.1. Fetch User data (needed for `Header.vue` and possibly other components or for auth checking)
  2.2. Fetch Workflows data via WebSocket (or polling for now)
3. Once the data has been fetched and the Vuex store is populated, the components are ready to be displayed, so we remove the progress bar and display the application

I thought (could be wrong, sorry) that there would be an initial "heavy/expensive" GraphQL query to fetch the workflows (2.2). But subsequent queries would receive only incremental updates.

With that in mind, I thought for this scenario, of the user viewing the dashboard, the query would bring the status too (as we gray out stopped workflows), and possibly the events (I think so? The latest message or something similar).

So that data would be used by `GScan.vue` to create the tree of workflows, and could be used by any number of components, like `Dashboard.vue`, which would summarize the stats and also display events.

kinow:
Similarly, when the user visits `/#/workflows/five`, if he adds several views with Tree, Graph, Gantt, Dot, etc, I wouldn't expect each component to interact with WebSocket/GraphQL/etc. Instead, I thought the data would be put in the store for these components to simply display - in a reactive way.

kinow:
Said all that, we can go with the current approach, as this discussion could drag over time, and I think we have a good momentum on Cylc UI, and would prefer to continue the development of new components/views/etc instead of focusing too much on lengthy discussions around this (not too hard to change it later). And as I could very well be wrong too hehe :)

[...]

PR #199, opened 2019-08-21
Fix toolbar displaying logic

kinow:
Since the `Toolbar` component we created, the navigation and routes changed a bit, and the logic got broken. Before, we only had `workflows` in the Vuex store when displaying a workflow. Now it's available throughout the application once one component that uses it has loaded it (part of the reason for #162).

This PR removes the `core/Toolbar.vue` toolbar component that came with the theme, as we are now using a custom `cylc/Toolbar` (based on the former), but with extra controls for the workflow being displayed.

Cheers
Bruno


kinow:
To review one could:

1. check out branch and run jupyterhub + ui
2. execute a long running workflow
3. open dashboard (toolbar should display only title)
4. go to workflows (toolbar should display only title)
5. open the running workflow screen (toolbar should display title and controls)
6. go back to dashboard (toolbar should display only title)
7. go to workflows (toolbar should display only title)

PR #200, opened 2019-08-21
Handle unknown statuses (custom icon, and console error log)

kinow:
Not really well-versed in SVG, but tried adding a path (via Inkscape) for a question mark (adapted from https://en.wikipedia.org/wiki/File:Icon-round-Question_mark.svg, license OK).

Then added an array (to be replaced by an `Enum` at some point) that checks if the current supported statuses include the status of the `Task`. If not, the question mark is displayed, and the status is logged to the browser console as error.

Cheers
Bruno


kinow:
Added @hjoliver as one of the reviewers as he suggested this enhancement, and @oliver-sanders as he knows the component code - and SVG as well.

@hjoliver here's a preview

Screenshot_2019-08-21_15-41-57

To review this branch, one could start JupyterHub + UI from this branch and run a workflow with some status not supported (e.g. retry/retrying, hold/held). Or use the mock data, which contains two invalid statuses (i.e. `NODE_ENV=offline npm run serve`).

src/components/cylc/Task.vue
```diff
@@ -102,6 +102,19 @@
...
...15.282,2.063 -4.699,1.346 -9.126,3.484 -12.876,6.219 -3.238,2.362 -6.333,5.391 -8.687,8.531 -4.159,5.549 -6.461,11.651 -7.063,18.687 -0.04,0.468 -0.07,0.868 -0.062,0.876 0.016,0.016 21.702,2.687 21.812,2.687 0.053,0 0.113,-0.234 0.282,-0.937 1.941,-8.085 5.486,-13.521 10.968,-16.813 4.32,-2.594 9.808,-3.612 15.778,-2.969 2.74,0.295 5.21,0.96 7.38,2 2.71,1.301 5.18,3.361 6.94,5.813 1.54,2.156 2.46,4.584 2.75,7.312 0.08,0.759 0.05,2.48 -0.03,3.219 -0.23,1.826 -0.7,3.378 -1.5,4.969 -0.81,1.597 -1.48,2.514 -2.76,3.812 -2.03,2.077 -5.18,4.829 -10.78,9.407 -3.6,2.944 -6.04,5.156 -8.12,7.343 -4.943,5.179 -7.191,9.069 -8.564,14.719 -0.905,3.72 -1.256,7.55 -1.156,13.19 0.025,1.4 0.062,2.73 0.062,2.97 v 0.43 h 21.598 l 0.03,-2.4 c 0.03,-3.27 0.21,-5.37 0.56,-7.41 0.57,-3.27 1.43,-5 3.94,-7.81 1.6,-1.8 3.7,-3.76 6.93,-6.47 4.77,-3.991 8.11,-6.99 11.26,-10.125 4.91,-4.907 7.46,-8.26 9.28,-12.187 1.43,-3.092 2.22,-6.166 2.46,-9.532 0.06,-0.816 0.07,-3.03 0,-3.968 -0.45,-7.043 -3.1,-13.253 -8.15,-19.032 -0.8,-0.909 -2.78,-2.887 -3.72,-3.718 -4.96,-4.394 -10.69,-7.353 -17.56,-9.094 -4.19,-1.062 -8.23,-1.6 -13.35,-1.75 -0.78,-0.023 -1.59,-0.036 -2.37,-0.032 z m -10.908,103.6 v 22 h 21.998 v -22 z"
```
oliver-sanders: As a sidenote I am slightly concerned as to how efficient these icons are to render as each SVG contains all of the required elements. If it is a problem there are some things we could do to improve it.

This question mark is the most expensive to render component of the Task icon, but as the "unknown" state isn't an expected one we could potentially just switch this on/off `<g`> with a `v-if` to avoid rendering unnecessarily.

kinow: Good point. I have very little experience with rendering many SVG elements in a UI application. Last time I ran the app through the Firefox performance monitoring, it already raised the alarm for the time to render some of the pages, but I didn't dig much to investigate what was the problem(s).

PR #202, opened 2019-08-22
add preparing icon

### LINKED ISSUE #201: task status icon and naming
Supercedes #194 

**[UPDATE]: AMENDED** (groan!) - see below https://github.com/cylc/cylc-ui/issues/201#issuecomment-523696622

We semi-agreed yesterday that the following server-internal task proxy statuses can all be absorbed into a single task status in the UI, because they all come under the category *"prerequisites satisified, and queued for execution"*:
- **queued** (prereqs satisfied, still in an internal named queue)
- **ready** (prereqs satisfied, still in the sub-process pool queue, for job submission)
- **submitted** (job submitted, still in the batch system queue)

Under normal circumstances users do not need to care about the above distinctions, just that their task's prerequisites are now satisfied and the job is "queued for execution".

However:
- they can easily infer "submitted" from the additional presence of a job status icon and batch system job ID
- they can query a queued task to find out if it is in a particular named queue, or in the sub-process pool queue (but they should only need to do this if their named queues are strongly limiting, or if the subprocess pool is too small).

**Additional proposals**
- we call this queued-for-execution UI status icon **queued** (for obvious reasons)
- for descriptive purposes, if needed we can make the following distinction:
  - queued/*submitted* (has a job ID and job status icon)
  - queued/*submitting* (queued for job submission; in the server either internal-queued or process-pool queued).

- we should *change the names of the server-internal "queued" and "ready" statuses*, because "queued" clashes with the combined UI status, and "ready" is a highly ambiguous name. These are internal so we can be verbose:
   - queued --> **internal-queued** or **named-queued**? (the latter is probably clearer because of course the subprocess pool queue is also "internal").
   - ready --> **process-queued**

## UI task status icons and names

task-status-icons

## UI task and job status icons and names
task-job-status-icons

(Note the two red-label cases here are purely descriptive, no need for different task icons, because the job status icon clearly shows that the task has moved from an internal cylc queue to the external batch queue).

oliver-sanders:
close #201

The renaming of the ready/queued state(s) will have to come in a future PR after the change is made to Cylc Flow.

### DIFF
--- a/src/components/cylc/Task.vue
+++ b/src/components/cylc/Task.vue
@@ -39,6 +39,16 @@
           stroke-dasharray="157"
           v-bind:style="progressStyle"
         ></circle>
+        <!-- dot in the middle (small hub)
+               * position in the middle
+               * radius can be changed independently
+               * we can't just change the radius with CSS (SVG2 only).
+        -->
+        <circle
+          id="dot"
+          cx="50" cy="50"
+          r="8">
+        </circle>
         <!-- circle in the middle (hub)
                * position in the middle
                * radius can be changed independently
@@ -50,7 +60,7 @@
         </circle>
         <!-- outer circle
                * position in the middle
-               * let radius = 50% - (2*stroke-width)
+               * let radius = 50% - (stroke-width/2)
         -->
         <circle
           id="outline"
@@ -138,6 +148,12 @@
             }
         }
 
+        @mixin dot() {
+            #dot {
+                fill: $foreground;
+            }
+        }
+
         @mixin hub() {
             #hub {
                 fill: $foreground;
@@ -195,8 +211,15 @@
             }
 
             /* status */
-            &.waiting {
+            &.waiting, &.queued {
+                /* NOTE: queued tasks may acquire an xtrigger or similar */
                 @include outline();
+            }
+
+            &.ready, &.preparing {
+                /* TODO: by cylc 8.0.0 - remove ready state */
+                @include outline();
+                @include dot();
             }
 
             &.submitted {


src/components/cylc/Task.vue
```diff
@@ -39,6 +39,16 @@
           stroke-dasharray="157"
           v-bind:style="progressStyle"
         ></circle>
+        <!-- dot in the middle (small hub)
+               * position in the middle
+               * radius can be changed independently
+               * we can't just change the radius with CSS (SVG2 only).
```
oliver-sanders: **[svg rant]**

It's so tempting to do something like this:

```css
#hub {
    radius: 0;
}
&.submitted {
    #hub {
        radius: 8;
        transition: radius 0.5s;
    }
}
&.submitting {
    #hub {
        radius: 15;
        transition: radius 0.5s;
    }
}

```

But alas not. CSS3/SVG1.1 have limited coupling. Sadly the horrendusly delayed SVG2 is still in [draft status](https://svgwg.org/svg2-draft/)!

Doing something as simple as changing the radius of a circle requires JS. CSS3 animations are amazing so it's annoying but sadly we can't use them to animate the task status icons.

**[/svg rant]**

kinow: If using JS... I haven't had any case to use it, but Vue has a small API for animation/transitions: https://vuejs.org/v2/guide/transitions.html

In theory we could add hooks to the component lifecycle & listen on DOM operations (element added, or page loaded, or component created, etc), and instruct Vue to change statuses.

hjoliver:
> and checkpoint in "running" (similar, but the inner circle is larger, so I guess it's OK).

checkpoint has the "submitted" state, not "running" - right?

src/components/cylc/Task.vue
```diff
@@ -180,6 +196,12 @@
                 @include outline();
             }
 
+            &.ready, &.queued {
```
hjoliver:
```suggestion
            &.submitting {
```
Is there some reason not to do this now?  (I thought the CSS classes could reflect the icon name - "submitting" - and the server-side task states (ready and queued, but may change) could be mapped separately to the icons?)

kinow:
>checkpoint has the "submitted" state, not "running" - right?

I think its state is "running" - https://github.com/cylc/cylc-ui/blob/5405a5ba96586963df1e036f0726ad4ba40075c5/src/services/mock/checkpoint.js#L42-L55

Here's how I tested it (I could be missing something...):

1. `git fetch upstream refs/pull/202/head:pr-202 && git checkout pr-202`
2. `NODE_ENV=offline npm run serve`
3. Take a look at http://localhost:8080/#/workflows/one
4. Edited `src/services/mock/checkpoint.js`, looking for the first two tasks (so that they would appear in the screenshot) and changing their states. I don't remember changing checkpoint... but it's past 6PM Friday (and winter!!!), and my brain is going in sleep mode... so I could be missing something here :grimacing: 

Cheers
Bruno

src/components/cylc/Task.vue
oliver-sanders: Not sure how we want to go about this at the Cylc Flow end, might take a while before we see the "submitting" status coming through over the API.

hjoliver: We don't need to do that.  The server still has to distinguish the two finer-grained states (currently "ready" and "queued") that get combined at the UI into "submitting". So I think the server should just put its true states in the workflow summary data, and the UI can do what it wants with those (which we've decided is: map both server states into "submitting").

hjoliver:
> 

> >    checkpoint has the "submitted" state, not "running" - right?

> I think it has the "running" state

OK, I was confused because your screenshot shows task "checkpoint" has a `submitted` task status icon (large central dot) ... but I see now it has a `running` job icon.  Shouldn't it have the "radial progress bar" running status icon?

src/components/cylc/Task.vue
oliver-sanders: Simple solution!

Might lead to some documentation issues e.g. when the suite log says that a task is `waiting => queued`, or `cylc dump` output, DB, etc.

kinow:
>OK, I was confused because your screenshot shows task "checkpoint" has a submitted task status icon (large central dot) ... but I see now it has a running job icon. In that case, shouldn't checkpoint have the "radial progress bar" running status icon?

I am not sure if it has been implemented. I only noticed I had seen this icon near "checkpoint" while working on the Tree component, and that it was a bit similar to this new submitting icon. I need to catch up on what was decided for the icons now.

oliver-sanders:
> Shouldn't it have the "radial progress bar" running status icon?

We don't have progress in the API yet so all running jobs have a progress of 0 which means the progress bar is invisible.

When we have the progress field to play with we could probably make the minimum display value something like 5% so that a job that has just started running has a visible progress bar.

hjoliver:
Ah, OK - I just assumed that 0-% progress would be a vertical bar.  Sorry - my mistake!  Although, could we do that?  i.e. make the plain running icon have a narrow vertical bar, then the progress would get overlaid on that.

oliver-sanders:
> Ah, OK - I just assumed that 0-% progress would be a vertical bar

Yeah, we could do that, I think it's in one of the older sketches, just not implemented at present.

src/components/cylc/Task.vue
MartinRyan: a css pre processor such as stylus could also be used http://stylus-lang.com/docs/conditionals.html

oliver-sanders:
Blocked awaiting conclusion of https://github.com/cylc/cylc-admin/pull/47

hjoliver:
Sorry, intending to unblock soon!

hjoliver:
@oliver-sanders - 

> Blocked awaiting conclusion of [cylc/cylc-admin#47](https://github.com/cylc/cylc-admin/pull/47)

Done. So this issue can be unblocked, closed as won't-fix, and replaced with a new one to make the new `preparing` icon - OK?

oliver-sanders:
Will modify this PR.

hjoliver:
@kinow - I just requested a re-review from you (but not yet!) to avoid accidental 2nd approval and merge (the set of task statuses how they relate to the old states has been changed and finalized since Oliver did this PR, but I don't think he's updated it yet).

src/components/cylc/Task.vue
hjoliver: This conversation is now somewhat out-of-date due to the different task statuses that we finally settled on.

oliver-sanders:
> (probably just a two-minute change?)

Yep, just haven't got around to it yet, should do today!

oliver-sanders:
* Queued icon is now the same as waiting
  * We may choose another representation for queued tasks in the future e.g:
     * A queued icon
     * Something similar to an xtrigger label
* Preparing icon is the same as waiting but with a little dot in the middle

oliver-sanders:
Dammit, added a commit I didn't mean to.

hjoliver:
> Dammit, added a commit I didn't mean to.

Has that been un-added? (are we good to go?)

oliver-sanders:
Yep.

kinow:
I think this can be reviewed now then @hjoliver ?

hjoliver:
Yep :+1:

hjoliver:
> ready is not in that list, so instead of seeing the preparing icon, all I see if the "?". "queued" and "preparing" are also not there. "waiting" is there, but it only applies the outline. Should we fix that now, or move to a follow-up issue?

Hmm. yes.  I'm worried I've confused you - see this: https://github.com/cylc/cylc-ui/issues/201#issuecomment-550118005

So the valid UI states are now: waiting, expired, preparing, submitted, submit-failed, running, succeeded, failed (BUT with the slight more explicit/clearer "new names" in the tables given in the cylc-admin page).

The complication is that the back-end is not (yet) providing the data with those states, so you'll temporarily need to make the transformation. I.e. if the data says "ready" show the "preparing" icon.  If the data says "queued" OR "waiting" show the "waiting" icon. 

And anything not in the the new legal list PLUS queued and ready, should be shown as a question mark.

Does that make sense? :grimacing:

kinow:
> > ready is not in that list, so instead of seeing the preparing icon, all I see if the "?". "queued" and "preparing" are also not there. "waiting" is there, but it only applies the outline. Should we fix that now, or move to a follow-up issue?
> 
> Hmm. yes. I'm worried I've confused you - see this: [#201 (comment)](https://github.com/cylc/cylc-ui/issues/201#issuecomment-550118005)
> 
> So the valid UI states are now: waiting, expired, preparing, submitted, submit-failed, running, succeeded, failed (BUT with the slight more explicit/clearer "new names" in the tables given in the cylc-admin page).
> 
> The complication is that the back-end is not (yet) providing the data with those states, so you'll temporarily need to make the transformation. I.e. if the data says "ready" show the "preparing" icon. If the data says "queued" OR "waiting" show the "waiting" icon.
> 
> And anything not in the the new legal list PLUS queued and ready, should be shown as a question mark.
> 
> Does that make sense? grimacing

Ah, that's helpful. I was wondering why I had added that code that wasn't matching the component. Now I recall it was because of the values returned by the backend.

Happy to leave it like this, merge this PR, and then fix that later, as even if the PR was updated now to filter the right statuses, it wouldn't work until the backend is fixed - from what I understand.

Thanks!

PR #203, opened 2019-08-23
Tree Component

kinow:
A tree component for #145 (not closing, as there is still work to do for GScan).

This adds two new components, ` Tree` that holds `TreeItem`'s. A `TreeItem` may hold multiple `TreeItem`'s. This is based on how other components work (e.g. Vuetify TreeView).

It supports marking entries as active (single or multiple), changing CSS style when user hovers the title, but **this version is not using slots yet**. That means the only customization/parameterization available now are props.

The component does not use any other service, instead it expects the user to provide the data via props. This makes the component independent, and can be used with data from the central store Vuex.

It uses the existing `Task` and `Job` components #101. When you have a Cyclepoint, it will use the `Task` icon, and the old logic from Cylc 7 to choose the group state, based on the children states.

When a `TreeItem` with a `Task` is collapsed, it will show a summary of its jobs.

Unit tests included, using vue-test-utils. Commits organized per feature, and have also an example of how to use in the commit (`Tree2.vue`, removed in this PR).

It will require a few follow up issues, for slots, activate state, and to adjust the component so that it can be re-used to replace `GScan` (left side bar in our design).

Docs at the [Wiki](https://github.com/cylc/cylc-ui/wiki). I will continue documenting the `TreeItem` component next week, but the docs for `Tree` component are available already.

Cheers :vulcan_salute: 
Bruno


kinow:
Tested with `NODE_ENV=offline npm run serve`, and also with `npm run build:watch` plus `jupyterhub` in Cylc UI Server. Unit tests should be executed here by Travis soon too.

kinow:
Note: the requirement of using the data as-is returned from the GraphQL query was not possible, as we are using recursion for this component. So the data had to be changed. Initially, it was changed in a view, but it would have to be done again in each view that used the component (or even for the Dot view), so moved it to the Vuex action.

Whenever the `WorkflowService` updates the workflow data structure, we compute the hierarchical workflow data structure and set it as well in the global store.

kinow:
And another note on slots. With slots, we would be able to quickly use these components for the GScan visualization. However, slots cannot be handed down recursively (they are virtual nodes, so you cannot use them in the HTML `template` section of a Vue single file component).

I started adding slots yesterday (Thursday) around 3PM. Went home and to the gym to think how to implement it. Came back in the morning, and was near half of the way, when I gave up. It was getting quite long the new implementation, and I thought maybe `props` would do the trick.

Then I went as far as using the component in the `Tree` view (i.e. `/#/workflows/${workflow-name}`), and also using it in the sidebar (i.e. `Drawer.vue`). Had issues initially with #162 , as not all screens have the data necessary for component.

And then I realized that for the sidebar, the tree needs some extra customization for the active status. It must be active, when the user clicks on a workflow node (still haven't tested when we have `/part1/part2/workflow-name`, which would result in three nodes and extra hierarchy, only tested with a flat workflows list).

This is easier to achieve with slots... so in the end I may have another go at slots, in a few days. But first I really would prefer to get the component out for others to test and send feedback, while I think a bit more about the GScan part...  we also have the challenge of showing the Jobs of the Workflow, which won't work the same way as in the Tree view, because in the GScan we are not even displaying intermediary Cyclepoint and Task nodes...

kinow:
So, to review, just run it with either JupyterHub or with the mocked data, and have a read a the main issue (and comments if you really have time).

hjoliver:
Thanks @kinow :tada:

src/model/TaskState.model.js
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Cylc valid task states.
+ */
+class TaskState extends Enum {}
+TaskState.initEnum([
+  'RUNAHEAD',
+  'WAITING',
+  'HELD',
+  'QUEUED',
+  'READY',
```
matthewrmshin: These will need update soon, with recent changes and discussions.

kinow: Agreed, I was going to send this as a separate PR, but thought it was better to push the component as-is instead of trying to perfect it too much.

src/components/cylc/Tree.vue
```diff
@@ -0,0 +1,112 @@
+<template>
+  <!-- mx is margin on the x axis, and mb is margin-bottom, 3 is 3 our of 5 -->
```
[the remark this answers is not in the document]
kinow: Oh, nice catch, thanks!

src/model/TaskState.model.js
kinow: (later I will update the rest of the code to use this enum)

src/store/workflows.module.js
```diff
@@ -1,18 +1,108 @@
...
+ * @returns {[]}
+ * @private
+ */
+function _getWorkflowTree (workflows) {
+  const cycles = _getCycles(workflows)
+  const workflowTree = []
+  if (cycles.size > 0) {
+    for (const workflow of workflows) {
+      // add workflow minus taskProxies, with children
+      const simplifiedWorkflow = omit(workflow, 'taskProxies')
```
oliver-sanders: I'm guessing this makes a copy of the workflow minus the `taskProxies`.

Can we get away without creating a new object, could we just add a new field to the workflow?

src/store/workflows.module.js
```diff
@@ -1,18 +1,108 @@
...
+      }
+      workflowTree.push(simplifiedWorkflow)
+    }
   }
+  return workflowTree
 }
 
 const actions = {
   set ({ commit }, data) {
+    const workflowTree = _getWorkflowTree(data)
```
oliver-sanders: Maybe worth raising an issue for future work once we've got delta updates and websockets, this update can become incremental.

src/utils/tasks.js
```diff
@@ -0,0 +1,62 @@
+import TaskState from '@/model/TaskState.model'
+
+/**
+ * States used when comparing the parent with children nodes.
+ * @type {*[]}
+ */
+const orderedStates = [
```
oliver-sanders: Does the `TaskState` enum allow us to order the task states to avoid defining here?

src/utils/tasks.js
```diff
@@ -0,0 +1,62 @@
...
+  TaskState.READY,
+  TaskState.QUEUED,
+  TaskState.WAITING,
+  TaskState.HELD,
+  TaskState.SUCCEEDED,
+  TaskState.RUNAHEAD
+]
+
+/**
+ * States used when the parent is stopped.
```
oliver-sanders: Perhaps `s/stopped/completed/` ?

oliver-sanders:
Looks cool, exactly what the doctor ordered!

src/components/cylc/TreeItem.vue
```diff
@@ -0,0 +1,197 @@
...
+      return Object.prototype.hasOwnProperty.call(this.node, 'children')
+    }
+  },
+  created () {
+    this.$emit('tree-item-created', this)
+  },
+  methods: {
+    typeClicked () {
+      if (this.node.__type === 'task') {
+        // ok
```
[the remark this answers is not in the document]
kinow: Ops, nothing. Will remove it, good spot.

src/store/workflows.module.js
```diff
@@ -1,18 +1,108 @@
...
+ *
+ * @param workflows
+ * @returns {Map<any, any>}
+ * @private
+ */
+function _getCycles (workflows) {
+  const cycles = new Map()
+  for (const workflow of workflows) {
+    if (Object.prototype.hasOwnProperty.call(workflow, 'taskProxies')) {
+      console.log(workflow.taskProxies)
```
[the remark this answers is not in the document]
kinow: Removed.

src/components/cylc/TreeItem.vue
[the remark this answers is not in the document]
kinow: Removed.

src/utils/tasks.js
kinow: Fine by me, but I used stopped here because this states order was used when the `is_stopped` flag was set in the Python code: https://github.com/cylc/cylc-flow/blob/d66ae5c3ce8c749c8178d1cd53cb8c81d1560346/lib/cylc/task_state_prop.py#L101

Should I change it to completed?

src/store/workflows.module.js
kinow: Good idea! https://github.com/cylc/cylc-ui/issues/207

src/utils/tasks.js
kinow: Yup. When the `TaskState` enum is created, the array given to the `initEnum` function is saved to a read-only `Array`. Iterating the enum, is the same as iterating that array, so the values will be returned in order (tested [here](https://github.com/kinow/notebooks/blob/master/javascript/enumify/Enumify.ipynb)).

So we could remove `orderedStates`, but would have to keep `isStoppedOrderedStates`. Changing, added the following note to where the enum is created, so other devs will remember that maintaining the order of entries created in the enum is important:

```js
// NOTE: the order of the enum values is important to calculate the group states in the UI. Items at
// the top of the list have preference over the items below in the algorithm.
```

src/store/workflows.module.js
kinow: Sure, done! :+1:

kinow:
Feedback addressed, though left one item pending an answer from @oliver-sanders . Also rebased `master`.

src/store/workflows.module.js
```diff
@@ -1,18 +1,106 @@
...
+          state: '',
+          children: [],
+          __type: 'cyclepoint'
+        }
+
+        const childStates = []
+
+        for (const taskProxy of workflow.taskProxies) {
+          if (taskProxy.cyclePoint === cyclePoint) {
+            const simplifiedTaskProxy = omit(taskProxy, ['jobs', 'task'])
```
oliver-sanders: Same as with Workflows can we add these fields direct to the original TaskProxy rather than copying it?

src/utils/tasks.js
oliver-sanders: OK, not really important. If I understand the code correctly then I think "completed" is correct. At the moment the `isStopped` argument is always `false` anyway.

src/store/workflows.module.js
kinow: Done, edited previous commit to include this change too. No issues running the app again in the offline mode.

tests/unit/components/cylc/tree.vue.spec.js
```diff
@@ -0,0 +1,88 @@
...
+// eslint-disable-next-line no-unused-vars
+import * as vuetify from '@/plugins/vuetify'
+import Tree from '@/components/cylc/Tree'
+import { simpleWorkflowTree4Nodes } from './tree.data'
+
+const cycles = new Map([
+  ['user/workflow1', new Set([
+    '20100101T0000Z'
+  ]
+  )]
```
sadielbartholomew: The closing parenthesis is on the wrong line for logical grouping here, which made me think there was a syntax error initially.

tests/unit/components/cylc/tree.data.js
```diff
@@ -0,0 +1,51 @@
...
+        ]
+      }
+    ]
+  }
+]
+
+const simpleWorkflowNode = simpleWorkflowTree4Nodes[0]
+const simpleCyclepointNode = simpleWorkflowTree4Nodes[0].children[0]
+const simpleTaskNode = simpleWorkflowTree4Nodes[0].children[0].children[0]
+const simpleJobNode = simpleWorkflowTree4Nodes[0].children[0].children[0].children[0]
```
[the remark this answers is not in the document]
kinow: Happy to update it you think it is necessary. I am normally more lenient (to not say lazy 🤭) with test code.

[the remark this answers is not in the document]
kinow: Updated with the simpler

```
const simpleCyclepointNode = simpleWorkflowNode.children[0]
const simpleTaskNode = simpleCyclepointNode.children[0]
const simpleJobNode = simpleTaskNode.children[0]
```

approach (thanks!)

tests/unit/components/cylc/tree.vue.spec.js
kinow: Fixed! Though simply moving the `)]` back one indentation level breaks the linter. Had to move the `)` up, and separate from the `]`. Now it looks like:

```js
const cycles = new Map([
  [
    'user/workflow1',
    new Set([
      '20100101T0000Z'
    ])
  ]
])
```

kinow:
Rebased :+1:

hjoliver:
@oliver-sanders and @matthewrmshin - we need to get this in pretty quickly for the alpha-1 release this month.  @kinow will add the additional job data (start-times etc.) at lowest tree-expand level in a follow-up PR (also pre-release).

src/components/cylc/Tree.vue
```diff
@@ -0,0 +1,112 @@
...
+<script>
+import TreeItem from '@/components/cylc/TreeItem'
+
+export default {
+  name: 'Tree',
+  props: {
+    workflows: {
+      type: Array,
+      required: true
+    },
```
MartinRyan: You may want to add one empty line between multi-line properties https://vuejs.org/v2/style-guide/#Empty-lines-in-component-instance-options-recommended

kinow: Hadn't seen that part about styling in Vue docs, thanks. Do you think we need that here? It looks fine for me, but I use PyCharm on a large screen. It seems like they say either way is fine, but spaces may help devs with other editors such as Vim (paging @hjoliver as I think he uses vim?).

hjoliver: I use vim and/or VS Code; no strong opinion on this at this stage!

kinow: >no strong opinion on this at this stage!

Same here. We may have to adjust the other components to follow this pattern too. But perhaps we can do it later? Unless it is difficult to other to read the code, then happy to update this PR with an extra commit for styling.

src/store/workflows.module.js
```diff
@@ -1,18 +1,104 @@
...
+ *
+ * @param workflows
+ * @returns {Map<any, any>}
+ * @private
+ */
+function _getCycles (workflows) {
+  const cycles = new Map()
+  for (const workflow of workflows) {
+    if (Object.prototype.hasOwnProperty.call(workflow, 'taskProxies')) {
+      if (!cycles.get(workflow.id)) {
```
matthewrmshin: Is `.id` some kind of JavaScript built-in? (It gets syntax highlighted in blue here.)

src/utils/tasks.js
```diff
@@ -0,0 +1,42 @@
...
+  TaskState.RUNAHEAD
+]
+
+/**
+ * Gives a single state, based on a list of states of children nodes.
+ * @param childStates {Array} children nodes
+ * @param isStopped {Boolean} whether the parent node is stopped or not
+ * @returns {null|*} a valid Task State, or null if not found
+ * @link @see https://github.com/cylc/cylc-flow/blob/d66ae5c3ce8c749c8178d1cd53cb8c81d1560346/lib/cylc/task_state_prop.py
+ */
```
matthewrmshin: I guess this fixes us to a given version of cylc-flow?

src/store/workflows.module.js
kinow: Not that I am aware of. I think it is just the syntax highlight library getting confused.

src/utils/tasks.js
kinow: Yup, I think @hjoliver sent me the link on Riot to the the file in a branch for 7.x. Then I used it to write the JS code, and pressed "y" on my keyboard while focusing on the GitHub UI. That creates this link to the specific commit.

src/components/cylc/TreeItem.vue
```diff
@@ -0,0 +1,194 @@
...
+      }
+      return styles
+    },
+    nodeClicked (e) {
+      this.$emit('tree-item-clicked', this)
+      if (e.ctrlKey) {
+        console.log('Clicked with CTRL!')
+      } else {
+        console.log('Clicked without CTRL!')
+      }
```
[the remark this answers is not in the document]
kinow: Good catch. This was actually for the selection, to choose one or multiple, but I didn't implement the complex case where you can select multiple nodes, between different heights/distances.

And actually, I implemented the ugly way (but the one I have in my memory). I believe Vue has another special way, something like `@click.ctrl=methodWithCtrl` which is the equivalent of `@click=method`, but only fires when the CTRL key is used too.

Better remove for now to avoid any confusion. Thanks :+1:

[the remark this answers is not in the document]
kinow: Done.

src/components/cylc/TreeItem.vue
```diff
@@ -0,0 +1,194 @@
...
+      <!-- the node's left icon; used for expand/collapse -->
+      <v-flex
+        shrink
+        v-if="hasChildren"
+        class="type"
+        @click="typeClicked"
+        :style="getTypeStyle()"
+      >{{ isExpanded ? '&#9661;' : '&#9655;' }}</v-flex>
+      <!-- the node value -->
+      <v-layout @click="nodeClicked" row wrap v-if="node.__type === 'cyclepoint'">
```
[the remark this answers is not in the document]
kinow: IMO yes. Though I am not sure how to do that in a simple way. So would move it for later. Other components (e.g. Task, Job) may benefit of this improvement too. We have one `Enum` for the Task States, but that's hardcoded too.

Perhaps another Enum, or a constant file somewhere. Will add a `TODO` so that I don't forget to look into it later and raise an issue and/or a fix for it.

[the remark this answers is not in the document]
kinow: Done

src/components/cylc/Tree.vue
```diff
@@ -0,0 +1,112 @@
...
+    onTreeItemCreated (treeItem) {
+      this.treeItemCache.add(treeItem)
+      if (treeItem.isExpanded) {
+        this.expandedCache.add(treeItem)
+      }
+    },
+    onTreeItemClicked (treeItem) {
+      if (this.activable) {
+        if (!this.multipleActive) {
```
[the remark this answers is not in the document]
kinow: I think there are a few statements after the `if (!this.multipleActive)`, that are executed only for the `if (this.activable)`. I think - if I recall correctly - I did in that order to clear/empty the cache before adding. Link for the code mentioned: https://github.com/cylc/cylc-ui/pull/203/files#diff-d70e3541f48c5fdb3978c257a8726eceR103-R107

kinow:
>Some quick comments. Feel free to ignore.

Good comments - as always. I've pushed some small changes to adjust the code after your suggestions. And now will rebase, fix conflicts and push-force :+1: 

Thanks!

src/components/cylc/TreeItem.vue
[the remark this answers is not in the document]
kinow: Removed one-too-many lines, and broke activating nodes. Thankfully we have some unit tests covering that :pray: and Travis told me I had to fix these. Now it should be good to go.

src/components/cylc/Tree.vue
MartinRyan: > I use vim and/or VS Code; no strong opinion on this at this stage!

It's fine, a personal style choice really ;-)

PR #205, opened 2019-08-28
Update Vuetify and theme

### LINKED ISSUE #204: Update vuetify to 2.0.x
The material design theme we built the initial version on has been updated (about 1 month after the release of Vuetify 2? not too bad :)

https://github.com/creativetimofficial/vuetify-material-dashboard/blob/954b5451981ba0ebab80c7f11c6eeb3e19da6efb/package.json#L20

Soon we might be able to update the whole project and start using the latest Vuetify version.

kinow:
Closes #204 

I thought it would take longer, and be much harder, but updating Vuetify wasn't that bad. It involved several actions.

For each component used (we are probably using around 10 components of Vuetify?), we had to confirm we are not using any slot/prop/directive that was removed.

Or worse, confirm the behavior didn't change. 

## Some issues

(feel free to ignore this, this is more to log what was done)

- Updating the dependencies was trick (hence more than 1 commit for that), as some dependencies required other dependencies (Vuetifyjs requires `sass`, not `node-sass`, there is a new plugin to load Vuetifyjs component on the fly, etc).
- Vuetify is not being loaded directly as a plugin in the theme, now it's imported in the main application and initializes its configuration and tells Vue to use it.
- I've removed the theme that comes with Vuetify (what gives the green color to some elements/links), as we will have our own themes in Cylc.
- Several views were broken after the dependencies were updated. But fixing the views depended on having the component fixed first.
- Some components changed the name of directives and parameters. The IDE alerted most of these.
- But some components also changed how they behave. A data table, for instance, now requires a header. It used to work fine with only the data if you hid the headers (`Dashboard.vue` displays tables with no headers).
- The theme (material-dashboard by Tim Creative) removed the custom SCSS, and now we are using the Vuetify styles directly. Better IMO, but updating it was no fun.
- I modified a few files from the theme when the project was still a prototype, and only started to organize things with SCSS after others started working on the project. That is one of those issues that eventually come back to bite you, as I had to re-apply my changes now to SCSS, but finding them and then translating to Vuetify 2 was much harder than if I had them in a single place.

I tested it in the mode offline and with real data. In case some gets broken, it is very likely we should be able to locate what broke it in the following way:

- Is it broken because it is missing some dependency? Look at the initial commit that changed dependencies (the other are related to this oen)
- Is it broken in a specific view or component? Look at the commits related to that file (some are for HTML elements, others for the JS code that needed to be updated, or even for `<style>` tags that were moved to SCSS or updated)

Happy to rebase/fix this PR is others are merged first :+1: To review I think the easiest way is to run the application and confirm no error logs in the browser console, and that it looks OK.

Bruno


kinow:
Oh, if a reviewer wants to run it locally, remember to `npm install` first too, as the node modules are different here. Which is good, as this means we can now update all dependencies, and if there was a security issue in Vuetify 1.5 we would be OK (not very likely IMO).

kinow:
Hmmm, the linter passes locally with `npm` and in my IDE, but fails on Travis... :confused:

kinow:
The updates also brought a more recent linter, that failed in a file that was not touched in this PR. Already fixed.

But now unit tests are failing with errors loading sass files... :confused:

kinow:
A-ha! Learned more about Webpack and also acquired some troubleshooting skills, but it wasn't much fun. Two days to update to Vuetify, and one whole day investigating this webpack issue :unamused: 

Build passing, re-built it locally to have a look at the `npm run serve`, and everything looks OK. Ready for review.

kinow:
10k lines? I hadn't noticed it was this many lines haha. Thanks!!!

kinow:
Will fix conflict `package-lock.json` in a minute.

kinow:
Rebased onto `master` and fixed conflicts.

matthewrmshin:
This branch is in conflict.

kinow:
Oh, had forgotten about this one. I will update it later. The tree component was written in vuetify 1.x, so will update and test it, the update the PR.

kinow:
Conflict fixed, but I still need to confirm that the tree component will work. After that it will be ready for review again :tada:

kinow:
There was an issue with `package-lock.json` after merge. It retained the old result of `npm install` pre rebase, even though `package.json` had the new dependencies. Running `npm install` fixed it.

But it also brought a new version of the linter, which found a new issue in the `TreeItem.vue`, which is now fixed.

Then, running in offline mode first, it immediately crashed with a SASS issue.

```
@import "../../styles/material-dashboard/colors";
       ^
      Can't find stylesheet to import.

```

image

I had zero recollection of how to fix it :astonished: but luckily found the commit where I had done it before: https://github.com/cylc/cylc-ui/commit/ae0b57290bcfa185a570bb3a4cba5dc4bf158045 (organizing commits per change helped, though I could also have grep'ed a single commit I guess).

So that was fixed in the following way:

1. import the Vuetify 2.x style `@import '~vuetify/src/styles/styles.sass';`
2. used `map-get($grey, 'lighten-3');` which was used in the previous commit to replace `$grey-100` (old variable from Vuetify 1.x)

Then the component was rendered, though slightly broken:

image

Fixing this part was quite simple. First looked at the Vuetify components used: `v-layout`, `v-flex`. So just had to look at the previous commits and find how they were fixed. Or so I thought. Actually the `type` class was applying margin-right 10px, but that was not enough.

So instead of increasing to 15px or 20px, I used `mr-4` (margin-right number 4 from Vuetify, which seemed to work well).

Finally, adjusted the existing margin classes for the Tree component, to display the tree with better spacing. And now I think it is ready again for review (as long as Travis CI is happy).

kinow:
What it is looking like for me right now:

image

kinow:
Now broken with `sass` again (was fixed before with a webpack null-loader in this commit https://github.com/cylc/cylc-ui/pull/205/commits/3ca504bc98478ad316721aaaebabad0ecb95a6aa) :angry:

kinow:
The solution was to remove the rules first for `sass` and for `scss`, then add new ones with the `null-loader`. Ready for review.

kinow:
I have 50% of the #228 leaf node done. Just realized it is using Vuetify 1.5... meaning this PR will be yet again in conflict. Depending on which one gets merged first, this will require again to redo one or the other 😞

hjoliver:
@kinow - I think we need to get this massive PR in ASAP to avoid you running into the same problem again with every other PR. I don't think it needs a full proper code review, given that it just updates some underlying stuff that we depend on, and that we're still early in the process (once cylc-8.0.0 is out, we'll have to be more careful).  @matthewrmshin - are you wiling to upgrade your eye-balling to an approval on that basis?  (And others in @cylc/core - are you happy with this too?).

kinow:
@hjoliver +1 it's a bit boring the whole rework, but the worst part is the time spent on these conflicts/etc, which could be used to fix more things for the alpha release :disappointed_relieved:

kinow:
(oh, and I very much prefer to interface after the update, as it fixes the annoying green/blue colors on links/menu items, so the UI for the alpha release will look even nicer)

src/components/helper/Offset.vue
```diff
@@ -9,39 +9,43 @@
...
   }
-}
 </script>
 
 <style>
 .v-offset {
   margin: 0 auto;
   max-width: calc(100% - 32px);
   position: relative;
+  z-index: 1;
```
wxtim: Any particular reason for adding this? And for the choice of a low value?

matthewrmshin:
Branch in conflict!

src/components/helper/Offset.vue
kinow: Not sure. This file is from the theme (creative tom vuetify dashboard). So I just copied it as-is (I should have commented which files needed more attention in the review - sorry)

kinow:
Conflict fixed locally, though the leaf node now is a bit clumsy 

image

Let me fix this before pushing the updated version.

kinow:
Hmm, leaf node still broken. Will have to dig deeper at their style changes to see how to get it working nicely again with Vuetify 2

kinow:
Had to take a different approach, as the elements in Vuetify 2.x seemed to behave different. Moving it to the left with `margin-left: -00px` would reflect on the component `width` (as in the screenshot attached here earlier).

So instead, had to remove margin/padding from the whole content area, add this spacing around the nodes, making sure it still matched the design sketch, and then simply add a div for the leaf node, without altering anything (well, except the leaf's triangle, which now gets a positive margin-left to be under the job icon).

Also changed the font weight to "light", as otherwise it was calling too much attention IMO (ps: I believe we won't pass the contrast tests for accessibility with light grey text). One nit pick against Vuetify-2 was their change from naming text colors like "text-gray` now to `gray--text` 😕 

Ready for review, @hjoliver if you have time it would be great if you could double-check the leaf node, and tree view, just to make sure I didn't break anything, please 😬

hjoliver:
That's 3 approvals, as @matthewrmshin and @wxtim pair reviewed it.

Issue #155, opened 2019-07-30
Minimal UI tour of major components for new users


sadielbartholomew:
We could include a basic optional "tour"* of pop-up tooltips (or similar) providing descriptions of major components, to guide users around the UI when we detect they are using it for the first time.

Some users no doubt will not like this, & not want to use it, so a first step in the "tour" which I envision as popping up by default should be a welcome message & initiation question such as *'Hi there \<username\>. For your first time using the Cylc UI, would you like a quick guide around the components?'*.

Then the tour would provide descriptions such as *'This is how you can add a workflow view to the page.'*, *'You can add as many views as you like'*, etc. (obviously keep it informational & avoiding patronising users with obvious commentary). Perhaps we could ask to begin with whether the user used the old Cylc GUIs, & if so provide context to relate the two, e.g. *'This is the equivalent of the old GScan, but we have improved it such that... \<brief description of differences\>'*.

There are potential Vue components to use (partially) to implement this, though we could implement it ourselves via events without need for a special component:
* [vue-tour](https://github.com/pulsardev/vue-tour) with GIF example on the README there;
* [vue-popper](https://github.com/RobinCK/vue-popper) with demo [here](https://jsfiddle.net/Robin_ck/n840tvp2/).

hjoliver:
I like this idea, but unless it's super-easy to do I'd suggest not putting much effort into it until we're sure we have a viable product in terms of all the critical functionality.  Hopefully it's the sort of thing that can be bolted on later, or would it need to be designed in from the start?

kinow:
> Hopefully it's the sort of thing that can be bolted on later, or would it need to be designed in from the start?

There are libraries (@sadielbartholomew linked some) that make it easy to add to any application later. Unless we want something really sophisticated, and then we would have to create it from scratch.

So +1 for low priority, and for adding it later with an existing library.

sadielbartholomew:
> I like this idea, but unless it's super-easy to do I'd suggest not putting much effort into it until we're sure we have a viable product in terms of all the critical functionality

> So +1 for low priority, and for adding it later with an existing library.

I also agree, it is certainly not something to prioritise or necessarily have in initial versions. But it would be nice to include eventually, I think.

> Hopefully it's the sort of thing that can be bolted on later

[...]

Issue #11, opened 2019-02-04
Upgrade to Vue 2.6


kinow:
Just released https://medium.com/the-vue-point/vue-2-6-released-66aa6c8e785e

Time to upgrade 🎉

kinow:
Impressive.

```bash
$ npm update
npm WARN deprecated circular-json@0.3.3: CircularJSON is in maintenance only, flatted is its successor.
npm WARN eslint-config-standard@12.0.0 requires a peer of eslint@>=5.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN eslint-config-standard@12.0.0 requires a peer of eslint-plugin-node@>=7.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN eslint-config-standard@12.0.0 requires a peer of eslint-plugin-promise@>=4.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN eslint-config-standard@12.0.0 requires a peer of eslint-plugin-standard@>=4.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.7 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.7: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

+ vue-template-compiler@2.6.3
+ @vue/cli-plugin-eslint@3.4.0
+ @vue/cli-plugin-babel@3.4.0
+ vuetify@1.5.0
+ @vue/cli-service@3.4.0
+ vue@2.6.3
added 21 packages from 12 contributors, removed 40 packages, updated 58 packages, moved 10 packages and audited 24176 packages in 54.067s
found 0 vulnerabilities
```

And everything is still working! Well done Vue.js devs!

PR #215, opened 2019-09-05
guide: add beginnings of a user guide

oliver-sanders:


wxtim:
I notice that the "special task states" doesn't scale to full width on smallish screens. On very old phones it might look even worse:
Screen Shot 2019-09-05 at 14 52 26

sadielbartholomew:
#155 is the Issue I mentioned in the offline discussions on this just now. In short, ideally we can incorporate user guidance into the UI via means of e.g. a tour or (carefully designed & placed) tooltips or similar, rather than have it as a separate tab or page, where it essentially becomes documentation in my eyes. And I don't think any good UI should need documentation!

hjoliver:
@sadielbartholomew - 

> And I don't think any good UI should need documentation!

Maybe you're right in principle, but I tend to think we should have a documentation page as well in case we fail to quite live up to that lofty goal, and besides I think at least some users might prefer a minimal documentation page with information on (e.g.) task state meanings, what the various views are good for, etc.

So I'd suggest merge this now, and add to it as we go, and consider removing it in the future if our UI turns out to be so staggeringly intuitive that the documentation isn't needed!  Is that OK with you?

src/views/Guide.vue
```diff
@@ -0,0 +1,213 @@
...
+                </td>
+              </tr>
+            </table>
+          </v-card-text>
+        </v-card>
+      </v-flex>
+
+      <v-flex
+        md5
+        xs8
```
hjoliver:
```suggestion
        xs12
```
Actually this seems to fix it.

hjoliver:
Also: need a link from the dashboard page?

src/views/Guide.vue
```diff
@@ -0,0 +1,213 @@
+<template>
+  <v-container>
+    <h1 style="margin:0;">Guide</h1>
```
kinow: The margins from `h1`, `h2`, etc, are annoying at times. We can instead use `span` or `p`, and annotate with some typography class like `display-1` and `font-weight-thin`, etc.

So this gives similar result:

```
<p class="display-3 font-weight-thin mb-3">Guide</p>
```

(`display-3` is larger than `display-1`, there are multiple `font-weight-` classes, and `mb` stands for `margin-bottom`, you can also use `mx-`, `my-`, `mt-`, ditto for padding with `px` etc)

src/views/Guide.vue
```diff
@@ -0,0 +1,213 @@
...
+        md5
+        xs8
+      >
+        <v-card>
+          <v-card-title primary-title>
+            <p class="display-1 text--primary">Special Task States</p>
+          </v-card-title>
+          <v-card-text>
+            <div id="task-job-special-grid">
+              <div style="font-size: 2em">
```
kinow: This can be a Vuetify `List` with `avatar` and `three-line`. The avatar is to display an image to the left. And the `three-line` is just so the text is not wrapped on mobile display.

src/views/Guide.vue
```diff
@@ -0,0 +1,213 @@
...
+
+    div:first-child {
+      padding-right: 0.5em;
+    }
+
+    div > span:nth-child(3) {
+      color: $grey;
+    }
+  }
+</style>
```
kinow: Maybe we can use Vuetify [typography](https://v15.vuetifyjs.com/en/framework/typography) and [flex grid](https://v15.vuetifyjs.com/en/framework/grid) (will change a bit when we have vuetify 2)?

Removing the styles from here, and using the code from this PR https://github.com/oliver-sanders/cylc-web/pull/1

- Before & after diff desktop:

image

image

- Before & after diff mobile:

image

image

Might still need some adjustments, but that would be easier to be adjusted to Vuetify 2 too (WIP #11).

(Also added some review comments showing what I did basically)

kinow:
@hjoliver sorry, started reviewing in the morning and didn't see your review until I submitted mine.

>Also: need a link from the dashboard page?

There's a `TODO` for it somewhere I think. But if that's a new link in the dashboard, that will balance the links, and we will have 3 links on each side :grimacing:

hjoliver:
No need for sorry - you did a proper review! :+1:

oliver-sanders:
> I don't think any good UI should need documentation!

I know what you mean, a good UI should just be obvious, but the Cylc UI definitely needs explaining. At present most users are unaware of many of the useful features of the current Cylc GUI so there is definitely a documentation gap we need to cross.

There are good UIs which have documentation e.g. [GitHub](https://github.blog/changelog/2019-08-21-improved-visibility-for-related-pull-requests-in-issues/).

I don't think we should write a user manual for the GUI, perhaps "guide" is the wrong word but there are things like task-job separation which definitely need explaining.

The modern approach to GUI docs seems to be little bitesize chunks of easily digestible information which get slowly dribbled through to the user (very agile). As we add features I think we should write little "cards" to go along with the changes (like the ones in this PR) telling users about the new feature and how it works.

The "guide" page is just the place where these "cards" live. Users don't necessarily browse to this page for help, we could link to a "card" from within the GUI.

To re-use GitHub as an example they have a [chengelog](https://github.blog/changelog/) with "cards" containing information about new features, users get linked through to the changelog from the UI (they also have a blog for longer reads). The "Guide" page is a more "browsable" version of this changelog providing everything in one place for those motivated souls who want to find out more.

sadielbartholomew:
> So I'd suggest merge this (when it passes review), and add to it as we go, and consider removing it in the future if our UI turns out to be so staggeringly intuitive that the documentation isn't needed! Is that OK with you?

Yes, that's okay by me. Oliver clarified some of the points of the offline discussion we had in his comment above.

> I don't think we should write a user manual for the GUI, perhaps "guide" is the wrong word but there are things like task-job separation which definitely need explaining.

I agree that most of the aspects that will need explaining for certain relate to new facets such as task & job separation. And whilst these are represented by state icons etc., ultimately they are not pure UI aspects but Cylc logic ones.

> I think at least some users might prefer a minimal documentation page with all the necessary information in one place - e.g. task state meanings, what the various views are good for, etc.

Fair enough, & probably true, but in that case I would wonder why it should go in a tab of the UI & not in the documentation proper. In fact, I think as a UI tab we might risk wasting a significant amount of UI server resource, as users might open a dedicated tab just for reading the guide (or even multiple ones :worried:) & flick to-&-from it for advice? If necessary could we mitigate the server load for the guide endpoint in some way, maybe?

I'm not going to push for a different approach, but I'm raising some conerns :grin:! Certainly for now this is good enough way to incorporate the guidance content (which looks great so far, BTW).

src/views/Guide.vue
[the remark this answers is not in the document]
oliver-sanders: For reference the "top bar" should only display when we are viewing a suite.

I'm not sure where we want to fit the guide in, it's more similar to the dashboard than a suite...

oliver-sanders: So yes you can customise a `<p />` to behave as you would like `<hx />` to work, however, it's just as easy to customise `<hx />` to behave the way you want it to in the first place.

The big issue with using `<p />` or `<span />` for headings is that is is a problem for accessibility.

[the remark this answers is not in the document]
kinow: Got it. Let's fix the control bar display in 0.2

https://github.com/cylc/cylc-ui/issues/253

kinow: Oh good point. In that case, maybe this then?

`<h1 class='ma-0'>` 

`ma-0` is the class for no margins.

oliver-sanders: Ah, very nice!

src/views/Guide.vue
oliver-sanders: > Maybe we can use Vuetify typography and flex grid (will change a bit when we have vuetify 2)?

Any reason to change from using a table? Happy to use Vuetify components if they are higher quality, but in this case I don't think they are.

kinow: No special reason. I only used these Vuetify components because they translate to `flex`, `flex-wrap`, `nowrap`, and flex columns with `%`s, etc, and I was using them in the Tree view recently. I can try to achieve the same with a table :+1:

PR #217, opened 2019-09-09
Reset version to 0.1

### LINKED ISSUE #175: Organize releases in this repository
I was keeping track of the changes in the beginning of the project with some tags, to be able to go back in time after some complicated changes. I think these old tags can now be removed.

This is also going to be helpful on https://github.com/cylc/cylc-admin/issues/27. Not sure whether we should start on 0.1, or 1.0, or use alpha/beta, etc.

kinow:
Closes #175

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cylc-ui",
-  "version": "1.0.0",
+  "version": "0.1.0",
   "private": true,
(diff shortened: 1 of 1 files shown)


kinow:
@hjoliver would you have the time for a super quick review, please? One review should be enough IMO

PR #218, opened 2019-09-10
Update version in package-lock.json too

kinow:
@hjoliver forgot to include the change in the lock file for #175 . Just ran `npm install` so that it would update the file with the new version. Sorry, will try to remember it the next time.


kinow:
Thanks!

PR #223, opened 2019-09-13
Add function to compute percent progress for the Tree component

### LINKED ISSUE #220: Add progress to the Tasks in the Tree Component
For Cylc 7 GUI's, the progress is computed here: https://github.com/cylc/cylc-flow/blob/de7d938496e82dbdfb165938145670dd8e801efd/lib/cylc/gui/updater_tree.py#L248-L263

We want to have a similar behaviour in Cylc 8, and will adopt the same approach initially, tweaking it if we have any issues with DOM updates, styles, etc.

Now that the Tree Component is on master (#203), and the query has been updated to include `meanElapsedTime` and `startedTime` (#198), I think we are ready to update the component and display this information.

kinow:
Closes #220 

Implemented and with tests, but I am not sure if it is working or not :confused: 

For my workflow `five`, modified to run `script = sleep 2` for every task, and with a more distant stop point, it only displays progress as 0% or 100%.

I had a look at the code, and also at the response of the GraphQL queries... if I configured the script command to sleep for 2 seconds, the `meanElapsedTime` returned always 2 seconds.

Using that number with the old code ported to JS (using the job `startedTime` and the `Date.now()` time in JS), it looked to me like it had to always return 0 or 100 indeed.

Later, I changed the script to be: `script = "sleep $[ ( $RANDOM % 30 )  + 1 ]s`, in the hope the `meanElapsedTime` would be different. It was, at one point I was getting `12.4...`. But even then, it still returned 0% or 100%.

Any help to reproduce a scenario where the returned progress would be different than 0 or 100?

Here's the old code: https://github.com/cylc/cylc-flow/blob/de7d938496e82dbdfb165938145670dd8e801efd/lib/cylc/gui/updater_tree.py#L248-L263

Here's the code I ported to JS (which may have bugs, it's Friday 7PM, not the best time to write any code like this, but had some spare time...): https://github.com/cylc/cylc-ui/pull/223/files#diff-cef93d4e230a1d011a9de08fdf0a7b3cR52-R63

### DIFF
--- a/src/store/workflows.module.js
+++ b/src/store/workflows.module.js
@@ -41,6 +41,28 @@ function _getCycles (workflows) {
 }
 
 /**
+ * Compute percent progress.
+ *
+ * @see https://github.com/cylc/cylc-flow/blob/de7d938496e82dbdfb165938145670dd8e801efd/lib/cylc/gui/updater_tree.py#L248-L263
+ * @param {number} startedTime in milliseconds since 1970-01-01 00:00:00 UTC, e.g. 1568353099874
+ * @param {number} meanElapsedTime mean elapsed time (e.g. 13.5)
+ * @returns {number} the percent progress, e.g. 25 (meaning 25% progress)
+ * @private
+ */
+function _computePercentProgress (startedTime, meanElapsedTime) {
+  const now = Date.now()
+  // startedTime > now reportably possible via interaction with `cylc reset`
+  if (startedTime > now || meanElapsedTime === 0) {
+    return 0
+  }
+  const elapsedTime = startedTime + meanElapsedTime
+  if (now > elapsedTime) {
+    return 100
+  }
+  return 100 * (now - startedTime) / meanElapsedTime
+}
+
+/**
  * Compute a workflow tree where the root is the workflow node, followed by
  * the cycle points, and finally by task proxies.
  *
@@ -74,12 +96,19 @@ function _getWorkflowTree (workflows) {
             taskProxy.name = taskProxy.task.name
             taskProxy.children = []
             taskProxy.__type = 'task'
+            let startedTime = 0
             for (const job of taskProxy.jobs.reverse()) {
               job.name = `#${job.submitNum}`
               job.__type = 'job'
               taskProxy.children.push(job)
+              // we use only the highest submitNum startedTime
+              if (startedTime === 0 && job.startedTime) {
+                startedTime = Date.parse(job.startedTime)
+              }
             }
             simplifiedCyclepoint.children.push(taskProxy)
+
+            taskProxy.progress = _computePercentProgress(startedTime, taskProxy.task.meanElapsedTime)
 
             childStates.push(taskProxy.state)
           }
--- a/src/store/workflows.module.js
+++ b/src/store/workflows.module.js
@@ -99,7 +99,8 @@ function _getWorkflowTree (workflows) {
             taskProxy.children = []
             taskProxy.__type = 'task'
             let startedTime = 0
-            for (const job of taskProxy.jobs.reverse()) {
+            // the GraphQL query is expected to have `jobs(sort: { keys: ["submit_num"], reverse:true }) {`
+            for (const job of taskProxy.jobs) {
               job.name = `#${job.submitNum}`
               job.__type = 'job'
               taskProxy.children.push(job)
--- a/src/store/workflows.module.js
+++ b/src/store/workflows.module.js
@@ -1,6 +1,8 @@
 // raw GraphQL response data structure
 import { extractGroupState } from '@/utils/tasks'
 
+const STATES_WITH_PROGRESS = ['running']
+
 const state = {
   workflows: [],
   workflowTree: []
@@ -108,7 +110,9 @@ function _getWorkflowTree (workflows) {
             }
             simplifiedCyclepoint.children.push(taskProxy)
 
-            taskProxy.progress = _computePercentProgress(startedTime, taskProxy.task.meanElapsedTime)
+            if (STATES_WITH_PROGRESS.includes(taskProxy.state)) {
+              taskProxy.progress = _computePercentProgress(startedTime, taskProxy.task.meanElapsedTime)
+            }
 
             childStates.push(taskProxy.state)
           }
--- a/src/store/workflows.module.js
+++ b/src/store/workflows.module.js
@@ -45,21 +45,21 @@ function _getCycles (workflows) {
  *
  * @see https://github.com/cylc/cylc-flow/blob/de7d938496e82dbdfb165938145670dd8e801efd/lib/cylc/gui/updater_tree.py#L248-L263
  * @param {number} startedTime in milliseconds since 1970-01-01 00:00:00 UTC, e.g. 1568353099874
- * @param {number} meanElapsedTime mean elapsed time (e.g. 13.5)
+ * @param {number} meanElapsedTime mean elapsed time in seconds
  * @returns {number} the percent progress, e.g. 25 (meaning 25% progress)
  * @private
  */
 function _computePercentProgress (startedTime, meanElapsedTime) {
-  const now = Date.now()
-  // startedTime > now reportably possible via interaction with `cylc reset`
+  const now = Date.now() // milliseconds since 1970-01-01
+  // startedTime > now reportedly possible via interaction with `cylc reset`
   if (startedTime > now || meanElapsedTime === 0) {
     return 0
   }
-  const elapsedTime = startedTime + meanElapsedTime
-  if (now > elapsedTime) {
+
+  if (now > startedTime + meanElapsedTime * 1000) {
     return 100
   }
-  return 100 * (now - startedTime) / meanElapsedTime
+  return 100 * (now - startedTime) / (meanElapsedTime * 1000)
 }
 
 /**
--- a/src/components/cylc/TreeItem.vue
+++ b/src/components/cylc/TreeItem.vue
@@ -26,7 +26,7 @@
       </v-layout>
       <v-layout @click="nodeClicked" row wrap v-else-if="node.__type === 'task'">
         <v-flex shrink>
-          <task :status="node.state" :progress=0 />
+          <task :status="node.state" :progress="node.progress" />
         </v-flex>
         <v-flex shrink>
           <span class="mx-1">{{ node.name }}</span>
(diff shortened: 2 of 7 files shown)


kinow:
Question: can I skip calculating the progress based on the task state? While writing the code and testing the UI component, I realized that the progress bar is present only for certain states (where certain SVG classes are [applied](https://github.com/cylc/cylc-ui/blob/6788c81ce1aacc5906bbf29a2d0bffbabbf613c7/src/components/cylc/Task.vue#L210)).

So, does it make sense to calculate the progress for every task in the workflow, or could we limit it to the "progressable" tasks? Like `running`?

kinow:
The function is part of the Vuex workflow module. It is a private function to that module, not exported. It could be re-used for the Graph and for other components - if the component is not using the Vuex workflow module.

It receives a `startedTime` and `meanElapsedTime`, both JS `Number` types, and return another `Number` that should be between 0 and 100 :crossed_fingers:

matthewrmshin:
Yes, this is only relevant for tasks in the running state.

In the future, perhaps we can add n average queue time as well. ⌛️

hjoliver:
> So, does it make sense to calculate the progress for every task in the workflow, or could we limit it to the "progressable" tasks? Like running?

Yes as @matthewrmshin says, only running tasks need a progress bar.  It is supposed to show job execution progress only.  (Also noted in 3rd paragraph here, for the record: https://github.com/cylc/cylc-uiserver/issues/77#issuecomment-530260001)

hjoliver:
> Any help to reproduce a scenario where the returned progress would be different than 0 or 100?

@kinow - if you were still using "sleep 2" for the task job, that would just be so quick (in a 5 sec polling UI?) that you would only ever see 0 or 100 right?   You probably just need to do (e.g.) "sleep 30". (Or have I misunderstood?).

kinow:
@hjoliver 

>@kinow - if you were still using "sleep 2" for the task job, that would just be so quick (in a 5 sec polling UI?) that you would only ever see 0 or 100 right? You probably just need to do (e.g.) "sleep 30". (Or have I misunderstood?).

I tried with a sleep 20, and then with a random value. But I couldn't get it to give me any value that was not `0` or `100`. But will try again in a few minutes 👍

hjoliver:
In that case, maybe it is to do with how division and remainder are implemented in JS...

src/store/workflows.module.js
```diff
@@ -40,6 +40,28 @@ function _getCycles (workflows) {
...
+ * @returns {number} the percent progress, e.g. 25 (meaning 25% progress)
+ * @private
+ */
+function _computePercentProgress (startedTime, meanElapsedTime) {
+  const now = Date.now()
+  // startedTime > now reportably possible via interaction with `cylc reset`
+  if (startedTime > now || meanElapsedTime === 0) {
+    return 0
+  }
+  const elapsedTime = startedTime + meanElapsedTime
```
hjoliver: This variable is "expected time of completion" so "elapsedTime" isn't a good name for it. ("Elapsed time" is how much time has passed so far, while the job is running).

src/store/workflows.module.js
```diff
@@ -40,6 +40,28 @@ function _getCycles (workflows) {
...
+  if (now > elapsedTime) {
+    return 100
+  }
+  return 100 * (now - startedTime) / meanElapsedTime
```
hjoliver: Looks like it would work!

src/store/workflows.module.js
kinow: Aaaahh! I was trying to guess what was `tetc_unix `! Will change it, thanks!

src/store/workflows.module.js
kinow: Yeah, I changed very little the logic, only because this function is returning the value. But from what I could tell, it was supposed to give the same result. Just finishing some performance test on the complex suite + Cylc UI, then will jump back to this one.

src/store/workflows.module.js
hjoliver: I had to remind myself.  Bad "etc" was a bad variable name in the old code, sorry!

hjoliver: Found the problem: `startedTime` is **milliseconds** since 1970; meanElapsedTime is **seconds**!  Fixed in my environment, then:
shot

kinow: Thanks a ton @hjoliver ! Feel free to push to this branch if you'd like. Then I can sync and test and if it works and tests pass, move it to ready-to-review 🎉

hjoliver: Will do...

hjoliver: Done

kinow:
Thanks for the @hjoliver ! I will test it, update my unit tests, and mark as ready for review.

kinow:
Fixed unit tests to match the last change/fix by @hjoliver (thanks again). Doing a quick test now to see if I can get a screen shot of what it should look like, and finally will mark it ready to review.

kinow:
Appears to be working like a charm now! Thanks @hjoliver !!!

Screenshot_2019-09-16_15-15-07

To review, one could:

1. Start whole UI (from this branch, of course) + Hub
2. Run some suite that tasks take longer than 5 seconds, e.g.

```
[meta]
    title = "Inter-cycle dependence + a cold-start task"
[cylc]
    UTC mode = True
[scheduling]
    #runahead limit = 120
    initial cycle point = 20130808T00
    final cycle point = 20150812T00
    [[dependencies]]
        [[[R1]]]
            graph = "prep => foo"
        [[[PT12H]]]
            graph = "foo[-PT12H] => foo => bar"
[runtime]
[[root]]
script = "sleep 20; echo 'done!'"
```

3. Wait a little for the UI to render the tree and to have the progress displayed.

kinow:
Added one more commit to compute the progress only in tasks with a state in an array, that for now contains only `running`.

kinow:
Still got one more improvement before I change it to ready for review. @dwsutherland 's suggestion to remove this code

```
for (const job of taskProxy.jobs.reverse()) {
```

and instead ask the query to return jobs in order per submitNum (safer/faster). But gotta go, will push this change in a couple of hours.

kinow:
Before I go and forget, @cylc/core what about the cycle point icon? While testing in my local environment, I noticed that the tasks get their icon changed to running with a progress like 50% and 75%.

But what about the cycle point of that task? Should it contain any progress? (I think it would be useful when you have several tasks running in the same cycle point, but not sure if that makes sense...)

hjoliver:
> I noticed that the tasks get their icon changed to running with a progress like 50% and 75%.

What do you mean by that?  Initially at 50%?  (only if you have very fast-running tasks, compared to the data update?)

> But what about the cycle point of that task?...

I don't think progress is feasible at family or cycle point level in general, because we can't know what the concurrency will be of all the tasks inside (even if they are supposed to all run at once, some might be delayed by real-world resource contention or whatever).

kinow:
>What do you mean by that? Initially at 50%? (only if you have very fast-running tasks, compared to the data update?)

Sorry, wrote that in a hurry and it wasn't clear. I meant if you have a moment where a cycle point has two tasks, one at 50% and another one at 75%. Then should we try to calculate the progress of the cycle point task icon too?

If not, I think maybe a different icon for the cycle point would be better. At least when I was searching for running tasks, I had to skip the cycle points, and that wasn't easy/comfortable.

kinow:
Finished testing manually, it appears to be OK. Ran linter & unit tests locally. Both passing.

Sneaking a fix for the `checkpoint.js` file. This file is generated for the mocked data. So if we change the GraphQL query (as I did to order by job submitNum instead of doing that in JS), we have to re-run the example workflow.

But `checkpoint.js` always fails the linter, and someone needs to fix manually. Fix it with `sed` prepending `/* eslint-disable */ at the top of the file. As I assume we don't care about linting that file :+1: 

Ready for review now.

Thanks!
Bruno

hjoliver:
>  meant if you have a moment where a cycle point has two tasks, one at 50% and another one at 75%. Then should we try to calculate the progress of the cycle point task icon too?

We could compute a sensible "family (or cycle point) progress" only if all members were already running. If any members are not running yet, or worse have not even appeared in the task pool yet, then it's generally impossible because we can't possibly know for sure when those members will start running.

So I that basis I think we can't have family or cycle point progress.  We might need a special version of the "running" task status that can be distinguished easily from any particular percent progress, otherwise it will be confusing.

kinow:
>So on that basis I think we can't have family or cycle point progress (except in a rare circumstance that is probably so rare that it's not worth doing).

Sounds good to me.

>We might need a special version of the "running" task status that can be distinguished easily from any particular percent progress, otherwise it will be confusing.

:+1:

hjoliver:
At the moment the cycle point (and presumably family, when that works) icon is submitted (/zero percent progress) - I suppose that's OK for a alpha-1 release.  We can document that as one of the things to be improved in subsequent releases.

kinow:
>At the moment the cycle point (and presumably family, when that works) icon is submitted (/zero percent progress) - I suppose that's OK for a alpha-1 release.

I also think it should be fine for alpha-1.

>We can document that as one of the things to be improved in subsequent releases.

Agreed, and due to my goldfish memory, had to record it or would completely forget about it: #231

src/store/workflows.module.js
```diff
@@ -40,6 +42,28 @@ function _getCycles (workflows) {
...
+ * Compute percent progress.
+ *
+ * @see https://github.com/cylc/cylc-flow/blob/de7d938496e82dbdfb165938145670dd8e801efd/lib/cylc/gui/updater_tree.py#L248-L263
+ * @param {number} startedTime in milliseconds since 1970-01-01 00:00:00 UTC, e.g. 1568353099874
+ * @param {number} meanElapsedTime mean elapsed time in seconds
+ * @returns {number} the percent progress, e.g. 25 (meaning 25% progress)
+ * @private
+ */
+function _computePercentProgress (startedTime, meanElapsedTime) {
+  const now = Date.now() // milliseconds since 1970-01-01
```
oliver-sanders: How does time zone affect this?

kinow: Oh, good point!!

I think we have a bug! The new Date here will use client TZ. While the started time I think could be in any TZ?

matthewrmshin: The started time string always has a time zone. Does `Date.parse` recognise and handle that automatically? If so, the subsequent duration calculation should be correct regardless?

kinow: Oh, could be! I've added your comment here https://github.com/cylc/cylc-ui/issues/237 :+1: hopefully that issue can be closed as not-an-issue :grimacing:

hjoliver: Discussion moved to #237 ...

Issue #222, opened 2019-09-13
Performance placeholder issue


kinow:
A placeholder for performance issues. Set to milestone 0.1, but it will probably be moved to 1.0 I think.

Since the first prototypes that were able to display an e2e system, we have had issues running the workflow `complex`, an example suite from Cylc 7.

That workflow does not execute any command locally, simply sleeps, but it spawns heaps of tasks, with dependencies/requirements galore.

Another issue that we may have is with the progress percent calculation in the Tree View and Graph View.

Among other problems, so feel free to chime in and register any issue or progress here, or link with related issues.

## Outstanding issues:

- [x] Fix cached items retained after Tree component is destroyed #225 
- [x]  Memory leak in Tree/TreeItem after running complex workflow #227 

## Good resources

- [JavaScript performance beyond bundle size](https://nolanlawson.com/2021/02/23/javascript-performance-beyond-bundle-size/)

kinow:
I will start with a quick experiment I did today. I finished implementing the progress for the tasks in the Tree view. But the issue I had, I think, is not related to this new feature.

The complex suite was running fine, but the browser seemed to have some memory leak. I used an Ubuntu LTS, with an i7 16GB ram 200 GB SSD, running firefox with some 10 tabs, WebStorm, `npm` and `cylc`.

Using Chrome and its great memory tool (love Firefox, but their memory tool is a bit behind...), I took a few snapshots, as in the screenshot below (memory sizes below the snapshot name).

![image](https://user-images.githubusercontent.com/304786/64844392-c52dec80-d65b-11e9-8326-b8d325d62daf.png)

To populate the DOM, Vue.js uses a virtual DOM and creates VNode's to attach to it. Here's the progression of the VNode's:

_(Note: shallow size always <= retained size... shallow is simply the size for the object instances, retained is that shallow size plus whatever other objects are required/retained with that object)_

- Snapshot 1 234 MB (complex suite running)
- Snapshot 2 424 MB (~5 minutes later)
- Snapshot 3 721 MB (~5 minutes later) (Vue devtools crashed in Chrome)
- Snapshot 4 903 MB (~5 minutes later)
- Snapshot 5 903 MB (stopped complex suite, then after another 5 minutes)

The Snapshot 5 is very strange. I had stopped the suite already, but the memory hadn't reduced.

Then took yet another snapshot, now about 25 minutes later, while writing this paragraph.

- Snapshot 6 903 MB

[...]

PR #225, opened 2019-09-15
Fix cached items retained after Tree component is destroyed

kinow:
Part of #222 

Not sure if will be done before 0.1, but it is part of the problem in the memory performance issue.

The three caches were holding references to objects that were never removed or collected. Found by filtering for detached in the Chrome memory tool, then expanding the nodes until I located several that were retained due to the caches.

Screenshot_2019-09-16_00-07-46

(doesn't actually show the cache, but was around there where I located the issue)

It appears - after some initial test post midnight, so to be reviewed tomorrow morning - to fix the issue. However, still need to:

- [x] probably clear the cache when the Vuex store data gets re-populated, this could be the cause for not being able to run the complex suite for very long
- [x] check if the complex suite works

### DIFF
--- a/src/components/cylc/Tree.vue
+++ b/src/components/cylc/Tree.vue
@@ -105,6 +105,11 @@ export default {
           this.activeCache.add(treeItem)
         }
       }
+    },
+    clearCaches () {
+      this.treeItemCache.clear()
+      this.activeCache.clear()
+      this.expandedCache.clear()
     }
   }
 }
--- a/src/components/cylc/Tree.vue
+++ b/src/components/cylc/Tree.vue
@@ -35,6 +35,13 @@ export default {
       default: 0
     }
   },
+  watch: {
+    workflows: {
+      handler () {
+        this.clearCaches()
+      }
+    }
+  },
   components: {
     'tree-item': TreeItem
   },
--- a/src/components/cylc/TreeItem.vue
+++ b/src/components/cylc/TreeItem.vue
@@ -73,6 +73,7 @@
       <!-- component recursion -->
       <TreeItem
           v-for="child in node.children"
+          ref="treeitem"
           :key="child.id"
           :node="child"
           :depth="depth + 1"
@@ -169,7 +170,18 @@ export default {
   created () {
     this.$emit('tree-item-created', this)
   },
+  beforeDestroy () {
+    if (Object.prototype.hasOwnProperty.call(this.$refs, 'treeitem')) {
+      this.$refs.treeitem.map((treeitem) => {
+        treeitem.destroy()
+      })
+    }
+  },
   methods: {
+    destroy () {
+      // $destroy will trigger beforeDestroy and destroyed
+      this.$destroy(/* destroy from DOM as well */ true)
+    },
     typeClicked () {
       this.isExpanded = !this.isExpanded
       if (this.isExpanded) {
--- a/src/store/workflows.module.js
+++ b/src/store/workflows.module.js
@@ -126,6 +126,7 @@ function _getWorkflowTree (workflows) {
       }
       workflowTree.push(workflow)
     }
+    cycles.clear()
   }
   return workflowTree
 }
--- a/src/views/Tree.vue
+++ b/src/views/Tree.vue
@@ -100,7 +100,7 @@ export default {
   },
 
   beforeDestroy () {
-    this.$refs['tree0'].clearCaches()
+    this.$refs.tree0.clearCaches()
     workflowService.unregister(this)
   },
 
--- a/src/views/Tree.vue
+++ b/src/views/Tree.vue
@@ -6,6 +6,8 @@
       :activable="false"
       :multiple-active="false"
       :min-depth="1"
+      ref="tree0"
+      key="tree0"
     ></tree>
   </div>
 </template>
@@ -98,6 +100,7 @@ export default {
   },
 
   beforeDestroy () {
+    this.$refs['tree0'].clearCaches()
     workflowService.unregister(this)
   },
 


kinow:
There was another leak indeed when the `workflows` data changed in Vuex. For this case, the component doesn't have a way to tell "Do this before the reactivity kicks in and updates everything", the only way is through a `watch` (one of the few cases where you need a `watch`).

I am careful to use the `watch` instead of simply hooking up on the reactive values. But confirmed that Vuetify's `VTreeView` (which was used as reference for our `Tree` component) [also uses a `watch`](https://github.com/vuetifyjs/vuetify/blob/b0d7cf1a5d7c60899ec5183262d3b2e5c5f227c8/packages/vuetify/src/components/VTreeview/VTreeview.ts#L120) (though they are doing a somewhat better job but probably because they have more control in TypeScript).

kinow:
There seems to have another leak, in the ApolloData queries I think? Or the ApolloData response...

When the leak was in the `Tree` component, the top offender for retained data were `Object` instances. But at the moment, after running a workflow for a while and then stopping it (i.e. blank page for workflow), the top offender now are `string` instances.

Screenshot_2019-09-16_09-22-20

The strings at the top look like JSON objects returned from the GraphQL API. We shouldn't have so many being held back, especially now that there is nothing running.

Possibly caching configuration in Apollo.

kinow:
Not graphql queries, but some babel stuff. Building production mode now, will test the complex suite. Here's the base64 decoded:

```json
{"version":3,"file":"app.js","sources":["webpack/bootstrap","/home/kinow/Development/python/workspace/cylc-ui/node_modules/@babel/runtime-corejs2/core-js/array/from.js","/home/kinow/Development/python/workspace/cylc-ui/node_modules/@babel/runtime-corejs2/core-js/array/is-array.js","/home/kinow/Development/python/workspace/cylc-ui/node_modules/@babel/runtime-corejs2/core-js/get-iterator.js","/home/kinow/Development/python/workspace/cylc-ui/node_modules/@babel/runtime-corejs2/core-js/is-iterable.js","/home/kinow/Development/python/workspace/cylc-ui/node_modules/@babel/runtime-corejs2/core-js/object/create.js","/home/kinow/Development/python/workspace/cylc-ui/node_modules/@babel/runtime-corejs2/core-js/object/define-property.js"
```

kinow:
Started complex suite 09:50, first heap snapshot with 59MB.

Took another snapshot once the Tree components was rendered in the Tree view, with 120 MB.

Will let it run for 10-15 minutes, then take another snapshot while the component is rendered, then stop the workflow, give it some time for GC to work (or force it) and then another snapshot.

kinow:
Huh, another leak. Will re-run in development mode to see the objects without obfuscation.

Screenshot_2019-09-16_10-19-37

kinow:
There were a few dangling `HTMLDivElement` for "nodes" of the Tree component. After some investigation, I could confirm the children `TreeItem` were not removed, only the top level ones (i.e. it seemed to be killing only the top ones displayed and ceasing the destruction in the sub levels).

Forced it by calling the `$destroy` method from the parent `TreeItem` - which is not really recommended. Also passed `$destroy(true)` to force removal from the DOM. After that, there is an improvement but I cannot see the memory decreasing even after several minutes running, but it got better.

And more important, there are no more detached elements in the memory heap snapshots (i.e. nothing wrong with what we have not connected to the DOM / page, only expected services/utility objects/axios,graphql stuff, etc)

kinow:
Final memory leak in workflow "five" fixed. This last one did not demand any code. But good to post here so others are aware too (cc @hjoliver that I think might be using Vue dev tools browser extension too).

The last memory leak I was chasing consisted in constant increase of `Object` instances after 2-3 minutes. They were in a `_buffer` structure. Expanding on the chrome memory tool, I could see them in some sort of vue dev tools objects.

So started the app again in incognito (where I've disabled Vue) and the app is now behaving correctly.

Screenshot_2019-09-16_13-47-42

My **guess** after looking at some of these `Object` instances, is that that buffer is used by vue dev tools to keep the transaction log of the vuex commit/mutation/etc. There should be a way to clear it in the extension panel, but I haven't looked into that yet.

So if you realize that after a while your browser tab gets slow, and you have vue dev tools... well, you have a few options:

- use vue dev tools as I do at home, on Chrome (Firefox is my default browser, I use chrome for testing  accessibility/performance/vue/etc)
- leave long running tabs open on incognito without the extension
- find a way to limit the amount of information buffered in the dev tools (and tell others please :grimacing: )

hjoliver:
Nice; we should get this in for the alpha-1 release.

kinow:
Agreed! Moved to 0.1 milestone (wasn't sure if it would be ready this week to be honest)

kinow:
All done, that's the best I could get for the complex workflow for now. I started at ~25MB as it normally does in my computer.

Once the Tree component rendered the first nodes, the next snapshot got ~95MB. Important to note that taking snapshots while the workflow runs is much slower, probably because it's CPU intensive, and that complex workflow appears to be CPU bound too.

And 95MB is quite good - for this workflow. The next snapshot came after ~5 minutes running, and it marked ~150 MB. It is important to note here that the complex workflow does not behave like the five workflow, which has a constant pace, with the same number of tasks in each cycle point (with exception to the initial cycle point I think).

Shortly after that I stopped the workflow, and the tab memory was at about ~160 MB. Waited several minutes, forced GC, then went back to Dashboard. It dropped to 27.7 MB.

This means that the component was correctly destroyed, as well as its children nodes. There are still some dangling HTML divs that appear to have been created by the Tree/TreeItem component pair. But I will move it to a separate ticket.

Smaller workflows like five should be fine for now :tada: 

Screenshot_2019-09-16_14-43-51

kinow:
Ready for review. Unfortunately not easy to reproduce. Better to read the - lengthy, sorry - comments and code-review only :+1:

src/components/cylc/Tree.vue
```diff
@@ -36,6 +36,13 @@ export default {
       default: 0
     }
   },
+  watch: {
+    workflows: {
+      handler () {
+        this.clearCaches()
```
kinow: A `watch` because this is a case where it is necessary to act when the element changes, and that we have no easy way with Vue default approaches like `v-if`, `data`, `computed`, etc.

src/components/cylc/Tree.vue
```diff
@@ -106,6 +113,11 @@ export default {
           this.activeCache.add(treeItem)
         }
       }
+    },
+    clearCaches () {
+      this.treeItemCache.clear()
+      this.activeCache.clear()
+      this.expandedCache.clear()
```
kinow: These are cache instances that were not being cleared before. So they were increasing a lot over time, and leaving retained objects in the heap, preventing GC to collect them.

src/components/cylc/TreeItem.vue
```diff
@@ -117,7 +118,18 @@ export default {
   created () {
     this.$emit('tree-item-created', this)
   },
+  beforeDestroy () {
+    if (Object.prototype.hasOwnProperty.call(this.$refs, 'treeitem')) {
+      this.$refs.treeitem.map((treeitem) => {
+        treeitem.destroy()
```
kinow: The children / recursive TreeItem elements are not always being destroyed. Not sure if because it's happening too soon/late, or if because we are doing something that is not conventional for Vue components (the circular/recursive pattern is no that common). So we call a custom `destroy` method, which will then call `$destroy(true)` method from Vue. This `$destroy` takes care to trigger the `beforeDestroy -> destroy` lifecycle.

src/components/cylc/TreeItem.vue
```diff
@@ -129,7 +141,7 @@ export default {
     getTypeStyle () {
       const styles = {}
       if (this.hasChildren) {
-        styles['cursor'] = 'pointer'
+        styles.cursor = 'pointer'
```
kinow: Needed for the linter, happy to remove it if other PR fixes it first (I think I fixed it in the progress percent PR too)

tests/unit/services/user.service.spec.js
```diff
@@ -13,12 +13,13 @@ describe('UserService', () => {
   afterEach(() => sandbox.restore())
   describe('getUserProfile returns the logged-in user profile information', () => {
     it('should return user profile object', () => {
-      const userReturned = new Promise((resolve) => resolve({ data: {
-        name: 'cylc-user-01',
-        groups: ['root', 'wheel'],
-        created: '2019-01-01',
-        admin: true
-      }
+      const userReturned = new Promise((resolve) => resolve({
```
kinow: Linter

kinow:
Fixed conflicts, rebased `master`.

kinow:
@wxtim 

> Eyeballed, tried on a randomly generated horrible suite.

You actually managed to run the UI with this suite???

PR #240, opened 2019-09-18
Remove chartist and vue-chartist, as well as the theme component (not used) referencing it

### LINKED ISSUE #210: Remove chartist
It comes with the [theme](https://github.com/creativetimofficial/vuetify-material-dashboard) but we are not using. It has a considerable size, so removing it from `package.json` is probably good.

kinow:
Closes #210 

I think one review will do, but happy to wait for a second review if necessary.

It removes `chartist` and the `vue-chartist` dependencies. Both were used by the theme provided by creative tom.

If you remember our old dashboard, with charts/graphs; that was made with `chartist`.

As we are not using any of that, this PR simply removes those dependencies. Will wait for this one to be reviewed and then prepare another PR to remove vue-flag-icon (if I prepared that now, it would definitely have conflicts later if this one is merged first).

Cheers
Bruno


kinow:
Tested locally, running in offline and development mode (i.e. with hub + workflow five :grimacing:). Worked with no issues, and no console errors.

PR #241, opened 2019-09-18
Graph with live cylc data

MartinRyan:
The graph is now a component and is accessed by clicking a new graph action button on the Workflows screen. It uses the workflow id to instantiate the graph. The `graphData`  in the Vue data built-in is watched for changes and this initiates the graph update ( This same watch function is used by websocket functionality, which is still there in this draft, but disabled, so it can be used when websocket updates become available).

An initial empty cytoscape graph instance is created and all extensions are registered. The updated graphData is then used to create the new  node and edge elements with the desired values. (This is new as previously the desired values were already in the graphData,) These new elements values are used to generate the stylesheet dynamically. This is then applied to the cytoscape instance. The layout options are applied to the edge and node elements depending on the chosen algorithm, the default is dagre, and then run. A pie chart extension is used and for all state values except running is mapped to zero size. Once the `runpercent` value is available and mapped this will change the pie-chart to the nodes size and show a pie chart over running nodes. 

This is a draft pull request, some colour values may be incorrect and not all icons have been decided upon, default icons are then used. This version is only functional when used in the cylc system and not standalone. Collapsing nodes stilll has to be tested.


matthewrmshin:
Running this branch (with master merged in) right now on my box. Works as advertised. 🎉

src/views/GScan.vue
```diff
@@ -57,6 +57,15 @@
...
               </td>
+              <td class="justify-center">
+                <v-icon
+                    small
+                    class="mr-2"
+                    @click="viewGraph(item)"
+                >
+                  mdi-vector-polyline
+                </v-icon>
+              </td>
```
matthewrmshin: Need to add another column to the `headers` variable.

src/components/cylc/Graph.vue
```diff
@@ -0,0 +1,1329 @@
...
+            locked: false,
+            grabbable: true,
+            classes: ''
+          }
+          has(node, 'id') && !isEmpty(node.id) ? nodeObj.data.id = getUuid(node.id) : console.log('workflowUpdated - node id is empty')
+          has(node, 'label') && !isEmpty(node.label) ? nodeObj.data.label = node.label : console.log('workflowUpdated - node label is empty')
+          has(node, 'state') && !isEmpty(node.state) ? nodeObj.data.state = node.state : nodeObj.state = 'undefined'
+          has(node, 'runpercent') && !isEmpty(node.runpercent) ? nodeObj.data.runpercent = node.runpercent : nodeObj.data.runpercent = 0
+          has(node, 'parent') && has(node.parent, 'id') && !isEmpty(node.parent.id) ? parentId = getUuid(node.parent.id)
+            : console.log('workflowUpdated - node parent is empty')
```
matthewrmshin: Just a bit of a style here. Given that you are not assigning results of the ternary expressions, I would imagine that `if ... else` blocks would be easier to read here.

src/components/cylc/Graph.vue
```diff
@@ -0,0 +1,1329 @@
...
+            locked: false,
+            grabbable: true,
+            classes: ''
+          }
+          has(edge, 'id') && !isEmpty(edge.id) ? edgeObj.data.id = getUuid(edge.id) : console.log('workflowUpdated - edge id is empty')
+          has(edge, 'source') && !isEmpty(edge.source) ? edgeObj.data.source = getUuid(edge.source) : edgeObj.source = undefined
+          has(edge, 'target') && !isEmpty(edge.target) ? edgeObj.data.target = getUuid(edge.target) : edge.target = undefined
+          has(edge, 'label') && !isEmpty(edge.label) ? edgeObj.data.label = edge.label : edgeObj.label = ''
+          edgeObj.data.source !== undefined || edgeObj.data.target !== undefined ? edgesArray.push(edgeObj)
+            : console.log('skipping adding edge with empty source or target')
```
matthewrmshin: Ditto comment on ternary operator.

hjoliver:
@MartinRyan - my initial reaction after playing with this for a short while. Just considering functionality, not the code.

On the up-side:
1. Thanks for posting this as a draft PR - it is great to be able to try it out so easily
2. Well done for hooking Cytoscape up to real suite data - I'm sure that wasn't easy
3. I can certainly see some positives that will no doubt inform the eventual solution

On the down-side:
1. (This seems to be based on quite an old version of master and there may be a mess of conflicts when you rebase?)
2. Viewing a moderately sized cycling suite it seems to me there are quite a lot of problems and we are still quite a long way from a usable graph view. 

So for now, it's not a goer for the alpha-1 release this week (we knew that was a long shot, but it was worth checking).

Then for later, we probably need a collective brain-storming session on the graph view to:
- itemize current short-comings, compared to what we need
- (with reference to good aspects of the cylc-7 GUI graph view too)
- figure out if they can be fixed, or if Cytoscape is too heavy and inflexible for a live-update view with unusual display requirements (the job icons etc.) such as we need?  
- (and consider alternative ideas)

(It's worth noting that some of our plans for Cylc 8 will reduce the weight of the graph view compared to showing "the full graph" - but even so, not easy).

Sorry if that sounds negative. Note I haven't spent long with the branch and my first impression might be wrong in some respects.  We always knew the graph view would be hard (and that Cytoscape has to be proven in our context before we go ahead with it)!

matthewrmshin:
(I tried merging master into this branch. Conflict only in 2 files and relatively easy to solve in both cases.)

hjoliver:
(Good, I guess the graph view code is mostly orthogonal to the rest of the UI).

hjoliver:
(Note my comment on flexibility is partly inspired by @MartinRyan's comments on not being able to style the nodes in the way that we expected to be able to, i.e. in common with other views ... but also major node label crush issues which doesn't bode well for the job icons).

hjoliver:
Actually I just tried again with a much smaller suite, and it does work quite well with that:
```
[cylc]
     cycle point format = %Y
  [scheduling]
     initial cycle point = 2025
     [[dependencies]]
        R/^/P2Y = foo[-P2Y] => foo => FAM:succeed-all => bar

[runtime]
   [[root]]
       script = sleep 20
   [[FAM]]
   [[f1, f2, f3, f4, f5]]
      inherit = FAM
```

Screenshot from 2019-09-19 01-08-51

hjoliver:
Hmmm, if we had a bit longer before the alpha-1 release (which we don't), and could somehow prevent users from going too hard and big with the graph view (which we probably can't) then maybe that would have been a nice addition to the release.

hjoliver:
From observing a couple more small-ish suites, bad things happen when I do any zooming or panning, but the basic view actually looks pretty good.  So there you go, I have moved more in the positive direction :+1:  ... however we still have some major issues to contend with, to get what we need.

hjoliver:
I see much of the code change here - 6.8k lines - is `package-lock.json` :thinking:

kinow:
>I see much of the code change here - 6.8k lines - is package-lock.json 🤔

Probably Cytoscape dependencies, as long as there are no conflicts, we should be fine with this file. Only issue could be the impact on performance if it adds lots of large-ish dependencies (see #222)

src/components/cylc/Graph.vue
MartinRyan: I suppose it is subjective. I prefer ternary when it can be a simple one liner, the results are assigned to properties in the node object.

src/views/GScan.vue
MartinRyan: I thought this would be part of Actions?

MartinRyan:
> (Note my comment on flexibility is partly inspired by @MartinRyan's comments on not being able to style the nodes in the way that we expected to be able to, i.e. in common with other views ... but also major node label crush issues which doesn't bode well for the job icons).

Whatever graph technology is used, even with custom code, the styling will be require a different method than with other views.

MartinRyan:
> From observing a couple more small-ish suites, bad things happen when I do any zooming or panning (e.g. all nodes permanently disappear, leaving just edges; everything resets on every update; ...) BUT the basic view actually looks pretty good. So there you go, I have moved in the positive direction

I think this may be an issue with five second refreshes when using requests, rather than websocket updates only on data changes

hjoliver:
> Whatever graph technology is used, even with custom code, the styling will be more involved and require a different method than with other views.

I guess I'll have to take your word for that, but I don't really understand why yet.  Even a graph is just styled DOM elements isn't it? (unless it is a single "canvas" element - not sure what the terminology is - which I presume we can avoid?)  Isn't that the essence of D3 data-driven visualization, for example? Elements of the visualization should "react" automatically when the underlying data changes.  And then, why can't we style elements however we like, such as with our own custom SVG task and job icons?

The main difference for a graph view is the layout computation (i.e. compute the node positions and edge paths). Which should only need to be re-done - in principle - when new nodes appear or old nodes disappear.

hjoliver:
> I think this may be an issue with five second refreshes when using requests, rather than websocket updates only on data changes

Can you explain how the five second update would cause this? Are you suggesting that the graph layout takes longer than 5 seconds even for smallish suites, so it can't keep up?

Bear in mind that with websockets the data will often update much more quickly than every five seconds!

MartinRyan:
> > Whatever graph technology is used, even with custom code, the styling will be more involved and require a different method than with other views.
> 
> I guess I'll have to take your word for that, but I don't really understand why yet. Even a graph is just styled DOM elements isn't it? (unless it is a single "canvas" element - not sure what the terminology is - which I presume we can avoid?) Isn't that the essence of D3 data-driven visualization, for example? Elements of the visualization should "react" automatically when the underlying data changes. And then, why can't we style elements however we like, such as with our own custom SVG task and job icons?
> 
> The main difference for a graph view is the layout computation (i.e. compute the node positions and edge paths). Which should only need to be re-done - in principle - when new nodes appear or old nodes disappear.

The styling for a graph will have to be applied purely in javascript code and not using built-in vue framework constructs such as <v-icons> <v-cards> etc in html and top level referenced css.

MartinRyan:
> > I think this may be an issue with five second refreshes when using requests, rather than websocket updates only on data changes
> 
> Can you explain how the five second update would cause this? Are you suggesting that the graph layout takes longer than 5 seconds even for smallish suites, so it can't keep up?
> 
> Bear in mind that with websockets the data will often update much more quickly than every five seconds!

I think partial updates over websockets will definitely be more performant. The future graph will address only removing and updating changed nodes and edges of partial updates.

hjoliver:
> I think partial updates over websockets will definitely be more performant. The future graph will address only removing and updating changed nodes and edges of partial updates.

Ah, now I think we agree.  You second sentence there is more like what I had envisaged (I thought you were saying the layout had to be redone for *any* change every time...)

MartinRyan:
> > I think partial updates over websockets will definitely be more performant. The future graph will address only removing and updating changed nodes and edges of partial updates.
> 
> Ah, now I think we agree. You second sentence there is more like what I had envisaged (I thought you were saying the layout had to be redone for _any_ change every time...)

At the moment that is the case

kinow:
Not much time to play with it today, but just wanted to report that checked out the branch, ran npm install, and voila! Got the graph working locally with my running workflow and backend :tada: Well done @MartinRyan ! Will have another look at it tomorrow.

kinow:
Just took 2 heap snapshots, first immediately after loading the app. Second after loading the graph and waiting 1 minute. Removed the stop cycle point from my workflow, so it will keep running over night.

image

Will post the other snapshot tomorrow morning 👍

kinow:
Turned on computer, didn't touch anything else:

08:56 103 MB
09:00 109 MB
(pressed garbage collect button, waited a bit)
(memory tool crashed and disappeared, I think chrome was having some glitch, so restarted it)
09:32 29.9 MB

App is still running, the workflow as well.

src/views/GScan.vue
matthewrmshin: If so, then the `<v-icon>` element should live with the previous `<td>`?

src/components/cylc/Graph.vue
matthewrmshin: Don't get me wrong, I am a big fan of the ternary operator myself. In my life before Python, I have used it a lot to write simple logic tables, but I would always have the result assigned to a variable.

Yes, it is perhaps just a style preference.

dwsutherland:
> > > I think this may be an issue with five second refreshes when using requests, rather than websocket updates only on data changes
> > 
> > 
> > Can you explain how the five second update would cause this? Are you suggesting that the graph layout takes longer than 5 seconds even for smallish suites, so it can't keep up?
> > Bear in mind that with websockets the data will often update much more quickly than every five seconds!
> 
> I think partial updates over websockets will definitely be more performant. The future graph will address only removing and updating changed nodes and edges of partial updates.

I think the only way to do this with GraphQL subscriptions is to create/hold a "mini" data-store of only latest updated elements, and have the subscription resolvers hit that instead of the full data-store (which wouldn't be too hard)... Because the Subscription is just a query you define, and I don't think GraphQL knows whats been updated under the hood.. (Unless the UI knows what's changed when you update the data with a new set)

src/components/cylc/Graph.vue
```diff
@@ -0,0 +1,1284 @@
...
+            },
+            group: 'nodes',
+            removed: false,
+            selected: false,
+            selectable: true,
+            locked: false,
+            grabbable: true,
+            classes: ''
+          }
+          has(node, 'id') && !isEmpty(node.id) ? nodeObj.data.id = getUuid(node.id) : console.log('workflowUpdated - node id is empty')
```
kinow: Can't we just use `node.id` when creating `nodeObj`? @dwsutherland can correct me if I am wrong, but in the whole `nodes`, each node will have an unique ID.

e.g. Workflow ID `five`... the cycle point ID 'five/20100101T000000Z`, the task proxy `five/20100101T000000Z/foo`, and the job `five/20100101T000000Z/foo/01`. They should not repeat within the same workflow? If so, we don't need to calculate the UUID I think.

For the Tree view, the `v-for` loop demands a `key`, that must be unique. I am using the `.id` of each entry in the workflows GraphQL response.

ID's are also required in our schema (quite sure somewhere in GraphiQL it should appear with a `!` ? See this part of the Cylc Flow code: https://github.com/cylc/cylc-flow/blob/f9927a5c53f072e24f65bd0f2355bff393af5b42/cylc/flow/network/schema.py#L419

src/components/cylc/Graph.vue
```diff
@@ -0,0 +1,1284 @@
...
+    },
+
+    async preConfig (cytoscape) {
+      // cytoscape: this is the cytoscape constructor
+      try {
+        console.log('PRE-CONFIG')
+        cytoscape.use(cola)
+        cytoscape.use(dagre)
+        cytoscape.use(coseBilkent)
+        cytoscape.use(klay)
```
kinow: Maybe we could remove the different graph types, and use a single style?

This would reduce the code in the PR, making it easier to be reviewed. Then later we can add a PR that brings it back?

src/views/GScan.vue
MartinRyan: which do you prefer?

src/store/graph.module.js
```diff
@@ -0,0 +1,51 @@
+/* jshint esversion: 6, asi: true */
```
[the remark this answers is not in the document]
MartinRyan: This is ready for websocket functionality

src/views/GScan.vue
matthewrmshin: Don't mind. Probably easiest for them to live under the same `<td>`?

MartinRyan:
> Can you explain how the five second update would cause this? Are you suggesting that the graph layout takes longer than 5 seconds even for smallish suites, so it can't keep up?
> 
> Bear in mind that with websockets the dat

> > > > I think this may be an issue with five second refreshes when using requests, rather than websocket updates only on data changes
> > > 
> > > 
> > > Can you explain how the five second update would cause this? Are you suggesting that the graph layout takes longer than 5 seconds even for smallish suites, so it can't keep up?
> > > Bear in mind that with websockets the data will often update much more quickly than every five seconds!
> > 
> > 
> > I think partial updates over websockets will definitely be more performant. The future graph will address only removing and updating changed nodes and edges of partial updates.
> 
> I think the only way to do this with GraphQL subscriptions is to create/hold a "mini" data-store of only latest updated elements, and have the subscription resolvers hit that instead of the full data-store (which wouldn't be too hard)... Because the Subscription is just a query you define, and I don't think GraphQL knows whats been updated under the hood.. (Unless the UI knows what's changed when you update the data with a new set)
> 
> Also, how would it know which edges/nodes to delete? do you send a deletion set or send the ID skeleton of the whole data-store query each time?

The partial update data, or change set, should be controlled by the uiserver. Once received the included elements are mutated as specified. The included elements not in the graph are added, and the not included elements are removed from the graph.

src/components/cylc/Graph.vue
MartinRyan: is this the consensus?

src/components/cylc/Graph.vue
MartinRyan: I found it easier to scan for different nodes with unique ids while developing, but I can keep those just for the edges if required

src/components/cylc/Graph.vue
```diff
@@ -0,0 +1,1284 @@
...
+  },
+
+  beforeDestroy () {
+    workflowService.unregister(this)
+  },
+
+  mounted () {
+    console.log(`MOUNTED called, status: ${this.status}`)
+    this.handleMounted()
+    this.$store.watch((store) => {
```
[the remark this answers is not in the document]
MartinRyan: I prefer using the watch paradigm, and this will also be used when we move to websockets

src/components/cylc/Graph.vue
kinow: We are using `console.debug` in other parts of the code I think. Perhaps we could use that here too? My preference is also to have both sides of the ternary operator to be similar... i.e. `condition ? log(a) : log(b)` or `var x = condition ? a : b`. But not a blocker IMO if @MartinRyan prefers to keep it :+1:

src/store/graph.module.js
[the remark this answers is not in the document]
kinow: The websocket will take a while. We will need to prepare a PR for the tree view as well. I think it would be simpler to have this PR just to add the Graph component with the view changes, limiting the scope. Then add other features in subsequent PR's.

src/components/cylc/Graph.vue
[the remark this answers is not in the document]
kinow: The issue here is that the Tree component is adopting a different style. And I would be happy with either, but not both styles in the project, as I think it would be easier for maintainers if the components followed a certain pattern.

If this is better for the websockets, then I think we would have to discuss it when implementing websockets, and raise a PR to possibly fix it in both the Tree and the Graph components.

I really don't have a strong preference on either, but I use [this part of the Vue.js docs](https://vuejs.org/v2/guide/computed.html#Computed-vs-Watched-Property) as reference whenever I use a `watch` (which I just did recently in the Tree, then found out that it was not necessary and reverted again *):

> (...) it is tempting to overuse watch - especially if you are coming from an AngularJS background. However, it is often a better idea to use a computed property rather than an imperative watch callback.

So I think if we have the option to follow the documentation recommendations, we should. But as you pointed that it may be better for websockets, my suggestion here would be to be consistent with what we have for now in the other components, and move it to follow up pull requests.

* https://github.com/cylc/cylc-ui/pull/251/files#diff-d70e3541f48c5fdb3978c257a8726eceL38-R40 & https://github.com/cylc/cylc-ui/pull/251#issuecomment-532932593

src/components/cylc/Graph.vue
kinow: Gotcha, it was more in case it helped with performance by avoiding computing the UUIDs, but I believe this is probably nanoseconds and doesn't really matter :+1:

src/components/cylc/Graph.vue
kinow: Definitely not the consensus, just a suggestion. Someone from @cylc/core may chime in later and give their opinion too. So no need to change it if you don't think we need to do it now :+1:

src/components/cylc/Graph.vue
```diff
@@ -0,0 +1,1284 @@
...
+
+    layoutName: {
+      handler (value) {
+        this.updateLayout(value)
+      }
+    },
+
+    workflows: {
+      handler: function (newval, oldval) {
+        this.workflowUpdated(newval)
```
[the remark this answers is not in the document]
MartinRyan: With watch I can use `deep` to be aware of data changes, this is useful because the data is very nested. I was hoping it wouldn't be complex nested data, then I could use computed on just the graph data.

src/components/cylc/Graph.vue
MartinRyan: Every little helps :-)

src/components/cylc/Graph.vue
```diff
@@ -0,0 +1,1284 @@
...
+      try {
+        const config = {
+          autounselectify: true,
+          boxSelectionEnabled: true,
+          layout: {
+            name: 'dagre',
+            textureOnViewport: false,
+            hideEdgesOnViewport: true
+          },
+          style: [
```
kinow: Hmmm, looks like this is going to be tricky with Cytoscape. Spent some time today reading Cytoscape documentation, and playing with the code here. I believe now I understand what @MartinRyan said when he said we would have to re-compute the styles for each update.

## Not really CSS

There's some key information in their docs about styles https://js.cytoscape.org/#style:

>Style in Cytoscape.js follows CSS conventions as closely as possible. In most cases, a property has the same name and behaviour as its corresponding CSS namesake. However, the properties in CSS are not sufficient to specify the style of some parts of the graph. In that case, additional properties are introduced that are unique to Cytoscape.js.

So the style applied to nodes and edges is not necessarily CSS. Though it does look at lot like. This function that does part of the parsing of the style properties shows more or less what happens in the lower levels: https://github.com/cytoscape/cytoscape.js/blob/088995626b9cd69013da183980a3f178edc5a149/src/style/parse.js#L60

It has a list of properties that can be used... as in the docs :nerd_face: 

## But we can still define "classes"... though they must come during the initialisation

If we would like to have something like CSS classes, our best option is to use the Cytoscape [initialisation](https://js.cytoscape.org/#core/initialisation) and provide a `styles` to the Cytoscape object with the classes/groups that we want to have. e.g. (not tested)

```js
const styles = {
  '.running': {
    background-image: ...,
    background-color: states.RUNNING.color,
    ...
  },
  '.succeeded': {
    background-image: ...,
    background-color: states.SUCCEEDED.color,
    ...
  },
  ...
}
```

Doing this, in the `getNodes` and `getEdges`, we would then have to use the `classes: ['succeeded']`. It is similar to the current approach, but instead of providing functions that will receive the `node`, and then decide what to do, we specify the class and then it doesn't have to compute it again.

## cy.batch

We should probably try to use [`cy.batch`](http://js.cytoscape.org/#cy.batch) if we have performance issues due to recalculating styles.

## Theming

I think this affects theming. @oliver-sanders had had a look at some graph libraries before, and he has a very good understanding of how theming needs to work in Cylc, to allow us to support not only different nice UI's, but also attend users with accessibility issues (I know UK & NZ have users needing this).

So far, it seems to be that Cytoscape is not really compatible with our current approach for theming. Even though @oliver-sanders and I haven't agreed on how to implement it exactly, we both were looking at creating classes/variables in SCSS files, and allowing users to change it either in CSS or SCSS.

Looks like Cytoscape and SCSS won't work together in a component. So I wonder if @MartinRyan has an alternative for this, or if there's still time to investigate other libraries?

kinow:
Whew, finished reviewing it :tada: and again, kudos for making it work with live data and creating a draft PR @MartinRyan !

I've added some scattered comments in the code, but here's a simplified comment with things I think we need to consider for this PR:

1. Would be nice to get theming work as in other components with CSS/SCSS
2. I would prefer that it was more consistent with the tree view, job, and task components.
2.1. Receiving the data via props, and working with slots/events. At the moment the component is responsible for "driving the data". It watches the `workflows`, then calls `getNodes` and `getEdges` iff the workflow contains these data. But if we change that, we may have to change the component... IMO, the component needs to be made "dumber", and maybe just receive a prop for nodes, and another one for edges
2.2. Using a similar approach for reactive data, where the data gets put in the store, the view has the query and also know which data to extract from the response, then simply passes that to the component (as props or uses slots) - or we can open another ticket to plan how to implement  that for both components, if this approach with `watch` works better for websockets
3. I think it would be easier to review if this had only an initial implementation of the graph component and view. Without pan zoom, different algorithms, websockets realted code, and perhaps even without the tap feature on edges and nodes... these could be added later, and in case we need to investigate other implementations, it would also be easier to compare/revert/etc. I would focus more in achieving things like theming, matching the design sketch from @oliver-sanders, etc
4. Perhaps we could have `cylc/components/graph/` directory with modules like `options.js`, `styles.js`, etc, if that helps to separate the code?

This is just my opinion, and nothing of what is said here is consensus or a must-do @MartinRyan. More a comment so others can have a look at the code, and decide what to do next.

src/components/cylc/Graph.vue
hjoliver: We primarily want dagre (most like cylc-7 graphviz "dot" layout) but need to decide whether or not other layouts would be useful too (for full or partial graphs).  Let's leave it for the moment, I'll try to put some more thought into this PR tomorrow.

src/components/cylc/Graph.vue
hjoliver: These are definitely the big questions that we need to figure out. We do still need to investigate other libraries, particularly with consideration of our special requirements - that was always the intention, as we've noted a bunch of times.  I wonder why Cytoscape did styling this way?

hjoliver:
I agree with @kinow 's points above.  This PR is a bit too comprehensive given we do not yet know if Cytoscape can satisfy our primary/basic requirements (styling and data update in common with other views) and @MartinRyan seems to be suggesting that it can't (?).  Perhaps no other graph library can do what we want either, in which case we need to decide which compromises we can best live with ... but we just don't know yet. 

@MartinRyan - can you (if you are not doing so already) spend some time on these more fundamental questions?  It seems to me, we should be able to figure these things out with simple graphs (of live data) using several of the available graph libraries (no need to bother with panning and zooming etc. yet - they can presumably all do that).

hjoliver:
Just to add, some of the official cytoscape examples do look nice.  E.g. I can image that this sort of thing (even non-hierarchical) might be great for partial graphs around points of interest https://ivis-at-bilkent.github.io/cytoscape.js-fcose/demo-compound.html  ... however, I think we still have to figure out these fundamental issues before continuing with cytoscape.

src/components/cylc/Graph.vue
```diff
@@ -0,0 +1,1284 @@
...
+import Tippy from 'tippy.js'
+import getUuid from 'uuid-by-string'
+import VueCytoscape from '@/components/core/Cytoscape.vue'
+import { mixin } from '@/mixins/index'
+import { debounce, each, has, isEmpty, isUndefined } from 'lodash'
+// eslint-disable-next-line no-unused-vars
+// import graphservice from '@/services/graph.service'
+import { workflowService } from 'workflow-service'
+
+const QUERIES = {
```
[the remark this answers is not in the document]
MartinRyan: I have a graph service for websockets, but piggy backed on the workflowService when I had to change this to graphql requests as I didn't want multiple setIntervals firing for essentially the same data in workflows, but I can create a new graph service for requests a la workflowService if needs be?

src/components/cylc/Graph.vue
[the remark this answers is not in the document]
MartinRyan: The problem is the data is very nested. The Vuex recommendation is to have a flatter data structure, and as I have to reconstruct new node and edges data it works better to watch the store using `deep`. I'll take a look at i though :-)

src/components/cylc/Graph.vue
MartinRyan: You could probably have different stylesheet options linked to the different themes, as with changing graph layouts. The less that is computed in the stylesheet the greater the performance.

MartinRyan:
> I agree with @kinow 's points above. This PR is a bit too comprehensive given we do not yet know if Cytoscape can satisfy our primary/basic requirements (styling and data update in common with other views) and @MartinRyan seems to be suggesting that it can't (?). Perhaps no other graph library can do what we want either, in which case we need to decide which compromises we can best live with ... but we just don't know yet.
> 
> @MartinRyan - can you (if you are not doing so already) spend some time on these more fundamental questions? It seems to me, we should be able to figure these things out with simple graphs (of live data) using several of the available graph libraries (no need to bother with panning, zooming, collapsing, etc. yet - they can presumably all do that).

I'll move onto that phase tomorrow Hilary

src/components/cylc/Graph.vue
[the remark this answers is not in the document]
kinow: In that case, it could be that we both need to use `watch` + `deep`. The tree component also has a structure with multiple levels (tree hierarchy) :+1:

MartinRyan:
> Just to add, some of the official cytoscape examples do look nice. E.g. I can image that this sort of thing (even non-hierarchical) might be great for partial graphs around points of interest https://ivis-at-bilkent.github.io/cytoscape.js-fcose/demo-compound.html ... however, I think we still have to figure out these fundamental issues before continuing with cytoscape.

the current graph could do this using the cose-bilkent layout ;-)

src/components/cylc/Graph.vue
dwsutherland: > Can't we just use node.id when creating nodeObj? @dwsutherland can correct me if I am wrong, but in the whole nodes, each node will have an unique ID.

Yes all nodes and edges have unique IDs  (i.e. `sutherlander|noz|20190101T00|baa`), if they have the same ID then they are duplicate elements (circular query, or multiple queries) or reference IDs to the respective data element.

src/views/GScan.vue
oliver-sanders: It'll all change later so whatever works for now...

src/components/cylc/Graph.vue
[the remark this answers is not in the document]
MartinRyan: or have flatter data

src/components/cylc/Graph.vue
oliver-sanders: We probably only want Dagre, however, if there is a semi-usable alternative which can calculate complex layouts faster we may want to auto-switch for large graphs.

wxtim:
Tried it, and it worked very nicely for a simple suite:
```cfg
            graph = """
                Quokka => NUMBATS
                Numbat_a => Wambenger_a => Thylocene
                Numbat_b => Thylocene
                Numbat_c => Wambenger_b => Thylocene
            """
```
Unfortunately, when I tried it on a deliberately nasty suite I procedurally generated it didn't do my virtual machine much good at all and I had to reboot.
(I guess we haven't go max graph size limits sorted yet)

oliver-sanders:
> it didn't do my desktop much good at all and I had to reboot.

Do you know what took it out? Memory?

Would be good to find out if the issue is in the layout engine (presumably Dagre) or in the presentation layer.

hjoliver:
@MartinRyan - how can you look at your latest work, with labels and job icons?  Is that done on top of this branch?  (In which case, can you just update this PR branch?)

MartinRyan:
> @MartinRyan - how can you look at your latest work, with labels and job icons? Is that done on top of this branch? (In which case, can you just update this PR branch?)

Issue #238, opened 2019-09-18
cycle point order not sorted


hjoliver:
The GraphQL query on master (since yesterday) has:
```javascript
taskProxies(sort: { keys: ["cyclePoint"] })
```
 but the cycle points still aren't ordered correctly in the tree view:

![shot](https://user-images.githubusercontent.com/2362137/65088637-3f6cc100-da0e-11e9-9ff7-76aa09a514c2.png)

hjoliver:
We can bump this to 0.1 if it happens to be an easy fix (?)

kinow:
It should be easy to fix (I'm stuck with the leaf node... I should have done that in vuetify 2.x 😢). I will try to confirm using GraphiQL or maybe ask for help from @dwsutherland to check that the data is coming ordered as requested. If so, there must be something missing in the UI/Vue.js code that needs to be done to display it correctly.

dwsutherland:
Ah yes, use the snake_case version of `cyclePoint`  (`cycle_point`)  ... I could probably implement camel to snake on the back end though...

dwsutherland:
Also I should have set a default for `reverse` ... Heck I'll put a `cylc-flow` PR up...

kinow:
Aaahh. So it doesn't throw an error or warning in this case? That's a bummer, but looks simple to fix - hopefully

dwsutherland:
> Aaahh. So it doesn't throw an error or warning in this case? That's a bummer, but looks simple to fix - hopefully

Yeah, if the key is not a field of the data element.. then it ignores it.

PR #243, opened 2019-09-18
Order jobs per cycle point in mocked data

kinow:
Quick PR that does not change any source code, only a file that is generated by the `generate` script. Reason for this, is that the part of the query to order cycle points was not working (see #238).

Re-generated the mock data with the PR prepared by @dwsutherland (thanks!). This is useful only for screenshots and for future work, to make sure we don't confuse users - or ourselves with incorrectly ordered cycle points :)

Another 1 review issue IMO

Cheers
Bruno


kinow:
Here's what it looks like on `master` right now, FWIW:

image

PR #244, opened 2019-09-18
Use only Workflows table title

kinow:
From Riot today, @hjoliver 

> (...) the text in the grey box is a bit verbose and repetitive.  And the information is also redundant given the blue heading up top.
> (...)
> For the moment let's just make the existing text more concise.  i.e. "Your Workflows"

This PR changes the title text to "Your Workflows", removes the subtitle (text below it), and adjust margins after the change.

Here's before: 

image

Here's the end result:

image

Assigning @hjoliver to review as he requested the change. I think 1 review is enough, but otherwise anyone can feel free to jump in and have a second look at the change :+1:

### DIFF
--- a/src/components/material/Card.vue
+++ b/src/components/material/Card.vue
@@ -16,7 +16,7 @@
         :elevation="elevation"
         class="v-card--material__header d-flex align-center"
         dark
-        min-height="80"
+        min-height="60"
       >
         <slot
           v-if="!title && !text"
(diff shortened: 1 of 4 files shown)


kinow:
Here's how it looks when you have the subtitle, as in our UserProfile view (rest of the page missing because I am in the offline mode, and we have no mock for UserService).

image

kinow:
For what's worth, here's what it looks if we removed it:

image

I'm fine whichever way :slightly_smiling_face:

PR #245, opened 2019-09-18
Add tr to table row in Workflows/GScan view

kinow:
Reported by @hjoliver on Riot, but he mentioned @oliver-sanders was the one that spotted it

> (...)  have noticed one more problem. With two suites running, in the workflow list they now appear on the same line.

The image attached

Screenshot from 2019-09-18 17-41-40

Successfully reproduced locally

Screenshot_2019-09-18_19-55-13

Then confirmed that it was during the Vuetify migration. I missed adding a `<tr>`, so instead of having to rows (table row = tr), we get only a bunch of columns piled up together.

Screenshot_2019-09-18_19-57-33

Quick fix, here's the end result

Screenshot_2019-09-18_19-57-54

### DIFF
--- a/src/views/GScan.vue
+++ b/src/views/GScan.vue
@@ -45,19 +45,21 @@
               slot="item"
               slot-scope="{ item }"
             >
-              <td>{{ item.name }}</td>
-              <td>{{ item.owner }}</td>
-              <td>{{ item.host }}</td>
-              <td>{{ item.port }}</td>
-              <td class="justify-center">
-                <v-icon
-                    small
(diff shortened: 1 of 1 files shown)


hjoliver:
> but he mentioned @oliver-sanders was the one that spotted it 

No I didn't!  :rofl:

hjoliver:
(I see @wxtim is reviewing now, so I'll wait on the merge till he chimes in!)

wxtim:
Don't worry - 2 reviewers is more than enough for this.

PR #248, opened 2019-09-18
Remove fake-data Graph View and Cytoscape for alpha-1.

hjoliver:
(With apologies to @MartinRyan!)  The alpha1 release is to demonstrate a preview of the full system including initial view capability for real suites. A page showing a Cytoscape graph of fake data doesn't contribute much toward that goal and will likely confuse users/testers (why doesn't that look like my suite?...).  The relevant components etc. can easily be restored later if we go ahead with Cytoscape.  

@kinow - can you check that I haven't missed anything?  (No point in shipping unused bits).


kinow:
Sure. Will do some loval testing to confirm nothing is broken tomorrow morning.

hjoliver:
Tested in my environment and appears to be fully functional.

PR #249, opened 2019-09-18
Hide jobs by default in the Tree View

### LINKED ISSUE #247: Change default collapse/expand state for jobs in the tree view
From Riot chat with @hjoliver today

>can we default to jobs collapsed (with job icons in a row) - otherwise too busy with larger suites. This would be mitigated a bit by filtering, but as we don't have that yet, with jobs collapsed it is a more compact suite monitoring system where you can just drill down to the jobs if you need to.

kinow:
Closes #247 

Set to 0.2, so no need to merge/review now :+1: 

Changed instead of `expanded` in `TreeItem`, now it is `initialExpanded`, which denotes that that is only the initial state.

Then `beforeMount` is used to set the state to whatever was defined in the Vuex store. Meaning that before mounting the node in the DOM it will check if there is a desired `expanded` state in the data. If not, it will go with the `initialExpanded` :+1:

### DIFF
--- a/src/components/cylc/Tree.vue
+++ b/src/components/cylc/Tree.vue
@@ -7,7 +7,7 @@
         :node="workflow"
         :hoverable="hoverable"
         :min-depth="minDepth"
-        :expanded="expanded"
+        :initialExpanded="expanded"
         v-on:tree-item-created="onTreeItemCreated"
         v-on:tree-item-expanded="onTreeItemExpanded"
         v-on:tree-item-collapsed="onTreeItemCollapsed"
--- a/src/components/cylc/TreeItem.vue
+++ b/src/components/cylc/TreeItem.vue
@@ -79,7 +79,7 @@
           :depth="depth + 1"
           :hoverable="hoverable"
           :min-depth="minDepth"
-          :expanded="expanded"
+          :initialExpanded="initialExpanded"
           v-on:tree-item-created="$emit('tree-item-created', $event)"
           v-on:tree-item-expanded="$emit('tree-item-expanded', $event)"
           v-on:tree-item-collapsed="$emit('tree-item-collapsed', $event)"
@@ -119,7 +119,7 @@ export default {
       default: 0
     },
     hoverable: Boolean,
-    expanded: {
+    initialExpanded: {
       type: Boolean,
       default: true
     }
@@ -128,7 +128,7 @@ export default {
     return {
       active: false,
       selected: false,
-      isExpanded: this.expanded,
+      isExpanded: this.initialExpanded,
       leafProperties: [
         {
           title: 'host id',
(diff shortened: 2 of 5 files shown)


kinow:
GIFrecord_2019-09-19_095709

Tested on both offline mode and with real live workflow.

hjoliver:
In my opinion this is a significant improvement, especially when there are a lot of tasks.  And given that it is a small-ish code change to review and test, and we have not made the release yet, it should go in the alpha-1 release.

hjoliver:
It's an improvement because the default view is much cleaner and less overwhelming: just cycle points, and then task names with job icons lined up. Will post a good screenshot in a few minutes...

hjoliver:
@kinow - noticed one problem, probably not just on this branch - "latest message" is common to all jobs for a task (if a job fails and retries, the message will cycle from started through finished again in both job leaves as the second job progresses.  This might be coming from the back-end, not sure if @dwsutherland is online at the mo? - if so we should just remove the "latest message" item from the leaves for now.

kinow:
>@kinow - noticed one problem, probably not just on this branch - "latest message" is common to all jobs for a task...

Working on a different issue ATM, but will have a look at this and create a bug if I can reproduce it on my environment @hjoliver 👍

hjoliver:
Screenshot from 2019-09-19 11-43-15

@kinow - a good way to generate lots of failed-then-succeed jobs quickly:
```
[cylc]
   cycle point format = %Y
   [[parameters]]
       m = 0..5
       n = 0..2
[scheduling]
   initial cycle point = 3000
   [[graph]]
      P1Y = "foo[-P1Y] => foo => bar<m> => qux<m,n> => waz"
[runtime]
   [[root]]
      script = """
# sleep 15
# fail 50% of the time if try number is less than 5
if (( CYLC_TASK_TRY_NUMBER < 5 )); then
  if (( RANDOM % 4 < 2 )); then
     exit 1
  fi
fi"""
      [[[job]]]
          execution retry delays = 6*PT0S   # 6*PT5S
```
(With no sleep and zero retry delay, fast action!)

hjoliver:
(Note "latest message ... succeeded" on the failed task leaf above - symptom of aforementioned issue)

hjoliver:
Thanks for the review @MartinRyan - we can get this in now :grin:

dwsutherland:
> @kinow - noticed one problem, probably not just on this branch - "latest message" is common to all jobs for a task (if a job fails and retries, the message will cycle from started through finished again in both job leaves as the second job progresses. This might be coming from the back-end, not sure if @dwsutherland is online at the mo? - if so we should just remove the "latest message" item from the leaves for now.

The `latest_message` is a field of the `TaskProxy` not `Job`, and is updated in much the same way the old state summary did it:
```
       tproxy.latest_message = itask.summary['latest_message']
```
However, there will be some mechanism to update the `itask.summary['latest_message']` so no reason I why we can't add the field to each Job instead...

kinow:
>The latest_message is a field of the TaskProxy not Job, and is updated in much the same way the old state summary did it:

Oh, that makes sense. And in Cylc UI, we are simply setting the latest message in the job, to the same value from the task proxy.

```
job.latestMessage = taskProxy.latestMessage
```

image

So I think that explains why @hjoliver is seeing strange messages in the jobs.

PR #250, opened 2019-09-18
Avoid any calculation in print function when in production

kinow:
Small improvement for the `gquery` `print` function.

Noticed it the other day while working on some issues, and added a post-it so I wouldn't forget it.

When the mode is production, the function is currently doing some string manipulation, just to reach the end of the code and decide it won't print as it is production.

A better approach, IMO, is to just return immediately if production :+1:


kinow:
Set to 0.2. Small, one review should do.

PR #251, opened 2019-09-19
Collapse All toggle/switch

### LINKED ISSUE #246: Add Collapse All button to the tree view
From Riot chat with @hjoliver 

>can we have the collapse-all button? (you have the internals, right?)

kinow:
Closes #246

### DIFF
--- a/src/components/cylc/Tree.vue
+++ b/src/components/cylc/Tree.vue
@@ -33,6 +33,10 @@ export default {
     minDepth: {
       type: Number,
       default: 0
+    },
+    initialExpanded: {
+      type: Boolean,
+      default: true
     }
   },
   watch: {
@@ -50,7 +54,7 @@ export default {
       treeItemCache: new Set(),
       activeCache: new Set(),
       expandedCache: new Set(),
-      expanded: true,
+      expanded: this.initialExpanded,
       expandedFilter: null,
       collapseFilter: null
     }
--- a/src/components/cylc/Tree.vue
+++ b/src/components/cylc/Tree.vue
@@ -88,6 +88,7 @@ export default {
     },
     onTreeItemExpanded (treeItem) {
       this.expandedCache.add(treeItem)
+      this.$emit('tree-item-expanded', treeItem)
     },
     onTreeItemCollapsed (treeItem) {
       this.expandedCache.delete(treeItem)
--- a/src/components/cylc/Tree.vue
+++ b/src/components/cylc/Tree.vue
@@ -9,6 +9,7 @@
         :min-depth="minDepth"
         :initialExpanded="expanded"
         v-on:tree-item-created="onTreeItemCreated"
+        v-on:tree-item-destroyed="onTreeItemDestroyed"
         v-on:tree-item-expanded="onTreeItemExpanded"
         v-on:tree-item-collapsed="onTreeItemCollapsed"
         v-on:tree-item-clicked="onTreeItemClicked"
@@ -37,13 +38,6 @@ export default {
     initialExpanded: {
       type: Boolean,
       default: true
-    }
-  },
-  watch: {
-    workflows: {
-      handler () {
-        this.clearCaches()
-      }
     }
   },
   components: {
@@ -88,6 +82,7 @@ export default {
     },
     onTreeItemExpanded (treeItem) {
       this.expandedCache.add(treeItem)
+      this.expanded = true
       this.$emit('tree-item-expanded', treeItem)
     },
     onTreeItemCollapsed (treeItem) {
@@ -97,7 +92,14 @@ export default {
       this.treeItemCache.add(treeItem)
       if (treeItem.isExpanded) {
         this.expandedCache.add(treeItem)
+        this.$emit('tree-item-expanded', treeItem)
       }
+    },
+    onTreeItemDestroyed (treeItem) {
+      // make sure the item is removed from all caches, otherwise we will have a memory leak
+      this.treeItemCache.delete(treeItem)
+      this.expandedCache.delete(treeItem)
+      this.activeCache.delete(treeItem)
     },
     onTreeItemClicked (treeItem) {
       if (this.activable) {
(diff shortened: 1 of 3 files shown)


kinow:
Better leave this one for after #249 to check for conflicts.

@hjoliver I implemented it yesterday in 15 minutes... but then I commented (on GitHub or Riot I think?) that it wasn't that easy. But that's because I am trying to cover as possible corner cases... which may be worrying too much about this change.

The change is super simple. Just added a [`v-switch`](https://vuetifyjs.com/en/components/selection-controls#switches-boolean), which is a Vuetify 2.x (thankfully this is Vuetify 2.x now :tada: ).

Then if the user enable it, we call the previously developed `collapseAll` function. I covered one corner case in this PR, the one that annoyed me yesterday:

## User collapses all, and then manually expands one node

I found it weird if I enabled Collapse All, and then manually expanded a node. This would keep the switch in the position on/true for "Collapse All", even though the nodes are not all expanded.

Next time you click on the switch, it goes to expand all, and then it gets all weird as you lost track of what was expanded and what was collapsed.

This PR handles this case. If you expands any node after enabling "Collapse All", then it will change the status of the Collapse All :+1: 

# User manually collapses all

This is another corner case, that can be covered. We have a cache with what is expanded or not. So in theory we should be able to guess whether everything is collapsed or not, and update the switch state.

But will leave this one for later to avoid over-complicating this PR.

kinow:
Set to 0.2, matching the milestone of the issue related.

kinow:
## A new node is added to the tree

Ideally, IMO, if you selected "Collapse All", and a new node was added by polling/websockets, then that node must be collapsed too.

I believe it should be working, but needs to be tested too.

(@hjoliver this one deserves a bit more of testing than the other one for the job initial state IMO, especially if it behaves well with large-ish workflows :+1: no idea what's going to happen in that case)

kinow:
Hmmm, already found some weird behaviour when running the workflow `five`. This is a bit more risky to include in the 0.1 release as-is I think?

hjoliver:
OK, let's leave this for 0.2.  

It's not quite as critical anyway, with the default view having jobs collapsed after #249

kinow:
Problem fixed! And actually found an issue in the version on `master`.

The workflows watcher (remember, Vue docs say you rarely need a `watch`, turns out we also didn't need one here) was clearing the 3 caches we have, this way avoiding a memory leak.

However, the watched `Array` was not being completely re-created, only a few entries were being added/removed/updated.

This means we were clearing the whole cache, making it get out of sync with the watched `Array`. No impact on the visual output, the memory leak would disappear, but the Collapse All/Expand All functions would break, as they did here.

The fix is to listen to each node, and upon its removal, we simply remove from the caches. This way we have no memory leak, keep the caches synced with the data in the store.

It is ready for review for either 0.1 or 0.2 :+1: things to look for are the normal "does it work?", "does it break anything", and also "can I see any performance issue with this change?"

kinow:
Did some quick performance tests with my `five` workflow to confirm there was no regression (now I know where to look for that too, with Vue Dev tools first I check the cache sizes, then open incognito to take some snapshots).

- 14:13 24.4 MB
- 14:18 25.0 MB
- 14:23 25.3 MB
- 14:28 25.2 MB

No difference in the number of detached elements. No memory leak that I could identify with this simple workflow 🎉 the complex workflow has its own issue to tackle performance. But at least I don't think this PR should impact our current interface.

image

src/components/cylc/Tree.vue
```diff
@@ -84,6 +82,8 @@ export default {
     },
     onTreeItemExpanded (treeItem) {
       this.expandedCache.add(treeItem)
+      this.expanded = true
```
kinow: If any item is expanded, we know that the `Tree`'s `expanded` must be set to `true`. This will move the switch back to the off position, letting the user know the "Collapse All" is disabled now.

src/components/cylc/Tree.vue
```diff
@@ -112,11 +119,6 @@ export default {
           this.activeCache.add(treeItem)
         }
       }
-    },
-    clearCaches () {
-      this.treeItemCache.clear()
-      this.activeCache.clear()
-      this.expandedCache.clear()
```
kinow: This was removing all entries, without re-creating them (I thought they were being re-created). Now we are selective, and I also learned how/where to look for the size of the caches (which helps prevent performance issues).

hjoliver:
Seems to work well :+1: 

Now I think we need expand-all, collapse-all, and **collapse-jobs** (i.e. the new default view) though (sorry! :grimacing: )

kinow:
>expand-all, collapse-all

Ah, I see. The "Collapse All" switch might have been a bad idea. The switch is off by default.. once you click on it, it will call `tree.collapseAll()`... then if you turn it off by clicking again, it will call `tree.expandAll()`.

I think it makes more sense to have these two actions separated?

>and collapse-jobs (i.e. the new default view) though (sorry! grimacing )

Just jobs? Or would we have to allow users to choose something like "collapse/expand jobs|tasks|cycle points", or even by depth, or matching regex? We might have here an issue for 1.0 :rofl:

hjoliver:
Yeah separate might be good, if turning off "expand all" (say) restores the previous semi-collapsed state.

I was thinking just jobs at the moment. Because we now start up with the nice compact jobs-on-one-line view, but you can't get back to that after doing and expand-all.

kinow:
>but you can't get back to that after doing and expand-all.

Oh, good point. I would probably want that feature if I were a user with long running workflows with several 100s/1000s tasks.

hjoliver:
(But if separate, what if you turn on both collapse all and expand all? ... the entire universe might implode)

kinow:
>(But if separate, what if you turn on both collapse all and expand all? ... the entire universe might implode)

Very likely haha. Shall we close this issue and discuss how to implement this feature later?

There is just one fix for the tree component and the cache syncing that I would still like to get in for 0.2, so even if we close I may re-use some of this code and prepare a PR tomorrow.

hjoliver:
OK, closing for now (but make sure you don't lose the code you want to recycle :+1: )

Issue #227, opened 2019-09-16
Memory leak in Tree/TreeItem after running complex workflow


kinow:
See #222 and #225 for background.

For some strange reason the workflow `five` doesn't leave anything behind. But the `complex`, after running ~12 minutes left a few `HTMLDivElement` detached in the heap.

![Screenshot_2019-09-16_14-43-51](https://user-images.githubusercontent.com/304786/64932153-330f2980-d891-11e9-9a79-abe62a3c80b7.png)

kinow:
I created a branch and cherry-picked the commit from #251 PR. Then tested the workflow `five`.

![Screenshot_2019-09-24_09-02-15](https://user-images.githubusercontent.com/304786/65465644-f060d900-deb0-11e9-9edf-3edf07bdccf2.png)
![Screenshot_2019-09-24_09-01-44](https://user-images.githubusercontent.com/304786/65465646-f060d900-deb0-11e9-9d23-1a74a875ed79.png)
![Screenshot_2019-09-24_09-00-22](https://user-images.githubusercontent.com/304786/65465647-f0f96f80-deb0-11e9-85a8-9a1c109b657e.png)

Memory stayed within 24-27 MB before/after the tests. Which is fine.

The issue is still with the `complex` workflow. It now gave me a relatively OK handsaw pattern, and initial memory of 184MB at 09:15AM.

![Screenshot_2019-09-24_09-14-58](https://user-images.githubusercontent.com/304786/65465754-2900b280-deb1-11e9-8a76-491aff515e5a.png)

Didn't touch the VM for 10 minutes, then at 09:25 AM decided to take another snapshot. The browser stopped the snapshot at 39% progress, and then the VM became unresponsive, and I had to restart it :roll_eyes: .

kinow:
Not really a blocker for 0.2, so if necessary can be moved to 0.3 or 1.0 👍

kinow:
This issue hasn't occurred to me in a long time. I used to get browser tabs crashing with complex, and browser used memory going over the rooftop. Looking at the memory heap in the browser I'd see things like near 1GB for Cylc UI tab after running multiple workflows in a day.

But for the past few months since we have been improving the tree view (exploring infinite-tree, deltas, etc) this hasn't happened.

Closing :tada:

PR #254, opened 2019-09-24
Keep cache in sync, and use a global event bus

kinow:
Part of #222 and #227 

The first commit was cherry-picked from the #251 PR. This commit removes code that was `watch`'ing a data structure and clearing all caches. Now it listens to only one event for when the node is deleted, and then removes that treeitem from the cache.

Confirmed in the Vue Dev Tools browser extension that the cache was being updated correctly.

The second commit is the reason for this PR being WIP. I noticed we had a lot of `closure` object in the heap. And even some events in the detached state.

My guess was that a) using closures like `@on:click='$emit(...)'` was triggering the events too soon, or maybe invoking the events for parents too, and b) having to propagate the event back to parents wasn't efficient.

Now we have a global event bus for the tree component, which is [common in Vue](https://alligator.io/vuejs/global-event-bus/). Normally the global event is a solution when projects don't want to adopt Vuex. In our case we have Vuex, but I don't think we need to put a component's cache there (nobody else is going to use it).

I am running tests in the complex workflow now, but with the browser on my Windows host machine, and the workflow and hub running on my Linux VM.

Started at 3:36 PM, with 67.1 MB. It went up to 180, 240, 320, **and then went down to 214MB**. It later went down again to **127MB**, and back to ~400, then down again to 127MB.

Right now I looked at it, and it was near 800MB, then went down to 200, back to 342 MB now. I will leave it running until tomorrow, which should be enough time for the `complex` workflow to complete.

And also enough time to check the situation of my VM and host machine, and see if the browser didn't hang, if the garbage collector collected everything, and that once the workflow is done, the browser doesn't display anything and there are no extraneous elements detached in the heap.

Bruno

### DIFF
--- a/src/components/cylc/Tree.vue
+++ b/src/components/cylc/Tree.vue
@@ -9,6 +9,7 @@
         :min-depth="minDepth"
         :initialExpanded="expanded"
         v-on:tree-item-created="onTreeItemCreated"
+        v-on:tree-item-destroyed="onTreeItemDestroyed"
         v-on:tree-item-expanded="onTreeItemExpanded"
         v-on:tree-item-collapsed="onTreeItemCollapsed"
         v-on:tree-item-clicked="onTreeItemClicked"
@@ -33,13 +34,6 @@ export default {
     minDepth: {
       type: Number,
       default: 0
-    }
-  },
-  watch: {
-    workflows: {
-      handler () {
-        this.clearCaches()
-      }
     }
   },
   components: {
@@ -84,6 +78,8 @@ export default {
     },
     onTreeItemExpanded (treeItem) {
       this.expandedCache.add(treeItem)
+      this.expanded = true
+      this.$emit('tree-item-expanded', treeItem)
     },
     onTreeItemCollapsed (treeItem) {
       this.expandedCache.delete(treeItem)
@@ -92,7 +88,14 @@ export default {
       this.treeItemCache.add(treeItem)
       if (treeItem.isExpanded) {
         this.expandedCache.add(treeItem)
+        this.$emit('tree-item-expanded', treeItem)
       }
+    },
+    onTreeItemDestroyed (treeItem) {
+      // make sure the item is removed from all caches, otherwise we will have a memory leak
+      this.treeItemCache.delete(treeItem)
+      this.expandedCache.delete(treeItem)
+      this.activeCache.delete(treeItem)
     },
     onTreeItemClicked (treeItem) {
       if (this.activable) {
@@ -112,11 +115,6 @@ export default {
           this.activeCache.add(treeItem)
         }
       }
-    },
-    clearCaches () {
-      this.treeItemCache.clear()
-      this.activeCache.clear()
-      this.expandedCache.clear()
     }
   }
 }
--- a/src/components/cylc/Tree.vue
+++ b/src/components/cylc/Tree.vue
@@ -8,11 +8,6 @@
         :hoverable="hoverable"
         :min-depth="minDepth"
         :initialExpanded="expanded"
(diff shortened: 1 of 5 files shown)


src/components/cylc/Tree.vue
```diff
@@ -35,13 +32,6 @@ export default {
       default: 0
     }
   },
-  watch: {
-    workflows: {
-      handler () {
-        this.clearCaches()
```
kinow: This was clearing all cache. Even if we had expanded elements, or previously created elements, everything was destroyed, moving the cache to invalid states.

kinow:
While writing a comment here, at 3:52PM, the browser screen went blank, though memory was 452 MB (it's varying between 900 and 200 now). Then looked at the hub, and found several exceptions:

image

(@hjoliver @dwsutherland I suspect it was simply due to overload on my VM, and it maybe wouldn't happen in the HPC... or it may be related to https://github.com/cylc/cylc-uiserver/issues/65)

src/components/cylc/Tree.vue
```diff
@@ -55,6 +45,13 @@ export default {
       collapseFilter: null
     }
   },
+  created () {
+    TreeEventBus.$on('tree-item-created', this.onTreeItemCreated)
+    TreeEventBus.$on('tree-item-destroyed', this.onTreeItemDestroyed)
+    TreeEventBus.$on('tree-item-expanded', this.onTreeItemExpanded)
+    TreeEventBus.$on('tree-item-collapsed', this.onTreeItemCollapsed)
+    TreeEventBus.$on('tree-item-clicked', this.onTreeItemClicked)
```
kinow: Instead of propagating, we can simply listen to the events globally here.

kinow:
(Note-to-self: call `.$off` tomorrow when elements are destroyed, to clear the event listeners.... this may be impacting the performance of the current test, though at least the heap memory is doing OK)

src/components/cylc/tree/event-bus.js
```diff
@@ -0,0 +1,3 @@
+import Vue from 'vue'
+
+export const TreeEventBus = new Vue()
```
matthewrmshin: It is this simple? Wow!

kinow: Yup. From what I remember (no access to code right now), the components we create become `VueComponent`, which extend `Vue`.

In the same way that within a `VueComponent` we are able to simply do a `this.$emit`, and call `@on:click=this.method`, we can instead create a new instance of `Vue` (parent of `VueComponent`) and use just the event handling part.

kinow:
Computer logged in normally this morning, CPU fans were off. The `complex` workflow` had completed with no issues. JupyterHub also running OK.

Took another memory snapshot of the now blank page.

06:48 AM 154 MB

Waited a bit, forced garbage collect, and took a second snapshot

06:56 AM 154 MB

And a few detached div/span/SVG (Cylc components are the only ones adding SVG's to DOM).

image

kinow:
Leaving the workflows page and going to dashboard, it reduced the heap to 14.5 MB. So there is still a memory leak after/during the render of the complex workflow.

However, the caches are in sync after this PR. The global event bus also reduced the number of events and memory. So it should be ready for review now.

kinow:
To compare the number of events, have a look at `master`, with 44 normally for every update on the workflow `five`.

image

Now switching to this branch, most updates now have only 12 events (for created/expanded/destroyed nodes)

image

kinow:
@MartinRyan could you review this one whenever you have some spare time, please? Thanks!

kinow:
Rebased to pick up latest changes from `master`.

kinow:
Rebased and pushed one more commit with the changelog.

Issue #255, opened 2019-09-24
Fix coverage on Travis/codecov


kinow:
Badge reporting 0% now, I think I tried running coverage last week and it also report 0%. But I was on a branch with lots of changes and ignored as I thought it could be something with my code. And I could still use the IDE to report it.

Issue #234, opened 2019-09-17
Use query variables in GraphQL instead of string replace


kinow:
Right now our GraphQL queries in the workflow service start with

```graphql
      {
        workflows(ids: ["WORKFLOW_ID"]) {
          id
          name
          status
```

Then later in the code we 

```js
            workflowService.subscribe(
              this,
              QUERIES[queryName].replace('WORKFLOW_ID', this.workflowId)
            )
```

I am quite sure that actually we should use GraphQL query variables instead.

kinow:
Example from GraphQL docs: https://graphql.org/graphql-js/passing-arguments/

And from Apollo: https://www.apollographql.com/docs/react/essentials/queries/#caching-query-results

PR #259, opened 2019-09-25
Add unit tests for GQuery

kinow:
Started working on #234 today, which I thought would be a 30 minutes job. But then got stuck trying to add the variables to the query, and noticed I was changing a lot of code in `gquery.js`.

I realized it didn't have unit tests yet, so started adding a few ones here. Will add a few more tomorrow, but I digressed a little today (2 hours) trying to get the coverage to work (#255) with no luck.

### DIFF
--- a/tests/unit/services/gquery.spec.js
+++ b/tests/unit/services/gquery.spec.js
@@ -1,0 +1,51 @@
+import { expect } from 'chai'
+import { GQuery } from '@/services/gquery'
+import { parse } from 'graphql/language/parser'
+
+describe('GQuery', () => {
+  describe('constructor', () => {
+    it('should create an object correctly', () => {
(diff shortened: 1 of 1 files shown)


kinow:
Thanks for the review @MartinRyan ! Rebasing onto master to get coverage reports working (hopefully :crossed_fingers: )

kinow:
Codecov web interface shows +16%. Just waiting until their bot updates the comment here, then I believe one approval should be fine for the change as-is. Will prepare another PR later to add more tests for GQuery (easier this way with smaller iterations :+1: )

image

kinow:
Bot updated correctly saying it will increase to 16%, one approval should do (not changing files, just adding unit tests), so merging :tada:

PR #266, opened 2019-09-25
Color cylc logo.

hjoliver:
The new UI is looking pretty slick, but a bit **too grey** IMO.  Besides @oliver-sanders latest design doc does use the color logo: https://github.com/cylc/cylc-ui/issues/87#issuecomment-497382663

The Inkscape-generated color SVG file is stored on cylc/cylc-admin; I hand-edited the file to remove all the extraneous cruft for this PR.

@kinow - two questions:
- the UI is using inlined-SVG in `Header.vue` - do we need `public/img/logo.svg` as well?
- if we do, should we keep the grey-scale version too?  (or I could store that alongside the color one on cylc-admin, for safe-keeping).

shot


kinow:
>the UI is using inlined-SVG in Header.vue - do we need public/img/logo.svg as well?

No, but I added that image to be used by the JupyterHub - if I remember well.

kinow:
>if we do, should we keep the grey-scale version too? (or I could store that alongside the color one on cylc-admin, for safe-keeping).

If we won't use the logo for JupyterHub, then we can remove both. If we will use it, then I vote to keep only the file actually used and then move it to cylc-admin as you said.

kinow:
One review will do, or shall we wait for @oliver-sanders to have time to look at this too @hjoliver ?

hjoliver:
(oh yeah, I forgot about the hub logo usage).

One review will do, since it conforms to the design drawings.

PR #267, opened 2019-09-26
Fix coverage

### LINKED ISSUE #255: Fix coverage on Travis/codecov

kinow:
Closes #255 

Had to play with the settings a bit. First commit is re-enabling the coverage. Next commit is to match that istanbul plugin documentation, making it easier to check for differences. Then we reformat the JSON in the next commit, and finally had to enable `instrument` and `sourceMap`, as otherwise it would report only on covered file (i.e. exclude not covered files, giving a coverage report number of ~94%).

### DIFF
--- a/.nycrc
+++ b/.nycrc
@@ -18,5 +18,7 @@
   "reporter" : ["html", "lcov", "text-summary"],
   "cache" : false,
   "all" : true,
-  "check-coverage" : true
+  "check-coverage" : true,
+  "instrument" : false,
+  "sourceMap" : false
 }
--- a/.nycrc
+++ b/.nycrc
@@ -16,6 +16,6 @@
   "cache": false,
   "all": true,
   "check-coverage": false,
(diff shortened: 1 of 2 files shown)


kinow:
A good way to review this, would be to wait for Travis CI and Codecov. If it reports anything different than 0% (currently reported on master) then I'd say it's OK to be merged :grimacing:

kinow:
Or run `npm run coverage:unit` then look at the generated HTML under `./coverage`:

image

kinow:
There you, @codecov-io reported coverage increasing. Ready for review.

kinow:
Can you take a look, please, @hjoliver ? One review should do IMO.

PR #273, opened 2019-10-01
Replaces usages of `var` with `const`

### LINKED ISSUE #271: Use const or let instead of var
We have a couple of files that need some quick updates, replacing `var` by `const` or `let`.

- `gquery.js`
- `vue.config.js`

These should be quite easy to locate, or WebStorm can be used to quickly locate which ones need fixing, or VSCode or another IDE or a linter.

EDIT: just to note that one review should do. Minor change, the unit tests will cover some of what will be changed, but a quick manual test with env=offline won't hurt :+1:

kreatemore:
Fixes #271

Ran unit tests, and both `build` and `build:report` - everything seemed fine (except for some `console` usages reported by eslint).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,6 +39,7 @@ below.
  - David Sutherland
  - Martin Ryan
  - Tim Pillinger
+ - Alex Szabó
 
 (All contributors are identifiable with email addresses in the git version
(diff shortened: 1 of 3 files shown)


kinow:
(just waiting for Travis before merging)

PR #274, opened 2019-10-01
changed the var to let and const on the files gquery.js and vue.confi…

BeatRoy:
I have replaced var on the files `gquery.js` and `vue.config.js`.


kinow:
Hi @BeatRoy , this has just been fixed. Also, it would be better to use a `const` I think, as we are not changing the values again :+1: Take a look at the other open issues, or feel free to suggest other improvements. We have things like JSDoc, review the existing (though little) documentation, add more tests, etc, that do not have open issues yet.

PR #275, opened 2019-10-01
Fix size of dashboard links on hover

ashnewport:
Remove `row` prop from `<v-layout>` to align hover with sibling elements and maximise use of space.

Closes #270

PR #277, opened 2019-10-01
Upgrading Vuetify, testing material components

### LINKED ISSUE #276: Update to Vuetify 2.1
Vuetify announced the 2.1 release. We recently moved from 1.5 to 2.0. This next update should be much simpler. Change log: https://github.com/vuetifyjs/vuetify/releases/v2.1.0

kylerits:
-- Hacktoberfest contribution

Closes #276


src/views/Dashboard.vue
```diff
@@ -210,4 +210,21 @@ export default {
 .v-list__tile {
   padding: 10px 0;
 }
+
+.v-list-resize-left, 
+.v-list-resize-right {
+  padding: 0;
+  margin: 0 12px;
```
kinow: I think this has been fixed in https://github.com/cylc/cylc-ui/pull/275, just missing @ashnewport to add his name to contributors before being merged.

kylerits:
I changed the PR to just address the upgrade to Vuetify 2.1

src/views/Dashboard.vue
```diff
@@ -1,18 +1,15 @@
 <template>
-  <v-container
-    fluid
-    grid-list
-    class="c-dashboard mt-4"
-  >
+  <v-container fluid grid-list class="c-dashboard mt-4">
```
kinow: I think there's no change on this line, just formatting, could you undo the formatting changes, please? This way we can compare this diff again in the future when upgrading Vuetify again. So if we can have the bare minimum of what was required to upgrade from 2.0 to 2.1, that'd be helpful :+1:

src/views/Dashboard.vue
```diff
@@ -1,18 +1,15 @@
...
     <v-layout wrap>
       <v-flex xs6 md6 lg3>
         <p class="display-1">Workflows</p>
         <!-- TODO: link with data from the query -->
         <v-data-table
-            :headers="workflowsHeader"
-            :items="workflows"
-            hide-default-footer
-            hide-default-header>
+          :headers="workflowsHeader"
```
kinow: I think here too, it's only formatting?

src/views/Dashboard.vue
```diff
@@ -24,10 +21,11 @@
...
         <p class="display-1">Events</p>
         <v-data-table
-            :headers="eventsHeader"
-            :items="events"
-            hide-default-footer
-            hide-default-header>
+          :headers="eventsHeader"
+          :items="events"
+          hide-default-footer
+          hide-default-header
```
kinow: Formatting only?

src/views/Dashboard.vue
```diff
@@ -49,25 +47,17 @@
...
             </v-list-item-avatar>
             <v-list-item-content>
-              <v-list-item-title class="title font-weight-thin">
-                Settings
-              </v-list-item-title>
-              <v-list-item-subtitle>
-                View your Hub permissions, and alter user preferences
-              </v-list-item-subtitle>
+              <v-list-item-title class="title font-weight-thin">Settings</v-list-item-title>
+              <v-list-item-subtitle>View your Hub permissions, and alter user preferences</v-list-item-subtitle>
```
kinow: From what I could tell classes and tags are the same, so no need to change this one as well.

src/views/Dashboard.vue
```diff
@@ -49,25 +47,17 @@
...
             </v-list-item-avatar>
             <v-list-item-content>
-              <v-list-item-title class="title font-weight-thin">
-                Cylc Hub
-              </v-list-item-title>
-              <v-list-item-subtitle>
-                Visit the Hub to manage your running UI Servers
-              </v-list-item-subtitle>
+              <v-list-item-title class="title font-weight-thin">Cylc Hub</v-list-item-title>
+              <v-list-item-subtitle>Visit the Hub to manage your running UI Servers</v-list-item-subtitle>
```
kinow: No need to change this one either I think.

src/views/Dashboard.vue
```diff
@@ -120,7 +100,7 @@
 </template>
 
 <script>
-import { mixin } from '@/mixins/index'
+import { mixin } from '@/mixins/index';
```
kinow: This actually won't pass the linter. It's configured to fail on semi-colons...

kylerits:
I copied the Dashboard.vue file directly from the master branch. There should not be any changes to it. Package.json and package-lock.json were changed minimally to account for the upgrade to Vuetify 2.1

src/views/Dashboard.vue
```diff
@@ -41,7 +41,7 @@
       </v-flex>
     </v-layout>
     <v-divider />
-    <v-layout row wrap>
```
kinow: Hi @kylerits 

This has already been merged to master - https://github.com/cylc/cylc-ui/blob/46c757b92e9c344f2dcbc9c2ad1ca348dcc2f195/src/views/Dashboard.vue#L44

So it shouldn't appear here on your PR. You need to rebase onto master, dropping the commit where you added this file. Leaving only `package.json` and `package-lock.json` as changes here :slightly_smiling_face:

kinow:
While testing another PR for the same issue, I realized the upgrade isn't as simple as I expected. Some components' styles are actually broken after the upgrade, so I will have to go through each component used and compare making sure it doesn't break the rendered layout. Sorry @kylerits

PR #279, opened 2019-10-02
Removed unused theme components

### LINKED ISSUE #272: Remove unused theme components
We have now upgraded to Vuetify 2.x. In the process, we copied old files as well as new files. Among these files, there are some components used by the creative-tom's theme, but that are not used in our project.

So we can safely remove them, and add if necessary. They are:

- `src/material/Card.vue`
- `src/material/Notification.vue`
- `src/material/StatsCard.vue`

While these components are nice, relying on them would mean having to keep an eye on updates on both Vuetify and creative-tom's theme. So far we've been able to use just Vuetify, so I believe it would be safer to remove these for now :+1:

garrlker:
I ran unit tests, lightly tested in the browser, and checked the console for errors.
Everything seems to be running just fine!

Closes #272


kinow:
Excellent @garrlker ! Let me do a quick review. One review should do as this is a small change.

And thank you for taking your time to prepare this contribution :+1:

kinow:
@garrlker can you take a look at `CONTRIBUTING.md` and, if you are OK with that, please add your name to the contributors list? You can just push to this PR adding the extra change.

src/components/material/Card.vue
```diff
@@ -1,116 +0,0 @@
-<template>
```
kinow: Oh noes! Found that this one is actually used. Could you revert this change, please @garrlker ? My bad.

Here's where it is used: https://github.com/cylc/cylc-ui/blob/master/src/views/GScan.vue#L14

Found after seeing a browser console error.

kinow: Screenshot: 

image

kinow:
(just pushed to master, with one change to CONTRIBUTING.md, might be a good idea a quick rebase onto master :+1: )

src/components/material/Card.vue
garrlker: Oh dang!

I just put it back and added myself to the `contributing.md`.
I also went ahead and resolved a merge conflict with `contributing.md`.

Everything should be good to go! 👍

Issue #226, opened 2019-09-15
WorkflowService produces exception if there is no query to execute


kinow:
The current design in `WorkflowService` demands that once initialized, the singleton will keep running queries in the polling way (soon to be changed).

However, if after the singleton has been created, the user navigates to a page without the need for any GraphQL query, what happens is that the `WorkflowService` is still alive and asking the ApolloClient to execute queries.

But without a query ApolloClient throws an error as reported in #224 

Simply adding an `if` or trying to stop the polling could fix it. But I think we may need to actually improve this design a bit. If you create the singleton when looking at `/#/workflows/`, but then decides to spend the rest of the day on a page like `/#/bee` that produces 404, it appears that the `WorkflowService` instance is still around and live, even though it is not necessary.

hjoliver:
I vaguely recall we discussed something like this in Exeter: if user navigates away from a workflow, keep the subscription for say 30 seconds in case they quickly come back, else delete it (and recreate from scratch, as if for first visit, if they come back again later than that).  By extension, if there is no subscription at all (no current workflow views) then shut down (/hold?) the WorkflowService completely after 30 seconds.

kinow:
>I vaguely recall we discussed something like this in Exeter: if user navigates away from a workflow, keep the subscription for say 30 seconds in case they quickly come back, else delete it (and recreate from scratch, as if for first visit, if they come back again later than that). By extension, if there is no subscription at all (no current workflow views) then shut down (/hold?) the WorkflowService completely after 30 seconds.

Oh, alright. That's tricky, we may have to move this requirement to an issue. And fix this issue for 0.2 preventing it from throwing the error/exception.

kinow:
Issue gone! It was probably fixed either in the recent commits that fixed WebSockets, or some other earlier commit.

![image](https://user-images.githubusercontent.com/304786/72784414-a4bb1200-3c8d-11ea-8c1a-46b2eb00cd6a.png)

It's also possible to view the number of subscriptions in the console.

![image](https://user-images.githubusercontent.com/304786/72784501-d633dd80-3c8d-11ea-945a-096d73ca7d19.png)

Then visiting a page that doesn't need any query or subscription

![image](https://user-images.githubusercontent.com/304786/72784535-e8ae1700-3c8d-11ea-9104-4fa46878a68b.png)

Issue #164, opened 2019-08-02
Check how to reduce size of produced artefacts


kinow:
The vuejs threshold for raising a warning is around 244 kb. We have over 1 mb now, so we will have to try finding what is going on.

It could be that we need to filter what our theme is importing from Vuetify. Or could be Cytoscape (I remember seeing this warning before the Graph view, so probably not that). Could be the icons or fonts, etc.

kinow:
Checked the browser coverage tool already, and ~15% of the code loaded is not loaded. Looks like some Vuetify components are never used but still loaded by webpack.

Vue UI can probably cast some light on the issue.

![image](https://user-images.githubusercontent.com/304786/62355706-a4a24b00-b563-11e9-8504-0664bec3676f.png)

kinow:
After using A La Carte, the size is better, but still above the threshold.

![Screen Shot 2019-08-11 at 11 55 43-fullpage](https://user-images.githubusercontent.com/304786/62828016-1f3c2c00-bc30-11e9-9add-4cc720e47e1b.png)

```diff
diff --git a/src/plugins/vuetify.js b/src/plugins/vuetify.js
index bceb629..1368153 100644
--- a/src/plugins/vuetify.js
+++ b/src/plugins/vuetify.js
@@ -1,10 +1,89 @@
 import Vue from 'vue'
-import Vuetify from 'vuetify'
+import Vuetify, {
+  VAlert,
+  VApp,
+  VBtn,
+  VCard, // TBD: Card* used only by NotFound.vue, could be removed later?
+  VCardText,
+  VCardTitle,
+  VCheckbox,
+  VChip,
+  VContainer,
+  VContent,
+  VFadeTransition,
+  VDataTable,
+  VDivider,
+  VFlex,
+  VFooter,
:...skipping...
diff --git a/src/plugins/vuetify.js b/src/plugins/vuetify.js
index bceb629..1368153 100644
--- a/src/plugins/vuetify.js
+++ b/src/plugins/vuetify.js
@@ -1,10 +1,89 @@
 import Vue from 'vue'
-import Vuetify from 'vuetify'
+import Vuetify, {
+  VAlert,
+  VApp,
+  VBtn,
+  VCard, // TBD: Card* used only by NotFound.vue, could be removed later?
+  VCardText,
+  VCardTitle,
+  VCheckbox,
+  VChip,
+  VContainer,
+  VContent,
+  VFadeTransition,
+  VDataTable,
+  VDivider,
[...]

PR #280, opened 2019-10-07
Add subscriptions

### LINKED ISSUE #207: Update components' data structure incrementally via websockets
From: https://github.com/cylc/cylc-ui/pull/203#discussion_r319464561

When we add WebSockets to Cylc UI Server backend, and to Cylc UI, we need to review the data structures updates.

kinow:
Closes #207 
Closes #266 (noticed this while testing this PR, will add a comment where it fixes this issue)

This pull request adds subscriptions to Cylc UI. The `ApolloClient` has a feature called __link__, which defines the transport link. It can be an HTTP link, a WebSockets Link, a mocked link, or whatever transport link you need to transport the data.

We were using only an `HTTPLink` in the `ApolloClient`. Now the link is actually a `split` function (from `apollo-link` module).

This function acts as a ternary operator, and will decide based on the query definition which link to use. If the query definition specifies it is a subscription, it will use a `WebSocketLink` (from new package added in this PR `apollo-link-ws`), otherwise it will use the existing `ApolloLink` (from `apollo-link-http`).

Unit tests passing on my environment, managed to fix IDE issues, now it is only missing to make it work with live workflow.

Sibling PRs:
- https://github.com/cylc/cylc-flow/pull/3390
- https://github.com/cylc/cylc-uiserver/pull/82


kinow:
Build is passing, tests passing fine. Spent most of the morning working around dependencies that needed to be imported or removed, then rest of morning and part of afternoon fixing tests.

Now the application is loading fine, but the `GScan.vue` view when accessed - first place where Vue.js will actually try to use WebSockets - is failing with:

image

Not sure why. Could be because we are using the `http` protocol instead of `ws` (the WebSockets are being served by Tornado, which I think is expecting it to be http). But having graphiql working helps a little as we know it should be possible to talk from this browser to the server. 4PM, not likely to find a solution today, but it may be working tomorrow morning and the PR would be ready for review (this is the last part required to update the Vuex data structure via WebSockets)

kinow:
Oh, actually solved the problem. I think it's about 90% done now.

The first issue was that I had to use `ws://` as in the GraphiQL template (extremely helpful, whew).

The second issue was that looks like the `ApolloClient` doesn't use the URI from the links... if you don't specify a URI for the client, when using the HTTP Link it will create a dummy URL simply appending `/graphql` to the current server URL.

Both fixed, now needs just a modification in the `WorkflowService` to remove the polling and, instead, use WebSockets with a `subscription` query. Will start working on that tomorrow morning with :coffee: and more calm.

kinow:
Apparently done. Happy to take any reviews now, but can't remove the Draft mark, as this PR if merged without the WebSockets on the backend, it would break Cylc UI.

For me it looks much faster now with WebSockets, and the application is more interactive for users. So if curious, have a try. Hopefully it will work well as long as the other 2 PR's for backend are checked out and built in a venv 🤞

package.json
```diff
@@ -44,10 +43,16 @@
     "@vue/cli-plugin-unit-mocha": "^3.11.0",
     "@vue/cli-service": "^3.11.0",
     "@vue/test-utils": "^1.0.0-beta.29",
+    "apollo-cache-inmemory": "^1.6.3",
+    "apollo-client": "^2.6.4",
+    "apollo-link": "^1.2.13",
+    "apollo-link-http": "^1.5.16",
+    "apollo-link-ws": "^1.0.19",
```
kinow: These were dependencies of ApolloBoost, a helper that simplifies creating Apollo clients. However, when you provide a link that can handle both HTTP and WebSocket, ApolloBoost ignores it and unless you provide a URI, it will create and use an HTTP Link, simply ignoring the WebSocket.

kinow: https://github.com/apollographql/apollo-client/blob/9c19c5810a8908d7034a384d06fdb76eec02b3b6/packages/apollo-boost/src/index.ts#L164

Where it shows that ApolloBoost is not compatible with WebSockets. Hence the removal, and the need to add these dependencies.

package.json
```diff
@@ -44,10 +43,16 @@
...
     "axios": "^0.19.0",
     "babel-eslint": "^10.0.3",
     "chai": "^4.2.0",
     "cross-env": "^5.2.0",
+    "cross-fetch": "^3.0.4",
```
kinow: We were using the node-fetch for tests, but it didn't work after removing the default ApolloClient (which would prevent Apollo and its libraries of trying to validate that the fetch global existed. This polyfill provides the global fetch for tests.

src/services/workflow.service.js
```diff
@@ -29,6 +31,70 @@ class LiveWorkflowService extends GQuery {
...
+  request () {
+    /**
+     * Perform a REST GraphQL request for all subscriptions.
+     */
+    if (process.env.NODE_ENV !== 'production') {
+      console.debug('graphql request:', this.query)
+    }
+    if (!this.query) {
+      return null
+    }
```
kinow: This `if` above fixes #226

src/services/workflow.service.js
```diff
@@ -29,6 +31,70 @@ class LiveWorkflowService extends GQuery {
...
+    const vm = this
+    return this.apolloClient.subscribe({
+      query: this.query,
+      fetchPolicy: 'no-cache'
+    }).subscribe({
+      next (response) {
```
kinow: We override `request` here because we call `.subscribe`, which returns an `Observable`. The next `.subscribe` is - by coincidence - a method used to listen whenever the observable is updated.

Without overriding it, we would break the existing behaviour of the `GQuery` and, consequentially, break `LifeWorkflowService` with no necessity.

src/services/workflow.service.js
```diff
@@ -29,6 +31,70 @@ class LiveWorkflowService extends GQuery {
...
+          'setAlert',
+          new Alert(err.message, null, 'error')
+        )
+      }
+    })
+  }
+}
 
-export { workflowService }
+export { LiveWorkflowService, SubscriptionWorkflowService }
```
kinow: We were exporting a singleton, that once imported would exist for the whole app.

Now we export classes, let users decide whatever they want to do with these (mock, test, extend, instantiate, create a singleton, etc).

Vue has a way to create application globals via `Vue.prototype.$applicationSingleton`, which has effectively the same result as exporting a singleton, with more flexibility if necessary.

Plus, it will be available to the whole application as `this.$workflowService`. The `$` is per convention.

ps: must be used with care this global objects in Vue, as they docs state

src/utils/graphql.js
```diff
@@ -1,35 +1,52 @@
...
+import { WebSocketLink } from 'apollo-link-ws'
+import { getMainDefinition } from 'apollo-utilities'
+import { InMemoryCache } from 'apollo-cache-inmemory'
 
-export function createApolloClient (uri) {
-  const authMiddleware = new ApolloLink((operation, forward) => {
-    operation.setContext({
-      headers: {
-        // FIXME: this is the random generated password, update every time it is generated for now!
-        Authorization: 'Basic Y3lsYzpiaFRTbnRpWjRqR2YzSWR3eTlVMg=='
```
kinow: From the old prototype, I suspect the PR for Cylc UI Server auth may require something like this, but we can cross that bridge when time comes, no need to leave an old Cylc Suite passphrase hard-coded here.

src/utils/graphql.js
```diff
@@ -1,35 +1,52 @@
...
-  fetch: buildAxiosFetch(axios)
-})
+  const wsLink = subscriptionUri !== null ? new WebSocketLink({
+    uri: subscriptionUri,
+    options: {
+      reconnect: true
+    }
+  }) : new ApolloLink() // return an empty link, useful for testing
+
+  const link = split(
```
kinow: Think of `split` as a ternary `?` operator. Given a condition, return either the next value, or the subsequent.

kinow:
Updated with one more commit for changelog.

CHANGES.md
```diff
@@ -24,9 +24,12 @@ workflows only.
 [#283](https://github.com/cylc/cylc-ui/pull/283) - Load user information
 on application startup.
 
-[#283](https://github.com/cylc/cylc-ui/pull/285) - Update Vuetify to 2.1,
+[#285](https://github.com/cylc/cylc-ui/pull/285) - Update Vuetify to 2.1,
```
kinow: Oops.

kinow:
Rebased, fixed conflicts, and added one small fix during rebase for the changelog entry of #285 (mea culpa).

kinow:
And fixed something I hadn't tested. I tested the subscriptions with live data. But after rebasing tried the `offline` mode and it failed. It was because I coded the new import, which supports mocking services, with the `SubscriptionWorkflowService` class.

Now, instead, we export everything as before, plus an `export default WhateverName`. Then in the `main.js` we can do `import MyFavouriteName from 'workflow-service'`.

In other words, user/caller can use whatever name s/he wants, as the module/callee is exporting a default implementation. `SubscriptionWorkflowService` for live data, and `MockWorkflowService` for the offline mode.

Tested both just now after pushing the last changes :+1:

kinow:
Huh, I thought it was ready after the PR's of David, but this actually is waiting on another PR of cylc-uiserver: https://github.com/cylc/cylc-uiserver/pull/82

hjoliver:
Merged :wink:

kinow:
I'm syncing all branches, will do a quick smoke test with this PR again. If everything works fine, will change it to ready for review. Thanks @hjoliver !

kinow:
Re-creating the virtual environments after syncing the repositories, got the following dependency error.

```
Successfully built cylc-flow
ERROR: graphql-relay 2.0.0 has requirement graphql-core<3,>=2.2, but you'll have graphql-core 3.0.0 which is incompatible.
ERROR: graphene 2.1.8 has requirement graphql-core<3,>=2.1, but you'll have graphql-core 3.0.0 which is incompatible.
```

It looks to me like there is a typo in the dependency requirement of `graphql-ws` (the typo is in their `setup.py`). But I will need to test it.

kinow:
image

I think this is valid for Conda packages, not sure if for setuptools/pip, where I think you need a comma between version ranges.

kinow:
PR submitted https://github.com/graphql-python/graphql-ws/pull/39/files

kinow:
Rebased, fixed conflicts, updated the other components & views (graph & gscan were updated after this PR was created).

Also tested with no workflows, with `five`, and then with `five` & `families2` simultaneously.

Found one problem. Independent of the running of running workflows. I had just `five` running, but now the Tree component was flickering. Disappearing, and then reappering. Quite annoying.

Looks like there are two messages received at the same time. My suspicion is that `GScan` is creating one query, and then the `Tree` view is creating another one. But the `Tree` view is - for some reason - reacting to both.

The `GScan` query won't have the `familyProxies`, which is required for the `Tree` component, resulting in a moment when the `Tree` component ignores the data updated in Vuex store. Will investigate a bit more, but very close to marking as ready for review.

kinow:
Ok, just finished with unit tests, and also some quick tests. Tomorrow will test it again with NIWA's notebook after re-creating the whole environment, and if it works there too, it will be ready for review.

kinow:
(One problem was that it had two queries, as I didn't call `.unsubscribe` on an Observable's subscription. Tricky to find that issue.)

kinow:
Apparently working OK on both my work & home environments. Used two workflows to test it. Found only one browser console message, but that it related to the workflow data structure returned, and how we are accessing attributes and creating the hierarchy (i.e. not related to this issue).

image

So ready for review. Please, remember that it is required to use the latest Cylc UI Server, which should also install the latest Cylc Flow (or install manually if you prefer to avoid the local cache).

Thanks
Bruno

kinow:
Conflicts fixed soon :+1:

kinow:
Fixed conflicts, but the `Toolbar` unit test is not passing due to another change, where we made the `Toolbar` visible only for workflows (as the title was redundant).

The issue is that the `Toolbar` is closely-related to the `Drawer` component. When a user switches to responsive mode (resizing browser, or when in mobile) the button to display the `Drawer` is in the `Toolbar`. So we need to `Toolbar` for every page, hidden by default - except when showing a workflow - and displayed when in responsive mode.

Blocked by #199. I started fixing it in this PR, but then it started to look like a change too big to accommodate in one PR only. Will fix #199 in another PR, then this PR's build should pass.

src/services/gquery.js
```diff
@@ -77,9 +77,13 @@ class GQuery {
    * queries.
    */
 
-  constructor () {
+  constructor (enableWebSockets = false) {
+    const baseUrl = `${window.location.hostname}${window.location.port ? ':' + window.location.port : ''}${window.location.pathname}`
+    const httpUrl = `${window.location.protocol}//${baseUrl}graphql`
```
oliver-sanders: Is `window.location.pathname` guaranteed to end with a trailing `/`?

src/services/workflow.service.js
oliver-sanders: Can probably remove the old `request` method, or potentially merge these classes since we shouldn't be using the old system for subscriptions after this.

src/services/workflow.service.js
oliver-sanders: Can we get rid of the `LiveWorkflowService` now?

src/services/workflow.service.js
```diff
@@ -29,6 +31,72 @@ class LiveWorkflowService extends GQuery {
...
+export default SubscriptionWorkflowService
```
oliver-sanders: Duplicate export statement?

src/utils/graphql.js
```diff
@@ -1,12 +1,52 @@
...
+    ({ query }) => {
+      const definition = getMainDefinition(query)
+      return definition.kind === 'OperationDefinition' && definition.operation === 'subscription'
```
oliver-sanders: What does this switch do?

src/services/workflow.service.js
kinow: With the first, you can choose `import { LiveWorkflowService } from workflow.service` or `import { SubscriptionWorkflowService } from workflow.service`.

The second is just the default/lazy import, `import AnythingYouWantToName from workflow.service`. Where `AnythingYouWantToName` will be actually `SubscriptionWorkflowService`. Normally it's imported with the same name (i.e. `import Vue from 'vue'`, though we could call it anything really).

src/utils/graphql.js
kinow: This one is to check if the GraphQL query starts with `subscription { workflow ... }`, or `query { workflow ... }` (or empty `{ workflow ... }`).

If it starts with `subscription`, then we will use the WebSocket link. Otherwise we will use the HTTP link.

src/services/workflow.service.js
kinow: I can't recall why I left. Not sure if it was still being used for testing/mocking, or if I was scared that the WebSockets could fail and we would have to revert to the old service :smile: will take another look to confirm.

src/services/gquery.js
oliver-sanders: In which case is there a risk `httpUrl` could be invalid?

src/services/workflow.service.js
oliver-sanders: Ahh, interesting, I didn't know you could mix the two approaches.

src/utils/graphql.js
oliver-sanders: Thanks for clarifying.

src/services/gquery.js
kinow: Yes, I think there is no way to guarantee it will work 100% times. As it would depend on how the user deployed it, and it also relies on the browser to return the value with that trailing `/`.

I think we used to have a `TODO` marker for it somewhere when the code was first created but I probably removed it. I will add a `TODO` and an issue on GitHub where we can discuss and see what's the best way to specify the URL or at least the basepath.

kinow: Done, both the `TODO`, and the [issue](https://github.com/cylc/cylc-ui/issues/343)

kinow:
Changed the `Dashboard` to use the `Vue.$workflowService` now too. So all the components should be using it.

I've noticed that the WebSockets appear to suddenly stop working, with no indication in the browser or workflow logs. I suspect it could be the component lifecycle. Maybe when we navigate the routes, somewhere we may be asking it to close the connection, or we are using the API incorrectly.

Will do a bit more of testing today/tomorrow :+1:

kinow:
@oliver-sanders removed the `LiveWorkflowService` and the old `request` method from `GQuery`.

kinow:
I saw a console error that I occasionally spotted when testing other branches. Decided to spend some time investigating it, and learned something new while fixing it.

In Vuex store, I had created a User object, with the user information and with a method to choose the Hub URL.

Screenshot_2019-12-19_19-55-35

I never understood why that method would not be available, if the Vuex store had the user. Turns out you can store only plain objects in Vuex store, and also in Vue data element - refs https://vuejs.org/v2/api/#data & https://vuex.vuejs.org/guide/state.html

Fixing it here as I'm investigating why sometimes my websocket is sending/receiving messages, but the UI is not updating :thinking: but might have to take a look if we are not expecting other Vuex objects to have functions or other features that are not supported.

kinow:
Done, I think now it's working better. There were a few changes, that even fixing conflicts still needed updating. For example, the `Toolbar` now was calling mutations, but using `this.$apollo` (from vue-apollo I think, or apollo-boost), but now we are creating our own client manually (not too complicated, and reduces the application size - see #164).

The unit tests are failing. This issue is blocked by #341, once that's merged, we can rebase this one, and the unit tests should pass :crossed_fingers: 

The only issue I had now was that sometimes I would open the page in Firefox, and it would still no data. Refreshing it, everything worked. Looks like the WebSocket is not working initially for some reason.

So besides the code review, might be a good idea to play with 2x workflows, running at the same time, calling the mutations, going back to the dashboard, checking if the job icons are updated about every 5 seconds, etc :+1:

kinow:
If GitHub action passes, then this PR is ready for review. Everything worked locally, tested with 2 workflows, also tested mutations. The toolbar buttons are a bit buggy at the moment, some times changing the enable/disable attribute incorrectly. But that is for another ticket.

kinow:
Thanks @dwsutherland !

Issue #258, opened 2019-09-25
'cylc hub' link doesn't respect base_url


ScottWales:
**Describe the bug**

When cylc 8 is running on jupyter hub with a base_url set, e.g.
```
# jupyterhub_config.py
c.JupyterHub.bind_url = 'http://127.0.0.1:8000/cylc'
```
the dashboard link points to `http://127.0.0.1:8000/hub/home` instead of `http://127.0.0.1:8000/cylc/hub/home`

**Release version(s) and/or repository branch(es) affected?**

8.0a1 installed through conda

**Steps to reproduce the bug**

Add above excerpt to `jupyterhub_config.py`, start the server and check the 'cylc hub' link on the dashboard

**Expected behavior**

Link should point to `http://127.0.0.1:8000/cylc/hub/home`

**Screenshots**

![Screenshot from 2019-09-25 11-38-06](https://user-images.githubusercontent.com/161133/65562247-fe822880-df88-11e9-996e-66e485647a95.png)

**Additional context**

The bind_url is set so I can run cylc behind an apache proxy that also runs other services

Manually going to the hub url works, it's just the link that's broken

**Pull requests welcome!**
This is an Open Source project - please consider contributing a bug fix
yourself (please read `CONTRIBUTING.md` before starting any work though).

kinow:
Successfully reproduced the issue :+1:

kinow:
In JupyterHub, this is done by a template file that initializes a global value at attaches it to the `window` object.

https://github.com/jupyterhub/jupyterhub/blob/5b13f96162b77b4ae23306327cd700b8faeb299c/share/jupyterhub/templates/page.html#L62-L64

In our case, our best option - other than replicating JupyterHub's approach - would be to query a service on startup I think.

kinow:
Here's a possible solution. If we add #260, the user information contains:

```json
{
  "kind": "user",
  "name": "kinow",
  "admin": false,
  "groups": [],
  "server": "/cylc/user/kinow/",
  "pending": null,
  "created": "2019-07-16T22:25:30.679281Z",
  "last_activity": "2019-09-25T05:18:47.041040Z",
  "servers": null
}
```

If this information gets stored in Vuex, we can use it to get the `server`, minus `kind` and `name`, to retrieve the Hub base_url. `kind` _could_ vary (need to check docs) for admin users, so perhaps fix that part as `/user/${name}/` and remove that from the `server`.

kinow:
[...]

PR #283, opened 2019-10-08
Load user information on application startup

### LINKED ISSUE #260: Load user information on application startup
When using JupyterHub, we have under the user settings an HTTP call to a web service that returns the following JSON:

```json
{
  "kind": "user",
  "name": "kinow",
  "admin": false,
  "groups": [],
  "server": "/cylc/user/kinow/",
  "pending": null,
  "created": "2019-07-16T22:25:30.679281Z",
  "last_activity": "2019-09-25T05:18:47.041040Z",
  "servers": null
}
```

This is the JupyterHub user information, extracted from the cookie, by JupyterHub code, being serialized as JSON.

This information is useful for other parts of the application, and for other components. We can query the service on startup, which should be very fast as it is the same server serving the HTTP that will reply, and information that should be on the cookie (not sure if there are any HTTP queries being done in the backend, or if querying DB directly during login).

The `Vue` instance, when created, can query this HTTP service, and put the user information (same as above) in the Vuex store.

kinow:
Closes #260 

This PR first adds a mock service for the `UserService`. The returned information should be the same for everybody, with the exception of created date, which is a `new Date()` object, meaning it will display different whenever called.

Screenshot_2019-10-09_12-14-28

So now this view should work on the offline mode.

Then we move the call to the `/userprofile` service from the `UserProfile.vue` view to a global route guard.

This guard will check if the Vuex store contains a user already in place. If not, it will fetch the user information from JupyterHub, and store in the Vuex store.

After this, the application should have the `store.state.user.user` information available to any screen/module/code/etc that may need it, e.g. #258 

Cheers
Bruno

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,6 +21,9 @@ None or N/A.
 [#257](https://github.com/cylc/cylc-ui/pull/257) - Display toolbar for
 workflows only.
 
+[#283](https://github.com/cylc/cylc-ui/pull/283) -  Load user information
+on application startup.
+
 ### Fixes
 
 [#275](https://github.com/cylc/cylc-ui/pull/275) - Fix size of dashboard
--- a/src/router/index.js
+++ b/src/router/index.js
@@ -12,6 +12,7 @@ import Router from 'vue-router'
 import Meta from 'vue-meta'
 import NProgress from 'nprogress'
 import store from '@/store'
+import { UserService } from 'user-service'
 
 import '../../node_modules/nprogress/nprogress.css'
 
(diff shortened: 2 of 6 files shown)


kinow:
Screenshot above was from the offline application. This one is from the app behind JupyterHub, immediately after startup without opening any other routes/views.

image

It should be possible to see the user information in the Vue extension panel.

kinow:
Here after opening the view, with no network activity.

image

Note: we may need to add a "Refresh" button on the User Profile to fetch the user information again after it has changed (instead of asking users to log out and log in again). But easier to do that later when we have everything in place and working (and being used/tested).

kinow:
Ready for review :+1:

Issue #187, opened 2019-08-14
cylc graph - investigate how closely cytoscape can fit graph requirements


MartinRyan:

oliver-sanders:
See also https://github.com/cylc/cylc-ui/issues/81#issuecomment-423555699

oliver-sanders:
Any progress here, I'm quite worried that Cytoscape won't be able to deliver the UI we require.

https://user-images.githubusercontent.com/16705946/60719622-36487980-9f20-11e9-8d90-69301420bd4d.png

hjoliver:
@oliver-sanders - I've sort-of been assuming that no existing components / frameworks will display the labels and job icons next to the graph nodes, so we might need to start with some temporary bodge before working on the "display engine" (as opposed to layout engine) ourselves.  But will see what @MartinRyan thinks.

MartinRyan:
if we decide on Cytoscape we can add ourselves to this list of usershttps://github.com/cytoscape/cytoscape.js/issues/914

maxkfranz:
See https://github.com/cytoscape/cytoscape.js-popper and the related demos

MartinRyan:
Using a html label and Tippy Popper
![Screenshot from 2019-10-23 10-23-54](https://user-images.githubusercontent.com/65905/67342961-712df600-f57f-11e9-866f-918089dfd8a9.png)

MartinRyan:
![Screenshot from 2019-10-23 10-24-05](https://user-images.githubusercontent.com/65905/67342993-915db500-f57f-11e9-98ed-0436be1e763d.png)

MartinRyan:
![Screenshot from 2019-10-23 10-28-34](https://user-images.githubusercontent.com/65905/67343079-e7325d00-f57f-11e9-949d-39dfc88e4812.png)

MartinRyan:
Using html labels and Tippy/ Popper.

MartinRyan:
**some cytoscape plus points**

**mutable graphs**
Cytoscape is built for dynamic graphs and does allow partial updates of layouts without re-renders of the whole graph, so adding and removing nodes and edges or updating styles would be more performant.

**render performance**
Cytoscape  is canvas based. Canvas and WebGL is much faster than svg with a large amount of nodes. I'm using svg icons at the moment, as background images for the nodes, I think this slows the graph down.

**compound nodes**
[...]

PR #289, opened 2019-10-15
graph with live data and node info

MartinRyan:
graph with node info
supersedes https://github.com/cylc/cylc-ui/pull/241

PULL REQUEST
**Requirements check-list**
- [x] No change log entry required (DRAFT).


hjoliver:
@MartinRyan - trying this out, I see the labels and job icons at all zoom levels, but I only see the node icons when zoomed in very close.  Do you see that?

MartinRyan:
> @MartinRyan - trying this out, I see the labels and job icons at all zoom levels, but I only see the node icons when zoomed in very close. Do you see that?

I see them at all zoom levels

hjoliver:
I wonder if it is some caching issue.  I see the node icons after a reload, but a bit of zooming causes them to disappear again.

MartinRyan:
> I wonder if it is some caching issue. I see the node icons after a reload, but a bit of zooming causes them to disappear again.
I've tested in chromium, firefox and chrome on linux

hjoliver:
Reinstalling and rebuilding the UI from scratch, just in case.

kinow:
Just tried on Firefox LTS, Ubuntu LTS, working for me :+1: 

GIFrecord_2019-10-16_182241

hjoliver:
@kinow - did you do a bunch of zooming, with the +/- controls?

kinow:
> @kinow - did you do a bunch of zooming, with the +/- controls?

Randomly, but just tried again focusing on zooming while it was updated, with no issues.

GIFrecord_2019-10-16_203232

hjoliver:
Looks good - I want that! (Will try what you suggested on Riot...)

hjoliver:
I had a broken Google Chrome - downloaded the latest and its working properly now :+1: phew.

hjoliver:
As discussed on Riot, I vote to merge this soon.  Minimal changes needed are:
- fix whatever is breaking the tree-view styling (screenshot above)
- remove Cose-bilkent and Cola layouts, at least for now
  - Cola appears a total mess too me, for non-trivial suites
  - Cose-bilkent might be nice for some things, but would have to turn off the "settling down" animation that happens in every update, to just display the final layout ... is that possible @MartinRyan ?)

hjoliver:
The freeze button is really good, and users probably will need it when inspecting the graph closely (as opposed to watching its evolution) but it forces the dagre layout.

hjoliver:
Gif showing cose-bilkent single-node update jump, and freeze -> dagre behaviour:
peek

MartinRyan:
fixed the cross styling and freeze button issues

hjoliver:
Fixes confirmed. Nice improvement on the job display :+1: :

Screenshot from 2019-10-22 22-40-12

hjoliver:
OK @MartinRyan do you want to hit the "Ready for review" button on this?

kinow:
Two files in conflict. Can we also update the description of the issue at the top :point_up: to remove the wip related text?

kinow:
If the node data has a `label`value that is too long, it will be displayed outside the popup window. Perhaps it needs some wrapping setting in CSS, or be truncated with `...`, etc?

kinow:
Some console errors in the browser. I think it is related to the workflow service implementation.

image

src/components/cylc/Graph.vue
```diff
@@ -0,0 +1,1644 @@
...
+    }
+    cytoscapeService.unregister(this)
+  },
+
+  mounted () {
+    console.debug(`MOUNTED called, status: ${this.status}`)
+    this.handleMounted()
+    this.$store.watch((store) => {
+      this.workflows = store.workflows
+    })
```
kinow: This is not very conventional. Any need to use `.watch` and be called whenever the store changes (which may or may not be for the `.workflows`) and then set the local value? Instead, we could use something like Vuex's [`mapState`](https://vuex.vuejs.org/guide/state.html#the-mapstate-helper) in a computed value?

src/components/cylc/Graph.vue
```diff
@@ -0,0 +1,1644 @@
...
+
+  metaInfo () {
+    return {
+      title: this.getPageTitle('App.graph', { name: this.workflowId })
+    }
+  },
+
+  beforeRouteLeave (to, from, next) {
+    if (tippy) {
+      tippy.hide()
```
kinow: This works now, but this is linking the component with the current route. Instead, I think we will have issues with this tooltip later.

If you look at the design sketch for the tabbed window:

image

It is possible to see that we may have multiple Graph components being used. Not all visible at the same time, though all existing in the same route/view. So the `beforeRouteLeave` would never be called.

Switching between Graphs (supposing we have 2 or more graphs open in the same panel), it could happen that a tippy tooltip would stay visible even after showing other graph (though it would depend on the z-index, and whether the issue happens when you have multiple graph components open).

That's why I had left a comment in a previous PR for the graph that the component should be completely separated from the route. Receiving everything it needed via parameters, and having methods that could be called when created/destroyed, or displayed/hidden, etc.

Probably a problem for later, but I think it would be a good idea to at least leave this comment so other reviewers will be aware of this (we had a bug fixed by Martin where the tippy tooltip would still be visible after you switched routes, from what I recall, that's why we have this `.hide()` call here).

src/components/cylc/Graph.vue
```diff
@@ -0,0 +1,1644 @@
...
+    this.handleMounted()
+    this.$store.watch((store) => {
+      this.workflows = store.workflows
+    })
+  },
+
+  created (cy) {
+    console.debug('CREATED')
+    this.suiteName = this.$route.params.workflowid
+    this.workflowId = this.$route.params.workflowid
```
kinow: If props are used, we can remove these two attributes and these two lines above.

src/components/cylc/Graph.vue
```diff
@@ -0,0 +1,1644 @@
...
+              })
+            })
+          }
+          console.debug('elements ==>>>> ', elements)
+          isEmpty(elements) ? console.warn('gdata is empty or undefined') : this.graphData = elements
+        }
+      } catch (error) {
+        console.error('workflowUpdated error: ', error)
+      }
+    },
```
kinow: IMO we could move this method elsewhere, and instead just receive the data that the component needs via props. Even the parent view would be a better place, or copy what's done in the tree view and add a hook in the vuex store too.

The Vuex store sets a global `.workflows` value. After the PR for the family tree, it will construct the `workflowTree` iff the `workflows.familyProxies` exist. That attribute - so far - is only ever present when you have a Tree view.

Similarly, we could call the method to create the Graph data iff the `workflows.nodesEdges` exist. This way, if you are on a view with only Trees, the graph method is not called, and vice-versa. And, if you are on the Dashboard, probably none will be ever called, as it will probably need just the `workflows`.

src/components/cylc/Graph.vue
```diff
@@ -0,0 +1,1644 @@
...
+            selected: false,
+            selectable: true,
+            locked: false,
+            grabbable: true,
+            classes: ''
+          }
+          if (has(node, 'task') && !isEmpty(node.task) && has(node.task, 'name') && !isEmpty(node.task.name)) {
+            nodeObj.data.label = node.task.name
+          }
+          has(node, 'id') && !isEmpty(node.id) ? nodeObj.data.id = node.id : console.warn('workflowUpdated - node id is empty')
```
kinow: Perhaps the tree component should also be more defensive. At the moment I'm trusting almost 100% on the GraphQL data to have some value, or at least be `undefined` or `null`. I think I had to check for the types returned in only a few places, but maybe the data returned is different. Just a comment, no action needed here :+1:

src/components/cylc/Graph.vue
```diff
@@ -0,0 +1,1644 @@
...
+          this.layoutName = value
+        }
+      } catch (error) {
+        console.error('changeLayout error: ', error)
+      }
+    },
+
+    async handleMounted () {
+      try {
+        this.status = 'success'
```
kinow: Was there some other code here before, that could throw errors? Right now I think this `handleMounted` could be removed? The `catch` above is probably never executed?

src/components/cylc/Graph.vue
```diff
@@ -0,0 +1,1644 @@
...
+        }
+
+        if (typeof cytoscape('core', 'undoRedo') !== 'function') {
+          console.debug('registering undoRedo')
+          undoRedo(cytoscape)
+        }
+
+        if (typeof cytoscape('core', 'expandCollapse') !== 'function') {
+          console.debug('registering expandCollapse with jquery')
+          expandCollapse(cytoscape, jquery)
```
kinow: Using the IDE, I couldn't locate where the `jquery` value was used.

image

They also had a PR some time ago to remove the JQuery dependency: https://github.com/iVis-at-Bilkent/cytoscape.js-expand-collapse/pull/75/files

So if that's correct, we could update this method, and also remove the JQuery dependency in the `package.json`. Or, if we decide to move the expand/collapse to a follow-up PR, then both dependency and this block of code could be removed?

MartinRyan:
> Some console errors in the browser. I think it is related to the workflow service implementation.
> 
> image
yes not sure about this as I'm using a seperate cytoscapeService?

src/components/cylc/Graph.vue
MartinRyan: well spotted, Id removed this before but it crept back!

src/views/Graph.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+    <graph></graph>
+</template>
+
+<script>
+import Graph from '@/components/cylc/Graph'
+
+export default {
+  name: 'Default',
```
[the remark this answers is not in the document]
MartinRyan: I removed the name

src/views/Graph.vue
```diff
@@ -0,0 +1,20 @@
...
+  components: {
+    Graph
+  }
+}
+</script>
+
+<style scoped>
+#core-footer {
+  position: relative;
```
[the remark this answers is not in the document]
MartinRyan: this is also gone :-)

src/services/cytoscape.service.js
```diff
@@ -0,0 +1,27 @@
...
+        this.request()
+      }, 10000)
+      this.request()
+    }
+    return ret
+  }
+}
+
+const cytoscapeService = new CytoscapeGQueryService()
+export default cytoscapeService
```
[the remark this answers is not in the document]
MartinRyan: lol, I was using workflowService previously but was told not too ;-)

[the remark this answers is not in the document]
MartinRyan: now using workflowService :-)

src/components/cylc/Graph.vue
MartinRyan: If we have flat data that doesn't need any manipulation then this method isn't needed, it can be sent straight to updateGraph().  I was using a watch on the store as using the 'deep' option can watch for changes deep in the data hierarchy.  Hopefully in the future  it will be just change data, so any change at all would be acted upon.

src/components/cylc/Graph.vue
MartinRyan: The graph is very sensitive to bad data, hence the checks, I would like to use more functional filters like monads, probably overkill here...

src/components/cylc/Graph.vue
MartinRyan: removed the dependency on jquery

src/components/cylc/Graph.vue
MartinRyan: yep this is a left over

MartinRyan:
> If the node data has a `label`value that is too long, it will be displayed outside the popup window. Perhaps it needs some wrapping setting in CSS, or be truncated with `...`, etc?
> 
> image

wrap added

src/styles/cytoscape/panzoom.css
```diff
@@ -0,0 +1,221 @@
...
+	top: 20px;
+}
+
+.cy-panzoom-zoom-only .cy-panzoom-zoom-in {
+	top: 45px;
+}
+
+.cy-panzoom-zoom-only .cy-panzoom-zoom-out {
+	top: 70px;
+}
```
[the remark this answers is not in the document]
MartinRyan: added

MartinRyan:
> There is some code to expand and collapse, but I think the buttons to actually execute it are commented out. Could we move this feature to later? This would also help reviewers, and instead focus here only on adding the graph + live data.

Those buttons are commented out as they need a graph with compound nodes to collapse or expand all nodes, otherwise they will cause artefacts.
Each compound node, when activated by having children with the parent property added, has a 'cueImage' in the centre of a plus or minus when hovering to activate collapse or expansion, if on a device without hover the compound node can just be tapped. https://github.com/cylc/cylc-ui/issues/187#issuecomment-547179333

MartinRyan:
> If the node data has a `label`value that is too long, it will be displayed outside the popup window. Perhaps it needs some wrapping setting in CSS, or be truncated with `...`, etc?
> 
> image

wrap added

MartinRyan:
> Do we need to support changing graph layouts now? If not, could we copy this branch somewhere else for reference later, and use a single layout? This would reduce the code of the component by a good amount (helpful in case we need to change it before adding more layouts, or use/test other graph libs)

I think they are useful, but perhaps others can decide that

MartinRyan:
> Some JS code is not used, sometimes because props are always set to false, not ever changing. Could we remove this code? It appears to be left-overs from previous code/tests, or other work-in-progress that can probably be done later.

yes, hopefully I've removed that now

>

MartinRyan:
> Two files with conflicts @MartinRyan. Can we also update the description of the issue at the top ☝️ to remove the wip related text?

done

MartinRyan:
> I think we don't need another service to fetch the data. The design proposed by Oliver for the workflow service should be able to merge queries, and serve data (via polling for now) for all views & components. This would make it easier to switch to the subscription service later on, plus, it works with offline/mocked data - there's a global alias already defined in vue.config.js.

I'm now using the workflowService, though the 5 second update is a little short for the graph. I was using a 10 second interval on the cytoscapeService.

oliver-sanders:
Hi, I think there is something wrong with the package-lock.json (causing the build failure).

I removed and rebuilt the lock but an getting this error in the console:

```
updateGraph error:  TypeError: "cy is undefined"
    _callee12$ Graph.vue:1139
    Babel 9
    Lodash 3
```

Perhaps the same issue.

MartinRyan:
> Hi, I think there is something wrong with the package-lock.json (causing the build failure).
> 
> I removed and rebuilt the lock but an getting this error in the console:
> 
> ```
> updateGraph error:  TypeError: "cy is undefined"
>     _callee12$ Graph.vue:1139
>     Babel 9
>     Lodash 3
> ```
> 
> Perhaps the same issue.

I'll take a look, I had to resolve conflicts with that file...

kinow:
For posterity, this PR has a change in package.json, adding svgo 1.3.2 as devDependency due to https://github.com/svg/svgo/issues/1180

src/components/cylc/Graph.vue
```diff
@@ -0,0 +1,1633 @@
...
+            },
+            group: 'edges',
+            removed: false,
+            selected: false,
+            selectable: true,
+            locked: false,
+            grabbable: true,
+            classes: ''
+          }
+          has(edge, 'id') && !isEmpty(edge.id) ? edgeObj.data.id = edge.id : console.debug('workflowUpdated - edge id is empty')
```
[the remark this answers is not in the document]
MartinRyan: These functions workflowUpdated() getEdges() and GetNodes() construct the required cytoscape data format and are a workaround until the graph is supplied with flat data, then they can be removed and the data supplied to the updateGraph() function. This will really speed performance back up again.

MartinRyan:
> Some CSS classes are being defined as top level elements, and may conflict with other elements in the UI (e.g. #holder div has several buttons and other HTML elements as children, that have their own classes like .frozen-button... I think it would be safer to move these styles under #holder if possible, or some other .cy-graph-something-etc parent element)
> 
css made more specific

MartinRyan:
corrected css class name

Issue #229, opened 2019-09-16
Tree component must correctly render families


kinow:
@hjoliver reported on riot families are not properly displayed with the current Tree Component. Adding as a tentative for the alpha release.

kinow:
Quickly hacked `five` to have a `BAR` family with two children `bar1` and `bar2`:

```
[meta]
    title = "Inter-cycle dependence + a cold-start task"
[cylc]
    UTC mode = True
[scheduling]
    #runahead limit = 120
    initial cycle point = 20130808T00
    final cycle point = 20150810T00
    [[dependencies]]
        [[[R1]]]
            graph = "prep => foo"
        [[[PT12H]]]
            graph = "foo[-PT12H] => foo => BAR"

[runtime]
[[root]]
script = "sleep 2; echo Hi!"

[[BAR]]
script = "sleep 2; echo Family!"

[[bar1]]
inherit = "BAR"

[[bar2]]
inherit = "BAR"

```

Here's the result:

![image](https://user-images.githubusercontent.com/304786/64936462-31059480-d8aa-11e9-9e1a-086a433a6b42.png)

The children are displayed. The family is never displayed in the tree.

kinow:
Here's the GraphQL query executed:

```
{
  workflows(ids: ["five"]) {
    id
    name
    status
    owner
    host
    port
    taskProxies {
      id
      state
      cyclePoint
      task {
        meanElapsedTime
        name
      }
      jobs {
        id
        host
        startedTime
        state
        submitNum
      }
    }
  }
}
```

And the response:

```
{
  "data": {
    "workflows": [
      {
        "id": "kinow|five",
        "name": "five",
        "status": "running",
        "owner": "kinow",
        "host": "ranma",
        "port": 43060,
        "taskProxies": [
          {
            "id": "kinow|five|20130829T0000Z|bar2",
            "state": "succeeded",
            "cyclePoint": "20130829T0000Z",
            "task": {
              "meanElapsedTime": 1.899999976158142,
              "name": "bar2"
            },
            "jobs": [
              {
                "id": "kinow|five|20130829T0000Z|bar2|1",
                "host": "localhost",
                "startedTime": "2019-09-16T05:51:23Z",
                "state": "succeeded",
                "submitNum": 1
              }
            ]
          },
          {
            "id": "kinow|five|20130829T1200Z|bar2",
            "state": "submitted",
            "cyclePoint": "20130829T1200Z",
            "task": {
[...]

PR #290, opened 2019-10-21
Prevent accessing user when still undefined

kinow:

It is possible that the `user` object will be accessed from the Vuex store while the `Promise` has not yet been fulfilled. This PR just changes the logic to call `next()` only if we already have a user, or after the `Promise` has completed.

**Requirements check-list**
- [x] Does not need tests (why? not sure how to test the `beforeEach` of routes, maybe with e2e tests?).


kinow:
Found in the console logs happening randomly while preparing to start working on #229

PR #292, opened 2019-10-24
Fix promise use in mocked data, and field being displayed in user form

kinow:

Noticed after rebasing the issue that I had a bug in my `Promise`. Was returning a `Promise` that returned a `Promise` (Vuex store method), instead of chaining it. But I think we also don't need to chain anything, instead we can just simply return the Vuex method/`Promise`.

But that showed that there was a bug (probably old) where the username was not displayed in the form. Probably nobody noticed because we rarely need that screen.

**Requirements check-list**
- [x] Does not need tests (why? mock data).

### DIFF
--- a/src/services/mock/user.service.mock.js
+++ b/src/services/mock/user.service.mock.js
@@ -4,10 +4,15 @@ import store from '@/store/index'
 export const UserService = {
 
   getUserProfile () {
-    return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


hjoliver:
(Did you mean to close this @kinow ?)

kinow:
Ooops. While helping Martin with the package.json issue I realized I had too many old branches and started removing the ones I thought were not in use.

Luckily GitHub has a useful "Restore Branch" button, and you spotted my mistake. Thanks!

PR #296, opened 2019-10-31
Ignore no-console eslint rule in a few places

kinow:

Simply prevents build warnings for calls to methods of `console`. The exceptions added are already protected by `if (process.env.NODE_ENV !== 'production') {` - meaning print only when not production environment - or are intentional - like in `Task.vue`.

**Requirements check-list**
- [x] Does not need tests (why? lint only).

### DIFF
--- a/src/components/cylc/Toolbar.vue
+++ b/src/components/cylc/Toolbar.vue
@@ -121,7 +121,6 @@ export default {
     },
     onClickAddView () {
       // TODO: implement adding views action
-      console.log('Adding views has not been implemented yet')
     }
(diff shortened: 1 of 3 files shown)


src/components/cylc/Toolbar.vue
```diff
@@ -121,7 +121,6 @@ export default {
     },
     onClickAddView () {
       // TODO: implement adding views action
-      console.log('Adding views has not been implemented yet')
```
kinow: I think the `TODO` should be enough

PR #298, opened 2019-11-01
Support any attributes defined in paths.js when creating routes

### LINKED ISSUE #293: Support any parameter used by routes

We are using vue-router in Cylc UI. But there is an extra layer created by the theme, where it uses a `paths.js` file to hold the routes/paths and its attributes (`name`, `view`, `meta`, etc).

The issue is that when you add a new attribute (e.g. `props`) then you have to update the function and the `.map()` call in `router/index.js` so that the attribute is used when creating the route.

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #293

Simply accept the complete `path` object, modifying it to use the `view` value when `name` is missing (i.e. keeping previous behaviour).

This allows the approach used by @MartinRyan in the Graph component PR to work, of [passing props to route components](https://router.vuejs.org/guide/essentials/passing-props.html). Easier than relying on the route parameters directly. Vue Router instead will inject the param to the view.

What happened in the Graph PR was that even though Martin added `props: true`, it was ignored as it was not expected by the function. The code now will use whatever we define in `paths.js`.

**Requirements check-list**
- [x] Does not need tests (why? not sure how to test this part).

### DIFF
--- a/src/router/index.js
+++ b/src/router/index.js
@@ -19,16 +19,15 @@ import '../../node_modules/nprogress/nprogress.css'
 // Routes
 import paths from './paths'
 
-function route (path, view, name, meta, alias) {
-  return {
-    name: name || view,
-    meta,
-    path,
-    alias,
+function route (path) {
+  const copy = Object.assign({}, path)
+  const view = copy.view
+  return Object.assign(copy, {
+    name: path.name || view,
     component: (resolve) => import(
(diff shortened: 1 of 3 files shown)


kinow:
To confirm it's working fine, I've adopted the approach of passing the param values as props in the Tree view in this PR as an example & improvement.

Issue #98, opened 2019-07-01
Layout: responsive sub-panels w/ tabs for core (views) panel


sadielbartholomew:
To implement as indicated in the mock-ups (e.g. [this one](https://github.com/cylc/cylc-ui/issues/87#issuecomment-501214388)) in #87, but ensuring that the result adapts to window size, for responsiveness even when only considering desktop usage.

An (implementation-)generic Issue for #90, focusing on the layout panelling & tabbing aspect (not further UI components such as menus, etc.) which is one implementation using a specific technology that may work.

matthewrmshin:
Swipe left/right for small touchscreen devices?

kinow:
As our tabbing system uses Lumino, I think we will either have to implement responsive tabbed ourselves, or wait/contribute to:

- https://github.com/jupyterlab/jupyterlab/issues/3275
- https://github.com/jupyterlab/lumino/issues/77

JupyterLab is rendered OK in responsive mode.

![image](https://user-images.githubusercontent.com/304786/86546113-2d511480-bf87-11ea-84a1-1bb57f1698e3.png)

However, it indeed responds only to mouse events. Simulating gesture/touch, it doesn't work (as reported in the issues above).

oliver-sanders:
We went with Lumino, it works nicely.

Issue #90, opened 2019-06-20
Create a Workflow component


kinow:
This is related to #87 and #75 .

@oliver-sanders showed us the JupyterLab application, which allows users to add as many notebooks as they like to a sort of dashboard panel.

This looks like an experiment worth a try - with the risk of having some performance issues due to the number of workflows open simultaneously. So I think a _Workflow_ view could be used for that.

JupyterLab [uses](https://github.com/jupyterlab/jupyterlab/blob/c952a47ee1a2d3ed5a223478cd91b87f96d5b87d/jupyterlab/staging/package.json#L67) [PhosphorJS](http://phosphorjs.github.io). From their about:

>PhosphorJS provides a rich set of widgets, layouts, events, and data structures. These enable developers to construct advanced, production-quality, desktop-like web applications that would be otherwise impossible using CSS alone. 

In order to include it in our current application, we would need something similar to the following steps:

- [ ] Create a `Workflow` component, that simply loads the Phosphor main view
- [ ] Create a `WorkflowTreeView` component (simpler than the graph for this experiment)
- [ ] Make sure the `WorkflowTreeView` can be added multiple times to the `WorkflowTreeView` component panel - that probably means using props or some other mechanism in the component to pass the workflow we want it to display

With that I think we would be able to have something similar to JupyterLab.

kinow:
Examples:

- [Tabs](http://phosphorjs.github.io/examples/tabs/)
- [Menu](http://phosphorjs.github.io/examples/menus/)
- [DockPanel](http://phosphorjs.github.io/examples/dockpanel/)
- more [here](http://phosphorjs.github.io/examples.html)

@oliver-sanders the `Workflow` component would simply have a `template` that loads the `DockPanel`... and each dock would be a `WorkflowTreeView` (for now, but later we can add graph view too). What do you think?

oliver-sanders:
DockPanel! Awesome!

sadielbartholomew:
While it looks powerful, I am not sure that PhosphorJS is the right way to approach creating our UI frame components (panels, tabs, menus etc,). Would it not be much simpler to use Vue component(s) designed for this purpose, so we can create a solution most customised to our own requirements, & so it/they are simpler to incorporate being Vue-based?

[...]

PR #301, opened 2019-11-05
Add gscan component

kinow:
These changes partially address #139

Initial version of the `GScan` component. Will need more iterations later until the requirements in #139 are fully addressed.

**Requirements check-list**


kinow:
Note: it takes users to `/workflows/$workflow-name`. Which at the moment is rendered by the `Tree` view. Later, we should replace it by a `Workflow.vue` I think, which adds the initial area to add tabs as in #98

kinow:
And a GIF of what it looks like in action:

GIFrecord_2019-11-05_165021

kinow:
Build failed due to sass-loader and vuetify. Removing node-sass to see if it will pass now (off to home, will log in again later to check if build passed). Only tested with the workflow five, so it would be good if a reviewer could test with other workflows and see if there's anything broken :)

kinow:
Build passed on Travis :tada:

src/App.vue
```diff
@@ -7,13 +7,106 @@
 </template>
 
 <script>
+import { workflowService } from 'workflow-service'
+
 const defaultLayout = 'empty'
 
+const QUERIES = {
```
oliver-sanders: Why is this stuff here rather than in `GScan.vue`?

src/App.vue
```diff
@@ -7,13 +7,106 @@
...
+          id
+          name
+          state
+          cyclePoint
+          latestMessage
+          task {
+            meanElapsedTime
+            name
+          }
+          jobs(sort: { keys: ["submit_num"], reverse:true }) {
```
oliver-sanders: Should be able to the summary thinggy (is it `stateSummary`) rather than requesting full info on each job.

src/components/cylc/GScan.vue
```diff
@@ -0,0 +1,102 @@
+<template>
+  <div>
+    <div v-if="workflows && workflows.length > 0">
+      <div v-for="workflow in workflows" :key="workflow.id">
+        <v-list-item :to="`/workflows/${ workflow.name }`">
+          <v-list-item-action>
+            <!-- TODO: the workflows are always running now, but later we will need https://github.com/cylc/cylc-ui/issues/90 adn update the icon displayed here -->
```
oliver-sanders: Note workflows can be running but held.

src/components/cylc/GScan.vue
```diff
@@ -0,0 +1,102 @@
...
+            <v-icon>mdi-play-circle</v-icon>
+          </v-list-item-action>
+          <v-list-item-title>
+            <v-layout align-center align-content-center wrap>
+              <v-flex grow>{{ workflow.name }}</v-flex>
+              <v-flex shrink ml-4>
+                <!-- task summary tooltips -->
```
oliver-sanders: Nice.

I think we might want to factor out some of this functionality into a common component so all views can benefit from these features (e.g. tooltips).

I was thinking of something like `JobBox` which either holds:
* State totals (as in this case).
* A collection of jobs (which collapses down to state totals if there are too many jobs).

Food for later thought...

src/App.vue
```diff
@@ -7,13 +7,106 @@
...
 
 <script>
+import { workflowService } from 'workflow-service'
+
 const defaultLayout = 'empty'
 
+const QUERIES = {
+  root: `
+    {
+      workflows {
```
oliver-sanders: BTW: I think we will need to improve the whole `gproxy` thing so it can merge this query properly, my bad I should have opened a follow on issue at the time.

kinow: There are two or three issues created that involve changes in the `gproxy`, so there's room for improvement, but it was a life saver. Without that early in the project, we would have a lot built on top of - probably - multiple queries spread over the project.

src/components/cylc/GScan.vue
kinow: Aaahhh, good point! I added this icon to fill in the space as otherwise the text would be left-aligned without anything before it.

Could we leave this icon for now? I will work on #90 after this, and update GScan to use the new component, which should handle running, held, and.... stopped (?).

src/App.vue
kinow: Just had a look at the schema in `cylc.flow.network.schema`, the closest I found was `stateTotals`.

image

We can use that to give the numbers - which are missing - to the tooltip. I think there was a recent discussion on riot where @hjoliver and @dwsutherland mentioned something changes to the backend that I think is the same thing you are suggesting here. (I think.)

src/App.vue
kinow: I got used to having a component rendering, and the data fetching/massaging elsewhere. As it seems to make the components easier to maintain.

So for `GScan`, I tried to create it with only props, and then plug the data in. That's when I realized I didn't have a view/route for GScan, as it is displayed in pretty much every view of the system. It is also easier to write unit tests and check the coverage.

But if you find it confusing or that it is not how it was supposed to work with the WorkflowService, we can move it to `GScan.vue`. As that component it used throughout the application maybe it makes sense that it both fetches data and also renders it.

kinow:
Rebased, fixed conflicts. May still have to update the code based on further comments from @oliver-sanders (I think I replied with only more questions hehe)

src/App.vue
oliver-sanders: Here's the endpoint we used to call:

https://github.com/cylc/cylc-flow/blob/e3deb7d7049117e7c2a487f558f8b839a7d2ab25/lib/cylc/network/httpserver.py#L328

From memory I think it gave state totals and the five most recent task-jobs in each state?

src/components/cylc/GScan.vue
oliver-sanders: Should be really easy to implement, the field `workflow.status` field is either `running` or `held`:

```graphql
query {
  workflows{
    status
  }
}
```

src/App.vue
kinow: That'd be handy for the UI. Maybe we should open a new issue in Cylc UI Server - I think - to work on this?

hjoliver:
@kinow - have you tested this with more than one workflow?  When I do, the first one I click on (in the scan bar) appears in the tree view, but then second one results in a blank view.  The first view comes back if I click on the first one again in the sidebar.

kinow:
> @kinow - have you tested this with more than one workflow? When I do, the first one I click on (in the scan bar) appears in the tree view, but then second one results in a blank view. The first view comes back if I click on the first one again in the sidebar.

Oh! Interesting issue! I surely tested two workflows, `five`, and `families2` (an example you provided in some issue). But I probably did not test both workflows at the same time :smile: 

Let me try it and see what happens.

kinow:
I think it's fixed now!

- moved the logic of computing the workflow out of Vuex, into the view Tree.vue
- fixed the router title problem, which was not displaying the workflow name in the toolbar
- tested with two workflows at the same time
    - one workflow (`families2` source below) occasionally had some exceptions due to a value missin from the `lookup` table (i.e. a task `firstParent` was missing from `familyProxies`). After that, the application wouldn't allow users to switch workflows, being piratically broken. Added a try/catch and will log the exception to the browser console, and allow users to keep using the rest of the app - there may be a moment that the workflow tree is empty due to this exception

kinow:
(just need to fix the tests)

hjoliver:
That's got it, looks great.

kinow:
>Finally, we need some way to access to the cytoscape graph view!

Without #98 there's no way to access the graph now :disappointed: unless one knows the URL. Do you have any suggestion where we could include a link for the Graph view? Or perhaps we should try to display both components at the same time by default?

hjoliver:
As a temporary measure, could we just keep the old scan table view via a link from the dashboard?

kinow:
> As a temporary measure, could we just keep the old scan table view via a link from the dashboard?

Sure. Or should we add that link under the Dashboard entry???? It just occurred to me that I could have simply added it under the existing values (a bit of d'oh moment). If that's fine, it'll be a 5 min job.

hjoliver:
Yep that’s fine 👍🏻

kinow:
> Yep that’s fine 👍🏻

Done

image

src/components/cylc/Drawer.vue
```diff
@@ -36,19 +36,22 @@
...
+          :to="link.to"
+          active-class="primary grey--text text--darken-3"
+          class="v-list-item"
       >
         <v-list-item-action>
           <v-icon>{{ link.icon }}</v-icon>
         </v-list-item-action>
         <v-list-item-title
-          v-text="link.text"
+            v-text="link.text"
```
kinow: Hm, I've added spaces to a few lines here without actually changing anything. Let me fix these.

src/App.vue
kinow: Moving it to `GScan.vue`. If necessary we can move it to somewhere else later 👍

src/components/cylc/GScan.vue
```diff
@@ -0,0 +1,204 @@
...
+    </div>
+  </div>
+</template>
+
+<script>
+import Job from '@/components/cylc/Job'
+import { workflowService } from 'workflow-service'
+
+const QUERIES = {
+  root: `
```
kinow: @oliver-sanders 👆

src/components/cylc/GScan.vue
```diff
@@ -0,0 +1,204 @@
...
+
+    getWorkflowIcon (status) {
+      switch (status) {
+        case 'running':
+          return 'mdi-play-circle'
+        case 'held':
+          return 'mdi-pause-octagon'
+        default:
+          return 'mdi-help-circle'
+      }
```
kinow: @oliver-sanders, @hjoliver also added an extra commit to choose the workflow icon. Given we have the icons loaded anyway, and they fit the CSS styles, would it be fine? Or do we need the component with SVG? They don't look exactly like the design sketch, but looks good to me.

image

(did a `cylc hold five` to test the other icon)

kinow: (note that the Workflows menu item is activated as it matches the "/workflows*" route)

hjoliver: They play icons look fine to me, and we can change to custom svg later on if we want to.

hjoliver:
I'm seeing a faint partial border (just the corners) around the task status icons though:
shot

kinow:
> 
> 
> I'm seeing a faint partial border (just the corners) around the task status icons though:
> shot

Your eyes are 20/20. I might need to visit my optician haha. Let me try to fix this, hopefully something simple. Thanks!

hjoliver:
> Your eyes are 20/20. 
 
Maybe my glass are :eyeglasses:

kinow:
Interesting, looks like there's nothing on Chrome, but on Firefox I can see these border, just like in your screenshot 😕

kinow:
Fixed (hopefully!). Gotta really buy some 🍻 to whoever worked on those browser developer tools. Found which class exactly was causing the element to render the borders. It was the `elevation-0`. I thought it would remove the elevation (buttons have have a fading grey border underneath it, to give the effect of elevation).

kinow:
One more commit, that is just adding some `if (Object.hasOwnProperty.call(object, 'property))` for the graph data.

We've added back the graph link, but the graph was broken due to properties being accessed before the object was populated/synced in/with Vuex store. Now it should

- add the gscan component under the existing menu items in the drawer/left sidebar
- fix the workflow title in the toolbar when browsing the tree/graph views
- fix the graph view

hjoliver:
Looking good.

Now wondering if gscan tool-tip reduction is feasible this week (or did that require new data from the back-end? I forget the result of that discussion...).

New:
shot

Old:
shot2

kinow:
>Now wondering if gscan tool-tip reduction is feasible this week (or did that require new data from the back-end? I forget the result of that discussion...).

I think @oliver-sanders pointed the [http endpoint](https://github.com/cylc/cylc-flow/blob/e3deb7d7049117e7c2a487f558f8b839a7d2ab25/lib/cylc/network/httpserver.py#L328) from where we used to get this data from, and in the discussion here and on Riot we were discussing that it would be possible to get state totals from the backend - though I believe it is not the same problem.

I think [this is the code in GScan that creates the message](https://github.com/cylc/cylc-flow/blob/de7d938496e82dbdfb165938145670dd8e801efd/lib/cylc/gui/gscan.py#L852) from your screenshot. It gets the data from [this method from the `state_summary_mgr`](https://github.com/cylc/cylc-flow/blob/de7d938496e82dbdfb165938145670dd8e801efd/lib/cylc/state_summary_mgr.py#L204).

image

I will try to reproduce this in JS :point_up:

hjoliver:
Thanks @kinow (and tomorrow is OK!)

kinow:
Rebased, and implemented:

- sorting the tasks within each state alphabetically (JS, when creating the summary data)
- display only N (default to 5) tasks (JS/Vue in the component, we use `array.slice(0, N)` to get up to N... then another `<span v-if="tasks.length > N">` that displays the text "And ${tasks.length - N} more" as before.

hjoliver:
Nice, thanks @kinow.  Looks great, works great!

We'll need sorting by "time" not name (e.g. when looking at succeeded tasks, want the most recently succeeded ones first).  But that can be a follow-up. This is great for my demo as-is (I won't tell anyone it's not really "most recent", he heh :grin: )

The Cylc 7 version also does this server side (I had forgotten) but it is probably better in principle that the UI does it (some future UI component might need to list all tasks in a state, not just a few.)

src/App.vue
hjoliver: This has been resolved below, with some follow-up needed. I'll update the main gscan issue #139 description.

src/components/cylc/GScan.vue
hjoliver: @kinow - I think you've addressed this, right? (gscan held icon).

src/components/cylc/GScan.vue
hjoliver: @oliver-sanders - can you post an issue for this, if you think it's needed.

src/App.vue
hjoliver: @oliver-sanders - ditto this - is an issue needed, it will it be obvious later on?

hjoliver:
@oliver-sanders - I'm going to merge this without your approval, as I need it in for demos soon. I believe all your comments have been addressed above, with some things punted to other issues (and/or updated the gscan Issue description). Note two questions on things that might need a new Issue page.

PR #311, opened 2019-11-11
Graph View style tweaks.

hjoliver:
Tidy up graph view styling, to at least make it more clean looking (if not actually conforming to our design at this stage, due to unresolved issues).
- use our job icon colours (mostly/sort-of) for graph nodes, and repeat some colours to approximate our new set of statuses better
- remove the gray "donut" shapes etc. inside each node (not sure what those were for, but they didn't look good ... maybe some experiment to approximate job progress or something in lieu of our svg task icons?)
- slightly less horrible font in the pop-ups
- also stops the animation from scratch on every update, for the other layouts, which is no good for a live monitor view (however, it still does a single double-jump for reasons that I couldn't figure out in the time available).

shot

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20631,9 +20631,9 @@
       "dev": true
     },
     "vue-cli-plugin-vuetify": {
-      "version": "2.0.1",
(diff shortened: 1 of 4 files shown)


kinow:
Small change, in a component that is new. The changes are contained within the graph component, so I'd be keen to merge it with a single review, unless you prefer to wait for a second reviewer @hjoliver

PR #314, opened 2019-11-12
Show ready/preparing task status.

hjoliver:
Get rid of the remaining `?` undefined task state icon that regular appears in the tree view.  The old "ready" status, still coming from the back-end, is to be renamed "preparing" - but we can use the new "open small dot" icon for it already. :+1: 

shot


kinow:
Small cosmetic change. One review is fine IMO. LGTM

PR #316, opened 2019-11-15
Make Job and Task functional components

kinow:
These changes partially address #222 
**Requirements check-list**

### DIFF
--- a/src/components/cylc/Task.vue
+++ b/src/components/cylc/Task.vue
@@ -40,7 +40,7 @@
           r="25"
           stroke-width="50"
           stroke-dasharray="157"
-          :style="`'stroke-dashoffset: ' + ${157 - (157 * (props.progress / 100))}`"
+          :style="`stroke-dashoffset: ${157 - (157 * (props.progress / 100))}`"
         ></circle>
         <!-- dot in the middle (small hub)
                * position in the middle
--- a/src/components/cylc/Task.vue
+++ b/src/components/cylc/Task.vue
@@ -3,7 +3,7 @@
          which can sit in text.
 -->
 
-<template>
+<template functional>
   <span
     class="c-task"
     style="display:inline-block; vertical-align:middle"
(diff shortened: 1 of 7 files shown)


kinow:
`Task` and `Job` are components that do not have data/state. They merely display whatever was passed down to them by the caller via props. So they are good candidates to be made into functional components, which require less resources.

kinow:
With the mocked data, doesn't give a huge improvement. But at least the number of Object and Vue components created appear to be consistently lower than comparing with `master` branch.

`master`

image

This PR

image

src/components/cylc/Task.vue
```diff
@@ -37,7 +40,7 @@
           r="25"
           stroke-width="50"
           stroke-dasharray="157"
-          v-bind:style="progressStyle"
+          :style="`'stroke-dashoffset: ' + ${157 - (157 * (props.progress / 100))}`"
```
oliver-sanders: I think this is how I originally implemented it but there was some problem with it not updating so it became a function.

tests/unit/components/cylc/task.vue.spec.js
```diff
@@ -7,47 +7,14 @@ import Task from '@/components/cylc/Task'
...
-    expect(wrapper.props().status).to.equal('')
-    expect(wrapper.props().isHeld).to.equal(true)
-    expect(wrapper.props().progress).to.equal(0)
-  })
-  it('should compute progress style', () => {
-    // local this gives the values used in the computed function, as we are not mounting the component
-    const localThis = {
-      progress: 25
-    }
-    expect(Task.computed.progressStyle.call(localThis)['stroke-dashoffset']).to.equal(157 - (0.25 * 157))
```
oliver-sanders: Why have these tests been removed?

kinow: Ah, yup. We cannot have `computed` values in functional components. A functional component is stateless, so it has no `data`, so nothing to compute. The code that was `computed` is still "computed", but it's done during the component creation, inlined with the backquotes. They are "computed" from the `props` values.

src/components/cylc/Task.vue
kinow: Hmmm, good to know. Let me test that.

kinow: Thanks for the comment @oliver-sanders ! That saved us some regression-bug-fix. The single quote would cause a problem here. Pushed one more commit that fixes this.

image

kinow:
Thanks @wxtim ! Two approvals, LGTM

Issue #295, opened 2019-10-31
Use vuetify directly without the creativetim theme


kinow:
**Describe exactly what you would like to see in an upcoming release**

We are not using most of what came with the theme. It was useful for prototyping the initial versions of the app. But after the #87 issue and the design sketch work, we have very little dependency on the theme.

Removing the theme will make it easier to update dependencies. When a new version of Vuetify came, we waited until the theme was updated. After dropping the theme, we can update the app as soon as Vuetify has been released - see #204, #276

**Additional context**

**Pull requests welcome!**

kinow:
The task consists in basically going through the theme (https://github.com/creativetimofficial/vuetify-material-dashboard), looking at what we are still using, and checking if we can use vue.js vanilla or vuetify directly.

PR #319, opened 2019-11-15
Move footer styles to SCSS file

kinow:
**Requirements check-list**

### DIFF
--- a/src/styles/cylc/_footer.scss
+++ b/src/styles/cylc/_footer.scss
@@ -3,3 +3,4 @@
   color: map-get($grey, 'darken-1');
   border-top: 1px solid #e7e7e7;
 }
+
--- a/src/components/core/Footer.vue
+++ b/src/components/core/Footer.vue
@@ -27,9 +27,3 @@ export default {
   }
 }
 </script>
-
-<style>
-#core-footer {
-  z-index: 0;
-}
-</style>


src/styles/cylc/_footer.scss
```diff
@@ -3,3 +3,7 @@
   color: map-get($grey, 'darken-1');
   border-top: 1px solid #e7e7e7;
 }
+
+#core-footer {
+  z-index: 0;
```
oliver-sanders: Out of interest why are we setting `z-index: 0`? Shouldn't have any effect?

kinow: Hmmm, good question. Had a look with Firefox, and removing this value, the `#core-footer` element still has `z-index: 0` (I guess that's the default value).

Not sure why the guys from creativetim added it (they still have this in the `master` version - https://github.com/creativetimofficial/vuetify-material-dashboard/blob/2a3ea658e72e7109b7aa65be73abb6dfc547e3ea/src/components/core/Footer.vue#L45).

Any objection to removing it? We can add it back later if necessary, and it would be helping towards #295

hjoliver: @kinow - do you want to just remove this line then (making the PR just a small deletion?)? Then we can merge it.

kinow: Sure, done in a couple minutes :zap:

src/styles/cylc/_footer.scss
```diff
@@ -3,3 +3,4 @@
   color: map-get($grey, 'darken-1');
   border-top: 1px solid #e7e7e7;
 }
+
```
kinow: Left the extra line as harmless, and was missing before.

src/styles/cylc/_footer.scss
kinow: Done

PR #324, opened 2019-12-04
Execute GitHub action on pull requests too and fix coverage

kinow:
This is a small change with no associated Issue.

Noticed that after marking a PR as ready for review it was not showing that the GItHub action had run successfully. Then found [a similar issue](https://github.com/sdras/awesome-actions/issues/98) (that repository has a lot of useful things for actions), and turns out I also did not include the `on: [pull_request]`, only `on: [push]`.

EDIT: added an extra commit to fix the coverage reports. Will have to add a secret token to the repo settings too, as documented in the Codecov github action docs: https://github.com/codecov/codecov-action

One review should to :+1: 

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Does not need tests (why? build only).
- [x] No change log entry required (why? e.g. invisible to users).
- [x] No documentation update required.

PR #325, opened 2019-12-06
Create Workflow component with Lumino

### LINKED ISSUE #90: Create a Workflow component

### LINKED ISSUE #323: Menu disappears when in responsive/mobile mode
**Describe the bug**

Menu disappears when the app viewport is resized to mobile/responsive screen size.

`master`

Resize viewport to less than 900px - more or less the mobile breaking point for our app, which should hide the drawer and bring the mobile/hamburger menu.

**Expected behavior**

From what I remember, we used to have a hamburger menu (that icon of the three patties stacked together).

**Screenshots**

image

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #90, and close #323 

**Requirements check-list**

### DIFF
--- a/src/components/cylc/workflow/Workflow.vue
+++ b/src/components/cylc/workflow/Workflow.vue
@@ -57,14 +57,12 @@ export default {
     })
   },
   methods: {
-    addTreeWidget () {
-      const id = `tree-widget-${new Date().getTime()}`
+    addTreeWidget (id) {
       const contentWidget = new ContentWidget(id, 'tree')
       this.dock.addWidget(contentWidget)
       this.treeWidgetIds.push(id)
     },
-    addGraphWidget () {
-      const id = `graph-widget-${new Date().getTime()}`
+    addGraphWidget (id) {
       const contentWidget = new ContentWidget(id, 'graph')
       this.dock.addWidget(contentWidget)
       this.graphWidgetIds.push(id)
--- a/src/components/cylc/workflow/Workflow.vue
+++ b/src/components/cylc/workflow/Workflow.vue
@@ -1,0 +1,75 @@
+<template>
+  <div>
+    <div ref="main" id="main" class="pa-4 fill-height"></div>
+    <div v-show="false">
+      <tree-wrapper
+          v-for="widgetId of this.treeWidgetIds"
+          :key="widgetId"
+          :workflows="workflowTree"
+          :widgetId="widgetId"
+      />
+      <graph-wrapper
+          v-for="widgetId of this.graphWidgetIds"
+          :key="widgetId"
+          :workflow-name="workflowTree[0].name"
+          :widgetId="widgetId"
+      />
+    </div>
+  </div>
+</template>
+
+<script>
+import { BoxPanel, DockPanel, Widget } from '@lumino/widgets'
+import { ContentWidget, EventBus, GraphWrapper, TreeWrapper } from '@/components/cylc/workflow/index'
+
+export default {
+  name: 'Workflow',
+  props: {
+    workflowTree: {
+      type: Array,
+      required: true
+    }
+  },
+  components: {
+    'tree-wrapper': TreeWrapper,
+    'graph-wrapper': GraphWrapper
+  },
+  data () {
+    return {
+      // create a box panel, which holds the dock panel, and controls its layout
+      main: new BoxPanel({ direction: 'left-to-right', spacing: 0 }),
+      // create dock panel, which holds the widgets
+      dock: new DockPanel(),
+      treeWidgetIds: [],
+      graphWidgetIds: []
+    }
+  },
+  created () {
+    this.dock.id = 'dock'
+    this.main.id = 'main'
+    this.main.addWidget(this.dock)
+    window.onresize = () => { this.main.update() }
+    BoxPanel.setStretch(this.dock, 1)
+    const vm = this
+    EventBus.$on('add:tree', () => {
+      const id = `tree-widget-${new Date().getTime()}`
+      const contentWidget = new ContentWidget(id, 'tree')
+      vm.dock.addWidget(contentWidget)
+      vm.treeWidgetIds.push(id)
+    })
+    EventBus.$on('add:graph', () => {
+      const id = `graph-widget-${new Date().getTime()}`
+      const contentWidget = new ContentWidget(id, 'graph')
+      vm.dock.addWidget(contentWidget)
+      vm.graphWidgetIds.push(id)
+    })
+    this.$nextTick(() => {
+      Widget.attach(vm.main, vm.$refs.main)
+    })
+  },
+  beforeDestroy () {
+    EventBus.$off('add:tree')
+    EventBus.$off('add:graph')
+  }
+}
+</script>
--- a/src/components/cylc/workflow/Workflow.vue
+++ b/src/components/cylc/workflow/Workflow.vue
@@ -11,7 +11,7 @@
       <graph-wrapper
           v-for="widgetId of this.graphWidgetIds"
           :key="widgetId"
-          :workflow-name="workflowTree[0].name"
+          :workflow-name="workflowTree.length > 0 ? workflowTree[0].name : ''"
           :widgetId="widgetId"
       />
     </div>
@@ -20,7 +20,7 @@
 
 <script>
 import { BoxPanel, DockPanel, Widget } from '@lumino/widgets'
-import { ContentWidget, EventBus, GraphWrapper, TreeWrapper } from '@/components/cylc/workflow/index'
+import { ContentWidget, GraphWrapper, TreeWrapper } from '@/components/cylc/workflow/index'
 
 export default {
   name: 'Workflow',
@@ -52,25 +52,23 @@ export default {
     window.onresize = () => { this.main.update() }
     BoxPanel.setStretch(this.dock, 1)
     const vm = this
-    EventBus.$on('add:tree', () => {
-      const id = `tree-widget-${new Date().getTime()}`
-      const contentWidget = new ContentWidget(id, 'tree')
-      vm.dock.addWidget(contentWidget)
-      vm.treeWidgetIds.push(id)
-    })
-    EventBus.$on('add:graph', () => {
-      const id = `graph-widget-${new Date().getTime()}`
-      const contentWidget = new ContentWidget(id, 'graph')
-      vm.dock.addWidget(contentWidget)
-      vm.graphWidgetIds.push(id)
-    })
     this.$nextTick(() => {
       Widget.attach(vm.main, vm.$refs.main)
     })
   },
-  beforeDestroy () {
-    EventBus.$off('add:tree')
-    EventBus.$off('add:graph')
+  methods: {
+    addTreeWidget () {
+      const id = `tree-widget-${new Date().getTime()}`
+      const contentWidget = new ContentWidget(id, 'tree')
+      this.dock.addWidget(contentWidget)
+      this.treeWidgetIds.push(id)
+    },
+    addGraphWidget () {
+      const id = `graph-widget-${new Date().getTime()}`
+      const contentWidget = new ContentWidget(id, 'graph')
+      this.dock.addWidget(contentWidget)
+      this.graphWidgetIds.push(id)
+    }
   }
 }
 </script>
--- a/src/components/cylc/workflow/Workflow.vue
+++ b/src/components/cylc/workflow/Workflow.vue
@@ -31,6 +31,7 @@ export default {
     }
   },
   components: {
+    // TODO: once our components (tree, graph, dot, etc) share a common interface (i.e. same props) we should be able to have a single wrapper
     'tree-wrapper': TreeWrapper,
     'graph-wrapper': GraphWrapper
   },
--- a/src/views/Workflow.vue
+++ b/src/views/Workflow.vue
@@ -1,35 +1,16 @@
 <template>
   <div id="workflow-panel" class="fill-height">
     <toolbar />
-    <div ref="main" id="main" class="pa-4 fill-height"></div>
-    <div v-show="false">
-      <tree-wrapper
-          v-for="widgetId of this.treeWidgetIds"
-          :key="widgetId"
-          :workflows="workflowTree"
-          :widgetId="widgetId"
-      />
-      <graph-wrapper
-          v-for="widgetId of this.graphWidgetIds"
-          :key="widgetId"
-          :widgetId="widgetId"
-      />
-    </div>
+    <workflow :workflow-tree="workflowTree" />
   </div>
 </template>
 
 <script>
-import { BoxPanel, DockPanel, Widget } from '@lumino/widgets'
 import { mixin } from '@/mixins/index'
-import { workflowService } from 'workflow-service'
-import Toolbar from '@/components/cylc/Toolbar'
 import { mapState } from 'vuex'
-import { convertGraphQLWorkflowToTree } from '@/components/cylc/tree'
-import Tree from '@/components/cylc/Tree'
-import Vue from 'vue'
-import Graph from '@/components/cylc/Graph'
-
-export const bus = new Vue()
+import Toolbar from '@/components/cylc/Toolbar'
+import Workflow from '@/components/cylc/workflow/Workflow'
+import { convertGraphQLWorkflowToTree } from '@/components/cylc/tree/index'
 
 // query to retrieve all workflows
 const QUERIES = {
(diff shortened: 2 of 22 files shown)


kinow:
I was just going to get my toes wet learning Lumino, then when I realized it wasn't really hard decided to start digging in the TypeScript code (no docs) and examples. Tried to match as much as possible the design sketch.

There are some limitations to how much we can tweak due to the way it builds screens with multiple widgets, each having some logic to decide the height/width. For instance, I wanted to increase font size and give it more space (padding/margins) to the tabs, but this is quite complicated without docs/examples.

One major issue now is reactivity. I am attaching the Vue component, and destroying it (I think, hard to spot the exact object in the browser memory, but will confirm later). But the component is not being updated. i.e., here's what it looks like at the moment:

image

So each time I add a new Tree component, it picks up the current workflow status, then freezes. Not being updated even though Vuex data is still changing. So I've got to sort this out before this is ready for review.

Then need to check which commits need to be squashed, check if it's possible to add any tests (don't think so at this stage), add the changelog, and it should be ready for review :confetti_ball:

kinow:
Also, almost forgot, not sure if we can use the Graph component. In order to add the Graph here, **we must be able to create the Graph component programmatically**.

If the Graph component (not the view) depends on VueRouter, extra objects, or if we are not able to create multiple instances of the Graph component, then we will need to refine the component code. But once that's done, adding a new Graph Widget (Lumino layout uses widgets, not components) shouldn't be too hard (<1 day work, couple hours perhaps).

kinow:
Eureka moment going for a walk. Insteas of creatijg a new tree component myself, it should be possible to instead simply add some value to an array, then use a v-for to get vue to create the component for me.

May have to hide the component, or find a way to tell vue where it should go in the dom tho. But a possible fix for reactivity.

kinow:
Came back from x-mas function and had to finish it otherwise I wouldn't be able to enjoy the weekend :smile: 

GIFrecord_2019-12-07_013024

Next week will review the commits, add changelog, and then wait until @oliver-sanders or somebody else has time to review the current draft, before adding any more code. But so far, I'm really happy with Lumino :+1:

hjoliver:
That looks brilliant, well done :bouquet:

kinow:
Ready for review. Will try to add a few unit tests, but that shouldn't alter the current code, so happy to take any reviews.

The last commit alters the `Graph` component. I changed as little as I could to get it working as a Lumino widget, and as a standalone Vue component. Tested the existing `Graph` view, and it appears to be working fine, but better if reviewers can check it too.

Here's what it should look like:

GIFrecord_2019-12-09_103557

A few known issues:

- Adding two graphs, only one is updated
- Adding graphs only works on Chrome and Ffox, IE almost works. In IE, I can see the graph in the mini-screen at the bottom right, but not in the main graph area
- There is an increase in memory usage. We will need to keep refining the code incrementally later.
- Eventually, you may see an error in the console saying something like "Cannot access name of undefined", this is caused by families and tasks not matching in the returned GraphQL data. I will need to isolate the issue, reproduce, and create an issue either in Cylc UI or in Cylc UI Server to handle that.

The issues above are not blockers IMO, as they are not directly caused by the new `Workflow` component.

kinow:
NB: No CI on this PR as #324 is still pending review :+1:

kinow:
Unit tests added, change log too. Commits are as organized as I could, without re-writing code. They are more or less grouped, there some CSS/style commits that could have been grouped together but I worked on separate parts of the code while trying to integrate Lumino, sometimes on the JS part, sometimes on the CSS.

Tested again on Ffox/Chrome. Ran the CI commands locally and everything passed. The CI passed on my repository when I pushed it, so it's possible to confirm it should have passed here too: https://github.com/kinow/cylc-ui/actions (only not running because we need #324 in order to trigger the action on pull requests and on push).

Coverage is not being uploaded in the action. It will reduce, probably, as we have some code for a view added here that is not being tested, only the component is. Will fix the coverage in another PR.

oliver-sanders:
Nice!

This is one of those really big things which brings the goal into closer view.

Few niggles:

* I've got a font issue on the close icons:
    Screenshot from 2019-12-10 13-35-09
* The graph view currently isn't showing anything for me (no console traceback).
* Resizing doesn't seem to work.

kinow:
>I've got a font issue on the close icons:

Ah, good to know. I chose a special character for the X close button, but should have done the simple first. I'm pushing one commit to use a simple letter X now.

>The graph view currently isn't showing anything for me (no console traceback).

Maybe if you try on a different browser. Though I am afraid the graph component needs a few more changes in order to be used as a widget with lumino (and we need to run a few more tests with cytoscape/etc, as IE Edge is not working, though I believe I've used Cytoscape apps with IE before)

>Resizing doesn't seem to work.

This is a bit tricky with lumino. If you position two widgets side by side, then you should be able to resize their width, respecting a minimum-width that I set (otherwise when opening on mobile or smaller screens you wouldn't see the widget content.

But as for height, the hard limit is the screen size I think. Initially I don't set a height for lumino widgets. I set a min-height, and then it computes the widget height. I couldn't pinpoint exactly where they are doing that in their TypeScript code... but the best I could do was to change the height of the content div to occupy as much as possible (see screenshot below) and also define a min-height.

image

oliver-sanders:
> This is a bit tricky with lumino

Sorry, I meant resizing the "panes" within Lumino by hovering over the gap between panes until you get the arrow icon, then using click+drag.

kinow:
> > This is a bit tricky with lumino
> 
> Sorry, I meant resizing the "panes" within Lumino by hovering over the gap between panes until you get the arrow icon, then using click+drag.

Ah, if that's working in JupyterLab maybe we can copy it from there. I used their examples to implement this PR. Here's how it works in one of their examples.

GIFrecord_2019-12-10_135806

kinow:
Will try to fix the conflicts here today.

kinow:
Will fix conflicts here after #199 and #280 :+1:

kinow:
PR updated. Conflicts fixed. Had to update the `GraphWrapper` component. This component simply gets the required parameters to create a `Graph` component from the *Add Tree* dialogue, plus data from Vuex store, and creates a `Graph` to be added as a Lumino widget/Vue component.

Tested with the recently merged subscriptions/websockets, and two workflows. Almost everything seems to be working as expected.

There is one thing that I may fix after the Tree component. If you have 2 workflows, and you open one and add some views (e.g. one tree, and one graph).

Then, when you switch to another workflow, the two views are kept, and the tasks/jobs are replaced by the new component. Here's what I mean.

GIFrecord_2019-12-23_161108

Between bug and feature, I think this probably falls in the category of bug/unexpected behaviour.

kinow:
Argh, one test failing :angry:

kinow:
GitHub action happy. This is officially ready for review! Will keep an eye for comments/reviews, but focusing now on the treeview as holiday homework (read: nerdy fun).

oliver-sanders:
Resizing Lumino tabs is now working brilliantly!

I think there is an issue with padding, the whitespace under the toolbar appears to be associated with the `<main>` element. The whole page is vertically scrolling by a similar amount so that might be related.

This could be booted to a follow-on PR.

Screenshot from 2019-12-23 17-35-52

src/components/cylc/workflow/Workflow.vue
```diff
@@ -0,0 +1,75 @@
...
+export default {
+  name: 'Workflow',
+  props: {
+    workflowTree: {
+      type: Array,
+      required: true
+    }
+  },
+  components: {
+    'tree-wrapper': TreeWrapper,
```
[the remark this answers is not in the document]
kinow: Oh, true! Well spotted. Initially the `Graph` component had no props, until a few days ago when we added the `workflowName`. But in the end having a single interface would eliminate the need of multiple wrappers :+1: (hadn't noticed it, I would probably let it slip later. Will add a TODO later)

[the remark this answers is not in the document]
oliver-sanders: Sure.

src/views/Workflow.vue
```diff
@@ -0,0 +1,194 @@
...
+    ...mapState('workflows', ['workflows']),
+    currentWorkflow: function () {
+      for (const workflow of this.workflows) {
+        if (workflow.name === this.workflowName) {
+          return workflow
+        }
+      }
+      return null
+    },
+    workflowTree: function () {
```
oliver-sanders: I think this interface kinda breaks the idea of views being able to request the data they subscribe to as the workflow component is subscribing to a superset of everything even if it is not needed (e.g. only job statuses/submit numbers required for the graph view).

Could we move this functionality into the store?

kinow: I think you are right. Will try to move to Vuex as a `getter` :+1:

kinow: The `currentWorkflow` was moved to the store recently too (above the code from this comment). Replacing both by `mapGetters`.

kinow: Done

src/components/cylc/workflow/Workflow.vue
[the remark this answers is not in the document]
kinow: TODO added.

kinow:
> Resizing Lumino tabs is now working brilliantly!
> 
> I think there is an issue with padding, the whitespace under the toolbar appears to be associated with the `<main>` element. The whole page is vertically scrolling by a similar amount so that might be related.
> 
> This could be booted to a follow-on PR.
> 
> Screenshot from 2019-12-23 17-35-52

It worked before, but I think I know where I broke it. The only recent change that could cause this was that PR for the Toolbar, where I added the `app` attribute to the `Toolbar` component, so that it would be positioned correctly in the UI.

The `Toolbar` component default height is `96px`, but as I've modified it with `dense`, it is `48px`. However, the padding seems to be ignoring that value... perhaps because of the `app` attribute. Let me try to update Vuetify, or check if it'd be a simple fix (have a few spare minutes now)

kinow:
Updating everything with `ncu -u` did not work. Even though it updated Vuetify. So it might be a configuration that is wrong or missing somewhere.

src/views/Workflow.vue
oliver-sanders: Wow, that was quick!

kinow:
Ahh!! Had to install the dependencies, try all possible combination of props in the `Toolbar` component in the `Default` layout.

Just to find that there were two toolbars after playing with the CSS.

image

Yup, one in the layout, and another one in the `Workflow` view. I removed the `Toolbar` from `Tree` and `Graph`, but never updated `Workflow` from this PR :man_facepalming: 

Fixed @oliver-sanders ! :laughing: 

image

oliver-sanders:
> Yup, one in the layout, and another one in the Workflow view

Aha! Great spot.

hjoliver:
(latest change tested as working -nice)

hjoliver:
Note lumino multi-pane resizing can appear to be broken if your browser window isn't big enough, because the minimum pane sizes are set quite large.

kinow:
> Note lumino multi-pane resizing can appear to be broken if your browser window isn't big enough, because the minimum pane sizes are set quite large.

Exactly. We can change the minimum height/width later if necessary. I chose values arbitrarily.

src/views/Workflow.vue
```diff
@@ -0,0 +1,171 @@
...
+  </div>
+</template>
+
+<script>
+import { mixin } from '@/mixins/index'
+import { mapState, mapGetters } from 'vuex'
+import Workflow from '@/components/cylc/workflow/Workflow'
+
+// query to retrieve all workflows
+const QUERIES = {
```
oliver-sanders: I think this query and the subscription logic can now be safely removed.

oliver-sanders:
Tested as working [nicely].

There is a bit of a vertical space issue the `#workflow-panel` doesn't seem to be utilising 100% of the height of its containing `#core-view`. I think this is caused by `.c-graph` which overspills its container vertically. Can move to a follow-on.

src/views/Workflow.vue
kinow: I think we still need this query? (though I spent some two minutes wondering if you were right first).

I think you mean we already have a query in the background for `GScan`?

The `GScan` query has no `familyProxies` and no `nodesEdges`. This query in the `Workflow` view is to bring the data required for the components/widgets that can be added.

Does that make sense?

oliver-sanders: The different views should be registering the subscriptions for the things they need. If we set the query here then we will always be requesting node-edges information even when no using the graph view.

kinow: Yup. But I couldn't find another place to add the query.

Tree and Graph components are pure components, not bound to routes and have a different lifecycle (i.e. the beforeDestroy may not be called depending on what you are doing).

kinow: That was not a very constructive comment. Let me try to adapt the current pattern to the tree and gtaph components and then try to move the queries over there.

It might just work 👍

kinow:
Squashed commits :+1: still pending one item from Oliver's review.

src/views/Workflow.vue
oliver-sanders: Aah, I didn't realise that the event hooks fire differently.

Perhaps we could make the subscription a `computed` element which could change based on the state of the component and make the workflow component responsible to collecting and issuing subscriptions?

kinow: What about leaving the queries here, but the initial state of the view is no query subscribed? Then when users click on "Add View", a new query will be added based on the user selected option (tree/graph).

oliver-sanders: I think the subscription should be controlled by the component (tree, graph etc):

* Closer association of subscription and consumption of data:
  * Possibility for auto unit-testing that consumption matches subscription.
  * Possibility to plug components into a different framework (e.g. run component as a standalone application) nice for testing/integration.
* More decentralised and plugin friendly:
  * Plugins are self-contained providing an interface to the subscription.
* More powerful allowing performance enhancements:
  * The subscription of a view is a function of its state...
  * The subscription may change during the life of the component as what the user chooses to see could effect the subscription we issue.
  * This allows for opportunistic optimisation, e.g. if the user is collapsing the view by family we only need to request `familyProxies` and can drop the subscription to `taskProxies` and with it `jobs` massively reducing load.

kinow: It is a good practice to separate control from the components. Vue itself is focused on the view layer only ([ref from their user guide](https://vuejs.org/v2/guide/#What-is-Vue-js)). Adding control and too much state to components is considered bad.

This is exactly why we have a service in our code - Angular, for example, does MVC, while Vue does MV only if you consider the state management as model. By adding the control of state into components, you run risk of code tangling, and lose the separation of concerns.

The same points you listed are used as arguments against using Axios directly in components, and instead simplifying them and externalizing the control to service/API objects. As it is easier to test, maintain, optimize, etc. Without interfering with template, layout, etc.

My goal with this separation is to have the Tree/Graph/Dot/etc components as simple to understand and maintain as possible, following what's considered a simple approach in Vue.

Another goal I have in mind, is combining "snapshots" of Cylc workflows with JupyterLab widgets. So that I could use a notebook to detail and document what I am doing, and even render the workflow with the component - without the need of Cylc UI or screen shots. Which should be simple enough if we have a component that just needs something like

```
import Vue from 'vue'
import Tree from 'cylc/tree'

// whatever code my widget needs, or any other Web app
// then instantiate a simple Vue app
<tree :data=snapshotJsonObject />
```

Now, if the `Tree` component tries to subscribe, or try to execute GraphQL queries, etc, it won't be easy to re-use it. Even within Cylc, if we ever want to provide a simple static visualization it wouldn't work.

This is a contrived example, based on what I want, and while it is not a core feature of Cylc UI, I think this can be used as an example to demonstrate benefits of having components being simplified, receiving `props`, `slots`, etc.

But I can see your point of view too. It is way easier to write components that call ApolloClient or Axios directly, but IMO it gets harder to maintain and to write tests.

But I think we are not making good progress here, so we could perhaps discuss this in the next group VC, in the Feb workshop, or wait for others from @cylc/core to chime in?

oliver-sanders: We really need subscriptions hammered down soon as this will block other work, I thought we had it sorted before.

> Now, if the Tree component tries to subscribe, or try to execute GraphQL queries

That's not what I was describing. I don't think the Tree component should subscribe or execute GraphQL queries at all. That's the job of the workflow service.

I do, however, think that the Tree component should provide a manifest of the information it requires (similar to other component patterns). This manifest is a GraphQL subscription string. 

The Tree component wouldn't control anything, just define its requirements. The logic to merge and execute subscriptions happens higher up in the system.

Fundamentally you cannot separate the subscription and the component, it would be like trying to separate the `package.json` file and the codebase, one defines the requirements for the other. You can try to move the subscription to another file, but it's still intrinsically linked to the component and will have to be updated with each change to the compoent.

oliver-sanders: I was describing a system which works like this:

* The Cylc "Views" (tree/dot/graph) each provide an interface for obtaining the subscription:

```js
SUB = `
    workflows {
         id
    }
`

computed: {
  subscription: () => { return SUB }
}
```

* When we "spin up" Cylc "Views" we extract the subscription from it.

```js
$workflowService.subscribe(cylcView.subscription)
```

This can be done in the Workflow Component or Workflow Service itself.

* So the Workflow Service is in charge, it *fetches* subscriptions, merges them, executes GraphQL, syncs the central data store.

kinow: >I do, however, think that the Tree component should provide a manifest of the information it requires (similar to other component patterns). This manifest is a GraphQL subscription string.

Ah, I see, sorry I misunderstood it. I think the `props` could be this manifesto then.

>Fundamentally you cannot separate the subscription and the component, it would be like trying to separate the package.json file and the codebase, one defines the requirements for the other. You can try to move the subscription to another file, but it's still intrinsically linked to the component and will have to be updated with each change to the compoent.

I see subscriptions as related to the view/route you are visiting. Not to the components you have on that view/route.

I could be wrong, but maybe this is a practical example of where we could use this separation. If we add a view/route in Cylc UI, to replace the old `cylc graph` command, we could re-use the same component. But instead of a GraphQL query, the `<graph />` component in the view/route would need just the dependency graph.

In this case, the view/route would not have any subscription. But the component would still work because it just needs to receive the data required via props. And in this case I think we would not have any GraphQL involved.

kinow: >I was describing a system which works like this:
>
>- The Cylc "Views" (tree/dot/graph) each provide an interface for obtaining the subscription:

The difference being that in this case we have a Workflow view, with these three components. It may have 0 or multiple instances of each of these components (tree/dot/graph).

What I mean by view, being a Vue component bound to a route, where we have lifecycle of the route (user entered, left the route, or before the user entered the route, etc). And that normally uses multiple components that may be present in other views.

And what I mean by component, is any instance of Vue that is re-used across views, and that is not bound to a route. Having a different lifecycle, and that can also be re-used by Vue - i.e. Vue may not destroy it, to re-use for performance, which again changes its lifecycle (e.g. `beforeDestroy` and `destroy` would not be called, instead we would need to hook `updated`)

oliver-sanders: `cylc graph` is a bit of a funny one, this needs to be done at the UIS (because we need to parse the suite.rc, form the dependency graph and use other Python code) so at the UI we can re-use the component, the subscription should actually be much the same too.

kinow:
Fixed #323 while testing this PR with the "Add View" link.

Screen Shot 2020-01-06 at 20 48 24

kinow:
Finished moving the subscriptions to the view. The `Workflow` component reacts to events, and adds new widgets. When a widget is added - alongside a Vue wrapper - we also create a subscription.

The widget has the same ID as the subscription. Later, when the user closes the widget, we capture the Lumino event, and emit a new Vue event. That event then triggers the deletion of the Vue components (wrapper et al) and the removal of the subscription.

This a compromise for the discussion around where to keep subscriptions.

The components are free of GraphQL details or queries. The subscription code (`gquery` in special) is not touched, and works as per design. Screenshot showing one way to confirm what subscriptions we have:

GIFrecord_2020-01-09_161017

Final notes: this PR has a commit that I picked from the Infinite Tree PR to fix the offline mode. Whichever goes first, the other PR will need amending. It is not really important, as they will cause conflicts anyway in the tree component. The conflicts will take a couple of days for fixing and testing, as I haven't experimented the infinite tree with lumino :+1:

oliver-sanders:
Okay, so you've copied the queries from the tree and graph view into the workflow component?

If so should the query in `src/views/Graph.vue` still be there?

kinow:
> 
> 
> Okay, so you've copied the queries from the tree and graph view into the workflow component?
> 
> If so should the query in `src/views/Graph.vue` still be there?

Replied on Riot, copying here for reference

Yes, I think @hjoliver  mentioned he'd like to keep the old table at /workflows working. That one where users can see a listing of workflows, and under Actions I think there are links for the Tree view and for the Graph view.

src/views/Workflow.vue
oliver-sanders: (resolved by compromise, see comments below)

oliver-sanders:
I think this is perfectly usable and further enhancements / tweaks can be punted as this is already a sizeable change - hence #358 which we can live with until `0.3`.

PR #338, opened 2019-12-16
task: support expired state

oliver-sanders:
Closes the "eccentric task states" component of #109

Proposal to handle the expired state:
  * ~Mark "expired" tasks as "succeeded", this is how they are handled in the graph.~
  * ~Most the time they should be easily distinguished from naturally succeeded tasks by the absence of a job.~
  * Treat "expired" as a top-level status with its own icon.

Questions:
  * [x] Should we do anything more to visually distinguish "expired" from "succeeded"? - *yes*

Screenshot:

Screenshot from 2019-12-23 09-43-59

**Requirements check-list**
- [x] Does not need tests - durp.
- [x] No change log entry required - document later
- [x] No documentation update required - documented within project

### DIFF
--- a/src/components/cylc/Task.vue
+++ b/src/components/cylc/Task.vue
@@ -17,7 +17,7 @@
       class="task"
       :class="[
         props.isHeld ? 'held' : '',
-        ['waiting', 'preparing', 'ready', 'submitted', 'running', 'succeeded', 'failed', 'submit-failed'].includes(props.status) ? props.status : 'unknown'
+        ['waiting', 'preparing', 'ready', 'expired', 'submitted', 'running', 'succeeded', 'failed', 'submit-failed'].includes(props.status) ? props.status : 'unknown'
       ]"
       viewBox="0 0 100 100"
     >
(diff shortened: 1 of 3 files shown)


hjoliver:
> Mark "expired" tasks as "succeeded" ...

I dunno about that. Expired means "waited too long, so give up and don't run it".  I don't see how that is equivalent to succeeded.  

> ... , this is how they are handled in the graph.

Is that right?  We don't trigger downstream tasks as if on success do we?  (tasks with an expire time would mostly not have downstream dependents, I expect).

hjoliver:
I vaguely recall we thought (or at least I did) that expired is the one "eccentric task state" (he heh, good term) that needs to remain a task state.

... which gels with my proposal here: https://cylc.github.io/cylc-admin/proposal-state-names.html#new-abstract-task-states

hjoliver:
A succeeded task with no jobs most likely indicates manual state reset, rather than expired (IMO)

oliver-sanders:
Fair enough, will dream up a new icon and update this PR.

hjoliver:
My icon suggestion from #109 
image

(But you may have a better idea).

oliver-sanders:
expired

kinow:
What about just a clock icon @oliver-sanders ? Similar to @hjoliver 's suggestion, without the warning symbol. I think we don't have a clock yet, and I think there are also some clock in unicode. Otherwise, I like options 2, 3, and 5. I find 1 reminds me too much of error, and 4 reminds me of the stop sign and of a failure.

hjoliver:
I think we will have a plain clock to signify a clock trigger, although it's not a task icon.

hjoliver:
https://www.google.com/search?q=expired+icon&client=firefox-b-e&sxsrf=ACYBGNSasCR1tPuIejUxd7oMsZFJElccXg:1576545302355&source=lnms&tbm=isch&sa=X&ved=2ahUKEwjNm-D1wLvmAhVOzTgGHXBuAZEQ_AUoAXoECA8QAw&biw=1920&bih=969

hjoliver:
I like this:
image

hjoliver:
(But the text would be too small, damn it).

oliver-sanders:
Did think about a clock but I think its meaning isn't especially obvious and it's graphically complicated.

hjoliver:
Technically “expired” does mean “clock expired” (expiry happens at some clock offset relative to cycle point). However this is so rarely used it doesn’t matter nearly as much as the other task state icons

oliver-sanders:
Here are four others which I didn't put up with the previous set:

expired-extra

hjoliver:
I prefer the last one, as a cleaner version of my original suggestion.  Expired is a clock setting after all, and when a task hits its expiry time that warrants an "!".  A massive letter "E" is kind of a cop out, as we've now managed to avoid resorting to that for all the other icons.

kinow:
The last one looks good to me too 👍

oliver-sanders:
Updated with clockface icon.

PR #341, opened 2019-12-18
Fix toolbar (responsive mode and drawer)

kinow:
These changes close #199 

Moved `Toolbar` component back to the `Default` layout, making it dependent on Vuex store `workflows.workflow` state, instead of `props`.

The components `Tree`, `Graph`, 

**Requirements check-list**

### DIFF
--- a/src/components/cylc/Graph.vue
+++ b/src/components/cylc/Graph.vue
@@ -546,7 +546,12 @@ const expandCollapseOptionsCola = {
 
 export default {
   name: 'Graph',
-  props: ['workflowName'],
+  props: {
+    workflowName: {
+      type: String,
+      required: true
+    }
+  },
   data: function () {
     return {
       isConnected: false,
@@ -570,7 +575,6 @@ export default {
       nodesEdges: [],
       workflows: [],
       subscriptions: {},
-      workflowId: '',
       // layout engines
       layoutEngines: [
         'dagre',
@@ -606,7 +610,7 @@ export default {
 
   metaInfo () {
     return {
-      title: this.getPageTitle('App.graph', { name: this.workflowId })
+      title: this.getPageTitle('App.graph', { name: this.workflowName })
     }
   },
 
@@ -633,7 +637,6 @@ export default {
 
   created (cy) {
     console.debug('CREATED')
-    this.workflowId = this.$route.params.workflowName
     workflowService.register(
       this,
       {
@@ -652,7 +655,7 @@ export default {
     subscribe (queryName) {
       const id = workflowService.subscribe(
         this,
-        QUERIES[queryName].replace('WORKFLOW_ID', this.workflowId)
+        QUERIES[queryName].replace('WORKFLOW_ID', this.workflowName)
       )
       if (!(queryName in this.subscriptions)) {
         this.subscriptions[queryName] = {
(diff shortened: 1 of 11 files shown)


kinow:
Hmmm, tests should be passing now. But I did a bit more of manual testing and I am not happy with the current behaviour when you refresh pages, resize it, etc. Will do a bit more of testing before declaring it ready for review - should have created a draft for just-in-case, sorry.

src/components/cylc/Graph.vue
```diff
@@ -633,7 +633,7 @@ export default {
 
   created (cy) {
     console.debug('CREATED')
-    this.workflowId = this.$route.params.workflowid
+    this.workflowId = this.$route.params.workflowName
```
oliver-sanders: The use of ID and Name is somewhat confusing.

tests/unit/components/cylc/toolbar.vue.spec.js
```diff
@@ -1,55 +1,53 @@
...
+    // v-app-toolbar when using "app" directive, must also have a v-app. So we need to initialize vuetify
     const wrapper = mount(Toolbar, {
-      store,
-      propsData: {
-        workflow: {
-          id: 'Workflow Test Toolbar',
-          status: TaskState.RUNNING.name.toLowerCase()
-        }
-      }
+      localVue,
```
oliver-sanders: What does this do?

src/components/cylc/Graph.vue
kinow: Agree, and I hadn't really thought too much about it until today.

If users will be able to access each other's workflows, then I think we will have to use IDs everywhere? i.e. kinow/five instead of simply five.

Happy to fix this in this PR too. Just a few lines changes.

tests/unit/components/cylc/toolbar.vue.spec.js
kinow: It is from vue-test-utils, and allows us to create a local vue instance that does not pollute global vue instance.

This allows multiple tests to hack the attributes of Vue without breaking other tests (changes are scoped to the test, function, closure, etc, where you created it).

In this case, we are hacking the local vue to add vuetify, and tests of other components should not be affected by that (at least that's the theory).

src/components/cylc/Graph.vue
oliver-sanders: Users will be able to access other's workflows but only via the other user's UI Server so I think we can just use name.

tests/unit/components/cylc/toolbar.vue.spec.js
oliver-sanders: Thanks for the info.

src/components/cylc/Graph.vue
kinow: Ah! Of course, we already have a scope with `users/kinow`, no need to do `users/kinow/kinow|five`. Will fix the `Graph` component. Thanks!

hjoliver: I take it that is not `/users/<authenticated-user>/<workflow-owner>/five`?

hjoliver: (I recall from back in J-Hub testing days, some discussion about two user names in the URL).

kinow: >I take it that is not /users/<authenticated-user>/<workflow-owner>/five?

That's right, just one user in the URL. I think if we both are running the workflow `five`, then I would be able to access it via

- `http://<server>/users/kinow/#/workflows/five`
- `http://<server>/users/hjoliver/#/workflows/five`

Assuming you granted me permission to access your UI Server/workflow.

kinow:
Now all tests passing, and managed to get the toolbar working OK locally, with both Tree and Graph components.

- Tested with 2 workflows, switching between both, changing their statuses from UI & command line (i.e. `cylc release...`, click on the hold icon, etc)
- Tested with the offline mode workflow
- Tested resizing browser to responsive mode. In this mode, the toolbar is always present, with the icon to expand the drawer with the menu.
- Tested the Graph component (it should now work with the Toolbar)

PR #355, opened 2020-01-06
Use wss when page protocol is https:

### LINKED ISSUE #353: use secure websockets
Support secure websockets (wss) at the UI so we can connect using https+wss.

> the fix is probably related to graphql-ws and/or apolloclient
> -- @kinow 

**Pull requests welcome!**

kinow:
These changes close #353

**Requirements check-list**
- [x] Does not need tests (why? not easy to test this part, we would need to mock the browser API to return a different protocol, but I don't think that's necessary?).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -41,6 +41,9 @@ hold, release, and stop workflows.
 [#280](https://github.com/cylc/cylc-ui/pull/280) - Add WebSockets client
 for GraphQL subscriptions.
 
+[#355](https://github.com/cylc/cylc-ui/pull/355) - Use wss when page
+protocol is http:
+
 ### Fixes
 
 [#275](https://github.com/cylc/cylc-ui/pull/275) - Fix size of dashboard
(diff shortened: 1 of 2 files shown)


kinow:
This is part of the solution I think. We tell Apollo Client that we want a link with `wss`. I don't think we need to do anything else besides specifying the protocol - at least that's what I expect from their API.

kinow:
This does not solve the issue. I still have the error, but that's very likely due to the Tornado server with `graphql-ws` serving HTTP only.

Note that this is not the JupyterHub SSL link that I am referring to. This is the Cylc UI Server HTTP server, behind the `configurable-http-proxy`. Just need to figure how to make that one talk SSL/HTTPS, and confirm `graphql-ws` works with that too :thinking:

kinow:
Will update this one and mark as ready for review later once I get it all working with no issues in my environment.

kinow:
Looks like it works with no change to the backend (cylc-uiserver/proxy/hub/etc). Once the SSL configuration is prepared in the hub, everything should work with this change 👍

kinow:
image

image

Note: https, and no errors or warnings in the console. The firefox websockets tool shows the messages being submitted successfully between client and server.

Issue #339, opened 2019-12-16
api-on-the-fly


oliver-sanders:
The Web-UI component of https://github.com/cylc/cylc-flow/issues/2123

* [x] Introspect the GraphQL schema to obtain a list of mutations. - #357 
* [x] Auto-generate any required forms. - #357 
* [x] Output form data in the appropriate structure for use with a GraphQL mutation request. - #357 
* [x] Use the argument types to work out what mutations relate to (i.e. workflow[s], task[s], job[s]). - #359 
* [x] Create an interface to list mutations relevant to an entity (i.e. workflow, task, job). - #359 
* [x] Support custom Cylc data types (see https://github.com/cylc/cylc-flow/pull/3469) - #382 
* [x] Perform basic input validation. - #382 
* [x] Add functionality to request a mutation using form output. - #385 
* [x] Centralise this functionality as a core service - #504
* [x] Put it all together in a tool-tip. - #504 
* [ ] Auto-complete workflow, task, family, cycle points, etc

![api-on-the-fly](https://user-images.githubusercontent.com/16705946/70877793-dbb15e80-2023-11ea-95ac-1252f638cbf7.png)

**Pull requests welcome!**

hjoliver:
I see you chose "insert" as an example above.

Note that the concept of "insertion" - which is hard to explain the need for to users - will go go away with spawn-on-demand (I'll update the new proposal!).  And even without spawn-on-demand we were hoping to hide it from users (automatic insertion). https://github.com/cylc/cylc-flow/issues/2143

So this *might* (remains to be seen) be one mutation that we do not actually need in the new UI.

oliver-sanders:
I've got as far as I can with this one for the moment as the next steps involve modifying the Workflow Component which is currently in flux.

My suggestion is that the mutations functionality lies in the Workflow Service and Workflow Component (we need some logic in the workflow component e.g. to filter by the workflow(s) currently being displayed). For this #376 will need to be attempted first.

oliver-sanders:
Bumped the remaining issue (auto-completion) to a new issue as it is a nice-to-have - https://github.com/cylc/cylc-ui/issues/602

PR #357, opened 2020-01-09
api on the fly - part 1 - generate forms from mutations

oliver-sanders:
Addresses the first three points of #339 

This PR adds a temporary "mutations" view which lists available mutations and auto-generates forms for them on request.

This is orthogonal to https://github.com/cylc/cylc-flow/pull/3469 which makes the mutations more declarative and brings them more inline with the Cylc CLI.

Getting the nested components and data model working correctly took nearly all the time. Thanks for the `<component :is="class" />` suggestion @kinow, I would have wasted days otherwise :+1:.

**Overview:**

* Introspect the GraphQL schema for information about mutations.
* Auto-generate forms to act as interfaces to these mutations.
* When complete output JSON from these forms to be used as arguments in the GraphQL request.

**Caveats:**

* ~Some of the default values in the GraphQL schema are presently set as string representations for list types~. Either Graphene or GraphQL seems to convert all default types to JSON which is added as a string in the JSON response? This results in traceback and errors. :eyes: 
* Styling not yet complete.
* The abstraction form generator is roughly independent of the component library being used but there is't a proper interface for implementing different libraries as yet. Not a big job but probably a bridge to cross if and when we get to it.

**Justification:**

There are some tools out there which could have been used, instead I wrote this from scratch. Just to dump my reasoning down here for the record:

There is a potential pathway where we convert GraphQL -> JSONSchema -> Web Form -> JSON -> GraphQL Arguments.

* There are a lot of interfaces here and it gets a bit messy/
* Most of the tools are immature
* None of them seem to work with mutations properly (they are seem focused on the data model)
* Most don't support Vue or are written for other frameworks.
* None work with Vuetify.
* Supporting GraphQL mutation schema natively seems more natural.

**TODO:**

* [x] Vuetify doesn't have an integer field, it has a "number" mode to the text field. This outputs strings rather than number which won't do.
* [x] Testing (basic).
* [x] Consider using an inheritance or composition model to keep things DRY.
* [ ] Fix persistence of the `FormGenerator` model.

**Requirements check-list**

### DIFF
--- a/src/components/graphqlFormGenerator/FormGenerator.vue
+++ b/src/components/graphqlFormGenerator/FormGenerator.vue
@@ -2,64 +2,34 @@
   <v-form>
     <h3>{{ mutation.name }}</h3>
     <p>{{ mutation.description }}</p>
-    <component
+    <form-input
      v-for="arg in getArguments(mutation)"
      v-bind:key="arg.label"
      v-model="model[arg.label]"
-     :is="arg.component"
+     :gqlType="arg.gqlType"
      :label="arg.label"
-    >
-      <component
-       v-if="arg.ofType"
-       :is="derive(arg.ofType).component"
-      >
-        <component
-         v-if="arg.ofType.ofType"
-         :is="derive(arg.ofType.ofType).component"
-        >
-        </component>
-      </component>
-    </component>
+    />
     <v-btn
       @click="meh"
     >
       Done
     </v-btn>
-    <pre>{{ model }}</pre>
+    <pre ref="output">{{ model }}</pre>
   </v-form>
 </template>
 
 <script>
-import Vue from 'vue'
-
-import { VTextField } from 'vuetify/lib/components/VTextField'
 import { VForm } from 'vuetify/lib/components/VForm'
-import GList from '@/components/graphqlFormGenerator/components/List'
-import GNonNull from '@/components/graphqlFormGenerator/components/NonNull'
+import FormInput from '@/components/graphqlFormGenerator/components/FormInput'
 
-const FORM_COMPONENTS = {
-  Form: {
-    component: VForm,
-  }
-}
-
-const TYPE_COMPONENTS = {
-  String: {
-    component: VTextField,
-  }
-}
-
-const KIND_COMPONENTS = {
-    LIST: {
-      component: GList
-    },
-    NON_NULL: {
-      component: GNonNull
-    }
-}
+// import Vue from 'vue'
+// Vue.component('form-input', FormInput)
 
 export default {
+  name: 'form-generator',
+
   components: {
+    'form-input': FormInput
   },
 
   props: {
@@ -73,68 +43,15 @@ export default {
     model: {}
   }),
 
-  created () {
-    // TEST DATA
-    this.mutation = {
-      name: 'My Mutation',
-      description: 'Test example.',
-      args: [
-        {
-          name: 'MyString',
-          type: {
-            name: 'String',
-            kind: 'SCALAR'
-          }
-        },
-        {
-          name: 'MyInteger',
-          type: {
-            name: 'Int',
-            kind:' SCALAR'
-          }
-        },
-        {
-          name: 'MyNonNull',
-          defaultValue: 'cant null this',
-          type: {
-            name: null,
-            kind: 'NON_NULL',
-            ofType: {
-              name: 'String',
-              kind: 'SCALAR'
-            }
-          }
-        }
-      ]
-    }
-  },
-
   methods: {
-    derive (type) {
-      const name = type.name
-      const kind = type.kind
-      if (TYPE_COMPONENTS[name]) {
-        return TYPE_COMPONENTS[name]
-      }
-      if (KIND_COMPONENTS[kind]) {
-        return KIND_COMPONENTS[kind]
-      }
-      return TYPE_COMPONENTS['String']
-    },
-
-    processType (type) {
-      return {
-        component: this.derive(type).component,
-        ofType: type.ofType
-      }
-    },
-
     getArguments (mutations) {
+      // TODO convert to computed?
       const ret = []
       for (let argument of mutations.args) {
-        const arg = this.processType(argument.type)
-        arg.label = argument.name
-        ret.push(arg)
+        ret.push({
+          gqlType: argument.type,
+          label: argument.name
+        })
         // spin up the model
         this.model[argument.name] = (
           argument.defaultValue
@@ -145,7 +62,9 @@ export default {
       return ret
     },
 
-    meh () {console.log(this.model)}
+    meh () {
+      this.$refs.output.innerText = JSON.stringify(this.model, null, '  ')
+    }
   }
 }
 </script>
--- a/src/components/graphqlFormGenerator/FormGenerator.vue
+++ b/src/components/graphqlFormGenerator/FormGenerator.vue
@@ -19,8 +19,7 @@
 </template>
 
 <script>
-import { VForm } from 'vuetify/lib/components/VForm'
-import FormInput from '@/components/graphqlFormGenerator/components/FormInput'
+import FormInput from '@/components/graphqlFormGenerator/FormInput'
 
 export default {
   name: 'form-generator',
@@ -108,7 +107,7 @@ export default {
           ret = []
           break
         }
-        if (pointer.kind == 'OBJECT') {
+        if (pointer.kind === 'OBJECT') {
           ret = {}
           break
         }
--- a/src/components/graphqlFormGenerator/FormGenerator.vue
+++ b/src/components/graphqlFormGenerator/FormGenerator.vue
@@ -1,5 +1,5 @@
 <template>
-  <v-form>
+  <v-form validate>
     <h3>{{ mutation.name }}</h3>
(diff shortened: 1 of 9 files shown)


oliver-sanders:
Screenshot from 2020-01-10 10-59-08

src/components/graphqlFormGenerator/FormInput.vue
```diff
@@ -0,0 +1,167 @@
+<template>
+  <!-- NOTE: the is field domes from `props` -->
```
kinow: s/domes/comes

src/components/graphqlFormGenerator/FormInput.vue
```diff
@@ -0,0 +1,167 @@
...
+     * and be managed by Vue correctly.
+     */
+    model: {
+      get () {
+        return this.value
+      },
+      set (val) {
+        this.$emit('input', val)
+      }
+    },
```
kinow: Oh, I bet it was not really fun to debug why the `v-model` was not working here ( I thought it would work, or `:value.sync`, which I tried but also didn't work :eyes: )

src/views/Mutations.vue
```diff
@@ -0,0 +1,152 @@
...
+    this.getSchema()
+  },
+
+  methods: {
+    getIntrospectionQuery () {
+      // we are only interested in mutations so can make our life
+      // a little easier by restricting the scope of the default
+      // introspection query
+      var fullIntrospection = gql(introspectionQuery)
+      var mutationQuery = gql(`
```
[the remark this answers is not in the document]
oliver-sanders: Yes, you're right.

I had originally implemented this by building the `gql` object then modifying it, but that had some kind of strange scoping issue.

src/components/graphqlFormGenerator/FormGenerator.vue
```diff
@@ -0,0 +1,120 @@
...
+          defaultValue = this.getNullValue(arg.type)
+        }
+        model[arg.name] = defaultValue
+      }
+
+      // done
+      this.model = model
+    },
+
+    deepcopy (obj) {
```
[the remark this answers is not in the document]
oliver-sanders: Good good, I don't like the stringify approach.

src/components/graphqlFormGenerator/FormInput.vue
```diff
@@ -0,0 +1,167 @@
...
+      // const ofType = this.gqlType.ofType
+
+      var componentProps
+      if (NAMED_TYPES[name]) {
+        componentProps = NAMED_TYPES[name]
+      } else if (KINDS[kind]) {
+        componentProps = KINDS[kind]
+      } else {
+        componentProps = NAMED_TYPES['String']
+        console.log(
```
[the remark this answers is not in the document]
oliver-sanders: Should probably be `console.warning` too...

src/components/graphqlFormGenerator/FormInput.vue
oliver-sanders: Yeah, the data model took much longer than all the rest of the work put together. Really confused the hell out of me!

src/components/graphqlFormGenerator/FormGenerator.vue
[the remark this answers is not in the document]
oliver-sanders: I'm still finding my way around contemporary JS so hints and tips like this are really helpful BTW.

oliver-sanders:
> I think you have experienced pretty much every kind of problem I had with Vue.js ￼

Yes, this was a pretty steep learning curve. I think I understand Vue a lot better now though.

oliver-sanders:
@kinow sneaking in one more commit to use mixins which make the code much **D**[n]**RY**er.

oliver-sanders:
@wxtim I've put you down for second review. This PR should work with the master branches of Cylc Flow and Cylc UIServer.

For second review I reckon:

* Functional test.
* Skim of code.
* Check requirements/todos.

To test:

* Cylc Flow: no suite needed.
* Cylc UIS: launch as usual.
* Cylc UI: build as usual.
* Browser: Navigate to `#/mutations`.
* Note some errors are to be expected (see caveats in the PR description).

The code is a little funky so some pointers here:

* `<component is="Foo" attr="value">` is effectively another way of writing `<Foo attr="value">`
* Mixins are a way of sharing configuration between components via composition rather than inheritance (similar to Cylc families) https://vuejs.org/v2/guide/mixins.html
* Components with names starting with a `V` come from Vuetify.

wxtim:
> > I think you have experienced pretty much every kind of problem I had with Vue.js ￼
> 
> Yes, this was a pretty steep learning curve. I think I understand Vue a lot better now though.

You can explain it to me now.

src/components/graphqlFormGenerator/FormGenerator.vue
```diff
@@ -0,0 +1,113 @@
...
+    <pre ref="output">{{ model }}</pre>
+  </v-form>
+</template>
+
+<script>
+import FormInput from '@/components/graphqlFormGenerator/FormInput'
+import cloneDeep from 'lodash/cloneDeep'
+
+export default {
+  name: 'form-generator',
```
wxtim: You've created a new object here and exported it - Can we have an over-arching docstring, or is that "just not very vue-js"?

src/components/graphqlFormGenerator/FormGenerator.vue
```diff
@@ -0,0 +1,113 @@
+<template>
+  <v-form validate>
+    <h3>{{ mutation.name }}</h3>
+    <p>{{ mutation.description }}</p>
+    <form-input
+     v-for="input in inputs"
```
wxtim: Testing my understanding - `inputs` is the list of buttons?

src/components/graphqlFormGenerator/FormInput.vue
```diff
@@ -0,0 +1,142 @@
...
+      default: () => {}
+    }
+  },
+
+  data: () => ({
+    defaultProps: DEFAULT_PROPS
+  }),
+
+  computed: {
+    /* The props to pass to the form input.
```
wxtim: props = properties? Is this an ES5/6 standard jargon, or a project jargon? If it's project can you put the full word in the docstring?

wxtim:
Am I supposed to _do_ anything with the object created by using this?

Can I do something with this, or should something happen? Should I paste it into the console?
```javascript
{
  "command": "\"nudge\"",
  "workflows": [
    "simple"
  ]
}
```

wxtim:
I have:
- [x] Read the code and spot any really obvious issues - at this rate my ability to read JS might overtake my abilities to read Fortran!
- [x] Tried running the UI - looks ok - I was able to nav to #mutations.

As far as I can tell this pull request does what @oliver-sanders intends it to do, and doesn't do other stuff. I've approved it, but I'd quite like answers to my ignorant questions. I'll merge if @oliver-sanders  doesn't.

src/components/graphqlFormGenerator/FormGenerator.vue
oliver-sanders: `export default` is a JavaScript pattern, export something in one file, import it in another.

The MDN pages are pretty good reference https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export

src/components/graphqlFormGenerator/FormGenerator.vue
oliver-sanders: Yeah, by input I mean a generic user input like a text box, radio buttons or whatnot.

These "inputs" are the arguments to the GraphQL mutation.

src/components/graphqlFormGenerator/FormInput.vue
oliver-sanders: "Props" is special Vue.js jargon. Props are like arguments that you pass to a component e.g:

```html
<component prop1="foo" prop2="bar" />
```

They aren't quite like "properties" in Python, the component shouldn't change them. Vue "computed" values are more like Python `@properties`.

In a way props kinda belong to their parent component:

```vue
<template>
    <component :foo="foo" :bar="bar" />
</template>

<script>
export default {
  data: () => ({
    foo: 'Foot',
    bar: 'Bart'
  }).
  methods: {
    whatever () {
      // this component is free to make changes to foo and bar because these values are set
      // in `data` whereas in `<component />` they will be set as `props`.
      foo = 'Fooooot'
    }
  }
}
</script>
```

oliver-sanders:
> Am I supposed to do anything with the object created by using this?

The JSON you see at the bottom of the form is just showing the data in the form. It should change as you type. The `mutations` page is a temporary view whilst we get all the form generation stuff sorted, the JSON is for development purposes.

Once we are finished that JSON will get attached to a GraphQL mutation and sent to the Workflow Server (via the UI Server).

oliver-sanders:
> I'd quite like answers to my ignorant questions

Keep asking, reviews are a great way of learning and even prompting authors to think more about what they've done.

src/components/graphqlFormGenerator/FormGenerator.vue
wxtim: I get that (I've been doing the codeacademy intro to JS and a course on vue) - I just wanted a docstring.

wxtim:
I'm happy that you've answered all my questions so I'm merging! Happy days.

PR #359, opened 2020-01-10
api-on-the-fly - part 2 - mutation association

oliver-sanders:
Addresses the fourth point of #339 

~**Note - Built on top of #357**~ now merged
**Note - Works with the data types defined in https://github.com/cylc/cylc-flow/pull/3469**

Use the argument types defined in the GraphQL schema to associate mutations with "Cylc Objects" by which I mean workflows, cycle point, tasks, etc.

This mapping will be used when you right-click on something (say a task in the tree view) to work out which mutations are relevant to the thing you clicked on.

At the moment this is built into the temporary Mutations view.

**Example**

Screenshot from 2020-01-10 14-41-32

**TODO**

* [ ] Test

**Requirements check-list**
- [x] No change long necessary, functionality not exposed to users (yet)

### DIFF
--- a/src/components/graphqlFormGenerator/FormInput.vue
+++ b/src/components/graphqlFormGenerator/FormInput.vue
@@ -109,7 +109,7 @@ export default {
       const kind = this.gqlType.kind
       // const ofType = this.gqlType.ofType
 
-      var componentProps
+      let componentProps
       if (NAMED_TYPES[name]) {
         componentProps = NAMED_TYPES[name]
       } else if (KINDS[kind]) {
--- a/src/views/Mutations.vue
+++ b/src/views/Mutations.vue
@@ -4,8 +4,8 @@
       Loading...
     </div>
     <div v-if="loaded">
+      <h1>Form Generator</h1>
       <v-select
-        v-if="mutations"
         v-model="selectedMutation"
         :items="mutationNames"
         label="Mutation"
@@ -19,6 +19,26 @@
         <!--FormGenerator :mutation='sampleMutation' /-->
         <FormGenerator :mutation='getMutation(selectedMutation)' />
       </v-card>
+      <h1>Associator</h1>
+      <v-select
+        v-model="selectedAssociation"
+        :items="Object.keys(cylcObjects)"
+        label="Cylc Object"
+      />
+      <ul
+        v-if="selectedAssociation"
+      >
+        <li
+          v-for="association in filterAssociations(selectedAssociation)"
+          v-bind:key="association.name"
+        >
+          <span><b>{{ association.name }}</b></span>
+          <span
+            v-if="association.multiple"
+          > (list)</span>
+          <span><i> - {{ association.argument }}</i></span>
+        </li>
+      </ul>
     </div>
   </div>
 </template>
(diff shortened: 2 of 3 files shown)


src/components/graphqlFormGenerator/FormInput.vue
```diff
@@ -0,0 +1,168 @@
...
+   :label="label"
+   :gqlType="gqlType"
+  />
+</template>
+
+<script>
+import Vue from 'vue'
+
+import { VTextField } from 'vuetify/lib/components/VTextField'
+import GList from '@/components/graphqlFormGenerator/components/List'
```
kinow: Probably for later. It's working now, but this brings a risk of running into some errors with webpack due to circular dependencies. This module (`FormInput`) has a dependency on the default export of `List`. And `List` has a dependency on `FormInput`.

There are other circular dependencies in this package, which could indicate that we need to move things into a separate module.

https://railsware.com/blog/how-to-analyze-circular-dependencies-in-es6/

(BTW, I might have introduced some circular dependencies in other components too, so that's something we can review later and fix when/if possible I think :grimacing: )

kinow:
(Or are associations going to be restricted later, and only for now I can associate anything with any mutation?)

oliver-sanders:
> Could you explain what are the associations

The "associations" work is simply about introspecting the arguments of mutations to find out what Cylc things the mutation operates on (i.e. workflows, cycle points, tasks, jobs).

The result is a mapping (e.g. the broadcast mutation works on multiple workflows and [optionally] multiple cycle points).

This will be used in a future PR to achieve this:

1. User right-clicks (or taps or whatever) on a cycle point node in the Tree View.
2. The UI works out which mutations take cycle point(s) as arguments (e.g. broadcast, stop, insert).
3. The UI lists these mutations in a context menu.

oliver-sanders:
> Debugging, it shows that the this.value value is a String I think?

So this is the same traceback is the same issue mentioned in #357, I had thought this was caused by error in the GraphQL schema (from Cylc Flow) but it turns out it isn't.

GraphQL introspections provide "default values" in JSON format so lists come through as strings (e.g. `"[\"*\"]"`). The solution is to JSON.parse them which I've done in another branch.

This PR just adds associations and doesn't touch the form generation logic (which is where the error is coming from).

src/components/graphqlFormGenerator/FormInput.vue
```diff
@@ -0,0 +1,168 @@
...
+     * TODO: move to rule based system to allow changing
+     *       of parent components based on child types
+     */
+    props () {
+      // get the default props for this graphQL type
+      const name = this.gqlType.name
+      const kind = this.gqlType.kind
+      // const ofType = this.gqlType.ofType
+
+      var componentProps
```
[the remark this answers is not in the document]
oliver-sanders: Dammit, it's going to take me a while to buck the `var` habit :man_facepalming:!

src/views/Mutations.vue
```diff
@@ -0,0 +1,268 @@
...
+        print(mutationQuery.definitions[0]) +
+        // the fragments which power it
+        print(fullIntrospection.definitions[1]) +
+        print(fullIntrospection.definitions[2]) +
+        print(fullIntrospection.definitions[3])
+      )
+    },
+
+    getSchema () {
+      this.$workflowService.apolloClient.query({
```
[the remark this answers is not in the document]
oliver-sanders: Good thinking.

The mutation view is a quick proof of concept, this functionality is going to have to move somewhere more central, either into the workflow service or some new mutation service. Will move the query functionality across then.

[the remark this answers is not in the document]
kinow: Roger. And that isn't breaking the offline mode exactly. It only doesn't work in the offline mode if I navigate to `/#/mutations`. Everything else working fine, so not urgent.

src/components/graphqlFormGenerator/FormInput.vue
oliver-sanders: Hmmm, tricky.

FormInput and List are recursive components so one requires the ability to create the other.

Not sure how to get rid of that circular dependency.

kinow: It works for now, so we don't need to worry for 0.2. We can push to the 1.0 backlog.

oliver-sanders:
@wxtim I've put you down for review 2, this is a much quicker review than the last one.

src/components/graphqlFormGenerator/FormInput.vue
oliver-sanders: I've stuck up an issue tagged against `1.0`.

hjoliver:
Change log and tests?

oliver-sanders:
Added unit tests, change log not helpful as the mutations view (as is in this PR), will not make release.

hjoliver:
> change log not helpful as the mutations view (as is in this PR), will not make release.

It will make the next (alpha-2) release.  By 8.0.0 we should have another change long entry that removes it.  (At least, that's how we are managing pre-8.0.0 changes, in my mind!)

hjoliver:
You seem to have a bunch of minor quoting issues (via the linter) in the new tests.

oliver-sanders:
> It will make the next (alpha-2) release

I'll put a changelog in on the most contemporary aotf PR for a2.

oliver-sanders:
> You seem to have a bunch of minor quoting issues (via the linter) in the new tests.

Doesn't fail in my env, different linter versions perhaps, should be fixed.

PR #360, opened 2020-01-13
drop IE support?

oliver-sanders:
Quick question, should we be setting the `X-UA-Compatible` flag?

Probably no harm done, Vuetify apparently [supports IE11 via pollyfill](https://vuetifyjs.com/en/getting-started/browser-support).

**Requirements check-list**

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -2,7 +2,6 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link href='https://fonts.googleapis.com/css?family=Roboto:300,400,500,700,300italic,400italic' rel="stylesheet" type="text/css">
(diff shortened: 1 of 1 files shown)


kinow:
>Quick question, should we be setting the X-UA-Compatible flag?

Not sure. We got that from the theme that we initially used: https://github.com/creativetimofficial/vuetify-material-dashboard/blob/2a3ea658e72e7109b7aa65be73abb6dfc547e3ea/public/index.html#L23

From what I understood searching, it appears to be a header for Internet Explorer and its compatibility setting. Doesn't look really useful to me. Just tested and it worked the same for me on IE Edge with/without this change. So +1 to removing it.

oliver-sanders:
My understanding is that `X-UA-Compatible` tells Internet Explorer what compatibility spec to use for the page.

Setting `X-UA-Compatible="edge"` is a signal to IE11 to use latest spec.

So it shouldn't effect edge but might effect IE11.

kinow:
>So it shouldn't effect edge but might effect IE11.

I thought that too. But testing locally, both IE edge and 11 had the same behaviour.

11 not showing anything, and no error message before after this change.

PR #366, opened 2020-01-15
Add a tree component by default when accessing the workflow page

kinow:

**Requirements check-list**
- [x] Does not need tests (why? let's with some e2e later for this view).
- [x] No change log entry required (why? e.g. the workflow component is new; so there is no change for the user, except for the workflow view, already in the changelog).

### DIFF
--- a/src/views/Workflow.vue
+++ b/src/views/Workflow.vue
@@ -191,6 +191,13 @@ export default {
       this.$workflowService.unsubscribe(subscriptionId)
     })
   },
+  mounted () {
+    // Create a Tree View for the current workflow by default
+    const subscriptionId = this.subscribe('tree')
+    this.$nextTick(() => {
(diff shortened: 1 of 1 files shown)


src/views/Workflow.vue
```diff
@@ -191,6 +191,13 @@ export default {
       this.$workflowService.unsubscribe(subscriptionId)
     })
   },
+  mounted () {
+    // Create a Tree View for the current workflow by default
+    const subscriptionId = this.subscribe('tree')
+    this.$nextTick(() => {
```
kinow: If added to `created`, it would fail as the component was created, but nothing was added to the DOM, not even the Lumino panel holder.

On `mounted` we will have the component mounted on the DOM, but no guarantee that the Lumino objects are there too. So we wait one tick, then we can add to the widget :+1:

PR #371, opened 2020-01-16
Tweak workflows component sizes.

hjoliver:
Address comments (somewhere) on unused space at the bottom of the workflows component.  It was still sized as if the old footer existed.

Also shrunk the minimum tab size a bit.

shot

### DIFF
--- a/src/styles/cylc/_workflow.scss
+++ b/src/styles/cylc/_workflow.scss
@@ -4,10 +4,10 @@
 #workflow-panel {
   #main {
     display: flex;
-    min-height: calc(100vh - 82px - 56px);
+    min-height: calc(100vh - 48px);
     .content {
-      min-width: 500px;
-      min-height: 400px;
+      min-width: 300px;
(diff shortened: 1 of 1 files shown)


src/styles/cylc/_workflow.scss
```diff
@@ -4,10 +4,10 @@
 #workflow-panel {
   #main {
     display: flex;
-    min-height: calc(100vh - 82px - 56px);
```
hjoliver: 82px was the size of the footer.

56px seemed to be a bit wrong - not sure where that came from.

kinow: >56px seemed to be a bit wrong - not sure where that came from.

:me hiding:

hjoliver: hehe, extremely minor issue :+1:

kinow:
Had a look at the code, change looks sensible, screen shot looks OK too. :+1:

PR #379, opened 2020-01-22
Add ConnectionStatus component

### LINKED ISSUE #93: Inform the user whether the backend is offline or not
Apache Jena, Riot, and other applications are able to tell you when there is a connection issue with the backend.

We should be able to do the same. Progressive web apps frameworks offer ways to do that, but maybe we don't need to use anything but Vuejs.

Examples:

- https://github.com/filrak/vue-offline
- https://pusher.com/tutorials/pwa-vuejs
- https://alligator.io/vuejs/vue-detect-online-offline/
- https://github.hubspot.com/offline/docs/welcome/

kinow:
These changes close #93 

Found one way to implement it while working on the websockets issue, so using the momentum to write docs and implement small changes such as this one.

**Requirements check-list**

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -16,13 +16,23 @@ import router from '@/router'
 import store from '@/store'
 
 // GraphQL client
+/**
+ * @type SubscriptionWorkflowService
+ */
 import SubscriptionWorkflowService from 'workflow-service'
+import { createGraphQLUrls, createSubscriptionClient } from '@/utils/graphql'
 
 // Sync store with router
 sync(store, router)
 
 // WorkflowService singleton available application-wide
-const workflowService = new SubscriptionWorkflowService()
+// On the offline mode, we do not have a WebSocket link, so we must create a null SubscriptionClient to use an empty link
+const graphQLUrls = createGraphQLUrls()
+let subscriptionClient = null
+if (process.env.NODE_ENV !== 'offline') {
+  subscriptionClient = createSubscriptionClient(graphQLUrls.wsUrl)
+}
+const workflowService = new SubscriptionWorkflowService(graphQLUrls.httpUrl, subscriptionClient)
 Vue.prototype.$workflowService = workflowService
 
 Vue.config.productionTip = false
--- a/src/utils/graphql.js
+++ b/src/utils/graphql.js
@@ -4,6 +4,69 @@ import { HttpLink } from 'apollo-link-http'
 import { WebSocketLink } from 'apollo-link-ws'
 import { getMainDefinition } from 'apollo-utilities'
 import { InMemoryCache } from 'apollo-cache-inmemory'
+import { SubscriptionClient } from 'subscriptions-transport-ws'
+import store from '@/store'
+
+/**
+ * Create the HTTP and WebSocket URLs for an ApolloClient.
+ *
+ * @private
+ * @return {{wsUrl: string, httpUrl: string}}
+ * @private
+ */
+export function createGraphQLUrls () {
+  // TODO: revisit this and evaluate other ways to build the GraphQL URL - not safe to rely on window.location (?)
+  const baseUrl = `${window.location.hostname}${window.location.port ? ':' + window.location.port : ''}${window.location.pathname}`
+  const httpUrl = `${window.location.protocol}//${baseUrl}graphql`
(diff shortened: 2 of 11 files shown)


kinow:
Based on the component with the same name, from [Vue UI](https://github.com/vuejs/vue-cli/blob/8fcea225b7b4406ff7c03e34d4e962b45d78ef9f/packages/%40vue/cli-ui/src/components/app/ConnectionStatus.vue).

Instead of running a GraphQL query, we simply check if the WebSocket connection has been established or not. And instead of `div`s with some styles, we use Vuetify's [snackbars](https://vuetifyjs.com/en/components/snackbars).

kinow:
Need to write some tests, but here's the component in action.

2020-01-22 13-26-18 flv

Note that the UI Server is not automatically started by `jupyterhub`. But once it's up, the Apollo Client should reconnect. Though I noticed that sometimes it appears that the browser would have trouble reconnecting. But nothing wrong in ApolloClient that I could tell. Looks more like it tried so many times that even after focusing on the browser with the mouse, it still ignores.

Opening a new browser normally solves it. If it is confirmed as an issue, this can be tracked separately - still better to have an indicator that the connection is broken I guess?

kinow:
Also note that Vue has a lifecycle for pages/components that include things like `beforeDestroy` that are invoked when you refresh a page (never found it in the docs, but quite sure it's invoked).

So if you reload the page, for 1 second you should see the offline notification, then it disappears when page is loaded again.

hjoliver:
Nice :+1: 

> Opening a new browser normally solves it. If it is confirmed as an issue, this can be track ed separately - still better to have an indicator that the connection is broken I guess?

IMO, yes.

src/main.js
```diff
@@ -16,13 +16,23 @@ import router from '@/router'
...
 sync(store, router)
 
 // WorkflowService singleton available application-wide
-const workflowService = new SubscriptionWorkflowService()
+// On the offline mode, we do not have a WebSocket link, so we must create a null SubscriptionClient to use an empty link
+const graphQLUrls = createGraphQLUrls()
+let subscriptionClient = null
+if (process.env.NODE_ENV !== 'offline') {
+  subscriptionClient = createSubscriptionClient(graphQLUrls.wsUrl)
+}
```
kinow: I tried to avoid having any import from `utils/graphql` here, but unfortunately I found no other way to prevent the ApolloClient of trying to create the WebSocket connection - which is not available in offline mode.

src/utils/graphql.js
```diff
@@ -4,6 +4,69 @@ import { HttpLink } from 'apollo-link-http'
...
+ *   reconnect: boolean,
+ *   lazy: boolean
+ * }} options - SubscriptionClient options (only two main options added here, see their doc for more)
+ * @param {*} wsImpl
+ * @return {SubscriptionClient} a subscription client
+ */
+export function createSubscriptionClient (wsUrl, options = {}, wsImpl = null) {
+  const opts = Object.assign({
+    reconnect: true,
+    lazy: false
```
kinow: We need these two default values. But callers (especially unit tests) are free to change. For unit tests, you want to have `lazy: true`, so that the `SubscriptionClient` doesn't try to create the connection, as node/mocha don't have a WebSocket implementation, resulting in runtime error.

kinow:
Ready for review. I've added a changelog entry under the 0.2 release, but that doesn't mean that's the version it will be released.

So no hurry reviewing it. If not 0.2, we can review it for 0.3, and I just amend the commit to move the entry under 0.3, or 0.4, etc :+1: 

To review, take a look at the code & tests. Most of the changes in the `.js` code (not in the `.vue`) are for unit tests. The easiest path, the first which I implemented, was have everything under `createApolloClient()`.

That method was getting bloated, and getting too "smart". It was creating the `ApolloClient`, creating the GraphQL URL's, and now creating a `SubscriptionClient`. The last would result in an error as WebSockets are not available with node/mocha, without a browser or some polyfill.

Mocking ES6 import dependencies is not possible with Sinon. There are other modules like proxyquire that do that, but I thought it was better to change the objects a bit, instead of adding another dependency to the project.

Once running the app, stop the Hub. You should now see a red warning at the top. If you restart the hub, and maybe in another window or even another browser launch the UI server, then that warning should disappear, and both window/browsers should be showing the UI with no issues.

oliver-sanders:
Nice.

Works as expected.

Issue #381, opened 2020-01-22
Firefox 60.8 compatibility


hjoliver:
Placeholder to record any browser compat issues that we discover during Cylc 8 development.  Eventually to be documented.

hjoliver:
Firefox 60.8 via yum package manage on Centos 7:
- POC cytoscape graph view in Lumino tabs behaves very badly (but seems to work fine sans Lumino)

kinow:
IE (not edge) on win10pro:

- blank page, browser console errors

kinow:
Fixed the Ffox ESR68 with a change in the Graph component.

![image](https://user-images.githubusercontent.com/304786/72953467-14e6a680-3dfa-11ea-8801-a98d747a8df5.png)

hjoliver:
I think my browser version was 60.8 not 68 - was that a typo @kinow ?   Either way I'll test your PR on my laptop tonight.

kinow:
It was 68, but when I installed it on my Linux box (you can just unzip on Linux, handy) and started `./firefox` from that folder and confirmed the version, it was broken just like your screenshot. With the graph area all blank, except for controls. And the navigator area was displaying the miniature graph.

hjoliver:
OK, I guess it was not just one version then (and, not just the Centos-packaged build of it).

kinow:
> OK, I guess it was not just one version then (and, not just the Centos-packaged build of it).

Yes, hopefully. Tomorrow I will test again with 60.8. Even though that's an older version, as it's a release from July 2019 (I think you mentioned it?) I think everything in Cylc UI should work with that version (especially if it works with IE!)

kinow:
The issue that I have now, is that everything works once you resize the browser on Firefox 60.8. Before digging into the code, I searched with not much hope of finding anything useful... however:

https://github.com/cytoscape/cytoscape.js/issues/1748
https://stackoverflow.com/questions/40307570/a-hidden-cytoscape-graph-cannot-be-displayed-later
https://js.cytoscape.org/#cy.resize

Looks like this issue happened to others :grimacing: From the last link:

>If your code resizes the graph’s dimensions or position (i.e. by changing the style of the HTML DOM element that holds the graph, or by changing the DOM element’s position in the DOM tree), you will want to call cy.resize() to have the graph resize and redraw itself.

So it looks like it might be a good idea to call `resize()` when the graph objects (nodes/edges) are created, and also when the widget is added.

kinow:
[...]

Issue #74, opened 2019-05-23
Graph View


kinow:
The GraphQL data model is stable now, so we can create the initial views for the Vue.js app. ~~The first one will be the Graph View, listing all the tasks found in the task pool.~~

It is important to list all the tasks, without filtering at this point.

It is also important to simulate the change in the tasks, as we need to confirm that the view can be used with a dynamic graph.

While other web interfaces may be used as reference, there are things in the Cylc 7 GUI that can be used for the initial implementation, like the task colors.

- [ ] write mocked data
- [ ] add a JS library to plot a graph
- [ ] respond to changes in the mocked data, updating the graph

kinow:
Python has a project called `vcrpy`: https://github.com/kevin1024/vcrpy, which I believe was based on the Ruby `vcr`: https://github.com/vcr/vcr.

And like the VCR, you can use it to record HTTP request/responses. It is used to record interactions client-server, and replay them later (like a VCR). Which is useful when writing integration tests for web applications.

Looks like there are JS ports too:

- https://www.npmjs.com/package/fetch-vcr
- https://github.com/linkedin/sepia

So we should be able to start a suite with some tasks and interaction (e.g. the old `complex` found in the tutorial or examples of Cylc 7 I think). That suite also takes a while to complete.

And then change the existing `SuiteService` to return the current workflow state. And write a `setTimeout` call, to advance the state every 5 or 10 seconds.

oliver-sanders:
> he first one will be the Graph View

I would strongly advise implementing the Tree View first. It is the simplest view and for most users [here at least] it is "the Cylc GUI".

kinow:
Cylc 7 Graph view: https://cylc.github.io/doc/built-sphinx-single/_images/gcylc-graph-and-dot-views.png

To the right of the screen shot, is the Led view.

oliver-sanders:
*tldr;* From a quick look I think dagre.js with a custom presentation layer is probably the way to go but some further research would be a good idea. Performance is likely to be poorer than the Cylc7 graphviz approach.

The graph view is a combination of:

* A layout engine
   * This is the hard part, proper nasty algorithms, we should use an existing library.
   * The prime candidate is dagre.js
* A presentation layer
   * I've called this the "graph engine" elsewhere as some of them are large "graph theory" packages with algorithms for graph traversal etc (which we don't actually need).
[...]

PR #383, opened 2020-01-23
Tidy up the Graph component (part #2)

kinow:
These changes partially address #74 and #381, see also #302 

Initial view of the Graph component when using this branch:

image

After dismissing the warning, and expanding the layout options:

image

Now left with a frozen graph, and dagre engine. While right has dismissed the warning, hierarchical layout, layout options hidden, and it is showing live data.

image

Tests above were done with Firefox 72.0.1 (64 bit) on Ubuntu LTS, using my home laptop.

@hjoliver it kinda works with 60.8. Will take another look tomorrow, but I suspect it's the Lumino widget and how it is handling updates and resizing.

For me, the `/#/graph/families2` view worked fine.

image

But got the same issue reported by @hjoliver when using Lumino.

image

**Requirements check-list**
- [x] Does not need tests (why? it's simply deleting or moving code).


kinow:
Notes to self:

- I think multiple Graph components with Lumino is broken due to the ID of the DIV container being used... maybe refs will fix it?
- Need to fix the CSS styles that were "fixed" but broke the layout

hjoliver:
> Need to fix the CSS styles that were "fixed" but broke the layout

What do you mean by "the layout" here - the graph layout, or more general page layout of the UI?

hjoliver:
Technically the "graph layout" is the computed relative positions of all the graph nodes - the shape of the graph, sort of ... so maybe you mean the graph styling (and/or the viewport or whatever the term for that is)

kinow:
Oh, yup. Graph styling, sorry. I am just trying to remember where each style is used, and what CSS styles need changing. Should be done by today/tomorrow.

src/components/cylc/graph/Graph.vue
```diff
@@ -0,0 +1,631 @@
+<template>
+  <div class="c-graph">
+    <SyncLoader
+      :loading='loading'
+      :color='vueSpinner.color'
+      :size='vueSpinner.size'
+      class='spinner'
+    ></SyncLoader>
+    <div class='switch-layout ma-0'>
+      <v-expansion-panels
```
kinow: Using an expansion panel, I think we have more space to display the graph. It's hidden by default, user can click on the expand button to visualize the settings now.

src/components/cylc/graph/Graph.vue
```diff
@@ -0,0 +1,631 @@
...
+      tippy: null,
+      /**
+       * Graph data. These values are empty by default, then are populated later via watchers.
+       */
+      graphData: {
+        nodes: [],
+        edges: []
+      },
+      cytoscapeInstance: null
+    }
```
kinow: Removed data that was not used, or javascript variables declared outside the data (i.e. not reactive, accessible from outside the component), and grouped them with some simple comments for future developers.

src/components/cylc/graph/Graph.vue
```diff
@@ -0,0 +1,631 @@
...
+    this.updateGraph = debounce(this.updateGraph_, 100)
+    this.workflowUpdated = debounce(this.workflowUpdated_, 100)
+  },
+
+  mounted () {
+    // create a div for cytoscape, this is an approach similar to vue-cytoscape
+    // but the vue-cytoscape component offers more than what we need, so we imitate
+    // part of its behaviour here, and ignore the rest and handle the graph and cytoscape
+    // settings ourselves.
+    const element = document.createElement('div')
```
kinow: This is the approach of `vue-cytoscape`. We didn't need the whole library, so removed it from `package.json`, and added a few lines here to set up the `<div>` for Cytoscape.

src/components/cylc/graph/Graph.vue
```diff
@@ -0,0 +1,631 @@
...
+    element.setAttribute('class', 'cytoscape')
+    this.$el.appendChild(element)
+    this.cytoscapeInstance = Object.freeze(cytoscape(({
```
kinow: Spent half day on this one.

The `data` elements in a Vue app are simple data. No promises of methods being accessible or not.

Furthermore, adding a deeply nested object, means Vue needs to iterate it to add proxies/observers.

This call was crashing my browser :open_mouth: By using `Object.freeze`, Vue is later able to tell you don't want reactivity, doesn't bother setting up reactivity, creating proxy objects, etc.

src/components/cylc/graph/Graph.vue
```diff
@@ -0,0 +1,631 @@
...
+    this.setupHtmlLabel(this.cytoscapeInstance, states)
+    // set the initial zoom-level and pan (position) of the graph
+    // to set only the zoom level, call: this.cytoscapeInstance.zoom(0.5)
+    this.cytoscapeInstance.viewport({
+      zoom: 0.5,
+      pan: {
+        x: 100,
+        y: 100
+      }
+    })
```
kinow: Also spent a long time searching for this one. Reaaaally bothered by having to always zoom out when opening the graph for my two simple `five` and `families2` workflows :smile:

src/components/cylc/graph/Graph.vue
```diff
@@ -0,0 +1,631 @@
...
+  },
+
+  beforeDestroy () {
+    // before the component is destroyed, we need to remember to remove the tooltip, otherwise it is still visible
+    // in the next view/route.
+    if (this.tippy !== null) {
+      this.tippy.hide()
+    }
+    this.cytoscapeInstance.panzoom('destroy')
+    document.removeEventListener('keydown', this.keydownEventHandler)
```
kinow: These were not being removed. So we had possibly some detached objects & memory leaking somewhere. (Found about the Panzoom clean up call while looking at their docs to set up default values, by accident).

src/components/cylc/graph/Graph.vue
```diff
@@ -0,0 +1,631 @@
...
+  },
+
+  computed: {
+    // we map all the workflows here, but we have the workflowName prop to filter and know which one is being displayed
+    ...mapState('workflows', ['workflows'])
+  },
+
+  methods: {
+    /**
+     * @param {[
```
kinow: GitHub syntax parser bug I think.

src/components/cylc/graph/Graph.vue
```diff
@@ -0,0 +1,631 @@
...
+        actions: {}, // actions to be added
+        undoableDrag: true, // Whether dragging nodes are undoable can be a function as well
+        stackSizeLimit: undefined, // Size limit of undo stack, note that the size of redo stack cannot exceed size of undo stack
+        ready: function () {
+          // callback when undo-redo is ready
+          this.loading = false
+        }
+      }
+      instance.undoRedo(undoRedoOptions)
+    },
```
kinow: Not quite sure undo-redo is working... but left it anyway. Lumino also has some key listeners...

src/components/cylc/graph/Graph.vue
```diff
@@ -0,0 +1,631 @@
...
+      instance.panzoom(panzoomDefaults)
+    },
+
+    /**
+     * Sets up the navigator plugin in the instance.
+     * @param {cytoscape} instance - the cytoscape instance
+     */
+    setupNavigator (instance) {
+      instance.navigator({
+        container: this.$refs['cytoscape-navigator-overlay'],
```
kinow: This was tricky too.

Before, we were accessing the element by CSS class.

When using it with Lumino, you would have two widgets & components. The `.cytoscape-navigator-overlay` would return an array with two elements. So only one would be used by both tabs, the other would appear blank.

src/components/cylc/graph/Graph.vue
```diff
@@ -0,0 +1,631 @@
...
+              const jobId = `<span class="jobid">${job.batchSysJobId}</span>`
+              const jobSquare = `<div class="box-job-graph ${job.state}-graph"></div>`
+              const details = `<details><summary>${jobId}${jobSquare}</summary>`
+              const submitNum = `<br><strong>submit number:  <span style="color: #555;">${job.submitNum}</span></strong>`
+              const batchSysName = `<br><strong>batch sytem name: <span style="color: #555;">${job.batchSysName}</span></strong>`
+              const batchSysJobId = `<br><strong>batch sytem job id: <span style="color: #555;">${job.batchSysJobId}</span></strong>`
+              const host = `<br><strong>host:  <span style="color: #555;">${job.host}</span></strong>`
+              const startedTime = `<br><strong>started time:  <span style="color: #555;">${job.startedTime}</span></strong>`
+              const submittedTime = `<br><strong>submitted time:  <span style="color: #555;">${job.submittedTime}</span></strong>`
+              const finishedTime = `<br><strong>finished time:  <span style="color: #555;">${job.finishedTime}</span></strong><br>`
```
kinow: Several string concatenation operations replaced by backticks.

However, we still have a lot of `innerHTML` and styles and attributes being changed directly. This is quite bad for performance, possible to cause reflows too.

- https://coderwall.com/p/o9ws2g/why-you-should-always-append-dom-elements-using-documentfragments
- https://www.sitepoint.com/10-ways-minimize-reflows-improve-performance/

src/components/cylc/graph/Graph.vue
```diff
@@ -0,0 +1,631 @@
...
+    /**
+     * Sets up the event listeners for the instance.
+     * NB: remember to remove the event listeners to avoid memory leaks.
+     */
+    setupEventListeners () {
+      document.addEventListener('keydown', this.keydownEventHandler)
+    },
+
+    keydownEventHandler (event) {
+      if ((event.metaKey && event.key === 'Z') || (event.ctrlKey && event.key === 'Z')) {
```
kinow: (`which` is deprecated - https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which, `key` is a possible replacement... TL;DR: `which` behaviour varies per browser)

kinow:
I think I got too excited, so didn't squash the commits so that I can revert some if requested. Will add some comments to the new code, but here are the main changes:

- There was some code from `vue-cytoscape` ported to the project.
- Replaced string concatenation in the component by backtick (template literal)
- Moved configuration objects to an `index.js` file
- Moved component and related files to its own folder
- Removed unnecessary files
- Fixed typos and unnecessary or wrong CSS styles
- Added comments to code
- Used a `v-switch` and a `v-select` for layout controls (I think @oliver-sanders suggested it in some issue on github or message on riot :tada: )
- Used a `v-expansion-panel` to display the layout options only when requested (more visible space for the graph)
- Set transparency to the minimap for the same reason above
- Defined initial zoom level, and pan (position of graph)
- Changed styles to use as much viewport as possible (both Graph view and on Lumino as widget)
- Separated DOM elements, so that Lumino widgets are now able to display multiple Graph components
- Added some code to clean up event listeners and Cytoscape objects
- A few more changes here and there where I thought the code could be simpler or the IDE raised some warning

src/components/cylc/graph/Graph.vue
```diff
@@ -0,0 +1,631 @@
...
+  }
+}
+</script>
+
+<style lang="scss">
+@import '~@/styles/cytoscape/html-label.scss';
+@import '~@/styles/cytoscape/cytoscape-custom.scss';
+@import '~@/styles/cytoscape/panzoom.css';
+@import '~tippy.js/themes/light-border.css';
+</style>
```
kinow: There were 2 blocks of `css` and `scss`, but `scss` can handle `css` too. The Toppy CSS was imported in the JS part (webpack is able to handle that, but easier to have that here I guess)

public/index.html
```diff
@@ -4,7 +4,6 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link href='https://fonts.googleapis.com/css?family=Roboto:300,400,500,700,300italic,400italic' rel="stylesheet" type="text/css">
-    <link href="https://use.fontawesome.com/releases/v5.0.13/css/all.css" rel="stylesheet">
```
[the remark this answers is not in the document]
oliver-sanders: FYI we import fontawesome in index.html https://github.com/cylc/cylc-ui/blob/master/public/index.html#L7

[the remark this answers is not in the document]
kinow: I think you mentioned it somewhere else too. That got me thinking what could be using it, and by coincidence I believe it's only the `Graph` component that was using it for icons.

This code removed in the highlight above is exactly to remove the `public/index.html` import :point_up:

src/components/cylc/graph/Graph.vue
```diff
@@ -0,0 +1,631 @@
...
+                elements.nodes.push({
+                  data: node
+                })
+              })
+            }
+          }
+        }
+        // clear existing data
+        this.graphData.edges.length = 0
+        this.graphData.nodes.length = 0
```
oliver-sanders: Wow I didn't know you could modify an array by messing with its length parameter!

src/components/cylc/graph/Graph.vue
oliver-sanders: What is is meant to do?

src/components/cylc/graph/Graph.vue
oliver-sanders: It's nasty but this code won't make it to 1.0.0 so we can ignore it for now.

src/components/cylc/graph/Graph.vue
kinow: After you modify the position of nodes, for example, pressing `CTRL + Z` should undo that, and move your node back to its original position.

Demo: https://rawgit.com/iVis-at-Bilkent/cytoscape.js-undo-redo/master/demo.html

src/components/cylc/graph/Graph.vue
oliver-sanders: Vue has some support for modifier keys too https://vuejs.org/v2/guide/events.html#System-Modifier-Keys

src/components/cylc/graph/Graph.vue
kinow: Couldn't find a more elegant way to clear lists in JS, but I guess that's normal for JS, I think :grin:

src/components/cylc/graph/Graph.vue
oliver-sanders: Could do with sticking a comment about this in the code (we have a bad history with forgetting important information like this).

kinow: Good idea. Done!

kinow:
Thanks for the review @oliver-sanders !

kinow:
@oliver-sanders sorry, pushed two more commits :point_up: 

The first is to make the widget wrapper to invoke the `resizeGraph()` function of the `Graph` wrapped component, once the widget has been activated. This fixes the Firefox 60.8 issue reported by @hjoliver in #381 

The second commit is to fix a console error I noticed could happen when the user closed the widget while the layout was "running" (`cytoscape.layout().run()` more or less). There is nothing that the user or us can do about it, so the only option for that case is to ignore the error — though it's still possible to capture the error with the browser console by checking the option to break on any exception.

src/components/cylc/graph/Graph.vue
oliver-sanders: That's a bit nicer than the (100, 100) hack.

oliver-sanders:
> The second commit is to fix a console error I noticed could happen when the user closed the widget while the layout was "running"

Interesting, I didn't see that.

Code looks good.

PR #405, opened 2020-02-17
Remove duplicate computed/getter code, simplify current tree code avoiding data duplication

### LINKED ISSUE #232: Use __typename instead of creating extra __type
When the `WorkflowService` runs and retrieves the response for a GraphQL query, for the Tree Componet (#145) we re-create the hierarchy and set the nodes types in `__type`.

The query response in GraphiQL looks like this:

```json
{
  "data": {
    "workflows": [
      {
        "id": "kinow|five",
        "name": "five",
        "status": "running",
        "owner": "kinow",
        "host": "kinow-VirtualBox",
        "port": 43011,
        "taskProxies": [
          {
            "id": "kinow|five|20131226T0000Z|foo",
            "state": "succeeded",
            "cyclePoint": "20131226T0000Z",
            "task": {
              "meanElapsedTime": 20.100000381469727,
              "name": "foo"
            },
            "jobs": [
              {
                "id": "kinow|five|20131226T0000Z|foo|1",
                "batchSysName": "background",
                "host": "localhost",
                "startedTime": "2019-09-16T21:56:04Z",
                "state": "succeeded",
                "submitNum": 1
              }
            ]
          },
...
```

But, the response from ApolloClient looks like this:

```json
{
  "data":{
    "workflows":[
      {
        "id":"kinow|five",
        "name":"five",
        "status":"running",
        "owner":"kinow",
        "host":"kinow-VirtualBox",
        "port":43011,
        "taskProxies":[
          {
            "id":"kinow|five|20140118T1200Z|foo",
            "state":"running",
            "cyclePoint":"20140118T1200Z",
            "task":{
              "meanElapsedTime":20.200000762939453,
              "name":"foo",
              "__typename":"Task"
            },
            "jobs":[
              {
                "id":"kinow|five|20140118T1200Z|foo|1",
                "batchSysName":"background",
                "host":"localhost",
                "startedTime":"2019-09-16T22:14:31Z",
                "state":"running",
                "submitNum":1,
                "__typename":"Job"
              }
            ],
...
```

Not sure where the `__typename` gets added. Whether GraphiQL has this info but is simply hiding it, or if it is exclusive to ApolloClient (in which case they could change and break our UI, not so nice).

Here's the part of the code we currently set the type: https://github.com/cylc/cylc-ui/blob/6788c81ce1aacc5906bbf29a2d0bffbabbf613c7/src/store/workflows.module.js#L59

kinow:
These changes partially address #145, and closes #232 

The first three commits in this PR were:

- 02391dccba6cbbfe9e5096f21096a270d208e4f0
    * We had the same code in the View `computed`, and in the Vuex `getter`. This commit removes the duplicate code.
- 012139e86ffed9a7c944f81c678b717d7c1a818e
    * The code of other components is already under their respective packages (`graph`, `workflow`, etc), but not for `tree`. This commit simply moves `Tree.vue` and `TreeItem.vue` under the existing `tree` package.
- 58e20dbda903ee704351d40580a4db2b179f9b5a
    * This is the main change here I think. Tried to - hopefully - do as @oliver-sanders suggested, avoiding duplicated data, just re-using the existing data by reference and adding the bits necessary like `children: []` array, `expanded: true`, etc.

EDIT: added more commits later to finish the suggestion of @oliver-sanders , which keeps the JSON data as-is, and just attaches it to `treeItem: { node: jsonObject, state, id, etc }`. Had to update tests, and some other files for that, thus the extra commits.

If GitHub Actions pass, then will check with @oliver-sanders if the last change makes sense? Then will try to re-build the Infinite Tree on top of this, without duplicating any data (we may still need to discuss the deltas, but at least this should :crossed_fingers: address the part about duplication?)

**Requirements check-list**


src/components/cylc/tree/Tree.vue
```diff
@@ -6,15 +6,14 @@
         :key="workflow.id"
         :node="workflow"
         :hoverable="hoverable"
-        :min-depth="minDepth"
```
kinow: Not using `min-depth` any longer. We are now simply giving the component an array with the cycle points. Each cycle point will contain families or tasks, each family contains tasks, and so on.

This was a left-over from the first versions of the tree component.

src/components/cylc/tree/index.js
```diff
@@ -1,6 +1,132 @@
...
+ *
+ * @param familyProxy {Object} family proxy
+ * @return {{children: [], __type: string}}
+ */
+function createFamilyProxyNode (familyProxy) {
+  return {
+    __type: 'family',
+    ...familyProxy,
+    children: []
+  }
```
kinow: @oliver-sanders I think this is what you meant with your previous example from the Graph component?

I tested and for the infinite tree we can use the same approach re-writing the state.

```js
return {
  __type: 'family',
  ...familyProxy,
  children: [],
  state: {
    open: true,
    size: 32,
    // etc, then
    state: familyProxy.state
  }
}
```

kinow: And promise to keep the same for the infinite tree. Then there would still be the issue of how to handle the deltas :+1:

oliver-sanders: Doesn't the splat operator duplicate the data?

The way the graph component does it is to put a reference to the object under it's own namespace:

https://github.com/cylc/cylc-ui/blob/697d383d281f6e158411c9128c38873fb18533c6/src/components/cylc/graph/Graph.vue#L253

So perhaps something like this:

```js
function createFamilyProxyNode (familyProxy) {
  return {
    __type: 'family',
    node: familyProxy,
    children: []
  }
```

Then you get a clear separation of Cylc data and Tree data.

```js
> treeItem.state
{
  open: true,
  size: 32
}
> treeItem.node.state
'succeeded'
```

kinow: The `...` spread operator works similarly to `Object.assign`, so we could end up with copies. Changed to `node: { }`.

The `TreeItem.vue` funnily had a `node` prop already. So it became `node.node.state`... added 9 new commits to change the code and update required tests/other related code. Should we add another commit to rename it the `TreeItem` prop to `treeItem`? Or should we do that later @oliver-sanders ?

Quite happy with the result for now. It got simpler, and better organized this way, which should help with the infinite tree work later. (I've added a few TODO comments to help me remembering what goes where/why.)

oliver-sanders: This looks good to me.

kinow:
Set to 0.2 as we haven't released and I **think** it would be OK to add this one (@hjoliver we can kick it to 0.3 if it's better too - just let me know and I'll update linked issues too).

hjoliver:
I think we can take this as approved by @oliver-sanders - given his comments here and on Riot.

PR #406, opened 2020-02-18
Use Webfont Roboto offline

### LINKED ISSUE #386: Remove files that require Internet connection
**Describe the bug**

Off the top of my head, we have FontAwesome being removed in #383 , but also the Roboto webfont.

These files will result in a 404 in environments such as HPC with restrictions to access external resources.

`master`

Access the UI and check the resources downloaded.

**Expected behavior**

Everything being served from wherever Cylc UI was installed only.

**Screenshots**

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #386 (I believe this is the last of the files being downloaded now).

**Requirements check-list**
- [x] Does not need tests (why? changed location of font, there should be no visual difference).


kinow:
Many new files added, but they were referenced anyway in the CSS file we were downloading. Checked in browser network tab that it was not fetching the webfont CSS file any longer.

First tried to use a webpack plugin that promises to download the webfonts, but it failed twice. Some incompatible plugins or dependencies that we use perhaps?

Then used https://google-webfonts-helper.herokuapp.com/fonts/roboto?subsets=latin to create a ZIP with the fonts we use (yup, we use even the font-weight: 100 and 300, when certain styles of Vuetify are applied - note: if we limit the typography styles later, we can reduce a bit more the app size, if necessary).

PR #408, opened 2020-02-19
Set large modifier to same as medium/small for Dashboard elements

### LINKED ISSUE #407: Dashboard responsiveness problem
The Workflows column gets wider (and pushes the Events column across) as the window gets narrower:
peek

kinow:
These changes close #407 

Set to milestone 0.2, but can be pushed to 0.3. I haven't added a changelog, but will do if necessary :+1: 

**Requirements check-list**
- [x] Does not need tests (why? only modifies a CSS style applied for responsive layout).
- [x] No change log entry required (why? e.g. it's visible to users, but I think tis such a small change that it doesn't need a changelog entry?).

### DIFF
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -5,7 +5,7 @@
     class="c-dashboard mt-4"
   >
     <v-layout wrap>
-      <v-flex xs6 md6 lg3>
+      <v-flex xs6 md6 lg6>
         <p class="display-1">Workflows</p>
         <!-- TODO: link with data from the query -->
         <v-data-table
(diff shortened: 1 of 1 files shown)


kinow:
@hjoliver I've set it to six columns by six columns. The grid has a total of 12 columns. So that divides Workflows & Events equally.

But I had a look at the design sketch, and the Workflows area appear to be slightly smaller than Events. What do you reckon? Should we make it 4/8? 3/9? Or leave 6/6?

2020-02-19 14-38-06 flv

hjoliver:
Much better, thanks!

(But I'll put up another issue for really narrow-screen widths, because that's really a different problem...)

PR #414, opened 2020-02-20
Use old simple CSS to load fonts, instead of trying to use webpack loaders

### LINKED ISSUE #413: Offline webfont not working behind the proxy+hub
**Describe the bug**

The URL for the fonts are `/fonts/....`, which works fine without the hub.

Behind the proxy+hub, the URL is blocked, and redirected to `/hub/fonts/...`. The JS content is served from `/user/kinow/js/...`. So we would need to do the same for fonts. Except that it is a bit trickier, apparently, as the URL is created when the SCSS styles are processed... :thinking: 

`master`

Run the UI + Hub, and check the browser console. I have the font in my OS, so I couldn't see before that it was broken, only after looking at the browser console while I was working on the infinite tree I noticed the issue,

**Expected behavior**

No errors, nice font used.

**Screenshots**

image

**Additional context**

Related issue: #406 

**Pull requests welcome!**

kinow:
These changes close #413 

So I spent some time this afternoon trying to get the webpack loaders to work with the font, and with a relative path.

When I used a relative path, webpack would crash complaining that it couldn't load it. Searched vue-loader and sass-loader issues, and there are some workarounds. But the closest I found, was to have one prefix for production, like `/`, and another one for dev, like `../../`.

This doesn't solve out issue, as the path for our fonts is dynamic, and can be `/user/kinow/font/...`, or `/user/hjoliver/font/...`.

Went to the gym, and then realized we could use the oldest approach. Just load a CSS like a normal file, without using any of the loaders or webpack.

This is, in practice, the same we were doing before my previous change. But instead of loading it from Google's servers, we are using our local site.

To test, here's what I did:

- `sudo apt install font-manager`
- removed all Roboto* fonts
- restarted firefox
- refreshed, and confirmed `master` was broken, then switched to this branch
- confirmed the new font was loaded

image

**Requirements check-list**
- [x] Does not need tests (why?adding fonts to site, no other code change).


kinow:
And tested with offline + `npm run serve`, and hub+proxy with `npm run build:watch`.

Issue #356, opened 2020-01-08
Move GraphQL queries to their own modules


kinow:
**Describe exactly what you would like to see in an upcoming release**

We have `graphql-tag` in our dependency tree, and we are using its `gql` for queries. One feature of `graphql-tag` is allowing for queries to be imported, just like normal modules.

This reduces cluttering and dependencies, with queries in a single place, and that can be shared among multiple components.

**Additional context**

Vue CLI UI app is an example of how that can be done: https://github.com/vuejs/vue-cli/tree/9ec2d34e652e4a7cdb9c8c956a4d67a5ea91af92/packages/%40vue/cli-ui/src/graphql

**Pull requests welcome!**

PR #418, opened 2020-03-01
Move queries to a module

kinow:
These changes partially address #356 

Moving all GraphQL queries from views and components to a single place. This removes some duplication that we had, for example, the workflows tree and graph queries were defined in both `Workflows.vue` and `Graph.vue`/`Tree.vue`.

Also renamed the `GScan.vue` view, to `WorkflowsTable` - it used to be the `GScan` view some time ago.

**Requirements check-list**
- [x] Does not need tests (why? moving things around only).


kinow:
This change is present in the Infinite Tree branch too, but I think it would be simpler to move some of the changes there to other PR's.

kinow:
In a future PR, we can also move `src/utils/graphql.js` code over that `src/graphql/` package too.

PR #419, opened 2020-03-01
Fix GraphQL alias simplify code, create cycle points from GraphQL, and revert __typename change

### LINKED ISSUE #403: WorkflowService/GQuery seem to have issues when fields are aliased in the query
**Describe the bug**

A query such as:

```graphql
subscription {
  cyclePoints: familyProxies(ids: ["root"]) {
    cyclePoint
  }
  familyProxies (exids: ["root"], sort: { keys: ["firstParent"]}) {
    id
  }
}
```
Ends up with two `FamilyProxy` selections in the GraphQL Query JS object. One of them contains an "alias" `cyclePoints` as specified in the query.

image

But the Vuex store ends up with an object missing `cyclePoints`. Probably just needs to use the `alias` when it is present, or the `name` otherwise.

image

Use a GraphQL query with an alias.

**Expected behavior**

A JS object is added to the store with both `familyProxies` and `cyclePoints`.

image

**Screenshots**

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #403, and revert #232 (so that offline mode works again, as `__typename` is created by `ApolloClient` only - i.e. unless you are querying a real endpoint with the client, no `__typename`s are created).

It also creates cyclepoints using GraphQL, instead of iterating over family proxy data in JS (thanks to @dwsutherland).

**Requirements check-list**

### DIFF
--- a/src/components/cylc/gscan/index.js
+++ b/src/components/cylc/gscan/index.js
@@ -9,15 +9,18 @@
 function getWorkflowSummary (workflow) {
   const states = new Map()
   for (const taskProxy of workflow.taskProxies) {
-    for (const job of taskProxy.jobs) {
-      // TODO: temporary fix, as the backend is sending ready jobs, but they will change in cylc flow&uiserver in the future
-      if (job.state === 'ready') {
-        continue
+    // a task in waiting, may not have any jobs
+    if (taskProxy.jobs) {
+      for (const job of taskProxy.jobs) {
+        // TODO: temporary fix, as the backend is sending ready jobs, but they will change in cylc flow&uiserver in the future
+        if (job.state === 'ready') {
+          continue
+        }
+        if (!states.has(job.state)) {
+          states.set(job.state, new Set())
+        }
+        states.get(job.state).add(`${taskProxy.name}.${taskProxy.cyclePoint}`)
       }
-      if (!states.has(job.state)) {
-        states.set(job.state, new Set())
-      }
-      states.get(job.state).add(`${taskProxy.name}.${taskProxy.cyclePoint}`)
     }
   }
   for (const [stateName, tasksSet] of states.entries()) {
--- a/src/services/gquery.js
+++ b/src/services/gquery.js
@@ -23,7 +23,11 @@ function getSelections (a) {
   const selections = {}
   for (const selection of a.selectionSet.selections) {
     if (selection.kind === 'Field') {
-      selections[selection.name.value] = selection
+      let key = selection.name.value
+      if (Object.hasOwnProperty.call(selection, 'alias') && selection.alias) {
+        key = selection.alias.value
+      }
+      selections[key] = selection
     }
   }
   return selections
--- a/src/components/cylc/tree/TreeItem.vue
+++ b/src/components/cylc/tree/TreeItem.vue
@@ -14,15 +14,15 @@
       >{{ isExpanded ? '&#9661;' : '&#9655;' }}</v-flex>
       <!-- the node value -->
       <!-- TODO: revisit these values that can be replaced by constants later (and in other components too). -->
-      <div class="node-data" @click="nodeClicked" v-if="node.node.__typename === 'CyclePoint'">
+      <div class="node-data" @click="nodeClicked" v-if="node.type === 'cyclepoint'">
         <task :status="node.node.state" :progress=0 />
         <span class="mx-1">{{ node.node.name }}</span>
       </div>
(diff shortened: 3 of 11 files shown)


kinow:
These changes are in the infinite branch too, but can be applied earlier, which also simplifies the PR for the infinite tree. I left three comments in the code, 2 `TODO` for the infinite tree, and one comment about how something will work with the infinite tree. Happy to remove if others think it'd be confusing to add them now.

Now I will start working on the deltas using a new branch off the `master` branch. The 2 PR's I created, plus the delta one, once merged, will make integrating the infinite tree much easier :+1: (and I'm always OK with `git rebase`ing and fixing some conflicts).

src/components/cylc/gscan/index.js
```diff
@@ -9,15 +9,18 @@
 function getWorkflowSummary (workflow) {
   const states = new Map()
   for (const taskProxy of workflow.taskProxies) {
-    for (const job of taskProxy.jobs) {
-      // TODO: temporary fix, as the backend is sending ready jobs, but they will change in cylc flow&uiserver in the future
-      if (job.state === 'ready') {
-        continue
+    // a task in waiting, may not have any jobs
```
kinow: This appeared after I re-generated the mock data. Interestingly it got some tasks with no jobs, which caused console errors in the UI.

src/services/gquery.js
```diff
@@ -23,7 +23,11 @@ function getSelections (a) {
   const selections = {}
   for (const selection of a.selectionSet.selections) {
     if (selection.kind === 'Field') {
-      selections[selection.name.value] = selection
+      let key = selection.name.value
+      if (Object.hasOwnProperty.call(selection, 'alias') && selection.alias) {
+        key = selection.alias.value
+      }
+      selections[key] = selection
```
kinow: The alias in GraphQL queries is fixed here :point_up:

kinow:
Fixing conflicts in a bit.

kinow:
Rebased and did some manual testing locally with one workflow. Should be good to be reviewed assuming CI build passes.

src/components/cylc/gscan/index.js
```diff
@@ -9,15 +9,18 @@
...
+    if (taskProxy.jobs) {
+      for (const job of taskProxy.jobs) {
+        // TODO: temporary fix, as the backend is sending ready jobs, but they will change in cylc flow&uiserver in the future
+        if (job.state === 'ready') {
+          continue
+        }
+        if (!states.has(job.state)) {
+          states.set(job.state, new Set())
```
oliver-sanders: Can get this info from the task state summary thinggy, can do later.

src/components/cylc/gscan/index.js
oliver-sanders: Does the task task have jobs set to `null` rather than `[]`?

kinow: Funny, that's what I thought. That I had seen an `undefined`, not even a `null`. But trying now I only got empty `[]` jobs. Guess this is not necessary. Let me fix it.

kinow: Ah, actually, re-reading my comment above I remembered it was recorded in the `checkpoint.js`: 

https://github.com/cylc/cylc-ui/blob/8b8c3da7a1c054ae15a7b4f38ccd49957ceaca95/src/services/mock/checkpoint.js#L293-L308

Can't say what caused it. I re-generated it as I normally do, and then got this `undefined` error in the offline mode.

Do you think we should ignore as it shouldn't happen with live data? Maybe something with the way we produce the mock data, or something I could have done wrong @oliver-sanders ?

oliver-sanders: Mmm, I cant remember how we generate checkpoints, will have a dig...

kinow:
> Seems legit, tested production build, all seems fine.
> 
> I haven't run a development build in a while, seem to be having issues but these are also on master:
> 
> Screenshot from 2020-03-17 11-39-50
> 
> Just me?

That appears to happen with the `User` object, which comes from JupyterHub when it's not in the offline mode. I think we got an error somewhere and the object returned didn't have a `statusText`

https://github.com/cylc/cylc-ui/blob/f2a8c5571c9657e6a621b7fb35e0e38cd8ae98f7/src/services/user.service.js#L13

This file was not changed in this PR, so I am hoping this is not related :crossed_fingers: but if you don't see this error in other branches then I can have a better look to try and reproduce it.

PR #423, opened 2020-03-11
Enable e2e on GitHub workflow CI

kinow:
These changes partially address #170 (missing coverage)

**Requirements check-list**


kinow:
Had a post-it on my monitor to "add functional test for responsive mode on Cylc UI". That was from some time ago, when I broke the responsive mode while working on the Drawer component.

Our interface is stable now, so we can introduce a few e2e to parts that are not likely to change, like the Drawer, the footer, the user profile page, etc. I think we still need to wait a bit before adding tests to tree view, graph view, etc, as these are still changing, and running/editing/troubleshooting functional tests with Cypress is not fast nor easy, depending on the view & test.

kinow:
GitHub action was not running after I pushed new commits this morning. Marking as ready for review to see if that kicks a new build.

kinow:
Still battling against GitHub actions to get it to run Cypress and a headless browser, but have managed to add one more test that will verify that the responsive mode for the drawer & toolbar works.

When the viewport is desktop mode, it displays the application with the drawer.

image

And when the viewport is smaller and considered mobile or tablet, the drawer is hidden, but the user should see the button to display the drawer again (this is what I broke some time ago, now this test should catch it - as soon as GitHub action works).

image

kinow:
Functional tests enabled, and passing on GitHub. At the moment, three tests taking around 2 minutes to run. As we add more tests, this number will climb up. We just need to stay alert at the time each new test increases the total build time and, if necessary, use something like the Cypress GitHub Action [parallel](https://github.com/cypress-io/github-action#parallel) option.

src/views/UserProfile.vue
```diff
@@ -44,6 +46,7 @@
                       chips
                       multiple
                       disabled
+                      id="groups"
```
oliver-sanders: Should probably namespace these ids to prevent them conflicting with other components.

E.G. GScan will likely have an element `#groups`

Perhaps `profile.groups`, `profile-groups`, whatever.

tests/e2e/specs/drawer.js
```diff
@@ -0,0 +1,21 @@
...
+      .should('be.visible')
+  })
+  it('Is NOT displayed when mode is mobile', () => {
+    // when the window dimension is below a mobile-threshold, the app sets state.app.drawer as false
+    // and then the drawer is hidden
+    cy.viewport(320, 480)
+    cy.visit('/#/')
+    cy
+      .get('.v-navigation-drawer')
+      .should('not.be.visible')
```
oliver-sanders: Wow this stuff is powerful!

src/views/UserProfile.vue
kinow: Oh, good idea!

oliver-sanders:
@wxtim, @datamel if you get the time this would be a good flick through this / review to make you aware of the cypress test framework and the sort of things we can do with it (new to me too).

kinow:
Pushed one more test related to #427 . It simply checks that there is no toolbar in the dashboard view, but that there is one when you visit `/#/workflows/one` (the mocked/checkpoint workflow included). No other changes, unless GitHub actions fail.

kinow:
> 
> 
> Good.
> 
> Quick question, what is the `.visible` check testing for. Is it checking the CSS `display` attribute, presence of a DOM element or on-screen position?

It uses JQuery's [:visible selector](https://api.jquery.com/visible-selector/), so pretty much only what can actually be seen (e.g. a div with width/height set to 0 is not visible):

>Elements are considered visible if they consume space in the document. Visible elements have a width or height that is greater than zero.
>
>Elements with visibility: hidden or opacity: 0 are considered visible, since they still consume space in the layout.
>
>Elements that are not in a document are considered hidden; jQuery does not have a way to know if they will be visible when appended to a document since it depends on the applicable styles.
>
>This selector is the opposite of the :hidden selector. So, every element selected by :visible isn't selected by :hidden and vice versa.

hjoliver:
(Nice!)

kinow:
Fixed `git` conflicts.

PR #427, opened 2020-03-17
Use a metadata field in each vue route instead of a list of strings

kinow:

This should be more clear to quickly see what enables the toolbar or not. I also think this should be less error prone.

This morning I was going to write a e2e test to confirm that the toolbar is present when you visit the workflow route (the one with the Lumino component). I remembered I had broken the toolbar a couple times in the past, and it was also due to renaming routes (workflows, then tree, or treeview, etc).

By not relying on the route name, and instead using a flag in the Route record metadata (also present, a feature of VueRouter) we should reduce the risk of I or others breaking the toolbar again. Will prepare a e2e test for that anyway later as well :+1: 

**Requirements check-list**
- [ ] Appropriate tests are included (unit and/or functional). (I will add one test to #423)

### DIFF
--- a/src/router/index.js
+++ b/src/router/index.js
@@ -53,7 +53,7 @@ Vue.use(Meta)
 router.beforeResolve((to, from, next) => {
   NProgress.start()
   if (to.name) {
-    if (['tree', 'graph', 'workflow'].includes(to.name)) {
+    if (to.meta.toolbar) {
       // When a workflow is being displayed, we set the title to a
       // different value.
       store.commit('app/setTitle', to.params.workflowName)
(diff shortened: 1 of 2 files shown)


kinow:
No milestone set, but will set to 0.2 if merged soon, otherwise this can be done in 0.3, no hurry on this one.

oliver-sanders:
@wxtim, @datamel feel free to jump in on any of these if you get the chance, Hillary is snowed under ATM.

hjoliver:
>  if you get the chance, @hjoliver is snowed under ATM.

If my proposal goes through as-is, I may be about to emerge from the snow.  But yes, feel free anyway :+1:

PR #428, opened 2020-03-18
Use Vuetify only, dropping old theme components and functions

### LINKED ISSUE #295: Use vuetify directly without the creativetim theme

kinow:
These changes close #295

**Requirements check-list**
- [x] Does not need tests (why? removing elements only).


kinow:
Set to 0.3, but can be merged anytime. Just update the milestone if/once merged.

kinow:
Removed Vuex helper code, as it was used in only one place (and one function, `setImage`, would raise an error if called).

Removed `MaterialCard` and `Offset`, and used a `VAlert` instead (used already in the project). This should reduce the size of the artefacts produced, and we won't have to worry about it when Vuetify is upgraded.

Before: 

image

After:

image

Checked with WAVE, and it found no accessibility issues. The text and background colours can be changed later if necessary, not focusing on that right now.

Also removed `Filter.vue` from the theme, broken at the moment as we didn't update it as it was not in use. And also `Locales.vue` which I create some time ago copying a file from Vuetify. We can re-introduce these if necessary later. This also reduces the size of our final bundle.

hjoliver:
No need to both @oliver-sanders for a second review as this just removes unused stuff and doesn't affect the main workflow views.

PR #430, opened 2020-03-23
Add e2e tests for tree view (offline mode)

kinow:

The offline mode is broken again, no big news. **However**, as we now have e2e tests, it is much easier to have a test to confirm that the offline mode is working. Here's what we will have in this PR:

1. Right now it is broken because the GraphQL query used to create mocked data was not updated. So now we parse (Python `re` module) the query value from `src/graphql/queries.js` and use it when generating the mocked data.

2. Functional tests will confirm that the tree component is working in the offline mode. Meaning that if anything breaks it again, the CI will fail in GitHub, and then we can fix before merging the PR.

Hopefully no more PR's from me to fix the offline mode :crossed_fingers: :grimacing: 

Tried to test the things I remember breaking with more frequency during when changing the code of the current tree view:

1. No nodes are displayed by some accidental change
2. Nodes are displayed, but clicking on the expand/collapse icon nothing happens
3. Missing job node details (leaf nodes)
4. The triangle of leaf nodes is not displayed correctly (e.g. the `margin-left` CSS attribute is broken for some reason).

**Requirements check-list**

### DIFF
--- a/src/components/cylc/tree/TreeItem.vue
+++ b/src/components/cylc/tree/TreeItem.vue
@@ -224,7 +224,7 @@ export default {
     getNodeDataClass () {
       const classes = {}
       classes['node-data'] = true
-      classes[this.node.type] = true
+      classes[`node-data-${this.node.type}`] = true
       return classes
     }
   }
--- a/src/components/cylc/tree/TreeItem.vue
+++ b/src/components/cylc/tree/TreeItem.vue
@@ -14,15 +14,15 @@
       >{{ isExpanded ? '&#9661;' : '&#9655;' }}</v-flex>
       <!-- the node value -->
       <!-- TODO: revisit these values that can be replaced by constants later (and in other components too). -->
-      <div class="node-data" @click="nodeClicked" v-if="node.type === 'cyclepoint'">
+      <div :class="getNodeDataClass()" @click="nodeClicked" v-if="node.type === 'cyclepoint'">
         <task :status="node.node.state" :progress=0 />
         <span class="mx-1">{{ node.node.name }}</span>
       </div>
-      <div class="node-data" @click="nodeClicked" v-else-if="node.type === 'family-proxy'">
+      <div :class="getNodeDataClass()" @click="nodeClicked" v-else-if="node.type === 'family-proxy'">
         <task :status="node.node.state" :progress="node.node.progress" />
         <span class="mx-1">{{ node.node.name }}</span>
       </div>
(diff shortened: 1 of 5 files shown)


src/components/cylc/tree/TreeItem.vue
```diff
@@ -14,15 +14,15 @@
       >{{ isExpanded ? '&#9661;' : '&#9655;' }}</v-flex>
       <!-- the node value -->
       <!-- TODO: revisit these values that can be replaced by constants later (and in other components too). -->
-      <div class="node-data" @click="nodeClicked" v-if="node.type === 'cyclepoint'">
+      <div :class="getNodeDataClass()" @click="nodeClicked" v-if="node.type === 'cyclepoint'">
```
kinow: Adding the `node.type` as a class, so that tests can easily find `div.node-data` for all the nodes with data or, something like `div.cyclepoint-node-data` for just the cycle point node data.

src/graphql/queries.js
```diff
@@ -1,5 +1,8 @@
 // Code related to GraphQL queries, fragments, variables, etc.
 
+// IMPORTANT: queries here may be used in the offline mode to create mock data. Before removing or renaming
+// queries here, please check under the services/mock folder for occurrences of the variable name.
+
```
kinow: Just a note. FWIW, the offline mode is broken because the query in `get-checkpoint.py` was not excluding the `root` family, which was an improvement from another recent PR.

tests/e2e/specs/tree.js
```diff
@@ -0,0 +1,16 @@
...
+    cy.visit('/#/workflows/one')
+    cy
+      .get('.node-data-cyclepoint')
+      .should(($div) => {
+        // by default, in our expected viewport size for tests, both cycle points exist and are visible
+        expect($div).to.have.length(2)
+        expect($div.get(0)).to.contain('20000101T0000Z')
+        expect($div.get(1)).to.contain('20000102T0000Z')
+      })
+  })
```
kinow: These are the first e2e tests for the tree component :tada: which I hope will prove useful not only for the offline mode, but also pretty soon for the infinite tree. So that when merging we know it still works the same way (I hope I won't have to touch any tests for the infinite tree, as I am adding simple tests using the visible nodes only)

kinow:
Ready for review! Got well familiar with Cypress now, quite useful, and way better than Selenium.

I guess one review should be enough as these are only tests, though one file changed to add new classes. Maybe add someone from MO @hjoliver ?

PR #432, opened 2020-03-25
Add loading state to Lumino widgets with skeleton loaders

### LINKED ISSUE #431: Lumino tabs may display old content until new data is retrieved from server
**Describe the bug**

Was testing the UI + UI Server just now, and noticed a strange behaviour. When you switch between two workflows (e.g. five and families2), immediately after you switch you will see the same tabs in Lumino.

However, unless the JS `WorkflowService` has retrieved new data and updated it in Vuex store, the component won't react to changes. So it will keep displaying the old data for a few seconds, or indefinitely if the `WorkflowService` takes too long to fetch data (e.g. network issue, backend delay, etc).

I think we can do better, and instead display to the user an empty component, maybe with a placeholder item, or a loader, and then update the component once new data for the selected workflow is present.

`master`

1. start 2 workflows (e.g. five and families2)
2. open the Lumino component for one workflow, which will have a tree by default, then add a new graph
3. switch to the other workflow, and notice how you can see the graph and tree for the old workflow for a few seconds until the websocket has received data and updated the Vuex store

**Expected behavior**

User is informed about the change in state, and knows that the system is doing something. This could be via a loader, a spinning wheel, a placeholder skeleton element, etc. And then sees the new data for the workflow selected, but doesn't see anything from the other workflow.

**Screenshots**

image

The image above shows the workflow five selected in GScan, also the Toolbar displays five. The tree is updated almost instantaneously. But the graph takes a bit longer, so you can clearly see that that's the graph from the families2 workflows (workflow five has no families except for root).

In a few seconds the component then displays the graph for the workflow five.

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #431 

Use existing views data `isLoading` that is updated by the `WorkflowService`. When we subscribe a query/view to the service, we pass a callback. So far, the callback has been the same for every view, simply updating a data state for loading.

Then this value gets passed down via props to other components, all the way down to the Lumino widget wrapper component. Which uses that reactive value to display a Vuetify [Skeleton Loader](https://vuetifyjs.com/en/components/skeleton-loaders/). Once data has been received, the service calls the callback which updates the loading state (to false).

And then the skeleton loader disappears and the original component is displayed.

**Requirements check-list**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -78,6 +78,10 @@ view and move access to the dashboard.
 [#396](https://github.com/cylc/cylc-ui/pull/396) - Fix leaf and other tree
 nodes layout.
 
+[#432](https://github.com/cylc/cylc-ui/pull/432) - Add loading state to
+Lumino widgets with skeleton loaders, preventing it from displaying the
+data of other workflows for a brief moment.
+
 ### Documentation
 
 [#215](https://github.com/cylc/cylc-ui/pull/215) - guide: add beginnings
--- a/src/components/cylc/workflow/Workflow.vue
+++ b/src/components/cylc/workflow/Workflow.vue
@@ -7,18 +7,21 @@
           :key="widgetId"
           :workflows="workflowTree"
           :widgetId="widgetId"
+          :is-loading="isLoading"
       />
       <graph-wrapper
           v-for="widgetId of this.graphWidgetIds"
           :key="widgetId"
           :workflow-name="workflowName"
           :widgetId="widgetId"
+          :is-loading="isLoading"
       />
       <mutations-wrapper
           v-for="widgetId of this.mutationsWidgetIds"
           :key="widgetId"
           :workflow-name="workflowName"
           :widgetId="widgetId"
+          :is-loading="isLoading"
       />
     </div>
   </div>
@@ -43,6 +46,10 @@ export default {
     workflowName: {
       type: String,
       required: true
+    },
+    isLoading: {
+      type: Boolean,
+      default: false
     }
   },
   components: {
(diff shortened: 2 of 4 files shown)


kinow:
What it looks like after this change:

GIFrecord_2020-03-25_143837

(the graph component really takes much longer to render the first time, that's why it was possible to see the old graph for some seconds before, now when the WorkflowService is done, there's another loading, smaller, in the panzoom control, where three dots appear every other second until the component has finished loading)

kinow:
The UI looks slower to me after this change. But I think it is a consequence of the delay to update variables/state in Vue, due to the 5 seconds delay in our service. I believe it will be much better with incremental updates + websockets - the 5 seconds delay.

oliver-sanders:
Cool, looks good to me, we can come up with our own skeletons (e.g. for the graph view) later.

Thinking about how to standardise this across all views, could we make the skeleton and view components into slots so that workflow component can provide a default fallback?

kinow:
> Cool, looks good to me, we can come up with our own skeletons (e.g. for the graph view) later.

+1

> Thinking about how to standardise this across all views, could we make the skeleton and view components into slots so that workflow component can provide a default fallback?

That's a bit hard I think. The workflow component has only 1 `div`, which is the placeholder for the Lumino panel. It also loads invisible wrappers for each tree/graph/mutation that it must display.

But then we have the extra bit of code in the `ContentWidget` and in the -`Wrapper` components. These are the components that actually load data.

In order to have the skeleton in the workflow, we would need to change that first I guess?

kinow:
BTW, if you look at the [README](https://github.com/jupyterlab/lumino) of JupyterLab's Lumino, at the bottom there's a link to a project I sent with Vue + Lumino based on our work here.

The main difference is that I used a single wrapper component with a [default slot](https://github.com/kinow/vue-lumino/blob/3873a2232393282012354ea2574efbb4ac61c735/src/components/Lumino.vue#L112). I only thought about that after re-writing the code, but I think with that we could have `ContentWidget` and `ContentWrapper` components (no need to have `TreeWrapper`, `GraphWrapper`, `EtcWrapper`).

Planning to submit a PR for that soon. Not the final improvement for the workflow component, and I think you have other ideas to improve the existing Lumino/Workflow code, just one step to improve it that could be useful for later :+1:

Issue #420, opened 2020-03-02
Incremental updates with GraphQL deltas


kinow:
**Describe exactly what you would like to see in an upcoming release**

Avoid re-creating Vuex data unnecessarily. Subscribe to GraphQL delta queries, and update the Vuex store.

Components should react accordingly to the data changed.

**Additional context**

The main challenge now, is that everything is a subscription (we did that to replace the old polling service).

Maybe instead of polling or subscription, we can simply send one query, then another subscription to update what has changed.

This is a blocker for the infinite tree, because of the state being in the data, not in the component any more - #391 . Which means users may alter the state of the nodes in the tree, but once the data is updated with new state, that state overrides whatever the user has done, which is not desirable.

**Pull requests welcome!**

kinow:
Note: we need to use Apollo Client's `.watchQuery()`, that will create a kind of `Observable`. When we subscribe to that `Observable` (subscribe here has nothing to do with GraphQL), we can then retrieve the data.

The object created by `.watchQuery()` (an Apollo Client type `ObservableQuery`), besides being an `Observable`, has also the method `.subscribeToMore()`.

The documentation points users to this method, and this combination of a query via `.watchQuery`, and the GraphQL subscription via the `.subscribeToMore()`.

Note 2: first changing the `/#/tree/five` view, as that's simpler, and isolates the Tree component, without the Lumino tabs code. The GraphQL query used for now is:

```graphql
subscription {
  deltas (id: "kinow|five") {
    pruned {
      tasks
      jobs
    }
    workflow {
      id
      status
      tasks {
        id
        name
        meanElapsedTime
      }
      jobs {
        id
        state
        batchSysName
        batchSysJobId
      }
    }
  }
}
```

Will update the query above later if necessary. Also important to note that I'm using the pull requests from Cylc Flow and Cylc UI Server https://github.com/cylc/cylc-flow/pull/3500 & https://github.com/cylc/cylc-uiserver/pull/118 respectively.

kinow:
Work being done in the branch https://github.com/kinow/cylc-ui/tree/deltas-1

kinow:
Note to self: check in the UI Server & Flow PR's later if a warning seen in the browser is being caused by the client or by the server. I guess it's related to not all fields present in the schema being returned to the client.

![image](https://user-images.githubusercontent.com/304786/75641236-707b3e00-5c9c-11ea-8282-581c6a50e66e.png)

[...]

PR #433, opened 2020-03-27
Use workflow ID instead of workflow name in queries

### LINKED ISSUE #422: Use workflow ID instead of workflow name in GraphQL queries

The delta subscriptions are being built using workflow ID's. So you have to query them with:

```graphql
subscription {
  deltas (ids: ["kinow|five"]) {}
}
```

But in the queries we have been using the workflow name.

```graphql
query {
  workflow (ids: ["five"]) {}
}
```

Queries work with either, name or ID. Graph component was initially using the workflow ID too. I think it would make sense to use the workflow ID instead, as that will have to be done anyway in the future when we start work to allow users to visualize workflows of other users (which means we may have two "five" workflows running, `/kinow/five` and `/hjoliver/five` for example).

The views get the workflow name from the URL, e.g. `/#/workflows/five`. We can get the workflow ID from the URL too, but this requires changing the URL paths, and some views.

The other option is to get the workflow name from the URL, but the user name from the logged-in user, from the session.

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #422 

We have the user (same from the URL) that is pre-fetched in the Route's. So this PR is simply using that `user.username` to use as parameter in the query, instead of using the workflow name.

This is necessary for #420 , as there the deltas use the workflow ID, not the name.

Further work to support URL's with other user's username is part of another issue :+1: 

**Requirements check-list**

### DIFF
--- a/src/lang/en-GB/App.json
+++ b/src/lang/en-GB/App.json
@@ -1,7 +1,7 @@
 {
   "name": "Cylc UI",
   "dashboard": "Dashboard",
-  "graph": "Graph",
+  "graph": "Graph {name}",
   "workflow": "Workflow {name}",
   "workflows": "Workflows",
   "notFound": "Not Found",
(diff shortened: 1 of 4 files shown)


kinow:
An easy way to review, is going through `graph/$workflow`, `tree/$workflow`, and `workflow/$workflow`, and looking at the debug log in the browser console. The query must use the workflow ID.

image

PR #435, opened 2020-03-31
Move to yarn

### LINKED ISSUE #364: Migrate to Yarn?

Running `npm install` can lead to issues in the lock file, but that can be fixed by using `npm ci` instead. `yarn`, however, doesn't appear to have the same problem - at least the other projects where I have used `yarn` never had issues.

I also have the impression that `yarn` commands are faster.

It is also easier to have multiple source/packages in the same repo/monorepo. Which can be done with `npm` using external tools like [Lerna](https://github.com/lerna/lerna). With Yarn one can use [workspaces](https://yarnpkg.com/lang/en/docs/workspaces/).

Today I was upgrading dependencies and one test involved pinning one dependency at the project level. This dependency existed in the dependency chain, but it was a transitive dependency (dependency of another dependency).

It's also easier to pin down dependencies with Yarn using [Yarn resolutions](https://yarnpkg.com/lang/en/docs/selective-version-resolutions/). (the same is possible with pip by specifying the dependency in `setup.py`, or Java Maven, etc). For `npm`, we need an external library too like [`npm-force-resolutions`](https://github.com/rogeriochaves/npm-force-resolutions).

The cons for using Yarn is needing yet another tool in our build software stack, learning curve, and possible issues.

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #364 

**Requirements check-list**


kinow:
Simplest way to review is running `yarn install`. If familiar with `npm install`, it would be interesting to know if it is/looks faster than `npm install`. Then, running `yarn install` a second time should be even faster, and `git status` shouldn't show anything.

Docs for migration from npm: https://classic.yarnpkg.com/en/docs/migrating-from-npm/

First issue I hard was `lint`, which has an existing bug in `babel-eslint`. Interesting that we didn't have it before with `npm`.

```bash
yarn run v1.22.4
$ vue-cli-service lint
 ERROR  TypeError: Cannot read property 'range' of null
Occurred while linting /home/kinow/Development/python/workspace/cylc-ui/src/router/index.js:28
TypeError: Cannot read property 'range' of null
Occurred while linting /home/kinow/Development/python/workspace/cylc-ui/src/router/index.js:28
    at SourceCode.getTokenBefore (/home/kinow/Development/python/workspace/cylc-ui/node_modules/eslint/lib/source-code/token-store/index.js:298:18)
    at checkSpacingBefore (/home/kinow/Development/python/workspace/cylc-ui/node_modules/eslint/lib/rules/template-curly-spacing.js:60:42)
    at TemplateElement (/home/kinow/Development/python/workspace/cylc-ui/node_modules/eslint/lib/rules/template-curly-spacing.js:119:17)
    at /home/kinow/Development/python/workspace/cylc-ui/node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Array.forEach (<anonymous>)
    at Object.emit (/home/kinow/Development/python/workspace/cylc-ui/node_modules/eslint/lib/linter/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (/home/kinow/Development/python/workspace/cylc-ui/node_modules/eslint/lib/linter/node-event-generator.js:254:26)
    at NodeEventGenerator.applySelectors (/home/kinow/Development/python/workspace/cylc-ui/node_modules/eslint/lib/linter/node-event-generator.js:283:22)
    at NodeEventGenerator.enterNode (/home/kinow/Development/python/workspace/cylc-ui/node_modules/eslint/lib/linter/node-event-generator.js:297:14)
    at CodePathAnalyzer.enterNode (/home/kinow/Development/python/workspace/cylc-ui/node_modules/eslint/lib/linter/code-path-analysis/code-path-analyzer.js:634:23)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Fixed with the code from [this comment](https://github.com/babel/babel-eslint/issues/799#issuecomment-568195009).

Then found an actual bug :bug: I used [`.every`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) which should be used to apply a predicate to every element. What I wanted was actually to execute a function for every element (`.forEach`). I think that's why `master` branch is broken (it was probably introduced with the skeleton loader I think, or some other recent commit, but luckily functional tests detected it :+1: )

oliver-sanders:
Seems fine, if I understand correctly this doesn't actually lock us into using `yarn`, developers can still choose yarn/npm according to preference.

kinow:
> Seems fine, if I understand correctly this doesn't actually lock us into using `yarn`, developers can still choose yarn/npm according to preference.

That's correct. The static analysis and CI are run against `yarn`, and we wouldn't have a `package-lock.json`. But users could choose any build tool as long as it populates their `node_modules` folder correctly, and supports the `package.json` scripts/commands.

kinow:
>Seemed pretty slow to me 

That was the first `yarn install` I think? How long does it take if you try it a second time? For me it's <2s. And `yarn install <dependency-name>` is also faster than `npm`. But the first one took a bit while like `npm`.

>Got some error: 

That might be a blocker for this then :cry:

oliver-sanders:
> That might be a blocker for this then

It all still works. The error occurred whilst installing fsevents, but `build:watch` is still able to detect file modification and rebuild accordingly.

oliver-sanders:
$ yarn install
yarn install v1.22.4
[1/4] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 1.71s.

kinow:
Rebased

hjoliver:
On my Centos 7 host:

On master, with `npm install`
- 30s from scratch (deleted `node_modules`)
- 10s next time

On this branch, with `yarn install`
- 15s from scratch
- 1s next time

kinow:
Thanks @hjoliver !

Besides performance, my main drive to move to Yarn is that it is supposed to perform better installing dependencies.

If you remember some time ago I think someone updated something in `package.json`. Then when the other tried to run `npm install` it changed `package-lock.json` again. Which had to be fixed in yet another commit.

That should not happen with yarn (:crossed_fingers: )

hjoliver:
Brilliant, sounds good :+1:

kinow:
Rebased fixing conflicts

PR #436, opened 2020-03-31
tree: restore the isHeld task status attribute

oliver-sanders:
**Requirements check-list**

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,13 +4,14 @@
   "private": true,
   "license": "GPL-3.0-only",
   "scripts": {
-    "serve": "vue-cli-service serve",
     "build": "vue-cli-service build",
     "build:report": "vue-cli-service build --report",
     "build:watch": "vue-cli-service build --watch --mode development",
-    "lint": "vue-cli-service lint",
+    "checkpoint" : "./src/services/mock/generate",
     "coverage:unit": "nyc vue-cli-service test:unit",
     "dev": "npm run serve",
+    "lint": "vue-cli-service lint",
+    "serve": "vue-cli-service serve",
     "test:e2e": "vue-cli-service test:e2e",
     "test:unit": "vue-cli-service test:unit"
   },
--- a/src/services/mock/generate
+++ b/src/services/mock/generate
@@ -1,26 +1,29 @@
 #!/bin/bash
 
+if ! sed --version 2>/dev/null | grep -q 'GNU' 2>/dev/null; then
+    echo -e '** GNU sed required **\n' >&2
+    exit 1
+fi
 
+cd "$(dirname "$0")"
 CHECK_FILE='checkpoint.js'
 
-
 cylc reg "one" one
-
-
 if cylc run "one" --fcp=20000102T0000Z --no-detach; then
     cp \
         "${HOME}/cylc-run/one/work/20000102T0000Z/checkpoint/checkpoint" \
         "${CHECK_FILE}"
     rm -r "${HOME}/cylc-run/one"
 else
+    echo -e '** Suite run failed **\n' >&2
     exit 1
 fi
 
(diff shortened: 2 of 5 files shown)


hjoliver:
Presumably this is a temporary change.  Can we have an explanation @oliver-sanders ?

oliver-sanders:
> Presumably this is a temporary change. Can we have an explanation @oliver-sanders ?

Explanation? This restores the pause symbol which appears next to tasks when they are held. At some point this was lost from the tree view, this PR puts it back.

oliver-sanders:
> Three e2e/functional tests failing for the tree component

I'm guessing the snapshot used for testing the tree view doesn't have the `isHeld` field which is causing traceback resulting in missing nodes. I'm guessing this as cypress is being pretty useless for me ATM.

@kinow I've lost track of where we are at with data snapshots. How do we update these, are we still using `./generate`?

kinow:
>@kinow I've lost track of where we are at with data snapshots. How do we update these, are we still using ./generate?

Yup. The only difference is that there is no more query in `get_checkpoint.py`. Instead it now [parses the JS variable](https://github.com/cylc/cylc-ui/blob/d26d0e9abcea69ebb29637fd672359609801f3a1/src/services/mock/checkpoint/get_checkpoint.py#L8) from the tree component query in JS.

So in theory all you'd need to do is run the `get_checkpoint.py` and update the branch/PR (:crossed_fingers:)

hjoliver:
> Explanation? ...

oliver-sanders:
> Ah, OK, I kind of figured we were partially transitioned to the new task states. But no biggie

We are using the `is_held` state attribute (the old held state went away long ago) misleading title, apologies.

oliver-sanders:
Done,

I've also streamlined the checkpointing process a little and aliased it in the `package.json` (so that I don't forget again).

```
npm run checkpoint
```

package.json
```diff
@@ -4,13 +4,14 @@
   "private": true,
   "license": "GPL-3.0-only",
   "scripts": {
-    "serve": "vue-cli-service serve",
     "build": "vue-cli-service build",
     "build:report": "vue-cli-service build --report",
     "build:watch": "vue-cli-service build --watch --mode development",
-    "lint": "vue-cli-service lint",
+    "checkpoint" : "./src/services/mock/generate",
```
kinow, line 10: Brilliant idea!

oliver-sanders:
fixing...

src/services/mock/generate
```diff
@@ -1,26 +1,29 @@
 #!/bin/bash
 
+if ! sed --version 2>/dev/null | grep -q 'GNU' 2>/dev/null; then
+    echo -e '** GNU sed required **\n' >&2
```
oliver-sanders, line 4: POSIX `sed` sucks!

src/services/mock/generate
```diff
@@ -1,26 +1,29 @@
...
-# Finally we export it so we can use it from JS modules
-echo -e '\nexport { checkpoint }' >> "${CHECK_FILE}"
+sed -i \
+    -e "1s|^|const checkpoint = |" \
+    -e "1s|^|/* eslint-disable */\n|" \
+    -e "s|$USER|user|g" \
+    -e "s|$(hostname)|localhost|g" \
+    -e "s|$(hostname | tr '[:upper:]' '[:lower:]')|localhost|g" \
+    -e "\$s|$|\nexport { checkpoint }|" \
+    "${CHECK_FILE}"
```
kinow, line 29: :rocket: nice! Didn't think in combining everything. Hopefully will remember it for next time, but easier to read.

kinow:
I thought the Yarn one would be simple, but it looks like it doesn't work on macos. Good idea to create a separate one, doing that right now :+1:

PR #438, opened 2020-04-01
Use base URL provided by Python backend (or mocked if offline) instead of building via JS

### LINKED ISSUE #343: Revisit how we build the GraphQL service URL

Some JS applications have a fixed URL for the backend (e.g. github.com). Other APIs require users to change a file during deployment (e.g. alter `dist/config.js` and set the URL). In our case, we are making a few assumptions for now.

We use the `window.location` [API](https://developer.mozilla.org/en-US/docs/Web/API/Location), but even though there is a spec, it is a big vague on slashes, and in the end it is still up to browsers to implement it correctly.

This has worked well, but mainly for being development only. Looking at how users will deploy and use the UI Server and the UI applications may give better idea on how to implement this.

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #343 

**Requirements check-list**

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -11,6 +11,7 @@
     <noscript>
       <strong>We're sorry but Cylc UI doesn't work properly without JavaScript enabled. Please enable it to continue.</strong>
     </noscript>
+    <!-- the python_base_url is set by Python Tornado template engine, but ignored in JS when used in the offline mode -->
     <div id="app" data-base-url="{{ python_base_url }}"></div>
     <!-- built files will be auto injected -->
   </body>
--- a/public/index.html
+++ b/public/index.html
@@ -11,7 +11,7 @@
     <noscript>
       <strong>We're sorry but Cylc UI doesn't work properly without JavaScript enabled. Please enable it to continue.</strong>
     </noscript>
-    <div id="app" base-url="/users/john"></div>
+    <div id="app" data-base-url="{{ python_base_url }}"></div>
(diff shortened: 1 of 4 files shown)


kinow:
Requires a change on the server side / backend to have the base URL populated.

kinow:
The e2e tests are failing due to a bug fixed in the Yarn PR. Not sure why it only appeared now :bug: 

The test failure is also affecting @oliver-sanders 's PR for isHeld.

public/index.html
```diff
@@ -11,7 +11,7 @@
     <noscript>
       <strong>We're sorry but Cylc UI doesn't work properly without JavaScript enabled. Please enable it to continue.</strong>
     </noscript>
-    <div id="app"></div>
+    <div id="app" data-base-url="{{ python_base_url }}"></div>
```
oliver-sanders, line 14: Where does `python_base_url` come from?

kinow: From Cylc UI Server. The `index.html` is rendered by Tornado. In the offline mode this value will be as-is, but it is ignored in JS.

kinow:
Rebased.

public/index.html
oliver-sanders: Aaah

kinow: Added a comment explaining the above :+1:

oliver-sanders:
Had a go at running:

```console
$ npm run serve
```

Untitled

PR #440, opened 2020-04-02
Fix mocked WorkflowService

kinow:

The e2e tests are currently broken on `master`. This PR should fix that :crossed_fingers: . To review, just wait for GitHub actions and confirm both workflows pass (there's one that runs only the e2e tests).

**Requirements check-list**

### DIFF
--- a/src/services/mock/workflow.service.mock.js
+++ b/src/services/mock/workflow.service.mock.js
@@ -22,7 +22,7 @@ class MockWorkflowService extends GQuery {
    */
   subscribe (view, query) {
     const id = super.subscribe(view, query)
-    this.subscriptions.every(s => {
+    this.subscriptions.forEach(s => {
       s.active = true
     })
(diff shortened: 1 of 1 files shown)


kinow:
@hjoliver if you have time to review this one, one review should do. Just depends on GitHub actions passing.

The function used was this one: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

And has been replaced by this one: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

With that, it effectively now sets each view as `active`. In the mocked service, the data is loaded immediately in the constructor, so there's nothing else pending to mark views as `active`.

e2e tests were failing as the new skeleton-loader was never displaying the tree data, as the view was not active (e2e tests use the offline mode too, using mocked data).

Thanks!
Bruno

PR #442, opened 2020-04-05
Fix mutations in the Toolbar component buttons

kinow:

@hjoliver noticed that the toolbar buttons were not calling the mutations any longer. It was due to some recent changes in the backend (for api-on-the-fly if I recall correctly, to simplify the schema).

@hjoliver I started by trying to add the `Mutation` component as a dropdown menu (as in the design sketches). That way we would move one step closer to what we have in the sketches, and use the same code used in the Mutations view (lumino widget).

Alas I started changing too many files, as I needed specific mutations (`hold`, `stop`, `release`) and these were within the `Workflow` view at the moment. After that gets moved to a place where it can be shared by components (Vuex store?), it should be easier to add the mutations to our views and components.

**Requirements check-list**
- [x] Does not need tests (why? manual tests should be enough? Soon this code will be scrapped by different UI/components, so even an e2e could be scrapped later too).
- [x] No change log entry required (why? e.g. invisible to users, i.e. was not broken in 0.1, should be the same in 0.2).

### DIFF
--- a/src/services/workflow.service.js
+++ b/src/services/workflow.service.js
@@ -6,24 +6,24 @@ import gql from 'graphql-tag'
 
 // TODO: remove these once the api-on-the-fly mutations are hooked into the UI
 const HOLD_WORKFLOW = gql`
-mutation HoldWorkflowMutation($workflow: String!) {
-  holdWorkflow (workflows: [$workflow]) {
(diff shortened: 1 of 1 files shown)


kinow:
image

Tested, and can also confirm the waiting badge/icon is being displayed :+1:

Issue #335, opened 2019-12-16
store user preferences


oliver-sanders:
I thought we had an issue for this but couldn't find it.

Store user preferences for the UI.

We've discussed this in the past, I think the conclusions were:

* Store user preferences on the filesystem seen by the UI Server.
* We may be able to use JupyterHub to achieve this.
* (Ideally this would be stored in a file which users could SCP to other locations.)

Examples of preferences:
* Job themes (e.g. colour-blind) -#9
* App themes (e.g. dark/light)
* Default view(s)
* View settings (e.g. graph view transposed by default)

**Pull requests welcome!**
<!-- This is an Open Source project - please consider contributing code yourself
(please read `CONTRIBUTING.md` before starting any work though). --

oliver-sanders:
[cylc-con-2020]

* store preferences at the hub
* look into implications of using the hub DB
* look into using hub “services” if the hub db is not a good idea

PR #446, opened 2020-04-20
Change font size

### LINKED ISSUE #141: adjustable font size
Somewhat related to #9 

Allow users to adjust font-size independently of theme.

We have users for which this is a major functional requirement.

kinow:
These changes close #141

There is a TODO about #335, but for now it's storing the new font size in the `LocalStorage` (i.e. client-side only).

Under the user profile view, now there is a "Preferences" section, with the Font Size.

- Vuetify appears to be setting the fixed font-size only on `<html>` element, and computing the sizes of other elements relative to that value (rem).
- I've adjusted our code to do the same, except by the Graph component that has some fixed values that I don't dare to change as it would probably cascade the amount of changes necessary and I am not sure if these could be easily fixed
- Reset sets the value from Vuetify sass variable back to the `<html>` element
- The button to increase font size does so with a ratio of 20% for every click, and stores the new value in the `LocalStorage`
- The button to decrease does so with a 20% ratio too (decay), and also stores new value in the `LocalStorage`
- When the application is loaded, it takes care of setting the `<html>` element font size back to what is in `LocalStorage`. If there is nothing in `LocalStorage`, then it does nothing and the sass variable value is applied as before

**Requirements check-list**

Will add changelog and tests (unit + e2e) later.

### DIFF
--- a/src/utils/font-size.js
+++ b/src/utils/font-size.js
@@ -1,0 +1,79 @@
+import styles from '@/styles/index.scss'
+
+// Module for font size functions and constants.
+
+/**
+ * Initial font size, set in the Vuetify framework style.
+ */
+const INITIAL_FONT_SIZE = styles.fontRootSize
+
+/**
+ * Sets the font-size back to its default value. The default value is imported from Vuetify's scss style
+ * variables (at the time of writing, variable $font-root-size, set by default to 16px).
+ *
+ * @param {string} [initialFontSize] initial font size, defaults to the Vuetify value
+ */
+function resetFontSize (initialFontSize = INITIAL_FONT_SIZE) {
+  localStorage.fontSize = initialFontSize
+  document.getElementsByTagName('html')[0].style.fontSize = initialFontSize
+}
+
+/**
+ * Decrease the HTML element font size by 20%.
+ */
+function decreaseFontSize () {
+  const currentFontSize = this.getCurrentFontSize()
+  const newFontSize = `${currentFontSize * 0.8}px`
+  localStorage.fontSize = newFontSize
+  document.getElementsByTagName('html')[0].style.fontSize = newFontSize
+}
+
+/**
+ * Increase the HTML element font size by 20%.
+ */
+function increaseFontSize () {
+  const currentFontSize = this.getCurrentFontSize()
+  const newFontSize = `${currentFontSize * 1.2}px`
+  localStorage.fontSize = newFontSize
+  document.getElementsByTagName('html')[0].style.fontSize = newFontSize
+}
+
+/**
+ * Get HTML element (computed) font size.
+ *
+ * @returns {number} current font size
+ */
+function getCurrentFontSize () {
+  if (localStorage.fontSize) {
+    return parseFloat(localStorage.fontSize)
+  }
+  const html = document.getElementsByTagName('html')[0]
+  const htmlStyle = window.getComputedStyle(html)
+  return parseFloat(htmlStyle.fontSize)
+}
+
+/**
+ * Computes the expected font size, for a given number of clicks on the
+ * increase or decrease button. The increase flag tells whether it was on
+ * the increase, or decrease button. Returns the exponential function
+ * initial_value * ratio ^ n, where n is the number of clicks. Intended
+ * for testing.
+ *
+ * @param {boolean} increase true means increase, false decrease
+ * @param {number} clicks number of times a button was clicked
+ * @param {number} [initialFontSize] initial font size, defaults to the Vuetify value
+ * @returns {number}
+ */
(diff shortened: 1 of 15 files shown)


kinow:
image

image

(There could still be some elements where the change is not applied, e.g. "This is a read-only view of your user" text at the table top seems to be using fixed size. I'll try to tackle some more as I find, but we can also push it to the backlog and fix as we find them I think.)

src/utils/font-size.js
```diff
@@ -0,0 +1,80 @@
+import styles from '@/styles/index.scss'
+// TODO: const styles = { fontRootSize: '16px' }
```
kinow, line 2: This is a blocker at the moment. Interestingly, it works with Vue + webpack. You can import `.js` and `.scss` files with no issues. Except that if you run this file with Cypress in the e2e tests, then it complains that it cannot file any tests in your file. With the commented code I was able to get it to run, but it would be preferable to get the value from Vuetify, in case that changes in future releases.

tests/e2e/specs/userprofile.js
```diff
@@ -9,4 +16,63 @@ describe('User Profile', () => {
       .should('be.visible')
       .should('be.disabled')
   })
+  it('Increases the font size', () => {
+    resetFontSize()
+    cy.visit('/#/user-profile')
+    expect(parseFloat(INITIAL_FONT_SIZE)).to.be.equal(getCurrentFontSize())
+    const clicks = 3
+    // TODO: https://github.com/cypress-io/cypress/issues/695#issuecomment-379817133
+    // Also had to use Promises in order to locate right element with Cypress
```
kinow, line 25: Note to self: double check

tests/unit/utils/font-size.spec.js
```diff
@@ -0,0 +1,62 @@
+import { expect } from 'chai'
+import sinon from 'sinon'
+import * as fontSize from '@/utils/font-size'
+
+if (!global.localStorage) {
+  global.localStorage = {
+    getItem () { return '{}' },
+    setItem () {}
+  }
+}
```
kinow, line 10: No `LocalStorage` with JSDOM, interesting too.

kinow:
Cypress was skipping the newly added tests. In `userprofile.js` spec, it used one of our JS modules that imported Vuetify's `scss` file (in JS, not css). That resulted in no errors in Cypress GUI or terminal, and Cypress reporting no tests found.

After searching GitHub and StackOverflow, found some people mentioning things like turning off optimization. Tried it, and that fixed the issue. Now tests being executed. Ran the whole e2e tests locally, everything worked fine.

The issue was similar to this one: https://github.com/cypress-io/cypress/issues/6159

tests/e2e/specs/userprofile.js
kinow: The promise approach actually fixed it, so we don't need it any more. But good to know about this hack, as it could be useful later (p.s. comparing with Selenium, Cypress needs very little hacking)

oliver-sanders:
Looks good, session storage is a good stepping-stone.

I've part-tested, need a more functional setup to go further. Looks like one element isn't responding, I guess this is because the content isn't wrapped in `<p/>`, `<span/>` or whatever:

Screenshot 2020-04-23 at 00 00 27

kinow:
> Looks good, session storage is a good stepping-stone.
> 
> I've part-tested, need a more functional setup to go further. Looks like one element isn't responding, I guess this is because the content isn't wrapped in `<p/>`, `<span/>` or whatever:
> 
> Screenshot 2020-04-23 at 00 00 27

Ah, good idea! I was thinking how to tackle that dangling piece of text that wouldn't resize.

First I wrapped it in a paragraph. But then the font-size applied was the `16px` from the vuetify variable (guess it's [applied not only to `html` element](https://github.com/vuetifyjs/vuetify/blob/4140af4c776aa8cb2dbd52ed7a05c44268094ec8/packages/vuetify/src/components/VAlert/VAlert.sass#L13)).

But then I remembered Vuetify also provides classes for each font-size available, and it's documented in their [typography section](https://v15.vuetifyjs.com/en/framework/typography); and `body-1` is the default size, but not fixed as `16px`, but instead set as `1rem !important` (that's what we get in the computed style).

Pushed one more commit that wraps every piece of text that is not already in a `h3` or another element that defines the font-size, using the `body-1` class. Now it appears to be working OK.

Thanks for the testing and good suggestion!

kinow:
>Might want to squash some commits before merging to keep our commit history a bit more readable.

Rebased, and moved + squashed some commits together.

hjoliver:
@kinow - I'll it to you to deconflict and merge.

kinow:
> @kinow - I'll it to you to deconflict and merge.

Sure, thank you Hilary!

kinow:
Rebased on `master`. Also did a quick smoke test using offline mode, no issues found.

PR #448, opened 2020-04-23
Update enumify to 2.x

kinow:

**Requirements check-list**

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,5 +1,8 @@
 module.exports = {
   presets: [
     '@vue/app'
+  ],
+  plugins: [
+    ['@babel/plugin-proposal-class-properties', { loose: true }]
   ]
 }
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "build": "vue-cli-service build",
     "build:report": "vue-cli-service build --report",
     "build:watch": "vue-cli-service build --watch --mode development",
-    "checkpoint" : "./src/services/mock/generate",
+    "checkpoint": "./src/services/mock/generate",
     "coverage:unit": "nyc vue-cli-service test:unit",
     "dev": "yarn run serve",
     "lint": "vue-cli-service lint",
@@ -32,7 +32,7 @@
     "cytoscape-undo-redo": "^1.3.1",
     "dedent": "^0.7.0",
     "dotparser": "^0.4.0",
-    "enumify": "^1.0.4",
+    "enumify": "^2.0.0",
     "graphql": "^14.5.8",
     "graphql-tag": "^2.10.1",
     "jshint": "^2.10.2",
@@ -50,6 +50,7 @@
     "vuex-router-sync": "^5.0.0"
   },
   "devDependencies": {
+    "@babel/plugin-proposal-class-properties": "^7.8.3",
     "@cypress/webpack-preprocessor": "^4.1.3",
     "@mdi/font": "^4.5.95",
     "@vue/cli-plugin-babel": "^3.12.0",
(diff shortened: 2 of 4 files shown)


kinow:
This was a small change that was complicated by a series of factors. First, the library was updated to TypeScript, and using static fields in classes, which needs a Babel plug-in to work (https://github.com/rauschma/enumify/issues/21). However, adding that plug-in broke the instrumentation we have in place for tests.

That's because Istanbul was configured to run before Babel. So the Babel plug-ins had no effect (https://github.com/webpack-contrib/istanbul-instrumenter-loader/issues/89).

Fixing the first problem was easy, just a matter of installing and loading the plug-in.

The second one took me longer than expected, but also taught me more about Babel, Istanbul, and our current setup with Vue 2.

In order to fix the second problem, of Istanbul not taking into consideration Babel plug-ins, first I found that the latest releases (3.x and 4.) of `istanbul-lib-instrument` used by `istanbul-instrumenter-loader` (we use this loader to instrument for coverage) works well with Babel plug-ins.

So I remembered reading about resolutions in the [Yarn docs](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/#toc-how-to-use-it). This is similar to how transitive dependencies can be overridden in Maven. We basically tell Yarn to ignore the dependency used by the library, and instead trust us and use the version we give it.

Once that's done, everything works.

This same procedure would have to be done if we ever needed to use other Babel plug-ins, so I think it should be OK to add it now.

babel.config.js
```diff
@@ -1,5 +1,8 @@
 module.exports = {
   presets: [
     '@vue/app'
+  ],
+  plugins: [
+    ['@babel/plugin-proposal-class-properties', { loose: true }]
```
kinow, line 6: So that we support static fields in classes with ES6. (`loose` so that it is less strict with how we define the static fields)

package.json
```diff
@@ -7,7 +7,7 @@
     "build": "vue-cli-service build",
     "build:report": "vue-cli-service build --report",
     "build:watch": "vue-cli-service build --watch --mode development",
-    "checkpoint" : "./src/services/mock/generate",
+    "checkpoint": "./src/services/mock/generate",
```
kinow, line 10: Done by `yarn add...`

package.json
```diff
@@ -89,6 +90,9 @@
     "vuetify-loader": "^1.3.0",
     "webpack": "^4.41.0"
   },
+  "resolutions": {
+    "istanbul-instrumenter-loader/**/istanbul-lib-instrument": "4.0.1"
+  },
```
kinow, line 95: Overriding the dependency of the loader, as the new Istanbul instrument library loads Babel plug-ins. Alas it seems like a release with the new version might take a while over there.

oliver-sanders:
😭

```
TypeError: Super expression must either be null or a function
```

kinow:
> sob
> 
> ```
> TypeError: Super expression must either be null or a function
> ```

Did you get that running `yarn run serve`? Did you run `yarn install` too? (Forgot to mention that would be necessary)

src/model/TaskState.model.js
```diff
@@ -1,25 +1,34 @@
...
-  'QUEUED',
-  'WAITING',
-  'HELD',
-  'SUCCEEDED',
-  'RUNAHEAD'
-])
+class TaskState extends Enumify {
+  // NOTE: the order of the enum values is important to calculate the group states in the UI. Items at
+  // the top of the list have preference over the items below in the algorithm.
+  static SUBMIT_FAILED = new TaskState('SUBMIT_FAILED')
```
oliver-sanders, line 9: class TaskState has attribute which are TaskState instances, very meta!

PR #451, opened 2020-04-28
Copyright headers

kinow:

Copyright headers added. I wasn't sure where to add copyright to `.vue` files, then found out that @oliver-sanders had used simple HTML comments in some components, so used same approach, which appears to work fine.

**Requirements check-list**

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,20 @@
+/**
+ * Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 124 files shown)


kinow:
Doing some local testing, then once it's working OK locally, and GitHub actions are good too, will mark as ready for review.

kinow:
Ready for review @hjoliver . As we had no copyright headers, I started by copying the README entry from cylc-uiserver.

Then went with the IDE on each folder/file marked as source folder (in PyCharm other folders and filers get ignored, like `htmlcov`, those `.egginfo` folders, etc). Each file changed got a blue colour (again, from the IDE that marks whenever I change a file).

Went from top to bottom doing so, adding either a header with `/* */` for CSS and JS, or the simple HTML comment `<!-- -->` for HTML and .vue files. Once done, went bottom top reviewing.

Did a `yarn build` that found a couple problems from copy-and-paste, then once that passed, tried `yarn serve`. All good so far.

PR #454, opened 2020-05-04
Make sure subscriptions / queries are removed

### LINKED ISSUE #453: Confirm we don't have subscriptions outliving the views/use, to avoid performance issues, and add unit/e2e tests

The Views and Components used in Cylc UI tell the service about their queries. When the user navigates to another View, the queries are updated.

This update (merge) must add new fields from the queries, as well as remove fields from old queries that are no longer used (e.g. user goes from dashboard to tree view, then back to dashboard; at this point there is no need for the tree view query).

I found some cases where the query appears to be left behind after the views are no longer active. That happens because the views are using `beforeDestroy`, but they are rarely destroyed (for performance).

The PR #458 had a temporary version that fixed it. It was simply tethering the subscription to the view lifecycle. It has the disadvantage that when users switch quickly between views, they may have to re-request the whole data. So one possible scenario is keeping the subscriptions open for a while.

One way or the other, we need to address this, as otherwise using the application a few hours, and switching between views and workflows would certainly cause serious performance issues in the browser and in the server due to extra GraphQL queries running unnecessarily. 

A functional test should do the following:

- [ ] Navigate to some URL or URL(s) as per scenario (e.g. dashboard -> tree, or tree -> dashboad -> user profile, etc)
- [ ] For each scenario, it must confirm the service got the right views or components registered, that the subscribed queries are correct, and that the final query contains the expected fields (can use `graphql-tag` for that if necessary)

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #453 

I think with this the subscriptions (and thus the final merged query) should be correct after the user has left the views modified here.

But will spend some more time writing some e2e tests as this issue has already happened before.

**Requirements check-list**

### DIFF
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -246,8 +246,10 @@ export default {
     )
     this.subscribe('root')
   },
-  beforeDestroy () {
+  beforeRouteLeave (to, from, next) {
+    this.unsubscribe('root')
     this.$workflowService.unregister(this)
+    next()
   },
   methods: {
     subscribe (queryName) {
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -169,7 +169,6 @@ export default {
   data () {
     return {
       viewID: '',
-      subscriptions: {},
       isLoading: true,
       workflowsHeader: [
         {
@@ -247,7 +246,6 @@ export default {
     this.subscribe('root')
   },
   beforeRouteLeave (to, from, next) {
-    this.unsubscribe('root')
     this.$workflowService.unregister(this)
     next()
   },
@@ -257,25 +255,10 @@ export default {
        * Subscribe this view to a new GraphQL query.
        * @param {string} queryName - Must be in QUERIES.
        */
-      if (!(queryName in this.subscriptions)) {
-        this.subscriptions[queryName] =
-          this.$workflowService.subscribe(
-            this,
-            QUERIES[queryName].replace('WORKFLOW_ID', this.workflowName)
-          )
-      }
-    },
-
-    unsubscribe (queryName) {
-      /**
-       * Unsubscribe this view to a new GraphQL query.
-       * @param {string} queryName - Must be in QUERIES.
-       */
-      if (queryName in this.subscriptions) {
-        this.$workflowService.unsubscribe(
-          this.subscriptions[queryName]
-        )
-      }
+      this.$workflowService.subscribe(
+        this,
+        QUERIES[queryName].replace('WORKFLOW_ID', this.workflowName)
+      )
     },
 
     setActive (isActive) {
--- a/src/views/Workflow.vue
+++ b/src/views/Workflow.vue
@@ -56,7 +56,10 @@ export default {
     }
   },
   data: () => ({
-    subscriptions: {},
+    /**
+     * @type {Map<number, Object>}
+     */
+    subscriptions: new Map(),
     isLoading: true
   }),
   computed: {
@@ -88,7 +91,7 @@ export default {
     })
     EventBus.$on('delete:widget', (data) => {
       const subscriptionId = Number.parseFloat(data.id)
-      this.$workflowService.unsubscribe(subscriptionId)
+      this.unsubscribe(subscriptionId)
     })
   },
   beforeRouteEnter (to, from, next) {
@@ -109,6 +112,9 @@ export default {
     EventBus.$off('add:graph')
     EventBus.$off('add:mutations')
     EventBus.$off('delete:widget')
+    Array.from(this.subscriptions.keys()).forEach((subscriptionId) => {
+      this.unsubscribe(subscriptionId)
+    })
     this.$workflowService.unregister(this)
     next()
   },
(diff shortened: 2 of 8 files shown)


src/views/Dashboard.vue
```diff
@@ -246,8 +246,10 @@ export default {
     )
     this.subscribe('root')
   },
-  beforeDestroy () {
+  beforeRouteLeave (to, from, next) {
```
kinow, line 249: This is necessary, as the View is not just a component, but bound to a Vue Router router. So the lifecycle is different. Views don't get destroyed as Vue components do. So the `beforeDestroy` doesn't always get called.

kinow:
Ha! Wrote the first most basic e2e tests:

- User goes to Dashboard :heavy_check_mark: 
- User goes to Dashboard, then to Workflows view :heavy_check_mark: 
- User goes to Dashboard, then to Workflows view, then back to Dashboard :bug: 

In the third scenario instead of two subscriptions (i.e. two components or views subscribed), in my local environment it got **5** :open_mouth:

kinow:
Tests written for the number of subscriptions. Just need to improve the test. Instead of looking only at the number of subscriptions, it must make sure that the subscriptions present are the ones you expect.

For example, the third scenario again.

Dashboard -> Workflow -> Dashboard.

In this case, the GraphQL query that was merged, i.e. the final query, must **not** contain `familyProxies` nor `cyclePoints`, as these are not used by GScan or the Dashboard. After that this should be ready to review.

kinow:
Ready for review.

Testing the following scenarios:

- [x] User goes to the main page. We must have 2 (gscan/dashboard) subscriptions, and a query without cyclePoints. (i.e. GraphQL response doesn't have cyclePoints alias for familyProxies as in the tree query.)
- [x] User goes to the main page, then to the workflows page. We must have 2 (gscan/tree) subscriptions, and a query **with** cyclePoints.
- [x] User goes to the main page, then to the workflows page, then back to the dashboard. We must have 2 (gscan/dashboard) subscriptions, and a query without cyclePoints.
- [x] User goes to the tree page. We must have 2 (gscan/tree) subscriptions, and a query **with** cyclePoints.
- [x] User goes to the tree page, then back to the dashboard. We must have 2 (gscan/dashboard) subscriptions, and a query without cyclePoints.
- [x] User goes to the main page, then to the user profile. We must have **1** subscription (gscan), and a query without cyclePoints.

I used the cyclePoints as at the moment the tree view is the only one that uses it. Likewise, one could write similar tests later for the Graph view/component using `nodesEdges`. I didn't test the Graph component because at the moment I'm focusing on the deltas with the tree view, but plannign on doing so when we restart work on the graph component.

kinow:
What I learned with this pull request:

- It is tempting to add `cy.wait(500)` or similar, to wait a bit while a Vue router is loaded, **but that can easily create flaky tests**
- Instead it's a bit more of work, but you can use `cy.get($selector)` and `cy.contain($htmlElem, $text)` to use Cypress' default mechanisms to wait until elements are displayed
- When doing so, you may also accidentally assume certain components mean the view and data store are ready, while it's not correct. I assumed I could use the Toolbar (displayed only when you select a workflow) would be a good idea, but Vue loads each component in the view asynchronously, with no guarantee of order, and we have some components and views triggering different actions. See screenshot below from Cypress runner, where the Toolbar is briefly displayed with the Dashboard

image

- We can expose the Vue app to the JS world via the global `window` object, which is handy for testing/development, but a no-no for production (that's part of this PR)

src/views/Dashboard.vue
```diff
@@ -246,8 +246,10 @@ export default {
...
+    this.unsubscribe('root')
```
oliver-sanders, line 250: Surely this should be done by workflowSevice.unregister?

kinow: Ah, all this time I thought I needed both. Looks like the correct is

- Component/view wants to use the service:
  * `register()`
  * `subscribe()`
- Component is being destroyed or doesn't need the data anymore (e.g. a lumino widget removed)
  * `unregister` (which will do the same as `unsubscribe()`)

@oliver-sanders do we need the `unsubscribe methods`? Both in `GQuery.js`, and in all the views? Could we perhaps combine register/subscription in a single method? 

Updating the PR to use only `unregister`.

kinow: Done. I removed one unsubscribe call but left only some unsubscribe calls in `Workflow.vue` view. The way I built it, when you create that view, it "registers" the query. Then each time you create a widget it is "subscribed". Then when you remove the widget it just "unsubscribe"s. Then, when you leave the page it now only "unregisters".

Changing that here would increase considerably the amount of changes in this PR. But we can review it later (the goal here is just making sure we have the right amount of subscriptions when you navigate)

kinow:
Ready for review again.

src/views/Workflow.vue
```diff
@@ -136,20 +142,20 @@ export default {
...
      * Unsubscribe this view to a new GraphQL query.
-     * @param {string} queryName - Must be in QUERIES.
      * @param {number} subscriptionId - Subscription ID.
      */
-    unsubscribe (queryName, subscriptionId) {
-      if (queryName in this.subscriptions) {
+    unsubscribe (subscriptionId) {
+      if (this.subscriptions.has(subscriptionId)) {
         this.$workflowService.unsubscribe(subscriptionId)
+        this.subscriptions.delete(subscriptionId)
```
oliver-sanders, line 158: Looks good, though views may request multiple subscriptions.

hjoliver: Are you suggesting a change here?

kinow:
Note to self: this change came from #458 . Once/if this gets merged, I need to drop a commit there (otherwise I would have an extra subscription that was getting in the way)

oliver-sanders:
Code looks good, the only issue I see is what happens when subscriptions change.

The old system, though inelegant allowed views to change their subscription whilst running. Some reasons why subscriptions will have to change on the fly:

* User requests historical data.
* User requests extended data (e.g. expands view).
* User changes the query depth (e.g. from `n=1` to `n=2`).
* User changes the scope of a view (e.g. adding or removing workflows from a view).

kinow:
>Code looks good, the only issue I see is what happens when subscriptions change.

Good point :+1: 

Maybe we can create an issue to either bring back some of the old code, or work on an improvement later?

At least the part about starting, stopping/removing the subscriptions would stay anyway.

kinow:
Superseded by #458

Issue #416, opened 2020-02-27
WorkflowService/GQuery fetches more data than what it needs when GScan + a view/component are used


kinow:
**Describe the bug**

I was looking at the Vuex data store today, and at the JS console in the browser, when I noticed that the query for my tree view was bringing more data than necessary.

When you browse the `/#/dashboard`, we have - at the moment - two queries being merged, one for the `GScan` component, and another for the `Dashboard` view. The resulting query looks like:

```graphql
    subscription {
      workflows {
        id
        name
        status
        owner
        host
        port
        taskProxies(sort: {keys: ["cyclePoint"]}) {
          id
          name
          state
          cyclePoint
          latestMessage
          task {
            meanElapsedTime
            name
          }
          jobs(sort: {keys: ["submit_num"], reverse: true}) {
            id
            batchSysName
            batchSysJobId
            host
            startedTime
            submittedTime
            finishedTime
            state
            submitNum
          }
        }
      }
    }
```

That will bring the same data for all the workflows you may be running. (We need a few more changes in the backend to simplify this query more - see #401.) 

Now when you navigate to the Tree view, you will have two queries, one for the GScan component, and one for the Tree component.

The resulting query looks like:

```graphql
subscription {
      workflows {
        id
        name
        status
        owner
        host
        port
        taskProxies(sort: {keys: ["cyclePoint"]}) {
          id
          name
          state
          cyclePoint
          latestMessage
          task {
            meanElapsedTime
            name
          }
          jobs(sort: {keys: ["submit_num"], reverse: true}) {
            id
            batchSysName
            batchSysJobId
            host
            startedTime
            submittedTime
            finishedTime
            state
            submitNum
          }
          firstParent {
            id
            name
            cyclePoint
            state
          }
        }
        familyProxies(sort: {keys: ["firstParent"]}) {
          id
          name
          state
          cyclePoint
          firstParent {
            id
            name
            cyclePoint
            state
          }
        }
      }
    }
```

So after merging it, the query will bring the complete information for the tree view of all the workflows available.

I **think** it would be better to have the Tree View information (family proxies, cycle points, mode details about jobs, etc) data just for the workflow currently being displayed. And the GScan data (workflow info and preferably no jobs/tasks info after #401) for all the workflows available.

**Release version(s) and/or repository branch(es) affected?**

`master`

**Steps to reproduce the bug**

[...]

PR #457, opened 2020-05-12
Add GraphiQL view

### LINKED ISSUE #352: GraphiQL view
We can already access GraphiQL via the UIServer, should be easy to roll this into the UI?

**Pull requests welcome!**

### LINKED ISSUE #456: Render GraphQL error message in GraphiQL instead of object default string output

From: https://github.com/cylc/cylc-flow/pull/3500#issuecomment-627206304

We can just tweak the code in the current template. It shouldn't be too hard to display the error message. 

This saves developers/users time, as they won't have to look at the browser WebSocket messages for the error (pretty much only place this info is available).
**Additional context**

**Pull requests welcome!**

kinow:
Close #456 
Related to #352

**Requirements check-list**
- [x] Does not need tests (why? adding new view for third-party lib/component).


package.json
```diff
@@ -34,10 +34,13 @@
     "dedent": "^0.7.0",
     "dotparser": "^0.4.0",
     "enumify": "^2.0.0",
+    "graphiql": "^0.17.5",
     "graphql": "^14.5.8",
     "graphql-tag": "^2.10.1",
     "jshint": "^2.10.2",
     "nprogress": "^0.2.0",
+    "react": "^15.6.2",
+    "react-dom": "^15.6.2",
```
kinow, line 43: Looks weird to have this in a Vue app, but we wouldn't be the first ones to try

- https://github.com/vuejs/Discussion/issues/1301

And actually works. The generated app JS size will increase, but for the user it should still be loaded dynamically when the GraphiQL view is accessed.

The same already happened anyway, and via a CDN, instead of local files. So now GraphiQL should work even if accessed in an environment without external connectivity (related to #386).

src/components/cylc/Drawer.vue
```diff
@@ -51,6 +51,16 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         </v-list-item-action>
         <v-list-item-title>Dashboard</v-list-item-title>
       </v-list-item>
+      <v-list-item
+          to="/graphiql"
+          active-class="primary grey--text text--darken-3"
+          class="v-list-item"
+      >
+        <v-list-item-action>
+          <v-icon>mdi-graphql</v-icon>
```
kinow, line 60: Looks like a graph, not sure if we could use this in some other place.

Also, I placed the link under Dashboard, but perhaps it should go somewhere else?

src/services/workflow.service.js
```diff
@@ -54,6 +54,7 @@ class SubscriptionWorkflowService extends GQuery {
    */
   constructor (httpUrl, subscriptionClient) {
     super()
+    this.subscriptionClient = subscriptionClient
```
kinow, line 57: Needed this so that we could pass the same instance to the GraphiQL fetcher.

Tried to access the object from the instance of `$workflowService`, but it's hidden deep in another object. So the easiest approach was this one.

src/graphql/graphiql.js
```diff
@@ -0,0 +1,130 @@
...
+            query: graphQLParams.query,
+            variables: graphQLParams.variables
+          }, function (error, result) {
+            if (error) {
+              observer.error(error)
+            } else {
+              observer.next(result)
+            }
+          })
+          activeSubscription = subscription.subscribe((result, err) => {
```
kinow, line 81: This is what changed. `.subscribe` here is **not** from GraphQL. This is the method of a JS subscription, that will create an Observer. Later we are destroying it, as leaving Observers hanging around leads to memory leaks.

kinow:
What it looks like:

image

Another alternative is using the `empty` layout, which looks more like what we have, but doesn't have links to return.

image

kinow:
Notes: 

- This PR uses the latest version of `graphiql` JS module. This is the module responsible for rendering the application (which is a React component)
- The latest version of `graphiql` takes care to render the JSON returned in case of an error, instead of printing "[Object object]", which means you only have to look at the WebSocket messages if you really want/need to

image

- In this version of `graphiql`, they changed how the button to run queries work. When you have a subscription, it will change from a "Play" button to a "Stop". Meaning that you can stop it, freezing the output. I found it super useful for troubleshooting deltas, as I was taking screenshots or trying to copy the JSON response when I saw something wrong.
- If you do not stop the subscription and go to another view, Vue Router callback will take care to stop it for you :+1: 

(I really liked seeing GScan with the output of my deltas to compare the tasks/jobs listed there)

kinow:
Oh, one final note: if merged, we have the option of removing GraphiQL from Cylc UI Server. The UI Server has the logic to handle GraphiQL responses. But at the moment, it also has a template rendering the React App. That version is using older JS libraries, that are in a string (so no code analysis on that block of code).

hjoliver:
I'll come back to this after spending more time on the delta PRs ... but this looks really nice :+1:  (We might just need to agree exactly where to provide access and/or to hide from non-developers? ... but very cool and useful)

hjoliver:
I'd be happy for this to go in as it's useful, and we have previously agreed that several "developer views" including a graphiql one be worthwhile.  We can worry about whether or not users should see it later.  What do you think @oliver-sanders ?

hjoliver:
@kinow - would it be possible to treat this like our other workflow views (so it contributes to and is then driven by the merged single subscription)?  I'm guessing not as the component is presumably designed to manage its own subscription ... but just wondering.

kinow:
> @kinow - would it be possible to treat this like our other workflow views (so it contributes to and is then driven by the merged single subscription)? I'm guessing not as the component is presumably designed to manage its own subscription ... but just wondering.

Oh, interesting idea! With a little work we can do that in the future I think.

At the moment I think it could cause some confusion to other views, as the merging is not without risk. e.g. if you send a really complex query, and that gets merged, and say you have 100 workflows running, then all the 100 workflows **could** get some of the merged query because of #416 

But it could really useful if I could modify the query of a view dynamically, with a GraphiQL editor/view! We should add it to the backlog if/after this gets merged.

src/components/cylc/Drawer.vue
oliver-sanders: Good question about where to put it.

I would have thought it would be best implemented as a cylc "view", even though graphiql isn't actually a view of a single workflow we could pre-populate the query with:

```
query {
    workflows (ids: [<workflowid>] {
        id
    }
}
```

hjoliver: That's what I suggested below too, and this pre-populate idea seems an easy way to make that work (in terms of displaying the right workflow, if not being driven by the common subscription).

kinow: That might be doable. But first I would prefer to replace the current GraphiQL from UIS by this one.

What about:

- create `/$user/#/graphiql` view that shows GraphiQL in full screen
- remove GraphiQL from UIS
- remove "Close" and change by "Related to #352" from this issue
- create follow up for #352 to component-ize the GraphiQL, where we can discuss where to add it, how to load data, etc?

That way we get the nice error messages and the updated GraphiQL in Cylc UI.

oliver-sanders:
I think merging this would be pretty darned difficult, as it's really intended as a developer view the increased load shouldn't be important but you never know, users might actually start messing with queries...

kinow:
> I think merging this would be pretty darned difficult, as it's really intended as a developer view the increased load shouldn't be important but you never know, users might actually start messing with queries...

What about the alternative I mentioned before? Of just moving the JS code here to the UI, and removing the long Python string with a String template? That would at least 

- allow us to maintain that JS code updating libraries, linting, etc
- use latest graphiql library with the stop button
- decide later whether we want to add a view, maybe with more security, etc?

It would be a view, showing the graphiql in the whole page, just like the old one. Only difference that it would be rendered here.

oliver-sanders:
> moving the JS code here

Fine by me!

dwsutherland:
Maybe a link in the dashboard/home? And yes, moving the JS here is well worth it.

oliver-sanders:
> (P.S. I don't see an issue with users playing with GraphiQL, just doesn't need to be as prominent I suppose)

No problem at all, for our more advanced users I think this might be a pretty good feature.

When it comes to visibility we could probably handle this in the "Add View" context menu by putting Tree, Graph and Dot prominently at the top and having GraphQL, Queues, and other more technical views after a divider.

Alternatively we could do what some browsers do and hide developer options unless enabled for the benefit of simplicity.

Anyhoo an issue for later.

kinow:
PR updated.

Now it doesn't need Vuex (kept the subscription state in the GraphiQL view only).

The view link is not present in the menu, but users can navigate to http://localhost:8000/user/$USER/#/graphiql to visualize it. It queries exactly the same endpoint as the existing graphiql UI in the Cylc UI Server.

What it looks like:

image

The Play/Stop button is the only difference in the interface from what we have now I think. And a great feature that would help me a lot for the deltas tests. As I have had to copy messages or take screenshots to explain when I had issues or needed some help with deltas for the UI.

And the other improvement is that if you have an error, you won't receive the "[Object Object]" message. Instead the error is displayed in the UI.

image

I will prepare a draft PR for Cylc UI Server for after this gets merged :+1: 

Ready for review.
Thanks

kinow:
>In the meantime though, can we stick a temporary link in the Dashboard, just to avoid having to remember the URL?

Super easy to add this. Just wait to see if @oliver-sanders approves this too :+1: (in which case, I would probably just add it under Dashboard, unless someone has a suggestion as the location of the link, like below GScan items, at the very bottom near version, etc)

oliver-sanders:
We can add it under dashboard for now, it really shouldn't be there when we release 8.0.0 but that's a way off...

kinow:
image

Done, link added under dashboard.

oliver-sanders:
Approved with follow-on to turn this into a formal Cylc "view".

kinow:
> Approved with follow-on to turn this into a formal Cylc "view".

:+1: #352

PR #450, opened 2020-04-28
Use GraphQL subscription deltas in Cylc UI

kinow:
These changes close #420

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [ ] Appropriate tests are included (unit and/or functional).
- [ ] Appropriate change log entry included.
- [x] No documentation update required.

Issue #439, opened 2020-04-02
Use query fragments instead of strings to merge/combine queries


kinow:
**Describe exactly what you would like to see in an upcoming release**

Investigate using [`graphql-tag`](https://github.com/apollographql/graphql-tag) to merge queries.

That would allow us to use any sorting keys (#300), use variables (#234), and use `graphql-tag` imports instead of strings (#356).

And maybe with more work we can fix #209 and #417 too.

**Additional context**

i. Queries will have the same operation type `subscription`, and will/must share the [root level field](https://spec.graphql.org/June2018/#sec-Single-root-field). The root level will be `workflow`. For example:

```graphql
subscription A {
  workflow { id }
}

subscription B {
  workflow { nodesEdges { nodes { id } } }
}

subscription AB {
  workflow {
    id
    nodesEdges {
      nodes {
        id
      }
    }
  }
}
```

ii. Variables **must** match. Meaning, too, that we are not merging a query like GScan's, that doesn't have any variables, with a query that is limited for a single workflow like the Tree view query. So the following example will result in an error.

```graphql
subscription TreeView ($workflowID: ID) {
  workflows (ids: [$workflowID]) {
    id
    jobs { id }
  }
}

subscription GScan {
  workflows {
    host
    port
  }
}

// Otherwise if we had 500 workflows in Cylc, that would result in the query above returning the host+port **and** the jobs for all the 500 workflows; which was a bug in the previous implementation
// We want the 500 workflows for GScan to display int he sidebar, but only 1 workflow with the jobs data
// So, we don't want to merge these two queries
```

iii. Queries must reuse/share fragments. So that we can define what is a `Job` and use it in a Tree view query, or Graph view query, etc (this is partially done already).

**Pull requests welcome!**

kinow:
## Would it work with Cylc UI (i.e. Vue.js, ApolloClient, Babel, Webpack, etc, etc?)

First PoC, to confirm it works with Cylc UI.

https://github.com/kinow/cylc-ui/compare/master..graphql-tag-1

![GIFrecord_2020-04-03_141041](https://user-images.githubusercontent.com/304786/78314025-27791b00-75b5-11ea-8cb1-b63b67d6a58a.gif)

(disabled GScan as otherwise it would try to merge `string`+`gql`.)

So instead of parsing `String` to JSON, and then merging objects in JS, it would instead assume views want to subscribe a `Query` or `Subscription`, e.g. for the Tree View specific view (not lumino)

```js
[...]

Issue #417, opened 2020-02-28
WorkflowService/GQuery will merge subscription+query as a subscription


kinow:
**Describe the bug**

Not sure if really a bug, or more like a user error (probably the latter @oliver-sanders ? I think I might be doing something wrong when calling it) but if you have the following scenario:

- `src/components/GScan` has a `subscription { workflows ... }`
- `src/views/Tree` has a `query { workflows ... }`

The final result is: `subscription { workflows }` (I think it keeps the left side, and updates with different values from the right side of the operation?)

The issue happened while I was testing ApolloClient with the new deltas in @dwsutherland 's branch. I changed from `subscription` to `query` in the `Tree` view, and was expecting to see an HTTP request.

But instead ended up with a new operation in the WebSocket channel. Debugging it I noticed that the final query had been merged as subscription.

**Release version(s) and/or repository branch(es) affected?**

**Steps to reproduce the bug**

**Expected behavior**

Not sure what is the expected behavior here. I would expect to be able to use queries and subscriptions. Maybe the service needs two different methods, one to request data, the other to subscribe?

Or maybe just one to request data, with an extra option to also subscribe to updates with a subscription (that's what ApolloClient does I guess)

**Screenshots**

**Additional context**

I was changing it to experiment with the pattern from Apollo Client documentation, where the client asks for the large data in an HTTP request, then subscribes for updates via WebSocket.

![1*KwnMO21k0d3UbyKWnlbeJg](https://user-images.githubusercontent.com/304786/75502728-97d1d100-5a38-11ea-831a-d4dc68ebee0a.png)

My understanding is that it's better to avoid initiating a subscription with 1 message for a query. i.e. if you run a subscription with `query { workflows }`, it works, but you get all the tooling for a subscription, to get just 1 query response. The HTTP query is cheaper/faster, as it skips the operations/queues/extra syncio coroutines/etc and just goes straight to the resolver/schema.

**Pull requests welcome!**
This is an Open Source project - please consider contributing a bug fix
yourself (please read `CONTRIBUTING.md` before starting any work though).

kinow:
Not a blocker, as for now I've simply renamed everything to `query {}`. And am experimenting with the `subscribeToMore` to create a subscription for the query of the `Tree` view, handling deltas.

oliver-sanders:
Kinda user error in that we merge things we want to subscribe to, if we want a one-off query then we will require a different interface.

hjoliver:
[...]

PR #458, opened 2020-05-13
Add CylcTree and use Deltas

### LINKED ISSUE #420: Incremental updates with GraphQL deltas

kinow:
Closes #420 
Related to #453 

Related to previous (now closed) PR #450 

**CylcTree**

This branch contains the work derived from #450. Mainly the data structure `CylcTree`, which is simply an array+hierarchy, with a helper map for quick access of elements in the tree (useful to quickly add/remove with deltas).

**Requirements check-list**


kinow:
The current code produces:

image

That's in the Tree view. The data is built on the first response from GraphQL endpoint, using the initial burst of data to build the workflow hierarchical structure (i.e. workflow, with cyclepoints as children, with families as children, etc etc).

**It is not using the WorkflowService**. I tried using that in the other branch, and found some issues with how we were maintaining the state between Vue views/routes (fixed already).

And also had issues around query merging.

The query now is a `subscription { deltas {...} }` with information specific to the view. It could be merged with others I think, but due to some issues (#300, #417, and specially #416) I preferred to focus on the deltas.

I am also using query variables (see #234) and fragments as otherwise the query was getting really long and with repeated parts in the deltas (see #439).

kinow:
Next step: apply the deltas to the `CylcTree` object. This object was made so that deltas would be easier to be applied. :crossed_fingers:

kinow:
Alright, tests passing in GH actions now. Commits are organized. Pending here now:

- [ ] test and report here & Riot performance with the deltas PR's
- [ ] add unit tests for `CylcTree`
- [ ] discuss with @oliver-sanders and others about the changes here (especially not using the `WorkflowService` and using query fragments)
- [ ] add deltas to the `Workflow` component, having with that the tree component & views using deltas, unblocking work on Inifinite Tree #345 (depends on discussion in previous item)

kinow:
The `deltas { shutdown }` is received after a workflow is shut down. But you may still see delta messages after that.

image

I guess that could be the left-over in the channel/queue. So safe to just ignore these messages.

kinow:
Environment: Ubuntu LTS, browser firefox in background with tab opened
with the GitHub issue. Chromium running without the Vue dev tools extension
at http://localhost:8000/user/kinow/#/tree/${workflowName}, and memory tabs
open in dev console. Branches synced for deltas.

## branch add-cylc-tree-ds

### five

This branch with no workflows, consumes between 10 and 12 MB.

Starting `five`, the memory slowly started to increase, reaching 16MB,
and then dropped back to 11MB (probably a GC) in less than 2 minutes.
That happened again in about 1 minute, and continued repeating until
the test completed 5 minutes, never reaching 17MB.

Stopped workflow, waited a couple minutes, then pressed the GC button.
Memory usage went back to 10.7MB.

The workflow is quite simple, so it is not easy to spot nay performance
regressions. Below a screenshot of 1 minute profiling the page. It's possible
to see a consistency in the times the browser had to paint again (elements
added/removed/position changed). Most of the time it was idle, with very little
time in scripting and redering, excellent results.

image


### complex

The complex workflow started with 28 MB. Going up to 32 MB, then back to 28 MB.
Once it got around 39.6 MB, then went back to 26.5 MB. All in less than 3
minutes. That repeated a few times until the test completed 5 minutes. Without
ever reaching 40 MB.

Stopped the workflow, waited a couple minutes, then pressed the GC button.
Memory usage went back to 13.8 MB.

Below is the screenshot of 2 minutes profiling with the workflow. The workflow
is not a consistent as `five`. There are cycles where more families, tasks,
cycle points, or jobs are added/removed. You can see the blue graph with
the memory usage, going up and down. That's good, as that's the memory usage
going up, elements removed, and then GC acting.

image

Panting didn't happen a lot. And ~3 seconds scripting, and <200 ms rendering
for 120 seconds is not bad.

After this test, I did another test now scrolling the page, expanding/collpasing
items.

Memory kept under 40MB, and the app was responsive with no issues.

oliver-sanders:
Does this branch include the infinite tree work or is that to follow later?

src/graphql/queries.js
```diff
@@ -81,6 +86,116 @@ subscription {
...
+    }
+    pruned {
+      jobs
+      taskProxies
+      familyProxies
+    }
+  }
+}
+
+fragment WorkflowData on Workflow {
```
oliver-sanders, line 142: Nice way to avoid writing this out twice.

Could we use the above template for all subscriptions by just varying the fragments or does GraphQL not allow you to have empty fragments?

kinow:
>Does this branch include the infinite tree work or is that to follow later?

That will come later. Probably first simply doing something like `new InfiniteTree(cylcTree.root.children)`. Then another follow-up would be to investigate using the `InfiniteTree` directly, without intermediary data structure, though a bit harder.

src/graphql/queries.js
kinow: We could use string-replacement to build up the query+fragments. That's what I am working now (more thinking and reading than coding), for https://github.com/cylc/cylc-ui/issues/439

kinow:
Doing good progress on fragments. Mainly using GraphiQL and a bit of JS in the firefox console to play with ApolloClient.

I am thinking possible alternatives for this branch so that we can use deltas in the rest of the system. Some ideas end up with issues in how we use the ApolloClient and queries in Cylc UI. @dwsutherland helped me with some questions about GraphQL today on Riot (thanks again!).

So far the way forward appears to be similar to what we have here: 1 WebSocket connection, but 2 subscriptions.

One subscription belonging to GScan + Dashboard data. And another subscription for the Tree data.

The second subscription would exist only on Views that display workflow information, such as `/#/workflow/five` or `/#/tree/five` or `/#/graph/five`. Other views would continue to have a single subscription.

kinow:
Approach I am trying in the next days: https://gist.github.com/kinow/d505938ccb5f3dfe87c585ddb333956a

Use the Subscription from the Gist above. Instead of using the two fragments in the subscription, use it as a string template, i.e.

```js
export const WORKFLOW_SUBSCRIPTION = `
subscription OnWorkflowDeltasData ($workflowId: ID) {
  deltas (workflows: [$workflowId]) {
    $FRAGMENTS
  }
}
`
```

Then replace `$FRAGMENTS` by either Tree fragment, or Graph fragment, or both. Depending on what the user requested.

On the Tree view, `/#/tree/five`, it will be just the Tree fragment when the user navigates. When the user leaves the page, we stop the subscription.

On the Graph view, `/#/graph/five`, it will be just the Graph fragment. Same as above.

On the Workflow view, `/#/workflow/five`, it will be none. However, when the user navigates to that view, by default we are adding a Tree widget. So that will add the Tree fragment.

If the user adds another Tree widget, we check whether there is already a Tree fragment. If true, ignore. Otherwise we add the tree fragment.

If the user adds a Graph widget, we check whether there is already a Graph fragment. If true, ignore. Otherwise we add the graph fragment to the subscription.

When the user leaves the page, we stop the subscription.

With this I believe we would have deltas working in the UI and ready for review. Users would have 1 extra subscription, but that appears to be the only way, as I was not able to have a `subscription { workflows () {} deltas () {}` as that results in errors in GraphQL.

kinow:
Build might stay broken for a little while until I finish experimenting with query fragments in the WorkflowService :+1:

kinow:
Deltas working on the Tree view, and on the Workflow view, for the Tree component.

However, at the moment it is blocked by #454 , as I'm getting more and more subscriptions added. And adding/removing tree widgets, the UI is getting confused. That happens because the deltas are always applied, but if you have removed the widgets, then you must stop the subscription (hence #454).

Otherwise it will keep trying to update the tree, just to find it `undefined` as the widget is gone.

kinow:
Hmm, or maybe I should just merge that PR with this one, and rebase later if that one is merged first. :thought_balloon:

oliver-sanders:
Sorry @kinow that PR slipped through the gaps, will try to get in soon.

src/services/workflow.service.js
```diff
@@ -19,32 +19,13 @@ import { GQuery } from '@/services/gquery'
...
-}
-`
-
-const STOP_WORKFLOW = gql`
-mutation StopWorkflowMutation($workflow: WorkflowID!) {
-  stop (workflows: [$workflow]) {
-    result
-  }
-}
-`
```
kinow, line 47: Moved it out to the `graphql/query` where we have other query strings or `DocumentNode`s.

src/components/cylc/tree/index.js
```diff
@@ -15,21 +15,22 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { extractGroupState } from '@/utils/tasks'
-import { computePercentProgress } from '@/components/cylc'
```
kinow, line 19: The changes on this file are due to the move of the code to the `tree/deltas.js`, especially the code that was computing (really computing, as this was in a `computed` property of a component later :drum: ) the hierarchical tree, now in `CylcTree` and deltas module.

kinow:
Build passing again. Now just need

- [ ] write unit / functional tests
- [ ] add change log
- [ ] include the tests for #454 and close that PR as superseded by this one
- [ ] review once more to make sure we have only deltas changes here

Should be able to finish most by tomorrow, maybe leaving the tests for later, but at least other would be able to review it :+1:

kinow:
(rebased)

kinow:
Pending unit tests for 

- [ ] Deltas (new code in `tree/deltas.js`)
- [ ] CylcTree
  - [x] Workflow
  - [x] Cycle points
  - [x] Family proxies
  - [ ] Task proxies
  - [ ] Jobs

And of course, pending too that the PR's for deltas are merged first in Cylc UI Server and Cylc Flow. If this one gets merged first, the Tree view & component won't work, as there won't be any deltas in the schema/response.

oliver-sanders:
Going to try and give this some real time over the next few days.

As you said earlier, the Cylc "tree" is being built in the "Tree" component rather than the workflow service.

Will we be in a position to move this functionality up to the workflow service in future PR? Do we need to come up with a strategy for centralising this functionality, merging, etc.

src/components/cylc/tree/deltas.js
```diff
@@ -0,0 +1,201 @@
...
+ * @param {{
+ *   taskProxies: [string],
+ *   familyProxies: [string],
+ *   jobs: [string]
+ * }} pruned
+ * @param {CylcTree} tree
+ */
+function applyDeltasPruned (pruned, tree) {
+  // jobs
+  if (pruned.jobs) {
```
oliver-sanders, line 39: It might be a nice to reduce these branches to a mapping e.g:

```js
PRUNED = {
    'jobs': 'removeJob',
    'taskProxies': 'removeTaskProxy',
   // ...
}
```

Just to simplify and formalise the solution and prevent surplus logic getting added to these functions.

src/components/cylc/tree/deltas.js
```diff
@@ -0,0 +1,201 @@
...
+ *   workflow: Object,
+ *   cyclePoints: [Object],
+ *   familyProxies: [Object],
+ *   taskProxies: [Object],
+ *   jobs: [Object]
+ * }} added
+ * @param {CylcTree} tree
+ */
+function applyDeltasAdded (added, tree) {
+  if (added.cyclePoints) {
```
oliver-sanders, line 71: Same sort of thing here:

```js
ADDED = {
    'cyclePoints': [createCyclePointNode, 'addCyclePoint']
    // ...
}
```

src/components/cylc/tree/deltas.js
```diff
@@ -0,0 +1,201 @@
...
+ *
+ * @param updated {{
+ *   familyProxies: [Object],
+ *   taskProxies: [Object],
+ *   jobs: [Object]
+ * }} updated
+ * @param {CylcTree} tree
+ */
+function applyDeltasUpdated (updated, tree) {
+  if (updated.familyProxies) {
```
oliver-sanders, line 108: And here...

src/components/cylc/tree/deltas.js
```diff
@@ -0,0 +1,201 @@
...
+    if (deltas.shutdown) {
+      tree.clear()
+      return
+    }
+    if (tree.isEmpty()) {
+      // When the tree is null, we have two possible scenarios:
+      //   1. This means that we will receive our initial data burst in deltas.added.workflow
+      //      which we can use to create the tree structure.
+      //   2. Or this means that after the shutdown (when we delete the tree), we received a delta.
+      //      In this case we don't really have any way to fix the tree.
```
oliver-sanders, line 166: Thanks for the great docs! 👏

src/graphql/queries.js
```diff
@@ -17,75 +17,155 @@
...
+# TREE DELTAS BEGIN
+
+fragment WorkflowTreeDeltas on Deltas {
+  id
+  shutdown
+  added {
+    ...WorkflowTreeAddedData
+  }
+  updated {
+    ...WorkflowTreeUpdatedData
```
oliver-sanders, line 46: I think the added and updated fragments should always be identical for any rational subscription.

src/graphql/queries.js
```diff
@@ -17,75 +17,155 @@
...
-        name
-        cyclePoint
-        state
-      }
-      task {
-        meanElapsedTime
-        name
-      }
+    taskProxies(sort: { keys: ["cyclePoint"] }, ghosts: true) {
+      ...TaskProxyData
```
oliver-sanders, line 60: I like these fragments 👏 

This is a really nice way to get flat subscriptions.

Getting ahead of myself and thinking about what the subscription interface might look like in the future:

* Components could provide lists of fields to the workflow service.
* This would give us total separation of components and GraphQL implementation.
* We can potentially do some clever things like:
  * Validating the requested fields against the GraphQL schema.
  * Providing sensible fallback values in the event that the schema doesn't contain the required fields.

kinow:
@oliver-sanders 

>As you said earlier, the Cylc "tree" is being built in the "Tree" component rather than the workflow service.
>
>Will we be in a position to move this functionality up to the workflow service in future PR? Do we need to come up with a strategy for centralising this functionality, merging, etc.

I created the CylcTree in the view, because its lifecycle is the same as the view, i.e. when you visit a view that uses deltas, you need an instance of the tree, but once you navigate away, you don't want it lingering around.

To avoid duplication, it's using a mixin. So the functionality is kind of centralised at the moment I think.

But if you think it would be better in the WorkflowService, no issue for me. We will just need to make sure to call the right methods of the service from components and/or view/vueroute callbacks (e.g. component destroyed, before entering route, route was updated, etc - not sure if that makes much sense without the code, sorry)

src/graphql/queries.js
kinow: They really simplify queries, kudos to @dwsutherland who showed me fragments first (quite a while ago I think?).

>This would give us total separation of components and GraphQL implementation.

+1 !

>Components could provide lists of fields to the workflow service.

Like with `props` as it's doing now? Or something else?

I think it will be easier to discuss the improvements to the subscription interface with code, but +1 to any improvement over the current code. It is mostly a "let's get deltas working" approach. 

In the same sense, the Workflow view & component were a "let me see if lumino would work here..."... that just worked and is still in use 😄

src/graphql/queries.js
kinow: The updated fragment is a subset of the added fragment now, i.e. the only simplification that we could have I think is using the updated in the added fragment, but I think that could be confusing.

The reason for that is that the added data contains the initial burst of data. And the added data contains cycle points, while cycle points are not in the updated as I think a cycle point is not updated?

src/components/cylc/tree/deltas.js
kinow: @oliver-sanders updated! Only downside of the change is that I lost ability to trace/refactor the functions `removeJob`, `createJobNode`, etc, from within the IDE. As we are using strings, and the IDE is not clever enough to identify the function usage.

I can still search-by-string, and with enough tests it should be OK I think?

oliver-sanders:
> To avoid duplication, it's using a mixin.

I hadn't spotted that! If you have two views which both require a tree structure (e.g. Tree, Dot, Graph, etc) can they share the same in-memory tree? I'm guessing they would each end up with their own?

kinow:
> 
> 
> > To avoid duplication, it's using a mixin.
> 
> I hadn't spotted that! If you have two views which both require a tree structure (e.g. Tree, Dot, Graph, etc) can they share the same in-memory tree?

My first implementation worked that way. The CylcTree object was in Vuex, and the service was updating it.

I had the following issues with this approach:

1. When I went back to Dashboard, the tree was still being updated (that's why I started those issues about making sure the subscriptions were stopped)
2. Vuex cannot be used to persist code, only data... which means if you have something like `{ id: 1, addChildNode: function () {}`, that `addChildNode` may disappear after it was added to Vuex. Instead, we would need to move the functions to a separate module, pass the instance of the tree, and make sure to commit/dispatch mutation/actions in Vuex

That's why I went with this approach of moving the CylcTree object out of the data store.

>I'm guessing they would each end up with their own?

That's right. Even though the views use the same structure, I think there's no guarantee the user will visit both views in order. i.e. if the user goes to Workflow, then to the Dashboard, and then to the Tree view, while the user is looking at the Dashboard, there's no way to keep the tree object up to date.

---

So I think we are able to share the CylcTree between multiple views if necessary. We just need to use Vuex for the data, and make sure that the views lifecycle hooks are responsible to start/stop deltas (i.e. the tree would exist only while the views that use the deltas exist, which is similar to what we have?)

kinow:
But of course multiple components can share the same tree if in the same view, or create their own trees too; i.e. if in the workflow view, we add 1 dot, and 1 tree component, the view object could create a `cylcTree` object (as it's doing now) and pass it to each component via props (that's how it's working, but only the tree component uses it right now).

oliver-sanders:
> I had the following issues with this approach

Ok, the way I had imagined it is that the workflow service (which does all of the subscriptions etc) *always* builds the tree as and when deltas arrive in much the same way that under the old model the workflow service *always* built the vuex store as subscription data arrived. The tree would use pointers into the flat store to avoid duplicating data.

This way we don't have to worry about creation and destruction events for the tree, if a delta arrives, apply it to the tree. And the tree shouldn't need any functions in it (as it has a service providing all this functionality).

kinow:
> > I had the following issues with this approach
> 
> Ok, the way I had imagined it is that the workflow service (which does all of the subscriptions etc) _always_ builds the tree as and when deltas arrive in much the same way that under the old model the workflow service _always_ built the vuex store as subscription data arrived. 

But we need multiple trees I think? When you browse `/#/tree/five` and `/#/tree/complex`, the trees created are different, and get different updates/deltas.

I think even with this model, we would end up creating the tree when the view is created, and destroying it when the view is not used any longer (it won't be destroyed in practice).

At the moment on `master` we use a tree from Vuex **in the view**: https://github.com/cylc/cylc-ui/blob/dc23995cc2d0e8251ba1ce7a2a424b17fbee6515/src/views/Tree.vue#L72

This means that the view is the only user of that central data store object. It then passes that down to its components. That's similar to what we have now, but without the central store.

>The tree would use pointers into the flat store to avoid duplicating data.

This part is working as you expected :+1: The `CylcTree` node contains things like 

`{
  id: 'kinow|five|20100101|1`,
  node: { $DATA_FROM_GRAPHQL },
  children: [...], <-- for hiearchy, also pointers here
  state: {} <-- coming soon, needed for the infinity tree
}`

So it re-uses the data, adding a bit more for hierarchy and infinite tree.

> This way we don't have to worry about creation and destruction events for the tree, if a delta arrives, apply it to the tree. And the tree shouldn't need any functions in it (as it has a service providing all this functionality).

I think we still need to destroy the tree? When you leave `/#/tree/five` back to `/#/`, you need to stop the websocket subscription, and destroy the tree to free the memory I think? Not sure if I understood your point.

The methods in the `CylcTree` are related to the tree hierarchy and building the tree nodes. While the methods in the Workflow Service now are only related to subscriptions (GraphQL, timeout, URL, queries, ApolloClient, WebSocket, etc), which looked simpler to me.

Still happy to move everything to the WebSocket if you find it simpler. I will just need some time to move the methods over there, and create a separate JS module with the methods from `CylcTree`, so that it contains only data.

That would be similar to master, but it would still be bound to the view lifecycle, as otherwise we would have a huge memory leak without deleting trees & stopping subscriptions. What do you think? Should I finish the last unit tests and update this PR that way?

kinow:
Hmmm, testing a workflow with families (families2, that used to work) resulted in a tree with no tasks or jobs. Workflow five is still OK, so I missed something in one of the last commits. Moved to draft until I finish this manual testing 👍

oliver-sanders:
Going to have to do some thinking, my brain isn't working very well at the moment.

> **Definition:** "delta update functions" are the things which update the tree(s) e.g. `applyDeltasPruned`. (I couldn't work out what to call them).

I think there are two options:

1) Decentralised

    * Keep the tree as a mixin.
    * The tree registers its delta update functions with the WorkflowService.
    * The workflow service pushes only the relevant deltas to the tree.
    * This way we don't create one subscription per tree.

    > **E.G:** The GScan component might only subscribe to workflow deltas whereas the Tree component might subscribe to workflow, taskProxy and job deltas.

    Pros:

    * One tree per component means that we can perform filtering in the delta update functions.
    * Components get notified of "events" (e.g. task proxy removed) via these delta update function thinggies.

    Cons:

    * Extra compute and memory overheads in maintaining multiple trees.

2) Centralised

    * Have one tree in the vuex store.
    * Always keep it up to date.

    Pros:

    * Efficient storage and fewer trees to update (therefore less CPU).
    * Only one tree to manage.
    * Separates the components from GraphQL

    Cons:

    * Filtering has to be done in the component rather than in the tree.
    * Components don't notified of "events" (e.g. task removed) via delta update function thinggies.

I'm not sure which way we should go, both have their merits.

> But we need multiple trees I think?
 
We should be able put everything in one tree with `workflow` as the top level item:
 
```
+ root
   + workflow
        + cycle
            + family
                + task
                    + job
```
 
(which looks suspiciously similar to Tui)
 
> we would end up creating the tree when the view is created, and destroying it when the view is not used any longer
 
I kinda imagined the vuex store would have the flat data, and a tree. All data would be tethered to the lifecycle of the *subscription* which requested it rather than the *component*. When a subscription is unregistered we wipe the data. 
 
The simplest approach to this housekeeping (which would be fine for alpha release) would be to wipe all data whenever a subscription is unregistered (we can optimise this later).
 
Note we have to implement this housekeeping one way or another.
 
> This part is working as you expected
 
👍

oliver-sanders:
> the simpler/faster way for now, I think, is with a single filter in the view for all components

> if we had a filter that changes the GraphQL query it would be applied to all widgets

I don't think that global filtering is going to be a viable option for the Cylc UI.

The main use cases for filtering are probably:

* Having a separate tab filtering for "important" information:
  * E.G: A tree view showing only failed tasks but the main tree showing all tasks.
  * E.G: A tree showing critical tasks and another tree showing all tasks.
* Limiting what we show in the graph view (because it can't handle many tasks):
  * E.G. Restrict just the graph to the n=0 window.
  * E.G. A tree showing archive data (the n=max window) with a graph showing the n=1 window.

> The InfiniteTree also allows filtering in the component, without changing the data.

👍

> But there would be a lot more of traffic due to multiple GraphQL subscriptions (which now I think I understand why you had that in WorkflowService)

Absolutely!

Having one graph and one tree open would double the load on the UIS and the network traffic. That's why I centralised all subscriptions and data management at the beginning.

The default layout should probably be:

* GScan
* Tree
* Graph

So that's three subscriptions out-of-the-box. Once users start adding extra tabs that number escalates. Chuck browser tabs on-top of that and we could end up with quite a lot of subscriptions. This is why I wrote the subscription merging logic, it doesn't just combine multiple subscriptions into one request, it combines them into one subscription removing any duplication in the process.

Also when a user navigates away from a workflow it would be good to keep the subscriptions going for a short period of time so that if the user navigates back to that workflow the data is still live (to encourage the use of the GScan sidebar for tabbing rather than browser tabs which are less efficient from our end).

We currently have users who will have one GCylc window open for each workflow (often over ten windows), this is the problem the GScan column was designed to address.

kinow:
Hmmm, gotcha, I was going in the wrong direction, sorry.

The `WorkflowService` just needs some changes then to accommodate the new code... 

>So that's three subscriptions out-of-the-box. Once users start adding extra tabs that number escalates. Chuck browser tabs on-top of that and we could end up with quite a lot of subscriptions. This is why I wrote the subscription merging logic, it doesn't just combine multiple subscriptions into one request, it combines them into one subscription removing any duplication in the process.

This might need some further work, I had issues when I tried to add the deltas query to the `WorkflowService`—I started working on `WorkflowService` before moving things to the view, and also started with Vuex until I moved it to the view.

I think we just need to remove the query-merging. Things I remember I had problem when using the existing service:

- I couldn't use query variables (I had at one point an extra variable in my query for the strip null, but I think now I have only the workflow ID and define strip null at one level, as @dwsutherland made it inherit/re-use in the children nodes of the query) #234 
- I couldn't use a graphql-tag object with fragments (queries are strings, but I was trying to combine the fragments programmatically to re-use them; probably don't need that now) #439 
- I didn't find a way to prevent it from merging my `subscription { deltas {} }` query; as the server returns an error if you send a document with something like `subscription { deltas {} workflow {} }`
- The service is programmed to set the data received in the Vuex store with something like `response.workflows`, but that fails as the response for deltas will be `response { deltas { added {} removed {} ... }}` #209 
- The produced query was fetching more data than necessary #416 meaning that the final query (which didn't work) had

```graphql
subscription {
  workflow () { <--- no filter!
    $GSCAN_DATA
    $GRAPH_DATA <--- gets added if you have the graph widget, but notice that if you have 100 workflows, every workflow gets the graph data too
  }
  deltas { <--- rejected by server as explained above
    added ...
  }
}
```

I chatted with @dwsutherland over RIOT trying to find a way to use the deltas within the same subscription, but in the end the only solution was to have a separate one.

I found a problem with the backend today, which I reported in a comment in Cylc UI Server, where apparently I received deltas messages out of order. And also found another error in the browser console when I was going home, but will analyze it tomorrow and see whether it was in the UIS or WUI code.

That means that this issue will need a bit more of testing and time before being ready to be merged. I will use this time to start thinking how to either move everything to `WorkflowService` or find another way to achieve something similar, that allows for

- [ ] multiple active subscriptions
- [ ] in the future allow filters/parameters to be applied to each query
- [ ] start/stop subscriptions in the view lifecycles (even though I agree that leaving them running a bit longer to allow users to re-use them without having to start subscriptions from the beginning, I'd prefer to move that for the future as that needs more attention to avoid memory leaks; we could simply use `setInterval`, or even go big and think about `ServiceWorkers`, which I think RIOT uses too)

I **think** I know what I would need to change in the `WorkflowService` to support the above. Will just need a bit more of time :+1: 

How does that sound?

oliver-sanders:
@kinow sorry to throw a spanner into the works there 😔.

I'm not sure what the best way forward is, very much open to ideas, we just need to lay down a good efficient framework now that we can easily build on in the future.

Feel free the strip out the old query merging stuff, it was temporary anyway.

Here's an approach which *could* work, not saying it's the best way, I haven't a clue, but thinking aloud...

*some_component.vue*
```js
export default {
  // views provide flat lists of the fields that require (which can change)
  // the workflow service then promises to make that data available and to keep
  // it up to date
  computed: {
    taskData: [
      'id',
      'name',
      'status'
    ],
    jobData: [
      'id',
      'submitNum',
      'status'
    ]
  }
}

// the component could potentially use an observer pattern to recieve delta events
```

*some_service.js*
```js
// we then loop over subscriptions and merge these lists
function getSubscription(sub_sets) {
  const taskData = []
  for (const set of sub_sets) {
    for (const item of set.taskData) {
      if (item not in taskData) {
       taskData.push(item)
      }
    }
  }
  // ...

  return `
    subscription (ids=[$ids]) {
      deltas {
        added {
          taskProxies {
            { taskData.join('\n') }
          }
         updated {
           taskProxies {
            { taskData.join('\n')
           }
        }
        removed { id }
    }
  }
  `
}
```

We would probably need to write out the subscriptions per-workflow something like:

```js
subscription one {
  workflows (ids: ['one']) {
    // ...
   }
}
subscription two {
  workflows (ids: ['two']) {
    // ...
   }
}
// ...
```

oliver-sanders:
The Achilles' Heel, of any centralised approach is going to be filtering!

If the GUI was just the Tree view it would be so easy, we could request only the data we need and request the rest when users expand or collapse nodes. If you want to filter for say failed tasks you would only have to request failed tasks.

But having multiple views makes this tricky, I guess I'm focussed on improving worst-case scenario more-so than the best-case. From the worst-case perspective, it's better to request all of the data once, then filter it in the UI then to request the data multiple times with different filtering.

I think the main use-cases for filtering will be for temporary searching and having separate tabs with specific filters (e.g. a tree view containing all failed tasks), in which case filtering in the UI is preferable.

Something I was thinking about when we started this, though it felt a bit over the top, was to start a GraphQL server in the UI and have the components request data from this local server. This allows the views to get data in the preferred format and with filtering, but has obvious overheads and potentially data duplication.

But this diagram just looks ridiculous:

```
Scheduler (PB Serv) => UIS (PB Client) => UIS (GQL Serv) => UI (GQL Client) => UI (GQL Serv) => UI Component (GQL Client)
```

kinow:
>@kinow sorry to throw a spanner into the works there pensive.

No problem. I'm pointing these things after using the service with Vue/JS, but had no idea about them before. Always easier in hindsight.

>Feel free the strip out the old query merging stuff, it was temporary anyway.

:+1:  I spent some time thinking on this issue this weekend, and I may disable/remove the query merging. But before doing that, I still want to find the minimum changes possible to get the deltas in the `WorkflowService`, so that this issue can be review/merged, unblocking work on infinite tree.

>Here's an approach which could work, not saying it's the best way, I haven't a clue, but thinking aloud...
>
>some_component.vue

I think that could work if in components that are views (e.g. `views/Workflow.vue`), but not sure about the components like `components/cylc/workflow/Workflow.vue`. We could perhaps also use fragments instead of fields.

If a view declares it needs the job data, that is likely to be used by the `Job` component. Any view/component using the `Job` component must provide the same minimum data (state, messages, etc). And that can be assembled into something like `fragment JobData { ... }`.

With `graphql-tag`, we can create also some objects with `gql\`fragment JobData { ... }\`` which compiles the fragment into a JS object, which can be used for a `compiled`/`data`/`props` in Vue, instead of a `string`.

>We would probably need to write out the subscriptions per-workflow something like:

If you mean to have these subscriptions separately, i.e. one JavaScript `Observable` created for each of these with `apolloClient.subscribe(query, variables, etc)`, then I agree. 

If you mean that to be in the same query, then I think that is not supported in our GraphQL backend (not sure if due to our tooling/dependencies, or by GraphQL spec).

image

Moving two deltas within a single subscription with different filters is also not permitted.

image

When I started working on this issue, I spent some time thinking how this could be used with Graph, Tree, Dot, and even Dashboard in the UI. And tried some queries in GraphiQL and chatted with David on Riot to see if there were other alternatives, but only alternative I found was to have separate GraphQL subscription/queries.

>The Achilles' Heel, of any centralised approach is going to be filtering!

Agreed! And sorry for not keeping that in mind. I haven't thought much about that at all until now :disappointed_relieved: 

>I think the main use-cases for filtering will be for temporary searching and having separate tabs with specific filters (e.g. a tree view containing all failed tasks), in which case filtering in the UI is preferable.

Yeah, hard to say what's the best option. If we cannot make much progress on this, maybe we could move it to an issue, and use filtering in the UI for the meantime until we discuss/investigate the multiple-filters option.

kinow:
Almost done with a commit that would update the views and components all to use subscriptions, through the `WorkflowService`, and without query-merging.

This would fix a few of the existing issues with the `WorkflowService` (like using variables, being able to use fragments, using a graphql-tag object, etc).

Having some problems with the `Workflow` view and component. Blame is all on myself, as the current design of those two is far from  perfect (I would be happy if it was at least good-enough :-) ).

So trying to simplify it a bit. Basically, every widget added to the Workflow view will create a new subscription, that can be updated without changing other widgets. It won't have any UI element to trigger any change in the subscription, like filtering. But it should be a starting point to have filter for each widget.

If that works, will push that commit, and add one more that moves the subscriptions to the `WorkflowService`, out of the views/components. Then it should be ready for another round of reviews :+1:

dwsutherland:
> Moving two deltas within a single subscription with different filters is also not permitted.

The error message suggests using and alias for this one.

kinow:
> > Moving two deltas within a single subscription with different filters is also not permitted.
> 
> The error message suggests using and alias for this one.

At least on GraphiQL that doesn't work. I will try with `ApolloClient` later.

kinow:
@oliver-sanders 

Updated in one commit so that we can edit/drop it later if necessary: https://github.com/cylc/cylc-ui/pull/458/commits/4f16766e309b7ec15bbeb44861305c88a1ebe9f9

In this commit:

1. removed query merging
2. moved subscriptions to views **(*)**, and the gscan component holds its own subscription, which lasts throughout the application lifecycle
3. in the `view/Workflow.vue`, for each widget, it creates a separate subscription, with its own `isLoading` value. In theory, I think we should be able to filter each subscription separately later. Does that make sense?

**(*)**: if this looks OK, I can add one more commit that moves the subscriptions to the `WorkflowService`. Either way is fine for me. The only difference is that instead of keeping the `subscription` in the View/Component, I would need to generate a random subscription ID, and use it to pass to the `WorkflowService`.

kinow:
Just had a chat with @hjoliver (thanks!) and realized that the way I implemented the subscriptions, we would end up with multiple subscriptions, one for each widget added to the Workflow tab.

Let me fix that, and then try to write a summary on the changes. Ready for review in one or two days 👍

kinow:
Whew, had some issues figuring out how to remove widgets in Lumino programmatically. But now the PR should be ready for review.

1. no query merging, though the deltas is using fragments, so we can build on top of that and merge/combine fragments for other queries with deltas, e.g. for the graph or dot components
2. every view will have 2 subscriptions at least, 1 for gscan, and another one for the view's data
3. the workflow view is a special case as it has tabs/widgets. For each widget type, you will have only ever 1 subscription active. So if you have 10 tree tabs open, you will still have 1 active GraphQL subscription query for deltas. If you close all, then you have 0 tree subscriptions (in this case this view would have only 1 subscription)
4. going to a 404 page uses the `Empty` layout, which doesn't have any subscription, same as before
5. each view that uses deltas has a `tree` object, which holds the current tree object, and to which deltas are applied. This tree is backed by an `Array` and a `Map`. The array is used in Vue for reactivity. The map is used by the tree methods to access elements in the tree more quickly and update with the new data from deltas. Vuex doesn't support objects with methods, only data. We could push the methods to external modules, but they would have to either be Vuex actions/mutations, or use the actions/mutations, which adds a bit of extra processing time (as Vuex centralizes everything) plus further time to figure out how to move everything there
6. the data in the `CylcTree.root.children` (the Array that Vue components use for reactivity) **in theory should be compatible with the Infinite Tree**... or at least I tried my best to include the data I remembered was necessary... so it should be easy to add that later 🤞

kinow:
Here's how I tested it, in case it's helpful to reviewers:

- Used workflows five (no families), families2 (any workflow with families will do), complex (what a nightmare)
- Always had the Network tab open, with the websocket filter, and keeping track of the operation ID in the subscription
- Confirmed that for each page we had the expected number of operations/subscriptions, that includes Dashboard, User profile, Tree, Graph, and Workflow views
- Confirmed there were no errors or warnings in the console (that we are not aware of)
- Using another browser looked at the Vue app with vue-dev-utils, and confirmed the Vuex store and the components had what was expected in each one
- With the browser memory tools confirmed we don't have any clear memory leaks, and that each workflow works fine after hours running
- If you test a view that contains parameters, that means that VueRouter route/view supports three lifecycles hooks at least, `beforeRouteEnter`, `beforeRouteUpdate`, and `beforeRouteLeave`. The first and the last are easy to test, just navigate to the page (e.g. `#/workflow/five`), and then visit another view (e.g. (`/`). Now you triggered both. But to trigger the update, you must navigate using a different parameter, like `/#/workflow/families2`. So it might be good to test the app with a) no workflows, b) 1 workflow, c) +2 workflows

src/components/cylc/tree/tree.js
```diff
@@ -0,0 +1,497 @@
...
+ * // and its children can be equally accessed with:
+ * const workflowChildren = rootNode.children
+ * const cyclePoint = workflowChildren[0]
+ * // and so it goes
+ * const familyProxy = cyclePoint.children[0]
+ * </script>
+ *
+ * This data structure also keeps a lookup `Map` object. This map
+ * contains the ID of each node as key, and the node itself as object
+ * reference.
```
oliver-sanders, line 110: 👌

And thanks for these docs!

src/components/cylc/tree/tree.js
```diff
@@ -0,0 +1,497 @@
...
+    /**
+     * @type {Map<string, any>}
+     */
+    this.lookup = new Map()
+    if (!workflow) {
+      this.root = {
+        id: '',
+        node: {},
+        children: []
+      }
```
oliver-sanders, line 159: I'm guessing this is for testing purposes?

If no workflow is provided then the root node doesn't get added to the lookup?

src/components/cylc/tree/tree.js
```diff
@@ -0,0 +1,497 @@
...
+
+  /**
+   * @param {null|{
+   *   id: string,
+   *   node: Object,
+   *   children: []
+   * }} cyclePoint
+   */
+  updateCyclePoint (cyclePoint) {
+    if (cyclePoint) {
```
oliver-sanders, line 235: Out of interest why do we need to defend against `null` values in each method?

src/components/cylc/tree/tree.js
```diff
@@ -0,0 +1,497 @@
...
+      }
+      // add if not in the lookup already
+      const existingFamilyProxy = this.lookup.get(familyProxy.id)
+      if (!existingFamilyProxy) {
+        this.lookup.set(familyProxy.id, familyProxy)
+      } else {
+        // We may get a family proxy added twice. The first time is when it is the parent of another
+        // family proxy. In that case, we create an orphan node in the lookup table.
+        // The second time will be node with more information, such as .firstParent {}. When this happens,
+        // we must remember to merge the objects.
```
oliver-sanders, line 286: Ooh, nasty. Must remember to make sure views are resilient to this.

src/components/cylc/tree/tree.js
```diff
@@ -0,0 +1,497 @@
...
+        merge(existingFamilyProxy, familyProxy)
+        this.lookup.set(existingFamilyProxy.id, existingFamilyProxy)
+        // NOTE: important, replace the version so that we use the existing one
+        // when linking with the parent node in the tree, not the new GraphQL data
+        familyProxy = existingFamilyProxy
+      }
+      if (!familyProxy.node.state) {
+        familyProxy.node.state = ''
+      }
```
oliver-sanders, line 295: Is this to handle ghost nodes?

Might be worth adding a comment here (like the earlier one) incase we handle this at the scheduler in the future.

src/components/cylc/tree/tree.js
```diff
@@ -0,0 +1,497 @@
...
+
+  /**
+   * @param {null|string} familyProxyId
+   */
+  removeFamilyProxy (familyProxyId) {
+    let node
+    let nodeId
+    let parentId
+    // NOTE: when deleting the root family, we can also remove the entire cycle point
+    if (familyProxyId.endsWith('|root')) {
```
oliver-sanders, line 347:
```suggestion
    if (familyProxyId.endsWith(`|${FAMILY_ROOT}`)) {
```

src/components/cylc/tree/tree.js
```diff
@@ -0,0 +1,497 @@
...
+   *   node: Object,
+   *   children: []
+   * }} taskProxy
+   */
+  updateTaskProxy (taskProxy) {
+    if (taskProxy) {
+      const node = this.lookup.get(taskProxy.id)
+      if (node) {
+        computeTaskProgress(taskProxy)
+        merge(node, taskProxy)
```
oliver-sanders, line 417: Could the `merge` overwrite the progress computed by `computeTaskProgress`?

src/components/cylc/tree/tree.js
```diff
@@ -0,0 +1,497 @@
...
+  /**
+   * @param {null|string} taskProxyId
+   */
+  removeTaskProxy (taskProxyId) {
+    const taskProxy = this.lookup.get(taskProxyId)
+    if (taskProxy) {
+      this.recursivelyRemoveNode(taskProxy)
+      // Remember that we attach task proxies children of 'root' directly to a cycle point!
+      let parent
+      if (taskProxy.node.firstParent.id.endsWith('|root')) {
```
oliver-sanders, line 431:
```suggestion
      if (taskProxy.node.firstParent.id.endsWith(`|${FAMILY_ROOT}`)) {
```

src/components/cylc/tree/tree.js
kinow: >I'm guessing this is for testing purposes?

Also because there are Vue components using the `.children` element. This way we don't have to add `v-if`/`v-show`. The `beforeRouteLeave` also clears the tree, resetting the `.children` array, so that Vue automatically removes the DOM/VDOM elements for us.

Though it should be possible to remove this I guess, and just check for null/undefined in tests & views/components if that's better.

>If no workflow is provided then the root node doesn't get added to the lookup?

That's right. Only values populated via deltas are added to the lookup.

kinow:
>Sorry for being absent for the last few days, ready to put some time into a proper code review. A few small comments so far, no blockers:

Thanks @oliver-sanders , I've finished moving the subscriptions back to the workflow service now. I'll take a break of a couple hours for lunch/gym. Then take a deep breath and start over reviewing what I did to make sure it's really ready for review.

- Subscriptions are in the `WorkflowService`
- There is an extra subscription for deltas in the `WorkflowService` because we cannot merge it with the other ones (yet, your suggestion from RIOT could work, but requires some GraphQL-FU and @dwsutherland will probably need to change/review the UIS code changes)
- I've added back the query merge; we can discuss later whether/how to update it to something like fragments if necessary
- I've removed the code that was fixing the number of subscriptions, but left the unit/e2e tests that verify the number of subscriptions. The tests are now checking for the wrong number (e.g. there's one test case where it checks for 4 subscriptions, which should never happen). I've added a `FIXME` marker, and am creating an issue in GH to track it later. This simplified the pull request and hopefully should make reviewing it simpler too.

Thanks!
Bruno

src/components/cylc/tree/tree.js
kinow: I don't think so. Here we are updating a task proxy. The `node` is an existing task proxy, that may be displayed in the UI too.

What we are doing, is given the new data for the task proxy in `deltas.updated`, we compute its progress (as UIS doesn't provide that info, but gives us the `meanElapsedTime`, `startedTime`, etc).

And then we `merge(object, sources...)`, where the left side object is the one receiving updates.

So I think `node` will get updated with the most recent progress from the `deltas.updated` data.

src/components/cylc/tree/tree.js
kinow: I think that's because `familyProxy`'s don't have states. So the UI (vue template/dom) would have to protect against `undefined`.

I found it easier to just define the state here. Further on, after the tasks/jobs are added, we then compute the family proxy state based on its children states (using old Cylc 7 logic).

(This is in theory, it's been implemented some time ago, and hopefully that's working OK... we have tests for that I think, but before 1.0 release we need to verify that the family statuses are being updated correctly I guess)

src/components/cylc/tree/tree.js
kinow: I tried to add as many comments as possible, for each situation where I got an error like "Cannot use undefined...", or "blabla is not a function", etc... this kind of runtime unexpected errors.

For each error I either added a comment, or took it back to @dwsutherland to review if the UIS was giving the right data (hence the amount of re-work/review/answering-silly-questions that David had to do some weeks ago :sweat_smile: )

src/components/cylc/tree/tree.js
kinow: I wrote this code before the deltas. I actually re-used a lot of the existing code that we had in the Vuex store or that was used by the tree Vuex module (not the best place to keep it).

And I wasn't sure what to expect from the deltas, so ended up coding defensively from the beginning.

Now I am not so sure it is necessary. I **think** we could get an undefined if I try to access a key that doesn't exist in the deltas, though that's hard. We should be able to remove this if you think that's better.

Just a matter of removing these `if` statements, removing the tests that were checking for `null`, running unit + e2e tests, and playing a bit with `five`/`families2`/`complex`. A 30-60 minutes job.

src/components/cylc/tree/tree.js
kinow:
```
      // See comment above in the else block. When we get family proxies out of order, we create the parent
      // nodes if they don't exist in the tree yet, so that we can create the correct hierarchy. Later, we
      // merge the data of the node. But for a while, the family proxy that we create won't have a state (as
      // the state is given in the deltas data, and is not available in the `node.firstParent { id }`.
```

(pushed in a separate commit)

kinow:
Ready for review again @oliver-sanders . @hjoliver I've added you as second reviewer, but feel free to re-assign if you'd like. @dwsutherland should be able to use the new UI and tell whether what is being displayed makes sense, or if there's something amiss.

- Removed as much changes as possible
- Changelog updated
- Tests updated
- Subscriptions in the workflow service
- Not bothering about closing subscriptions for other views (i.e. after a long time users may have performance issues, to be fixed later)

Tested with five in production mode without the Vue dev tools extension. Between 10 and 15 MB of memory.

image

families2 worked fine too, same memory footprint.

image

Mutations working fine as well. The memory was always between 20 and 30, I guess due to the query introspection and mounting the form/model in memory/etc.

image

Complex started at 75 MB at 02:47 PM. Around 03:30 PM, still under 100MB :tada: 

image

There were, however, errors in the console. One of them I'm aware of the reason (https://github.com/cylc/cylc-flow/issues/3658), not too sure about the other error. But I suspect it's another one to investigate later with calm, debugging UI & UIS/Flow, and with @dwsutherland 's help to pinpoint where exactly is the issue.

image

Tested switching views, opening dashboard, tree view, graph view, workflow view with multiple widgets and positioning widgets in different places. Seemed to work fine.

:+1:

src/components/cylc/tree/tree.js
oliver-sanders: Thought so, if not necessary any more it would be nice to remove at some point just to streamline the code but no biggie.

oliver-sanders:
> Complex started at 75 MB at 02:47 PM. Around 03:30 PM, still under 100MB 🎉

I'm measuring 250MB for this GH PR in Safari so I think we should see that as a pretty good win.

kinow:
> > Complex started at 75 MB at 02:47 PM. Around 03:30 PM, still under 100MB tada
> 
> I'm measuring 250MB for this GH PR in Safari so I think we should see that as a pretty good win.

With `yarn run build`, or `yarn run build:watch` (or another command that produces development build)? The former produces an optimized version, that should - in theory - consume less memory.

src/mixins/treeview.js
```diff
@@ -0,0 +1,69 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { mapActions, mapState } from 'vuex'
+/* eslint-disable no-unused-vars */
+import User from '@/model/User.model'
+/* eslint-enable no-unused-vars */
+
+const treeview = {
```
oliver-sanders, line 23: The name `treeview` is going to cause confusion with the "Tree View". Perhaps `datatree` or something like that?

src/views/Workflow.vue
```diff
@@ -93,15 +105,30 @@ export default {
   },
   beforeRouteEnter (to, from, next) {
     next(vm => {
-      // Create a Tree View for the current workflow by default
-      const subscriptionId = vm.subscribe('tree')
       vm.$nextTick(() => {
+        vm.subscribeToDeltas()
+        // Create a Tree View for the current workflow by default
+        const subscriptionId = new Date().getTime()
```
oliver-sanders, line 111: Is this necessary if the Tree Component initiates its own subscriptions?

kinow: That's because the Lumino widgets require an ID as argument when created.

src/mixins/treeview.js
kinow: Good point! Changing it tomorrow :+1:

src/views/Workflow.vue
oliver-sanders: Ah, sorry I was commenting more on the `vm.subscribeToDeltas()` bit.

kinow: Ah, that one. That's still necessary to start the GraphQL deltas subscription. Before, we were also calling `register`/`susbcribe`. Now we just `subscribeToDeltas`, which will tell `WorkflowService` to start the query and keep track of the subscription.

As there are no merging here, we have a single deltas subscription, so we don't need the ID as before.

src/components/cylc/tree/tree.js
kinow: Done! Tested a couple workflows and it appeared to have worked. We can re-add it later if necessary :+1:

src/views/Workflow.vue
kinow: Re-reading the code @oliver-sanders , I realized now that the deltas subscription is being created when the first widget is added, and then never removing it. Let me see if I can find an alternative for this.

oliver-sanders:
I've lost track of the Riot discussions a bit, where are we at with this PR?

Have the subscription issues been solved / moved to https://github.com/cylc/cylc-flow/issues/3658?

kinow:
> I've lost track of the Riot discussions a bit, where are we at with this PR?
> 
> Have the subscription issues been solved / moved to [cylc/cylc-flow#3658](https://github.com/cylc/cylc-flow/issues/3658)?

Yes @oliver-sanders , for now that's the closest explanation for that issue. I couldn't find anything else in the JS code that would be causing it. And what triggers it is a subscription being stopped, and a new one started (after the graph widget).

oliver-sanders:
Pretty much finished with the code review, testing going well, everything working as expected.

src/views/Workflow.vue
oliver-sanders: Any luck?

kinow: Ah that should be fixed (checking again now). I got distracted after I wrote this comment with the issue in the order of deltas messages and forgot to update this comment. Sorry.

kinow: It was closing the deltas subscription once you closed any tree widget. Added a check to confirm there are no more tree widgets before closing it. Should be good to go. Tested with a workflow running, and looking at the network tab in the browser. Confirmed there were only 2 subscription (gscan, deltas) and once all tree widgets were closed, the deltas is stopped. Then adding a tree widget, a new deltas subscription is created.

oliver-sanders:
Just tried to have another play with this but ran into an unexpected error during build:

```
 ERROR  Failed to compile with 1 errors                                      10:49:06

This dependency was not found:

* @/components/cylc/tree/Tree in ./src/components/cylc/workflow/index.js

To install it, you can run: npm install --save @/components/cylc/tree/Tree
 ERROR  Build failed with errors.
error Command failed with exit code 1.
```

I was able to build it before and GH actions can still build it, just me?

kinow:
Hmmm, I've seen that error, long time ago in the project. And more recently in another project I was having a play, but it was a problem in the configuration (vue.config.js I think). But I haven't changed the configuration - at least I cannot recall doing so in the last commits.

Could it be the version of node or npm/yarn? Just built in my environment after syncing branch to make sure I was on the latest commit.

```bash
yarn install
yarn run serve
```

It took a little while to build everything (I think I had another branch checked out). But then it loaded the offline mode successfully.

```
kinow@ranma:~/Development/python/workspace/cylc-ui$ node -v
node -v
v12.8.0
kinow@ranma:~/Development/python/workspace/cylc-ui$ npm -v # I actually use yarn, but FWIW
6.14.1
kinow@ranma:~/Development/python/workspace/cylc-ui$ yarn -v
1.22.4

```

oliver-sanders:
I deleted my `node_modules` dir just to make sure then `yarn install`ed them back again.

```console
$ yarn -v
1.22.4
$ node -v
v12.4.0
```

kinow:
> I deleted my `node_modules` dir just to make sure then `yarn install`ed them back again.
> 
> ```
> $ yarn -v
> 1.22.4
> $ node -v
> v12.4.0
> ```

Versions looking good. Good idea deleting `node_modules`. It's supposed to work now. 

The reason why it's complaining, I think, is that `@` is an alias to the location of `./src` folder. That alias is automatically set to us by Vue configuration. If you'd like to confirm that value is set and correct, you can run `vue inspect`. Somewhere in the output you should find the `@` alias set to your `src` path location.

oliver-sanders:
Ooh `vue inspect` is very nice:

```js
  resolve: {
    alias: {
      '@': '/Users/oliver/cylc-ui/src',
```

And just to check the obvious:

```console
$ ls /Users/oliver/cylc-ui/src
App.vue     graphql  lang     main.js  model	router	  store   utils
components  i18n     layouts  mixins   plugins	services  styles  views
```

kinow:
Hmm, that's strange (as strange as all these conda issues :disappointed: ).

Your alias is supposed to work fine. It's a feature from [webpack](https://webpack.js.org/configuration/resolve/#resolvealias).

What command did you use? `yarn run build:watch`? Any changes in your git clone?

oliver-sanders:
> Hmm, that's strange (as strange as all these conda issues 😞 ).

Damit.

I was using `yarn run build` but have also tried `build:watch`

I gave `yarn cache clean -f` a go too, no change :(

I may try rebuilding my conda env with a fresh node installation, hopefully that'll do it.

dwsutherland:
I found a fault in the data-store while testing this one, it was exposed here because the subscription is flat (so doesn't filter the results to those that have relationships).. The data store contains naked dummy tasks, and non-first-parent families:
image

```
          "familyProxies": [
.
.
.
            {
              "id": "sutherlander|baz|20200101T0000+12|poll_jin",
              "name": "poll_jin",
              "cyclePoint": "20200101T0000+12",
              "firstParent": {
                "id": "sutherlander|baz|20200101T0000+12|root",
                "name": "root",
                "cyclePoint": "20200101T0000+12",
                "state": "succeeded",
                "__typename": "FamilyProxy"
              },
              "__typename": "FamilyProxy"
            },
            {
              "id": "sutherlander|baz|20200101T0000+12|FAM5",
              "name": "FAM5",
              "cyclePoint": "20200101T0000+12",
              "firstParent": {
                "id": "sutherlander|baz|20200101T0000+12|root",
                "name": "root",
                "cyclePoint": "20200101T0000+12",
                "state": "succeeded",
                "__typename": "FamilyProxy"
              },
              "__typename": "FamilyProxy"
            },
.
.
.
```

I need to rework the family generation, to only have first-parent ancestors for non-dummy tasks.

kinow:
>I need to rework the family generation, to only have first-parent ancestors for non-dummy tasks.

:+1: I will need to have a look at your example again on Monday with calm to understand what's going on, and more importantly learn how to identify it in case it ever happens again (or another similar issue).

Thanks for the extra details in your comment @dwsutherland ! Will give it a try on Monday (and happy to review if you create a PR to UIS)

dwsutherland:
> happy to review if you create a PR to UIS

@kinow  - Here it is:
https://github.com/cylc/cylc-flow/pull/3707

No fundamental changes..

kinow:
Rebased following our 0.2 release :+1:

oliver-sanders:
Sorry, I'm still stuck building this:

```console
$ git fetch bruno
$ git checkout bruno/add-cylc-tree-ds 
HEAD is now at 8d58958 Keep 1 susbcription with deltas only, no matter number of tree widgets. Remove it when no more tree widgets.
$ rm -rf node_modules/
$ yarn install
yarn install v1.22.4
...
✨  Done in 21.49s.
$ yarn run build
yarn run v1.22.4
$ vue-cli-service build

⠸  Building for production...

 ERROR  Failed to compile with 1 errors                                      16:01:53

This dependency was not found:

* @/components/cylc/tree/Tree in ./src/components/cylc/workflow/index.js

To install it, you can run: npm install --save @/components/cylc/tree/Tree
 ERROR  Build failed with errors.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I've tried with an entirely new Conda env but get the same result.

Env info:

```console
$ conda env export | grep node
  - configurable-http-proxy=4.2.1=node12_hbce0df8_0
  - nodejs=12.4.0=h6de7cb9_0
$ which node
/Users/oliver/miniconda3/envs/cylc8/bin/node
$ node --version
v12.4.0
```

Anything obviously off?

hjoliver:
> Anything obviously off?

It works for me :thinking:

oliver-sanders:
> Could you run yarn list or npm list and compare with my output, please?

```diff
$ diff mine his
--- mine	2020-07-31 08:36:39.000000000 +0100
+++ his	2020-07-31 08:39:29.000000000 +0100
@@ -7264,4 +7264,4 @@
 │  ├─ tslib@^1.9.3
 │  └─ zen-observable@^0.8.0
 └─ zen-observable@0.8.15
-Done in 0.80s.
+Done in 1.46s.
```

Dammit.

oliver-sanders:
GOT IT! (I think).

Sooo, in a way it's kinda a Mac problem, or more specifically an AFS filesystem problem. AFS is not case-sensitive. There are other filesystems which have the same quirk (e.g. FAT32) and although NTFS is case-sensitive apparently Windows decides to shield users from this making it appear case-insensitive to the shell.

When we import files in Vue we don't specify the extension (there is some internal wrangling presumably handling `js`, `ts`, `.Vue` and any of the more esoteric JS flavours.

This creates a problem when you have two files with the same name but different extensions. Take a look as the FS tree:

```
src/components/cylc/tree
├── Tree.vue
├── TreeItem.vue
├── deltas.js
├── event-bus.js
├── index.js
└── tree.js
```

To AFS `Tree` looks the same as `tree` so whether Vue/Node/whatever attempts to import from `tree.js` or `Tree.vue` is down to luck/internals.

**tldr:** If we rename `tree.js` then I think we should be fine!

kinow:
@oliver-sanders renamed `tree.js`. Good to know about this issue. Will try to keep that in mind, not only for JS/Vue. The more environments supported, the merrier :+1: 

Did a quick test with `five` in my local computer. GitHub actions will be busy for some <10 minutes. If build succeeds, let us know if that fixes the issue for you too :+1: 

Thanks for the great detective work!

oliver-sanders:
✨  Done in 39.76s.

🎉

kinow:
Added a new commit that updates the offline mock service (i.e. no need to worry about the change here). 

```diff
commit ca68fa49ae8300707aaf48fe9c44ec661cd95db8 (HEAD -> add-cylc-tree-ds)
Author: Bruno P. Kinoshita <bruno.kinoshita@niwa.co.nz>
Date:   Mon Aug 3 10:36:47 2020 +1200

    Update offline mock service, no need to use a Promise

diff --git a/src/services/mock/workflow.service.mock.js b/src/services/mock/workflow.service.mock.js
index c385d5e..95e6f22 100644
--- a/src/services/mock/workflow.service.mock.js
+++ b/src/services/mock/workflow.service.mock.js
@@ -65,18 +65,15 @@ class MockWorkflowService extends GQuery {
     this.subscriptions.push({
       subscription
     })
-    return new Promise((resolve, reject) => {
-      subscriptionOptions.next({
-        data: {
-          deltas: {
-            added: {
-              workflow: checkpoint.workflows[0]
-            }
-          },
-          workflows: checkpoint.workflows
-        }
-      })
-      resolve(subscription)
+    subscriptionOptions.next({
+      data: {
+        deltas: {
+          added: {
+            workflow: checkpoint.workflows[0]
+          }
+        },
+        workflows: checkpoint.workflows
+      }
     })
   }

```

It matches the signature of the "online" service, and also avoids the unnecessary `Promise` - the constructor evaluates that function eagerly, which is the same as calling the function directly in the offline mode.

CHANGES.md
```diff
@@ -57,6 +57,9 @@ increase, decrease, and reset UI font size.
 view using graphiql.js that sends queries to the backend graphiql
 endpoint.
 
+[#458](https://github.com/cylc/cylc-ui/pull/458) - Add CylcTree and
+use deltas for the Tree view and component.
```
kinow, line 61: Just realized I need to move it to 0.3 version.

kinow:
Added a new commit moving the change log entry to 0.3 :+1: no other changes, shouldn't break CI or change any code/feature here.

hjoliver:
(Was using pre-SOD cylc-flow and cylc-uiserver, so this merge might make things a little broken for a while ...)

PR #460, opened 2020-05-18
Fix alert component colors

### LINKED ISSUE #459: Alert component not showing error color
**Describe the bug**

Found while testing #458. I added an `Alert` component to display a message, but it wasn't displayed in red.

From the Vuetify docs, looks like they changed their component. Instead of using `:color="error"`, now they are expecting `:type="error"`.

`master`

Add an alert to the store with `color="error"`. It is displayed as a transparent div.

**Expected behavior**

Div coloured according to the color/type specified, e.g. red for error.

**Screenshots**

Forgot to take a screenshot. Logging issue so I don't forget it :)

**Additional context**

https://v15.vuetifyjs.com/en/components/alerts

**Pull requests welcome!**

kinow:
These changes close #459

**Requirements check-list**

### DIFF
--- a/src/components/core/Alert.vue
+++ b/src/components/core/Alert.vue
@@ -19,12 +19,15 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
   <div v-if="alert">
     <v-alert
       :value="true"
-      :color="alert.getColor()"
+      :type="alert.getColor()"
       :icon="alert.getIcon()"
-      class="ma-0"
+      :class="getColor(alert.getColor())"
       dismissible
+      tile
+      light
+      colored-border
     >
-      <p class="body-1">{{ alert.getText() }}</p>
+      {{ alert.getText() }}
     </v-alert>
   </div>
 </template>
@@ -33,11 +36,29 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import {
   mapState
 } from 'vuex'
+
 export default {
   name: 'Alert',
 
+  data () {
+    return {
+      // TODO: remove later when https://github.com/vuetifyjs/vuetify/issues/11021 is fixed
+      colors: new Map([
+        ['error', 'red'],
+        ['success', 'green'],
+        ['warning', 'amber']
+      ])
+    }
+  },
+
   computed: {
     ...mapState(['alert'])
+  },
+
+  methods: {
+    getColor (type) {
+      return this.colors.get(type) || ''
+    }
   }
 }
 </script>
--- a/src/main.js
+++ b/src/main.js
@@ -55,7 +55,7 @@ Vue.prototype.$workflowService = workflowService
 Vue.config.productionTip = false
 
 /* eslint-disable no-new */
-new Vue({
+const app = new Vue({
   i18n,
   router,
   store,
@@ -68,3 +68,9 @@ new Vue({
     })
   }
 }).$mount('#app')
+// e2e tests use the offline mode, so here we expose the Vue.js app so Cypress can access it programmatically
+// e.g. window.app.$store and window.app.$workflowService.
+// Ref: https://www.cypress.io/blog/2017/11/28/testing-vue-web-application-with-vuex-data-store-and-rest-backend/
+if (['test', 'offline'].includes(process.env.NODE_ENV)) {
+  window.app = app
+}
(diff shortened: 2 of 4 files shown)


kinow:
Preview using `success`, `warning`, and `error` alerts.

image

image

image

kinow:
Was going to tick the box "Does not need a test...", but I think it might be testable actually. Let me give it a try.

src/components/core/Alert.vue
```diff
@@ -19,12 +19,15 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
       :value="true"
-      :color="alert.getColor()"
+      :type="alert.getColor()"
       :icon="alert.getIcon()"
-      class="ma-0"
+      :class="getColor(alert.getColor())"
       dismissible
+      tile
+      light
+      colored-border
```
kinow, line 28: `tile` removes the rounded borders.
`light` and `colored-border` to indicate no dark mode (due to upstream bug)

kinow:
What it looks like when you have an alert without the background color in an e2e test:

image

I am not checking whether the error is red, success is green, etc, because I couldn't get the exact values from the Vuetify style dynamically. So instead I just have two constants, one for black and the other for white.

And the tests just go through each type (error, warning, success) and check that the background color is neither black nor white.

kinow:
Spent a good amount of time trying to get the colors; not a good idea unless you want to dig deep into how Cypress and Babel/WebPack/etc work... also had a lot of work trying to write a test using Vuex state, then learned the hard way that Cypress is a bit opinionated on how you must write tests, some of their docs state how they believe tests should be written. Point taken, try to write simple tests, isolated from each other, without sharing states :+1:

kinow:
(coverage went down a bit without the e2e, but now it's back to green :tada: )

src/main.js
```diff
@@ -68,3 +68,9 @@ new Vue({
     })
   }
 }).$mount('#app')
+// e2e tests use the offline mode, so here we expose the Vue.js app so Cypress can access it programmatically
+// e.g. window.app.$store and window.app.$workflowService.
+// Ref: https://www.cypress.io/blog/2017/11/28/testing-vue-web-application-with-vuex-data-store-and-rest-backend/
+if (['test', 'offline'].includes(process.env.NODE_ENV)) {
+  window.app = app
+}
```
kinow, line 76: Same change in #454 . So once either this or that PR is merged, I will fix conflicts in the other PR :+1:

src/components/core/Alert.vue
```diff
@@ -19,12 +19,15 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
     >
-      <p class="body-1">{{ alert.getText() }}</p>
+      {{ alert.getText() }}
```
oliver-sanders, line 30: Shouldn't this text be wrapped, at least in a `<span/>`?

src/components/core/Alert.vue
```diff
@@ -19,12 +19,15 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
   <div v-if="alert">
     <v-alert
       :value="true"
-      :color="alert.getColor()"
+      :type="alert.getColor()"
       :icon="alert.getIcon()"
-      class="ma-0"
+      :class="getColor(alert.getColor())"
```
oliver-sanders, line 24: Are these colour classes are built into Vuetify?

src/components/core/Alert.vue
kinow: The component ends up wrapping the text in some HTML element later I think. I changed after I noticed some padding/margin, and saw that they don't use it in their examples: https://v15.vuetifyjs.com/en/components/alerts

src/components/core/Alert.vue
kinow: Yup, the material color palette https://v15.vuetifyjs.com/en/framework/colors

oliver-sanders: Matches the Cylc logo pretty well!

oliver-sanders:
@datamel quick UI activity, I've done a code review so a functional test and quick sanity check should be sufficient.

hjoliver:
Looks nice to me :tada: ; leaving final review for @datamel though.

kinow:
Thanks @oliver-sanders and @datamel !

PR #464, opened 2020-06-11
Fix Graph component CSS height, and remove cytoscape-navigator

kinow:

Fixing issues found by @hjoliver and posted on RIOT.

1. Scroll bar displayed in the Graph, when viewed in the tabbed/Lumino view (`/#/workflow/five`)

- The `views/Graph.vue` displays the Graph component in full screen, which computes the height as `calc(100vh - 48)`
- The `views/Workflow.vue` displays the Graph component within Lumino, with extra margin/paddings. So `calc(100vh - 48px)` gives a height that is not taking the margin/padding in account, so a scroll bar is displayed.
- This PR changes that so that the Graph HTML element simply does a `height: 100%`, and its parents are CSS-flex. Also sets the layout `core-view` element to have `height: calc(100 - 48px)`, although later we may move this to another element as the toolbar is not present in all views (?)

2. Cytoscape navigator is blank, not working

- See this issue about the navigator and Vue: https://github.com/cytoscape/cytoscape.js-navigator/issues/45
- We customized the navigator CSS in our `cytoscape-custom.scss`
- We updated the Cytoscape versions since the code was initially written
- In one of these updates, the library changed the CSS classes of the navigator, making our custom CSS invalid, hence the empty navigator
- Re-enabling it is not too hard, but then it breaks in the tabbed view:

image

image

So I think it's safer to go with the suggestion of @hjoliver , and remove the cytoscape-navigator for now? We can add it again later if we decide to keep using Cytoscape :+1: 

**Requirements check-list**
- [x] Does not need tests (why? visual/style changes).

### DIFF
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -33,6 +33,16 @@
   }
 }
 
+// The following two entries are used to set the height of the content view (core-view), under the
+// toolbar, to the maximum. And the skeleton needs to maintain the height 100% otherwise its children
+// nodes with height: 100% or auto won't be displayed at all.
+#core-view {
+  height: calc(100vh - 48px);
+}
+.v-skeleton-loader {
+  height: 100%;
+}
+
 :export {
   fontRootSize: $font-size-root;
 }
--- a/package.json
+++ b/package.json
@@ -26,7 +26,6 @@
     "cytoscape": "^3.9.0",
     "cytoscape-dagre": "^2.2.2",
     "cytoscape-expand-collapse": "^3.1.2",
-    "cytoscape-navigator": "^2.0.0",
     "cytoscape-node-html-label": "^1.1.5",
     "cytoscape-panzoom": "^2.5.3",
     "cytoscape-popper": "^1.0.4",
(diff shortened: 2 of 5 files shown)


src/styles/index.scss
```diff
@@ -33,6 +33,16 @@
   }
 }
 
+// The following two entries are used to set the height of the content view (core-view), under the
+// toolbar, to the maximum. And the skeleton needs to maintain the height 100% otherwise its children
+// nodes with height: 100% or auto won't be displayed at all.
+#core-view {
+  height: calc(100vh - 48px);
+}
```
kinow, line 41: I wonder if we should leave it here, or move it to another element. At the moment that removes `48px` from the area in `views/Dashboard.vue`, as we don't have a Toolbar visible. Though it doesn't seem to impact the user, that's still not good I think?

hjoliver:
Thanks @kinow, that was quick.  Nice description of the problems and causes too :+1:

src/styles/index.scss
hjoliver: (I would need a chat with you to understand the problem here, if it matters).

oliver-sanders:
👍

Issue #465, opened 2020-06-18
Create a global "catcher" for errors in the UI


kinow:
**Describe exactly what you would like to see in an upcoming release**

At the moment we have both promises and async code. Not all promises are being `.catch()`'ed, meaning the error is just being propagated until it appears in the console.

The async code has the `try/catch` pattern. And other promises have the `setAlert` or `SET_ALERT` action or mutation. But writing that over and over in separate components it a bit tedious. That's not something that can be easily fixed with mixins. HOC could work, but would require a common interface for views/components.

**Additional context**

This article has, among other good tips and examples, an `ErrorBoundary` component that could be useful as reference: https://michaelzanggl.com/articles/vue-cleaning-up-components/

![image](https://user-images.githubusercontent.com/304786/85077615-d1f8f580-b216-11ea-8678-d1fc7741aac5.png)

It uses Vue's `errorCaptured`.

* https://vuejs.org/v2/api/#errorCaptured

That should work well with components. Not sure if that will work with views (maybe adding to `Default.vue` layout).

We could also deprecate the `Alert` model and associated code, if we are able to get something like BootstrapVue's [`toast`](https://bootstrap-vue.org/docs/components/toast).

Instead of using Vuex and requiring code to call the mutation/action, in BootstrapVue there's a global object to create toasts, which takes care to position the alerts.

![image](https://user-images.githubusercontent.com/304786/85077801-40d64e80-b217-11ea-8e82-792fb6e4201b.png)

That's simpler than having to import and map the action or mutation

![image](https://user-images.githubusercontent.com/304786/85077926-79762800-b217-11ea-9249-073113bc51db.png)

to then call it

![image](https://user-images.githubusercontent.com/304786/85077898-6bc0a280-b217-11ea-86b7-f0c08a716c8e.png)

Plus the code for `Alert.model.js` and the Vuex code in `src/store/index.js`.

**Pull requests welcome!**

kinow:
Spent a few minutes working on this, then found that instead of using an `Alert` component, it might be better to use a `Snackbar`. The `Snackbar` is the same component used by the `ConnectionStatus` component, that puts the red tooltip at the top of the screen when the WebSocket connection breaks.

Snackbars appear to be unique on the screen, without being able to be stacked yet, as in Buefy or BootstrapVue, see https://github.com/vuetifyjs/vuetify/issues/2384

[...]

PR #466, opened 2020-06-19
Update Vuetify from 2.2.6 to 2.3.9

kinow:
These changes partially address #465 

Updates to the latest version of Vuetify. Looked at each view of our app, with JupyterHub and `five` running. Everything looks OK.

Kudos to Vuetify developers. They kept backward compatibility as it's not a new major release, and deprecated the values changed, without breaking the user's code. Furthermore, they also added warning messages to each deprecated item, with instructions how to fix it :pray: 

image

So after updating the version, I simply fixed as they suggested, and then no more warnings in the console.

**Requirements check-list**

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "vue-router": "^3.1.3",
     "vue-spinner": "^1.0.3",
     "vue-the-mask": "^0.11.1",
-    "vuetify": "^2.3.1",
+    "vuetify": "^2.3.9",
     "vuex": "^3.1.1",
     "vuex-router-sync": "^5.0.0"
   },
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "vue-router": "^3.1.3",
     "vue-spinner": "^1.0.3",
     "vue-the-mask": "^0.11.1",
-    "vuetify": "^2.2.6",
+    "vuetify": "^2.3.1",
     "vuex": "^3.1.1",
     "vuex-router-sync": "^5.0.0"
   },
--- a/src/components/cylc/ConnectionStatus.vue
+++ b/src/components/cylc/ConnectionStatus.vue
@@ -17,11 +17,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 <template>
   <v-snackbar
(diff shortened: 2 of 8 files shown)


kinow:
Really no idea how to fix the 2 tests failing. Managed to fix one, with a lot of changes to Vuex, which is not a good sign. Probably something in the new version that changed requires us to change something simple in the way we create the object for testing... hopefully.

kinow:
Failing test should be fixed now. Lessons learned:

- We were using `mapState` directly in the component, which is fine, but means that writing a unit test for the component requires you to properly handle Vuex in the test too. Our store has many entries already, even though it's far from comparing to large applications. Which means it's harder to maintain tests and make sure everything works.
- By moving the `mapState` to the caller of the component, the unit tests gets much simpler. All you need to pass to the component is a `prop`. Testing components with props in Mocha+Chai is easy-peasy.
- I thought the connection-status component was also hidden, but turns out now the component is present, but the SnackBar component is hidden; I guess Vuetify has some reason for doing it that way
- Mocha API for parameterized tests is documented under their "[Dynamically Generating Tests](https://mochajs.org/#dynamically-generating-tests)" docs section. Not as nice as in other test frameworks, but it works well.
- The output of failing tests is not really helpful if you have similar tests, meaning that you have to scroll up-down to find the line of the error, and it may still not match due to sourcemap... so you need to locate the function name. A helpful tip is to add a message to the last function in the assertion (e.g. `expect(predicate).to.be(true, 'This is a helpful message')`.

kinow:
Set to 0.2, but can be merged later and included in 0.3 too, no hurry on this one :+1:

kinow:
Pushed to 0.3

kinow:
2.3.9 is out, but we can update that later. Should be much easier than upgrading 2.2 to 2.3.

kinow:
No issues updating to 2.3.9, so pushed the changes to this PR in my last commit. Tested some views, and found nothing wrong with how components were displayed :+1: also no build errors.

oliver-sanders (approved):
Tested as working, all seems good.

PR #479, opened 2020-08-03
Prepare for 0.3 release

kinow:

p.s.: NPM semver schema is a bit more restrictive. I've seen projects using `-SNAPSHOT` or `-alpha.1`. But only the numeric parts are taking into account for the version. The modifier with `alpha.1` can be used too, but it could be confusing for us as we have that in the metapackage 8.0a3. So I just bumped up to the next release number.

**Requirements check-list**


kinow:
I think one review will be enough :+1:

PR #480, opened 2020-08-18
Add a Lumino component, replacing the old Workflow component.

### LINKED ISSUE #461: Use a single widget wrapper in the Workflow view

In the Workflow component (not view) every widget-ized component gets its own wrapper. So there is the Tree component, and the TreeWrapper. The Graph component, and the GraphWrapper.

Instead we should be able to use a single wrapper. 

**Additional context**

There are multiple ways to achieve this, I think.

We can wait until we have a uniform interface for components (i.e. Tree, Graph, Dot, etc, would all have the same props).

Or, we can use a default slot and allow each component to request whatever parameters are necessary.

I used the second approach when creating the Lumino Vue example app linked in the JuptyerLab project page: https://github.com/kinow/vue-lumino/blob/6f01f9b32717d267f1372838d1e0d2a73fa91ac4/src/components/Lumino.vue#L84

I think that `VueComponentWrapper` can be ported back to Cylc UI, fixing this issue.

**Pull requests welcome!**

kinow:
These changes close #461

The Lumino component works with a single wrapper, different than the Workflow component that needed one wrapper per each component displayed in tabs/widgets.

The Mutation view also needed to have a `.name` property, as that's now used to decide what tab title to display (lower-case'd).

**Requirements check-list**


src/components/cylc/workflow/index.js
```diff
@@ -16,223 +16,9 @@
...
-    </div>
-  `
-})
 
 const EventBus = new Vue()
 
 export {
-  ContentWidget, TreeWrapper, GraphWrapper, MutationsWrapper, EventBus
+  EventBus
 }
```
kinow, line 24: For Vue 3 we must get rid of the event bus. But that's for another pull request.

src/views/Mutations.vue
```diff
@@ -85,6 +85,7 @@ export function associate (mutations, objects) {
 }
 
 export default {
+  name: 'Mutations',
```
kinow, line 88: The name of the Lumino tabs is defined dynamically with the `name` attribute of the widget wrapped object.

src/views/Workflow.vue
```diff
@@ -17,24 +17,47 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+          :id="widgetId"
+          :workflow-name="workflowName"
+      />
+      <mutations-view
+          v-for="widgetId of mutationsWidgets"
+          :key="widgetId"
+          :id="widgetId"
+          :workflow-name="workflowName"
+      />
+    </lumino>
```
kinow, line 42: We could go even further and remove the need to have computed properties for the widgets. So that if a new component type is added, we don't have to modify anything in the code.

But that's for another PR. This PR is only about the multiple wrappers (IOW there's room for improvement in this view, but at least the mess of wrappers should be fixed :+1: )

src/views/Workflow.vue
```diff
@@ -72,24 +98,49 @@ export default {
...
     tree: new CylcTree(),
-    isLoading: true
+    isLoading: true,
+    // the widgets added to the view
+    /**
+     * @type {
+     *   Object.<string, string>
+     * }
+     */
+    widgets: {}
```
kinow, line 108: A single map, so that we don't need one function for each delete event. We could have one single function for the add event too, but our components have different "signatures" (the props passed down, as well as events, slots, etc).

In the future we should be able to use this `widgets` Object, without the need to have `treeWidgets`, `graphWidgets`, `mutationsWidgets`, `etcWidgets`. But that's for a future PR :+1:

kinow:
> +1 Mainly functional review with the usual disclaimers about my Vue competence. Nicely commented.

Thanks Hilary!

src/components/cylc/workflow/Lumino.vue
```diff
@@ -0,0 +1,166 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+-->
+<template>
+  <div>
+    <div id="workflow-panel">
+      <div ref="main" id="main" class="pa-4 fill-height"></div>
```
oliver-sanders, line 20: I think this `id` won't necessarily be unique since it is in a template.

src/views/Mutations.vue
oliver-sanders: Nice.

src/views/Workflow.vue
```diff
@@ -17,24 +17,47 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+          :key="widgetId"
+          :id="widgetId"
+          :workflow-name="workflowName"
+      />
+      <mutations-view
+          v-for="widgetId of mutationsWidgets"
+          :key="widgetId"
+          :id="widgetId"
+          :workflow-name="workflowName"
+      />
```
oliver-sanders, line 41: Out of interest if the `:workflows`, `:workflow-name` interface were standardised between the components could we do this:

```vue
<component
  v-for="widget of widgets"
  :is="widget['type']"
  :key="widget['id']"
  :id="widget['id']"
  :workflow-names="[workflowName]"
  :tree="tree.root.children"
/>
```

src/views/Workflow.vue
```diff
@@ -72,24 +98,49 @@ export default {
...
       // add widget that uses the GraphQl query response
-      this.$refs['workflow-component'].addTreeWidget(`${subscriptionId}`)
+      this.addTreeWidget(subscriptionId)
     })
     EventBus.$on('add:graph', () => {
       // subscribe GraphQL query
       const subscriptionId = this.subscribe('graph')
       // add widget that uses the GraphQl query response
-      this.$refs['workflow-component'].addGraphWidget(`${subscriptionId}`)
+      this.addGraphWidget(`${subscriptionId}`)
```
oliver-sanders, line 143: Any reason the `subscriptionId` is passed differently for different components:

```js
this.addTreeWidget(subscriptionId)
this.addGraphWidget(`${subscriptionId}`)
```

src/views/Workflow.vue
kinow: I think so! We could probably also have a unique `addWidget` method, instead of `addTreeWidget`, `addGraphWidget`, etc. So when the new `Dot` or `WorkflowTable` etc components are created, we don't need to remember to change this file (or if we do, it's a very small change in a single place I think)

src/views/Workflow.vue
kinow: Ops, I believe it needs to be `${}`, so that it's a `String`, and not a `Number` (from what I recall I had a comparison or some `String` method call somewhere that would cause an issue in runtime). Will fix it later :+1: Good catch!

src/components/cylc/workflow/Lumino.vue
kinow: Good point, I will confirm I don't have any CSS using `#main` (I think I had in the project where I wrote this component before moving here)

kinow: Oh, it's actually used in the styles

image

Should we try to remove it here? Maybe use a class, or try `#workflow-panel { div:first-of-type { ... } }` ?

src/views/Workflow.vue
kinow: Hmmm, the rest of the code was actually using `number`. So I've actually changed the rest to use a `number`, and added JSdocs as well. Ran the IDE linter and it found no issues in the file related to typing, and tested with `five` locally.

src/components/cylc/workflow/Lumino.vue
oliver-sanders: Yeah, this is technically invalid HTML, a class should suffice?

kinow: Done

src/views/Workflow.vue
hjoliver: (Are you going to do this here, or is it for a follow-up PR?)

kinow: Planning to leave for a follow-up PR if that's OK, and focus on just removing the multiple wrappers in this one.

kinow:
Pushed one more commit, where now we are cleaning the `tree` when the deltas are stopped. Also fixes a bug, as this code was still using the event bus for widget deletion, but that's actually handled in the view now (good step, as we need to get rid of event bus for Vue 3)

src/components/cylc/workflow/Lumino.vue
```diff
@@ -0,0 +1,166 @@
...
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+-->
+<template>
+  <div>
+    <div id="workflow-panel">
```
oliver-sanders, line 19: I think `workflow-panel` should be converted to a class as well.

src/views/Workflow.vue
```diff
@@ -17,24 +17,47 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 <template>
   <div id="workflow-panel" class="fill-height">
```
oliver-sanders, line 19: Here too.

oliver-sanders (commented):
Looks good, tested using the `cylc-scan-nt` branches on cylc-flow and cylc-uiserver, a few unrelated bugs.

Component tree looks good:

Screenshot 2020-08-19 at 10 38 22

There's a graph bug, which may or may not be related to these changes?

Screenshot 2020-08-19 at 10 39 22

oliver-sanders (approved):
Premature approval if the graph bug is un-related.

kinow:
> Premature approval if the graph bug is un-related.

Replaced the HTML ID's by CSS. Tested and nothing broke after the change. CI OK too.

Couldn't reproduce the graph issue, but I don't think this change should break the graph component. Could be a legit bug in the graph view/component, though as we are talking about getting rid of it, I think we should be OK to merge this one.

PR #481, opened 2020-08-19
Update dependencies

kinow:

I used `ncu` to get a list of dependencies that had updates available. Then went through these dependencies individually, updating each one and checking if the build was working OK.

`vuetify` was not updated as it is being updated in #466.

Updating `eslint-config-vuetify` resulted in some weird errors in `yarn install`:

>error An unexpected error occurred: "could not find a copy of eslint to link in /home/kinow/Development/python/workspace/cylc-ui/node_modules/eslint-config-vuetify/node_modules".

So I reverted it and will take care of that dependency later (it is a dev dependency).

**Requirements check-list**


kinow:
Huh, more dependencies missing now. Some `core-js`, `eslint-plugin-import`... :sigh:

Always an adventure updating JS dependencies :nerd_face:

package.json
```diff
@@ -17,87 +17,87 @@
...
-    "@lumino/datagrid": "^0.3.1",
-    "@lumino/default-theme": "^0.1.8",
-    "@lumino/widgets": "^1.9.3",
-    "apollo-boost": "^0.4.4",
-    "apollo-utilities": "^1.3.2",
+    "@lumino/algorithm": "^1.3.3",
+    "@lumino/datagrid": "^0.12.0",
+    "@lumino/default-theme": "^0.4.4",
+    "@lumino/widgets": "^1.13.4",
+    "apollo-boost": "^0.4.9",
```
kinow, line 24: When we first added Apollo, we used `apollo-boost`. Which is just the `apollo-client` wrapped by several utility functions, and features like caching.

Later, after realizing we didn't need any of the features in `apollo-boost`, I started moving the `WorkflowService` to use just `ApolloClient`. That also became necessary as we started customizing the client for the WebSockets.

I will have to confirm, but I believe I have finished getting rid of `apollo-boost`, but **added the dependencies for the non-apollo-boost setting in the `devDependencies`**.

If so, we can remove the Apollo dependencies from `devDependencies`, move them up here, and maybe reduce our list of dependencies, as ApolloBoost includes more code/dependencies.

package.json
```diff
@@ -17,87 +17,87 @@
...
+    "apollo-utilities": "^1.3.4",
     "axios-fetch": "^1.1.0",
-    "cytoscape": "^3.9.0",
+    "cytoscape": "^3.15.2",
     "cytoscape-dagre": "^2.2.2",
-    "cytoscape-expand-collapse": "^3.1.2",
-    "cytoscape-node-html-label": "^1.1.5",
+    "cytoscape-expand-collapse": "^4.0.0",
+    "cytoscape-node-html-label": "^1.2.0",
```
kinow, line 30: I'm updating the Graph dependencies, assuming they won't require us to change anything in the Graph component or view :+1: if they demand further work, I will remove these changes to simplify the PR, since we may choose a different library for handling Graphs in Cylc UI.

kinow:
`eslint-plugin-import` missing after updating dependencies. Added with `yarn add -D eslint-plugin-import`.

package.json
hjoliver: I think we may want to remove the graph view next release, pending @oliver-sanders massive re-do. Probably not good to have an barely-usable graph view with wrong look-and-feel alongside our nice (and now performant, post deltas) tree view. (n-distance windows might save graph view performance, but still have the other problems...).

kinow:
Now `vue-markdown` is giving me issues after updating the other dependencies. Related to this issue: https://github.com/miaolz123/vue-markdown/issues/18

Looks like `vue-markdown` require us to include another dependency now: `yarn add babel-runtime --save` :man_shrugging:

kinow:
I can now build it with `yarn run build:watch`, but one warning in the `./srv/views/Mutations.vue`:

```
"export 'introspectionQuery' was not found in 'graphql'
```

:thinking:

package.json
kinow: At least removing dependencies and code is easier. Just need to remember to remove not only `cytoscape-` dependencies, but also the `tippy` I think. Would be nice if there were dependency groups in NPM, or if at least we could add comments to `package.json` (JSON limitation)

kinow:
> I can now build it with `yarn run build:watch`, but one warning in the `./srv/views/Mutations.vue`:
> 
> ```
> "export 'introspectionQuery' was not found in 'graphql'
> ```
> 
> thinking

Tree view OK.

Graph view OK.

Mutations view is now broken. Looking into that.

image

From the warning message, it looks like the GraphQL library stopped exporting `introspectionQuery`?

package.json
```diff
@@ -17,87 +17,89 @@
...
     "istanbul-lib-coverage": "^3.0.0",
-    "lodash": "^4.17.15",
-    "null-loader": "^3.0.0",
-    "nyc": "^15.0.1",
-    "sass": "^1.26.3",
-    "sass-loader": "^8.0.2",
-    "sinon": "^7.5.0",
-    "standard": "^14.3.1",
-    "subscriptions-transport-ws": "^0.9.16",
+    "lodash": "^4.17.20",
```
kinow, line 87: I think `lodash` is actually a runtime dependency. Maybe it's working now if another dependency is loading it (the Apollo libraries do). But this needs to be moved to the `dependencies` and removed from the `devDependencies`.

kinow:
Fixed build, no more warnings, but for some reason the Mutations are not loading now.

image

kinow:
The `Mutations` view was wrapped in a skeleton loader that had a `isLoading` flag always set to true. Funny that I think it was supposed to happen on `master` too.

Anyhoo, have modified it to leave the loading to the `Mutations` view, as that view is using another flag for whether it is loading or not.

For tomorrow:

- [ ] move lodash to dependencies, and remove from dev dependencies
- [ ] move apollo dependencies to dependencies, and remove them from dev dependencies
- [ ] another manual test to make sure there are no regressions
- [ ] change log 

Then it should be ready for review

oliver-sanders:
Looks good to me.

kinow:
PR ready for review.

The `master` branch produces 8MB of assets.

image

This branch produces 7.9MB.

image

Most likely related to updates in libraries that have either reduced their code size, or did changes that permit better tree-shaking. Could also be related to the removal of ApolloBoost, replaced by the fine-grained Apollo libraries (ApolloBoost just wraps several libraries into a library for convenience, plus some constructors to get clients with cache, etc).

Bruno

src/components/cylc/workflow/index.js
```diff
@@ -218,15 +218,7 @@ const MutationsWrapper = Vue.component('mutations-wrapper', {
   },
   template: `
     <div>
-      <component
-        is="VSkeletonLoader"
-        :ref="widgetId"
-        :loading="isLoading"
```
kinow, line 224: This is always true if you have nothing in the UI but the mutation. With Graph and Tree views, the subscriptions when started can change the view's `isLoading` state.

But the mutation widget doesn't trigger any subscription.

So instead we simply add a Mutation view. The view has a loading state, and displays "Loading..." immediately when loaded. Furthermore, with work to add mutations in the right places, this view might be hidden or removed I guess.

src/views/Mutations.vue
```diff
@@ -39,7 +39,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 <script>
 import gql from 'graphql-tag'
-import { introspectionQuery, print } from 'graphql'
+import { getIntrospectionQuery as getGraphQLIntrospectionQuery, print } from 'graphql'
```
kinow, line 42: Changed in GraphQL 15. Luckily they didn't remove the functionality.

Note that we already have a function in this module called `getIntrospectionQuery`, by coincidence. So if we imported like this, we could end up with weird build issues (e.g. recursively calling itself instead of calling the function from the `graphql` module). So I've renamed the import :+1:

kinow:
Rebased

oliver-sanders (commented):
Tested in offline mode, graph broken but that might be expected.

I'm getting a lot of "incorrect peer dependency" warnings (though I got some of these before):

```
warning "@lumino/datagrid > @lumino/coreutils@1.5.3" has unmet peer dependency "crypto@1.0.1".
warning "apollo-client > apollo-link@1.2.13" has incorrect peer dependency "graphql@^0.11.3 || ^0.12.3 || ^0.13.0 || ^14.0.0".
warning "apollo-link > apollo-utilities@1.3.3" has incorrect peer dependency "graphql@^0.11.0 || ^0.12.0 || ^0.13.0 || ^14.0.0".
warning " > graphiql@1.0.3" has unmet peer dependency "prop-types@>=15.5.0".
warning " > @babel/plugin-proposal-class-properties@7.10.4" has unmet peer dependency "@babel/core@^7.0.0-0".
warning "@babel/plugin-proposal-class-properties > @babel/helper-create-class-features-plugin@7.10.5" has unmet peer dependency "@babel/core@^7.0.0".
warning " > @cypress/code-coverage@3.8.1" has unmet peer dependency "cypress@*".
warning " > @cypress/webpack-preprocessor@5.4.4" has unmet peer dependency "@babel/core@^7.0.1".
warning " > @cypress/webpack-preprocessor@5.4.4" has unmet peer dependency "@babel/preset-env@^7.0.0".
warning " > @cypress/webpack-preprocessor@5.4.4" has unmet peer dependency "babel-loader@^8.0.2".
warning "@vue/cli-plugin-eslint > eslint-loader@2.2.1" has incorrect peer dependency "eslint@>=1.6.0 <7.0.0".
warning " > eslint-config-standard@14.1.1" has unmet peer dependency "eslint-plugin-promise@>=4.2.1".
warning " > eslint-config-vuetify@0.5.0" has incorrect peer dependency "eslint@^6.7.2".
warning " > eslint-plugin-vue@6.2.2" has incorrect peer dependency "eslint@^5.0.0 || ^6.0.0".
```

kinow:
>Tested in offline mode, graph broken but that might be expected.

Yup, the graph never worked in offline mode. That's because the mocked WorkflowService only returns the tree data.

>I'm getting a lot of "incorrect peer dependency" warnings during installation (though I got some of these before):

Yup, caused by the move to Graphql 15. Most of our dependencies have some ranges like graphql 13 | 14, or something similar to that. But apparently everything that we use is still working.

But if you prefer to reduce the risk by moving it down to 14 I can downgrade it too (takes just a few minutes as I'm deleting the `node_modules` when updating just to make sure I don't have any vestige of previous installations)

oliver-sanders (approved):
> Yup, caused by the move to Graphql 15.

Ok, if you've got a handle on what's causing it, looks like the warnings are harmless.

kinow:
Rebased

hjoliver (approved):

PR #483, opened 2020-08-24
Filter tree item by tasks name and tasks states

### LINKED ISSUE #482: Tree component filtering

Allow users to apply filters to the Tree component.

- [ ] Display the filter menu in the UI (see #87 and #337 for reference)
- [ ] Apply filter to the local component
  - [ ] Filter by task name
  - [ ] Filter by task state

**Additional context**

Similar to Cylc 7's GUI feature:

image

**Pull requests welcome!**

kinow:
These changes close #482

**Requirements check-list**


kinow:
Aiming at keeping the state of each component instance well separated. First step was to remove the event bus that was shared by instances.

Then wrote the simplest approach to filtering I could think of. Not sure if the best, but testing against five, families2, and complex, might tell whether this performs good enough or not.

It is performing a simple `if (haystack.contains(needle))`, so searching for `needle=oo` in each of `haystack=['foo', 'boo', 'snoopy']` returns true for all, so all the elements are returned.

image

kinow:
Build passing, and filters working on `five` and with the `one` (offline mode).

GIFrecord_2020-08-25_122232

Time to test with other workflows.

kinow:
Worked OK with `families2` and deeply nested family structure.

Also no problems that I could tell with `complex`. Actually, it is a relief to be able to filter that long list of items in the tree. And the memory of the browser tab kept about the same. I even had the impression it was a bit less, maybe for having less nodes to update?

image

kinow:
Still need to write unit and functional tests, but before that, I think it'd be good to have more people testing it, so marking it as ready for review :+1:

hjoliver:
This seems to work well, even with two tabs of the complex flow (different filtering in each) :+1: 

I know the UI aspects are preliminary, so the following comments may be premature or useless!
- I think the "view toolbar" will need to stick to the top of the tab, not scroll with the view
- the only performance issue I noticed (with the complex flow): typing into the task name filter entry seemed a bit laggy
- task state filtering doesn't seem to work quite right, but that's probably because the datastore still has the wrong state info sometimes (since SoD on the back end).  (Also worth noting that since SoD, some states - like succeeded - are mostly gone until we have n-distance windows and the UIS populating historical tasks that have left the scheduler task pool).

hjoliver:
And:
- I'd expect hitting Return after entering a string would activate the filter (I was briefly confused that nothing was happening until I realized I had to press the Filter button)

kinow:
Hi @hjoliver !

>I know the UI aspects are preliminary, so the following comments may be premature or useless!

Definitely not useless. Not sure about premature, but I'm planning to improve the UI elements after I make sure I understand what it must look like with @oliver-sanders .

> I think the "view toolbar" will need to stick to the top of the tab, not scroll with the view

Also think that'd be useful. Nobody wants to scroll all the way to the top to change the filter, over and over again. +1

> the only performance issue I noticed (with the complex flow): typing into the task name filter entry seemed a bit laggy

That is probably due to the number of elements in the UI, not sure how much related to the filter. When I add a second tree view to the Lumino widget with complex, it also becomes laggy (master/this branch).

I'm hoping the improvements for the data store will improve performance for both the Lumino widget and for the filtering.

> task state filtering doesn't seem to work quite right, but that's probably because the datastore still has the wrong state info sometimes (since SoD on the back end). (Also worth noting that since SoD, some states - like succeeded - are mostly gone until we have n-distance windows and the UIS populating historical tasks that have left the scheduler task pool).

Initially I was looking for running and succeeded, but it never matched anything. Then I realized I needed to match whatever as in the task states, not in the job. I spent some time confirming, and at least I think the filter is applied correctly against the states of tasks returned in deltas.

These may need some adjustments later.

Another thing too, is that I'm using a hard-coded list of states, from an `Enum` that we use in other parts of the UI. While at least this gives us a central list of states, we could also consider fetching the list of states when the application starts from the backend, or just reviewing if this `Enum` needs to be updated post SoD/deltas.

> And:
> 
>     * I'd expect hitting Return after entering a string would activate the filter (I was briefly confused that nothing was happening until I realized I had to press the Filter button)

Yup, the first version of the filters (from yesterday :laughing: ) had `v-on:key.enter` I think, to capture the enter. I removed it after I added the combobox for task states. As it wouldn't be consistent?

Users could hit enter in the name. But what about the combobox? They would need to use the button. So would we have both, a button and support enter? Or just one option? In the end I went with one option only, the button to filter, to simplify, knowing that this wasn't the final UI.

So open for suggestions on how users should trigger the filter :+1: the effort to implement will be the same I guess.

Thanks a ton @hjoliver !

hjoliver:
> Another thing too, is that I'm using a hard-coded list of states, from an Enum that we use in other parts of the UI. While at least this gives us a central list of states, we could also consider fetching the list of states when the application starts from the backend, or just reviewing if this Enum needs to be updated post SoD/deltas.

We are heading toward the definitive list of task and job states as described here:
- [task/job states - see New column](https://cylc.github.io/cylc-admin/proposal-state-names.html#taskjob-states
- [New abstract task states](https://cylc.github.io/cylc-admin/proposal-state-names.html#new-abstract-task-states)

I'm pretty confident that those will stick - i.e. will change very infrequently if ever.  Also, if the UI fetched a list of states from the back end and found a new one, it wouldn't know how to display it anyway.

oliver-sanders:
Looks good!

Unfortunately there still seems to be some interaction of trees for running workflows.

filter-bug

hjoliver:
> Unfortunately there still seems to be some interaction of trees for running workflows.

What's the interaction? (I've stared at your gif for a bit without seeing it!)

oliver-sanders:
In the above demo I filter the left-hand tab to show only running tasks, which woks fine, however, as the workflow continues to run new running tasks do not appear in the filtered view (as they would if there were only one tab).

hjoliver:
Huh, you're right, good spotting.

kinow:
Ops! Good catch!!!

It was due to how I defined the reactivity between the workflow tree, and the filters being re-applied. It was configured to react on updates to the tree object, which never (rarely?) changes, as instead we change parts of that object with deltas.

Fixed by defining `deep: true`. Performance should not be that great now. I have something my mind that I **think** could improve the performance of deltas being applied. Instead of applying deltas with reactivity on each element of the tree, I would like to try applying the deltas first in batch, then once the `applyDeltas` function is done, triggering the reactivity programmatically.

That way Vue should be less busy keeping track of heaps of Observer/Proxy/etc objects, and have more time to handle DOM/VDOM. But that's for later, unless the performance of the filters is not good enough here :+1: 

GIFrecord_2020-08-26_235421

kinow:
Ah! On my computer, in development mode, the memory wasn't that bad, but I could tell there was a slight delay between updates to the filtered items, when comparing with the un-filtered three widget.

hjoliver:
Forgot to mention above, we've also lost the "grey tree blank" (forgot what it's called!) pre-loading graphic, in the tabs.

hjoliver:
Performance actually seems fine to me, with your latest update, on the complex flow :tada:    So that should be good enough for this PR, IMO :+1:

kinow:
> Forgot to mention above, we've also lost the "grey tree blank" (forgot what it's called!) pre-loading graphic, in the tabs.

I accidentally removed it in #461 . Preparing a new PR for that after this one is merged (needs to change the Workflow view too, just to avoid conflicts :)

oliver-sanders:
> forgot what it's called!

Skeleton.

src/components/cylc/tree/Tree.vue
```diff
@@ -17,21 +17,81 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+          dense
+          flat
+          hide-details
+          multiple
+          outlined
+          placeholder="Filter by task state"
+          v-model="tasksFilter.states"
+        >
+          <template v-slot:item="slotProps">
+            <Task :status="slotProps.item.value.toLowerCase()" :progress=0 />
```
oliver-sanders, line 52: Out of interest, I wouldn't have expected that we would need to convert task state to lower case?

src/styles/cylc/_tree.scss
```diff
@@ -0,0 +1,106 @@
+/**
```
oliver-sanders, line 1: I have to say I much prefer having code separated into files of the correct extension 👍.

src/components/cylc/tree/Tree.vue
```diff
@@ -45,36 +105,102 @@ export default {
...
+      let filtered = false
+      if (['cyclepoint', 'family-proxy'].includes(node.type)) {
+        for (const child of node.children) {
+          filtered = this.filterNode(child) || filtered
+        }
+      } else if (node.type === 'task-proxy') {
+        if (this.filterByTaskName() && this.filterByTaskState()) {
+          filtered = node.node.name.includes(this.tasksFilter.name) && this.tasksFilterStates.includes(node.node.state)
+        } else if (this.filterByTaskName()) {
+          filtered = node.node.name.includes(this.tasksFilter.name)
```
oliver-sanders, line 189: It would be good to upgrade this plain text search to a regex to match current `cylc review` functionality as task names can become very complex.

oliver-sanders (approved):
Wohoo, this is getting pretty cool.

Screenshot 2020-08-27 at 15 18 21

I really like the task status icons in the filter combo-box.

oliver-sanders:
(Because switching workflows is broken at the moment I can't test how the filters are effected by changing switching between workflows).

src/components/cylc/tree/Tree.vue
kinow: At the moment it's necessary. I can't recall exactly why, but the `TaskState` enum has values like `static SUBMITTED = new TaskState('SUBMITTED')`. Where the `String` value is set to `SUBMITTED`.

Here the component will use the `item.value` (`SUBMITTED`), which wouldn't match the CSS class.

image

I think we will have to review that enum, as there are task states in other parts of the code that are not using it yet, and I **think** there is some discussion somewhere about changing the states... or was it organizing jobs/tasks states?

kinow:
> (Because switching workflows is broken at the moment I can't test how the filters are effected by changing switching between workflows).

Fixed in the last commit. Switching between workflows should be working now :crossed_fingers: 

Thanks!

hjoliver:
Is this ready to go now @kinow ? (You haven't checked the "tests" checkbox in the description, or commented on tests?).

kinow:
> Is this ready to go now @kinow ? (You haven't checked the "tests" checkbox in the description, or commented on tests?).

Writing unit tests today, 50% done, just finishing the unit tests and then adding an e2e test.

kinow:
Writing these tests I'm finding a few small issues. @oliver-sanders would it be OK if I asked you to review this one again, probably Monday your time, please?

The issues I found are no blockers, but I'd prefer to fix them as I'm writing the unit tests anyway. Example issues (some could have been introduced when I updated code to fix linter issues too):

- If you filter by name, then while workflow is running, type another task name, then the filter is active without pressing the Filter button
- If you type something in the states field, it autocompletes for you, but hitting enter results in a browser console error
- When running the filters, hitting clear may apply the filter without hitting the Filter button

I'm spending more time with unit tests, and intend to write only one or two functional tests. That's because we are using the simple input and combo box UI elements from Vuetify, and I'm not sure if they match our UI design, nor if they work well with mobile, theming, or if they would have to be changed after we add new Dot/Graph views, etc. So I'm guessing these could change later, so no point in writing a lot of functional tests :+1:

kinow:
Most tests written, did some smoke tests, everything seems to be in order. But it's Friday 7PM, so this code comes with no guarantee :grin: will review the code, and probably write a few more tests to increase coverage on Monday NZ time. Then it should be ready for review after that :+1:

src/components/cylc/tree/Tree.vue
oliver-sanders: Ah I see, thanks for clearing that up.

I had a branch on Cylc Flow which converted task states to an enum which got exposed to GraphQL which would probably have made life easier, but unfortunately, serialising it for the DB/protobuf was a bit of a pain.

oliver-sanders:
> @oliver-sanders would it be OK if I asked you to review this one again

No probs.

kinow:
Last commit was to fix CSS styles. I was having a hard time looking at the button with a slightly different height than the other two input (text & select). Had to hard-set it to `40` (it's a `number` per the Vuetify docs, I simply matched the value from the input fields).

On mobile, the three elements (text & select & button) get squeezed together. Might need to display one per line. But didn't want to spend too much time working on this UI as it may change later :+1: Just enough to calm down my OCD :grin:

kinow:
Ready for review! :tada:

hjoliver (approved):
(My usual mainly-functional review on this part of this system) - looks really good, further improvements can be follow-ups IMO. :+1:

dwsutherland (approved):
Haven't delved into the code changes, but functionally it's a pass 🚀

oliver-sanders:
Looks ok, unfortunately I've totally lost the ability to view workflows at the moment, can't even view them by URL now!

Most likely unrelated?

Screenshot 2020-09-01 at 16 25 08

We've got two approvals anyway so happy for you to merge.

kinow:
The issue might go away if you delete your `node_modules` and re-install with `yarn install` @oliver-sanders :crossed_fingers: but shouldn't be related to the latest changes I think.

oliver-sanders:
Tried that I'm afraid.

PR #488, opened 2020-09-01
Add Vuetify skeleton loader component wrappers back to Workflows view

### LINKED ISSUE #484: Add UI loading state back to Workflow widgets

Accidentally removed in #461 , just need to wrap the component in a Vuetify skeleton loader, and work-around the widget tab name (dynamically generated with `wrappedComponent.name`, which is wrong as we have `view > lumino > wrapper > sleleton > wrappedComponent` with the skeleton loader, so the name displayed is something like `vskeletonloader0`)

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #484 

**Requirements check-list**


package.json
```diff
@@ -18,16 +18,16 @@
   },
   "dependencies": {
     "@lumino/algorithm": "^1.3.3",
-    "@lumino/datagrid": "^0.12.0",
-    "@lumino/default-theme": "^0.4.4",
-    "@lumino/widgets": "^1.13.4",
+    "@lumino/datagrid": "^0.14.0",
+    "@lumino/default-theme": "^0.5.0",
+    "@lumino/widgets": "^1.14.0",
```
kinow, line 23: Updating the dependencies too since it's a little related to Lumino. Everything appears to be working OK :+1:

src/views/Workflow.vue
```diff
@@ -20,24 +20,38 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
       <mutations-view
-          v-for="widgetId of mutationsWidgets"
-          :key="widgetId"
-          :id="widgetId"
-          :workflow-name="workflowName"
+        v-for="widgetId of mutationsWidgets"
+        :key="widgetId"
+        :id="widgetId"
+        :workflow-name="workflowName"
+        tab-title="mutations"
```
kinow, line 54: Needed to keep the name lower case. Now the default is to show whatever is the component name (which in this case is `Mutations`, with first letter capitalized), unless the user provides a prop for the title.

kinow:
To test, just start the UI, look at a tree view and check if you can see the skeleton (it may appear quickly, depending on how fast loading is in your env), then close the tab. Add a new tree view, and it should display the skeleton loader (again, since you closed all tree views, so there's no more deltas subscription running, adding a new one starts the loading process again).

oliver-sanders (approved):
Does the trick.

Sometimes the skeleton doesn't appear for some reason (see the first attempt in this screen-grab):

Kapture 2020-09-02 at 14 01 23

However, I'm not too worried about that.

hjoliver (approved):

Issue #389, opened 2020-01-29
sidebar: collapse should always be an option


oliver-sanders:
Presently the sidebar can only be collapsed in small windows. We should allow users to collapse the sidebar at all times.

**Pull requests welcome!**

hjoliver:
Is there a standard-ish way to do this? e.g. burger icon = "open the collapsed sidebar"; collapsed-burger icon = "collapse the open sidebar"?

oliver-sanders:
Burger seems to get used for this a bit.

Personally I think the slide-in nature of a sidebar is more clearly conveyed using an arrow of some kind as a burger bar typically implies a more in-context list (i.e. a dropdown), principle of least surprise and whatnot. I think I used an expand arrow attached to a "tongue" in the early design sketches.

hjoliver:
>  a burger bar typically implies a more in-context list  

I've been thinking of it as not-inappropriate, due to list-of-suites (gscan) sort of resembling a menu

But we could consider your arrow idea.

kinow:
There is a method in the drawer (the name of the component used for the sidebar) to toggle whether it's displayed or not.

That method is used in the responsive mode.

The pre-made layouts of Vuetify place the burger menu in the toolbar. But our toolbar is only displayed when looking at a view with a running workflow.

![image](https://user-images.githubusercontent.com/304786/87480886-5ad24800-c682-11ea-956c-2d423434f332.png)

We could display the mini version of the toolbar instead, and either expand on hover, or have a button to display it.

![image](https://user-images.githubusercontent.com/304786/87480944-79384380-c682-11ea-96b8-a3eea4c88e8f.png)

The issue with expand on hover is that it may trigger DOM reflow. So if you are using it on a view with lots of elements (e.g. graph, or tree, or lumino views) it could cause a brief lag.

MetRonnie:
Superseded by #1137

hjoliver:
Not sure which PR did it, but we can now drag to resize the side bar to zero, in which case the burger appears and can be used  to restore it.

PR #489, opened 2020-09-03
Tidying up styles, event bus, and fixing icons

kinow:

Not a major bug, since it is working in 0.2, but after updating Vuetify I noticed some icons are missing. That's due to icons being renamed in Vuetify (:man_shrugging:). This PR fixes that, and a few other things:

- [x] Fix Vuetify (material-design-icons) icons
- [x] Move the last styles out of components, into their own `_style.scss` file
- [x] Remove remaining event buses, passing the event listeners down from the parent (easier to migrate to Vue 3 after this)

**Requirements check-list**


src/components/cylc/Toolbar.vue
```diff
@@ -197,9 +197,3 @@ export default {
   }
 }
 </script>
-
-<style>
-#core-app-bar a {
-  text-decoration: none;
-}
```
kinow, line 204: Tested, and made no difference. I was googling for an image for the Dashboard for the icons, and found this old screenshot of Cylc UI :grimacing: 

D8Vr_CkUEAAtxdr

That was with Vuetify too, and we do have some stuff in our code reminiscent from the initial layout :sweat_smile: so might be that this was necessary before, and eventually after we modified the toolbar, the links didn't get an underline anymore :+1:  (at least that's my guess)

kinow:
After this change, I believe we don't have any more event buses around. They are deprecated in Vue 3 - https://dev.to/chenxeed/awesome-breaking-changes-in-vue-3-if-you-migrate-from-vue-2-3b98

And it's simpler to pass the listeners down. This way we don't have to worry about adding a `this.on('event', method)` accompanied by `this.off('event', method)` to avoid memory leaks. Vue will take care of cleaning up. So less code to read/maintain.

Had to split the existing `Toolbar` component. This PR has two toolbars, one that only takes care of toggling the display value of the Drawer (i.e. when you resize browser, we may display this toolbar). It is not displayed for Workflow views (i.e. Workflow view, Graph view, or Tree view).

Instead, those views display the Workflow View. That view has a Toolbar component and a Lumino component. This way, the view instantiates both, and tells the Toolbar "Hey, tell me when someone clicks 'Add Tree View' and then I will take care to tell the Lumino instnace about it" :sunglasses:

tests/unit/components/cylc/toolbar.vue.spec.js
```diff
@@ -15,46 +15,35 @@
...
-      return resolve(true)
-    })
-  },
-  stopWorkflow: function () {
-    return new Promise((resolve) => {
-      return resolve(true)
-    })
-  }
-}
-
```
kinow, line 52: Now the `cylc/Toolbar` is only for responsive mode. So these tests were moved.

tests/unit/components/cylc/toolbar.vue.spec.js
```diff
@@ -74,78 +63,25 @@ describe('Toolbar component', () => {
...
     })
     expect(store.state.app.drawer).to.equal(null)
     // empty wrapper before responsive is set to true
     expect(wrapper.find('button.default').exists()).to.equal(false)
     // let's make it responsive, so that the burger menu is visible
-    wrapper.vm.$data.responsive = true
-    await wrapper.vm.$nextTick()
+    await wrapper.setData({
+      responsive: true
+    })
```
kinow, line 82: The `setData` function does pretty much the same I was doing before, but it i) encapsulates the behaviour (so if Vue Test Utils changes it, I won't have to learn how to sync data in wrappers), and ii) takes care of refreshing the data state for the test (I was waiting for the next tick, but not sure if that's enough, or if I had to wait on more things) - https://vue-test-utils.vuejs.org/api/wrapper/#setdata

src/views/Workflow.vue
```diff
@@ -16,44 +16,51 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-        :key="widgetId"
-        :id="widgetId"
-        :loading="isLoading"
-        type="list-item-three-line"
-        tab-title="tree"
+  <div>
+    <toolbar
+      v-on:add-tree="this.addTreeWidget"
+      v-on:add-graph="this.addGraphWidget"
+      v-on:add-mutations="this.addMutationsWidget"
```
kinow, line 23: This is basically what replaces the Event Bus.

Instead of having the `Toolbar` component emitting global events, we give it a listeners for each custom event (`add-tree`, `add-graph`, and `add-mutations`).

This way `Toolbar` is able to use those listeners as `v-on:click=$listeners['add-graph]`, effectively sending an event to the listener of `add-graph` (this `Workflow` view in this case).

Further down, this view takes care to add a tree widget by communicating with the tree widget.

No global events, no event bus, no need to worry about cleaning up event listeners, and easier to understand how components are linked (good practice in Vue to send data top down this way :+1: )

kinow:
Coverage should increase a bit as I'm throwing some code away. There was one part that I had to duplicate code, it was in the templates of the two toolbars, as mixins can only change the template if we use functional components (which we should be able to, but that's for another discussion).

I think the template could be simpler depending on #389 , as that show/hide drawer burger button could be moved elsewhere.

Friday-code, so no issues if there's some improvements necessary. Will update it next week if necessary :+1: this is a low priority issue.

hjoliver (approved):
LGTM :+1:

oliver-sanders:
Looks good, I'm getting some traceback in the console at the moment.

This appeared when I loaded the tree view:

Screenshot 2020-09-15 at 13 34 59

This appeared when I loaded the graph view:

Screenshot 2020-09-15 at 13 34 03

kinow:
> Looks good, I'm getting some traceback in the console at the moment.
> 
> This appeared when I loaded the tree view:
> 
> Screenshot 2020-09-15 at 13 34 59
> 
> This appeared when I loaded the graph view:
> 
> Screenshot 2020-09-15 at 13 34 03

That might be because I created this PR before adding the fixes so that the UI works with stopped workflows (async PR's).

Just rebased it and force-pushed. Could you try again, please @oliver-sanders ? Thanks!

oliver-sanders:
>  Could you try again, please @oliver-sanders ? Thanks!

No probs. Tested again, the tree view and mutations views are working fine with no unexpected console errors. The graph view is now giving me a promise error:

Screenshot 2020-09-16 at 13 37 44

kinow:
Could it be some old files in your `node_modules` folder? It seems to be working fine for me.

image

image

hjoliver:
I get at least one of those errors on cylc-ui master too. Graph view: cannot create edge from non-existent source.

Then after closing the graph view I get errors from the tree view:
```
Uncaught TypeError: parent is undefined
    removeTaskProxy cylc-tree.js:426
    applyDeltasPruned deltas.js:56
    applyDeltasPruned deltas.js:53
    handleDeltas deltas.js:184
    applyDeltas deltas.js:257
    next Workflow.vue:215
    next workflow.service.js:85
    notifySubscription Observable.js:135
    onNotify Observable.js:179
    next Observable.js:235
    next Observable.js:327
    notifySubscription Observable.js:135
    onNotify Observable.js:179
    next Observable.js:235
    iterateObserversSafely iteration.js:4
    iterateObserversSafely iteration.js:4
    next Concast.js:23
    opId client.js:163
    processReceivedData client.js:532
    onmessage client.js:485
```

hjoliver:
(No errors from the tree view alone, but opening the graph view seems to screw things up even after closing it again).

kinow:
Thanks for testing and trying to reproduce the issue @hjoliver !

I think that's not related to the change here. This change only affects styles, icons, and replaces the event bus by Vue listeners. It shouldn't affect the deltas.

>(No errors from the tree view alone, but opening the graph view seems to screw things up even after closing it again).

Yup, found the issue in Cylc Flow. Adding the graph would trigger it. I think that's the same issue @oliver-sanders is having.

image

https://github.com/cylc/cylc-flow/issues/3658

hjoliver:
No errors from the tree view alone on this branch :+1:

hjoliver:
Left two tabs of tree view running for ages, with the "family" flow, no errors (on this branch).

@oliver-sanders - IMO we can blame the graph view (whether it's broken itself, or is triggering a bug in what the UIS sends) and merge this.

oliver-sanders:
Fair enough.

oliver-sanders (approved):

PR #490, opened 2020-09-06
Reduce the size of produced artefacts

### LINKED ISSUE #164: Check how to reduce size of produced artefacts

kinow:
These changes are related to #164  (does not close, as Graph/Mutation are still pending some diet too, oh, and the Vuetify [a-la-carte](https://vuetifyjs.com/en/customization/a-la-carte/) seems to miss a few components and bring imports from `src` instead of from `lib`, or twice from both)

**Requirements check-list**
- [x] Does not need tests (why? build only, plus e2e tests should detect some changes, plus manual smoke test).
- [x] No change log entry required (why? e.g. invisible to users; users won't notice visually…though performance-wise in a networked environment it would be noticeable).


kinow:
Size of artefacts on master

image

kinow:
Size after removing fonts

image

We are packaging several font files, and including those in the URLs. Some for compatibility with old browsers, though we cannot guarantee the rest—besides fonts—work in the UI (i.e. some functionality in the tree view could be broken in IE9, so no point in giving nice fonts, easier to just support IE Edge+).

woff2 is the newest format, but not supported by all the browsers. woff is the web format supported by most browsers. eof, svg, and ttf are for older browsers too.

kinow:
Size after shipping only the icons that we actually use in the UI, not all the collection of icons in Material Design Icons (**assets was 7.8MB, modules 4.5MB, dependencies 4.0MB**)

image

Normally you are able to apply [tree-shaking](https://en.wikipedia.org/wiki/Tree_shaking) with webpack (although there are alternatives like rollup that are a bit more powerful). While that works well with JS assets, the Material Design icons we were using were provided by `@mdi/fonts`, which includes every SVG file in the Material Design Icon collection.

The `@mdi/js` library requires a bit more of work. We have to define the icons we want to use, in our JS code. Which is no biggie IMO, as the JS has both template and JS. So instead of

```js
<template>
  <v-icon>mdi-stop</v-icon>
</template>
```

we have

```js
<template>
  <v-icon>{{ pathToSvgFile }}</v-icon>
</template>

<script>
import { mdiStop } from '@mdi/js'

export default {
  ...
  data () {
    pathToSvgFile: mdiStop // mdiStop has the path to the SVG, and here we set a variable in our JS code, to be used by the template
  }
}
</script>
```

This way, we have only the selected resources, as `@mdi/js` is tree-shakeable :man_dancing:

kinow:
The Graph component has quite a few dependencies. Removing it, I think the modules site would be reduced by near 0.7MB.

The Mutations component also brings a few due to Markdown and other dependencies. Couldn't find a way to shrink it down.

The GraphiQL brings React, React DOM, and CodeMirror. This one also has a good chunk, probably about 0.5MB.

So this is ~1.5MB of dependencies allocated for these components/views. :thinking: might think more about it later.

This PR has conflict changes with #489, so will wait for that one to go before moving forward with this one.

src/components/cylc/Drawer.vue
```diff
@@ -57,30 +55,16 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-        active-class="primary grey--text text--darken-3"
-        class="v-list-item"
-      >
-        <v-list-item-action>
-          <v-icon>{{ link.icon }}</v-icon>
-        </v-list-item-action>
-        <v-list-item-title
-          v-text="link.text"
-        />
-      </v-list-item>
```
oliver-sanders, line 79: What's going on here?

oliver-sanders (commented):
Looks good, a couple of changed in Drawer.vue I don't understand.

src/components/cylc/Drawer.vue
kinow: Unused code. We used to have more links in the drawer, before 0.1, before matching the design sketch.

This code is from the tim creative design, where it would add views links dynsmically.

hjoliver (approved):
LGTM, nice reduction :+1:

oliver-sanders (approved):

hjoliver:
(Still a Draft @kinow )

kinow:
> (Still a Draft @kinow )

Ah, yes. I remember why I left it as a draft. It's because it would be easier to get the #489 issue first in, with corrected icons. This PR will then replace the icons-library (while still keeping the correct icons).

kinow:
Almost done now. Just chasing the console error "vue.esm.js?a026:6796 Error: <svg> attribute width: Expected length, "large".".

It seems to be caused by something in the `ConnectionStatus.vue` component. :nerd_face:

kinow:
The problem was `<v-icon size="large" />` in the `v-icon` component. Looks like in the latest updates, they moved it to a directive `<v-icon large />`.

image

https://v15.vuetifyjs.com/en/components/icons

kinow:
Final numbers for now are **5.1MB** of assets (was 9.2MB), **4.2MB** modules (was 4.5MB), and **3.7MB** dependencies (was 4.0MB).

image

Also updates the packages (no major updates).

hjoliver (approved):
Re-approving LGTM :+1:

oliver-sanders (approved):

Issue #333, opened 2019-12-16
Tree: subtle refactor proposal


oliver-sanders:
A bit of spit and polish for the Tree view?

![tree-view-refactor](https://user-images.githubusercontent.com/16705946/70875744-e9fb7c80-201b-11ea-84c3-8557da745824.png)

Questions:

* [x] Change icon ordering?
* [x] Collapse multiple jobs for task nodes?
* [ ] ~~Chronological ordering top-bottom?~~
* [ ] Configurable chronological ordering: bottom-top (most recent cycle point at top) vs top-bottom

Progress:

* [x] Change icon ordering? #606 
* [x] Collapse multiple jobs for task nodes?
* [ ] Chronological ordering bottom-top? (most recent cycle point at top)

**Pull requests welcome!**

kinow:
> Change icon ordering?

+1 I like how jobs are aligned in the proposed new version. Also found the multiple job tidier, and probably easier to maintain in case a task gets something like hundreds of jobs (which would probably break our layout? Or at least trigger DOM reflow :disappointed: ).

> Collapse multiple jobs for task nodes?

+1 If it is really important to know how many jobs failed, and what are the status of these jobs, perhaps we could add a tooltip to the visible job - and display the icon as "decked" jobs, instead of single one, so users are aware it has multiple jobs.

>Chronological ordering top-bottom?
>Chronological ordering bottom-top?

I would prefer to see the most recent information first, but not sure if some users wouldn't want the other way round.

oliver-sanders:
> I would prefer to see the most recent information first

I think I'm leaning this way as well, it's familiar to many due to the rose bush / cylc review interface.

hjoliver:
OK, so we've agreed on chronological order bottom-top (i.e. most recent cycle point at top). I'll update the description...

hjoliver:
"Change icon ordering" and "multiple job collapsing" kind of go together.  job-icon-before-name requires collapsing jobs, or else names can get pushed off to the right.  The new order looks nice and clean, with good alignment. *But* I think it can be good to see if a job had multiple tries without having to expand the tree item.

Can we have our cake and eat it too?  (i.e. allows users to choose or toggle between new-order-collapsed-jobs, and old-order-expanded-jobs)

oliver-sanders:
> But I think it can be good to see if a task had multiple tries without having to expand the tree item.

[...]

PR #492, opened 2020-09-08
Sort tree view by cyclepoints and task names

### LINKED ISSUE #469: Consistent sorting of cycle points in the tree view
**Describe the bug**

Using the latest version of the deltas branch in the UI, and with the SoD branch, I got some cycle points out of order, apparently.

image

I think it should be either 8 - 9 - 10, or 10 - 9 - 8, but definitely not 10 - 8 - 9.

SoD cylc flow, and UI deltas branch. UIS on master.

**Expected behavior**

Consistent order of cycle points.

**Screenshots**

image

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #469 

These changes partially address #333 (see https://github.com/cylc/cylc-ui/issues/333#issuecomment-651600058)

Cycle points showing newest data first. Tasks in alphabetical order, by name. Jobs by submit number, reverse (we display the first only, which is the latest).

We do not need to sort the deltas when updating or removing data. As the data returned is used to update existing data and nodes in the UI.

We are also not sorting the cycle points. The reason is that the cycle points are added to the tree one after the other. So if we already have data in the UI, adding new data, even if in order to the JS data, the browser won't display in order. It will just append the new data (in order) after the existing data.

So because of that, the cycle points are returned in any order. And then we modify the existing cycle point sorting code (i.e. we were already sorting in JS, so no performance issue here IMO) to just show the newest cycle point first.

**Requirements check-list**

### DIFF
--- a/src/graphql/queries.js
+++ b/src/graphql/queries.js
@@ -53,41 +53,41 @@ fragment WorkflowTreeDeltas on Deltas {
 fragment WorkflowTreeAddedData on Added {
   workflow {
     ...WorkflowData
-    cyclePoints: familyProxies(ids: ["root"], ghosts: true) {
+    cyclePoints: familyProxies (ids: ["root"], ghosts: true) {
       cyclePoint
     }
-    taskProxies(sort: { keys: ["cyclePoint"] }, ghosts: true) {
+    taskProxies (sort: { keys: ["name"], reverse: false }, ghosts: true) {
       ...TaskProxyData
       jobs(sort: { keys: ["submit_num"], reverse:true }) {
         ...JobData
       }
     }
-    familyProxies (exids: ["root"], sort: { keys: ["firstParent"]}, ghosts: true) {
+    familyProxies (exids: ["root"], sort: { keys: ["name"] }, ghosts: true) {
       ...FamilyProxyData
     }
   }
-  cyclePoints: familyProxies(ids: ["root"], ghosts: true) {
+  cyclePoints: familyProxies (ids: ["root"], ghosts: true) {
     cyclePoint
   }
-  familyProxies (exids: ["root"], sort: { keys: ["firstParent"]}, ghosts: true) {
+  familyProxies (exids: ["root"], sort: { keys: ["name"] }, ghosts: true) {
     ...FamilyProxyData
   }
-  taskProxies(sort: { keys: ["cyclePoint"] }, ghosts: true) {
+  taskProxies (sort: { keys: ["name"], reverse: false }, ghosts: true) {
     ...TaskProxyData
   }
-  jobs(sort: { keys: ["submit_num"], reverse:true }) {
+  jobs (sort: { keys: ["submit_num"], reverse:true }) {
     ...JobData
   }
 }
 
 fragment WorkflowTreeUpdatedData on Updated {
-  taskProxies(sort: { keys: ["cyclePoint"] }, ghosts: true) {
+  taskProxies (ghosts: true) {
     ...TaskProxyData
   }
-  jobs(sort: { keys: ["submit_num"], reverse:true }) {
+  jobs {
     ...JobData
   }
-  familyProxies (exids: ["root"], sort: { keys: ["firstParent"]}, ghosts: true) {
+  familyProxies (exids: ["root"], ghosts: true) {
     ...FamilyProxyData
   }
 }
--- a/src/mixins/treeitem.js
+++ b/src/mixins/treeitem.js
@@ -1,0 +1,82 @@
+/**
+ * Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import orderBy from 'lodash/orderBy'
+
+const treeitem = {
+  methods: {
+    /**
+     * Sort children based on the type of the node.
+     *
+     * Each type may have a different sorting strategy. For example,
+     * for nodes of type workflow, we know our children will be cycle points.
+     * So we may sort them by their `id`, and in ascending or descending
+     * order.
+     *
+     * Ditto for other types. This way we can control how to sort the children
+     * nodes of each node in the UI.
+     *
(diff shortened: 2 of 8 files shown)


src/graphql/queries.js
```diff
@@ -53,41 +53,41 @@ fragment WorkflowTreeDeltas on Deltas {
 fragment WorkflowTreeAddedData on Added {
   workflow {
     ...WorkflowData
-    cyclePoints: familyProxies(ids: ["root"], ghosts: true) {
+    cyclePoints: familyProxies (ids: ["root"], ghosts: true) {
       cyclePoint
     }
-    taskProxies(sort: { keys: ["cyclePoint"] }, ghosts: true) {
+    taskProxies (sort: { keys: ["name"], reverse: false }, ghosts: true) {
```
kinow, line 59: That's why there was no order in the tasks. They were being sorted by cycle point (can't tell why, probably my bad :smile: ). Now sorting by name.

kinow:
Result after this PR.

image

kinow:
I think it would be nice to allow the user to choose whether the new or old cycle points are displayed first, but probably part of #333, and we would need to choose where to add a component showing that option. So work for another ticket/PR IMO :+1:

oliver-sanders:
> I think it would be nice to allow the user to choose whether the new or old cycle points are displayed first

Because of job ordering it might be best not to make this configurable to avoid the confusion of having opposite sorting in the same window.

oliver-sanders:
Can the tasks get out of order when deltas arrive containing new tasks not present before?

kinow:
> Can the tasks get out of order when deltas arrive containing new tasks not present before?

Only if the new tasks arrive in the `updated` delta. But I think they must always be in the `added`.

When we get these new tasks, their family and cyclepoint parents must either be included in this current `delta.added` payload, or, be already present (no sure if possible?).

I tried thinking of a workflow where I would have the parent of tasks (cycle point or family) present and with tasks, and then new tasks not already present in the UI arrived. But I couldn't come up with a good example. Is that possible @oliver-sanders ? 

If so I guess we could get the tasks out of order just like cycle points, and need to sort them in the frontend.

kinow:
Rebased, resolved conflicts.

oliver-sanders:
Sorry, I managed to get the tasks out of order:

Screenshot 2020-09-09 at 11 43 02

However, there was quite a lot going wrong with the UI for my example, many cycles are showing simultaneously with tasks in funny states, so it might not be the sorting that was going wrong?

I used this workflow:

```ini
[scheduling]
    initial cycle point = 1
    cycling mode = integer
    [[graph]]
        P1 = """
            a[-P1] => a
            b[-P2] => b
            c[-P3] => c
            d[-P4] => d
            e[-P5] => e
            f[-P6] => f
        """

[runtime]
    [[root]]
        script = sleep 1
```

kinow:
That's great @oliver-sanders ! With the example workflow I can try to reproduce this tomorrow or Friday and see what's going on. Thanks for testing again!

kinow:
@oliver-sanders I used your example workflow, and could successfully reproduce it.

I'm preparing Cylc UI to **maybe** demo it today in the RSE conference over here. So I've switched to Cylc Flow 8.0a2 and Cylc UI Server 0.2, and then I cannot reproduce the out-of-order issue with your workflow anymore.

I'll investigate a bit more to understand if the GraphQL queries are returning the nodes in the right order on `master`. If it is, then I will try to understand why I couldn't reproduce it with the latest release.

kinow:
Alright, understood what's going on.

image

image

The cycle point 395 got `b` and then `d`. Later, in another deltas.added payload, we receive the `395.c`. Which causes Vue to keep the `395.b` and `395.d` UI elements in-place (re-using when updates are received, for performance) and then appends `395.c` after `395.d`.

Now let me see how this can be solved… :nerd_face:

kinow:
All right, tests passing, ready for review again. Had to sort in the UI too due to some cases where the data could be displayed incorrectly in the UI (as @oliver-sanders 's example shows).

Didn't remove the backend sorting, as in the best case we might get the data correctly sorted, so the JS `Array.sort` algorithm would perform in the best-case-scenario—but happy to remove that if others prefer :+1:

oliver-sanders:
Tested with the same example again, tasks appear in the correct order 👍 

I am still seeing many more cycle points than I would expect to see, however, I've managed to replicate this on master so I'll raise a new issue for that.

src/mixins/treeitem.js
```diff
@@ -0,0 +1,82 @@
...
+          // sort cycle point children (family-proxies, and task-proxies)
+          // first we sort by type ascending, so 'family-proxy' types come before 'task-proxy'
+          // then we sort by node name ascending, so 'bar' comes before 'foo'
+          return orderBy(children, [child => child.type, child => child.node.name.toLowerCase()], ['asc', 'asc'])
+        }
+        case 'family-proxy': {
+          // sort family-proxy children (family-proxies, and task-proxies)
+          // first we sort by type ascending, so 'family-proxy' types come before 'task-proxy'
+          // then we sort by node name ascending, so 'bar' comes before 'foo'
+          return orderBy(children, [child => child.type, child => child.node.name.toLowerCase()], ['asc', 'asc'])
```
oliver-sanders, line 66: Follow on:

It might be nice to have the families appear before the tasks similarly to how some file listings put directories before files.

Because of the Cylc naming convention of lower-case for tasks, upper-case for families removing the `toLowerCase` would achieve this.

Do we think people would universally prefer this, should it be configurable?

src/mixins/treeitem.js
```diff
@@ -0,0 +1,82 @@
...
+        }
+        case 'task-proxy': {
+          // sort task-proxy children (jobs)
+          // sort by node submit  descending, so '4' comes before '3'
+          return orderBy(children, ['node.submitNum'], 'desc')
```
oliver-sanders, line 71: Follow on:

Would be good to include the natural sorting change made to the Cylc7 GUI so that number patterns are respected by sorting e.g:

* a_0
* a_5
* a_10
* a_15

Which currently get sorted as:

* a_0
* a_10
* a_15
* a_5

oliver-sanders (approved):
Ran the complex suite, couldn't find anything out of order!

src/mixins/treeitem.js
kinow: >It might be nice to have the families appear before the tasks similarly to how some file listings put directories before files.

That's how it should be working now. The family proxies should be displayed before the task proxies (because we order by 2 fields, first type, and then name).

>Do we think people would universally prefer this, should it be configurable?

Good question. Will leave that for follow-up question with @cylc/core

hjoliver: I don't see a great need to make that configurable.  A possible future change if people want it.

src/mixins/treeitem.js
kinow: Good idea! Done!

hjoliver (approved):
Seems to be all working properly now, including "natural" sorting. :+1:

src/mixins/treeitem.js
oliver-sanders: I think people should be happy enough with this.

Issue #332, opened 2019-12-15
placeholders and empty states


oliver-sanders:
Stand-in content for empty / null states to make users aware of the state (rather than just displaying a blank page).

Modern UI design seems to have settled on a greyscale background image for empty states these days. Clear, concise and simple to implement.

I believe material design has built-in support.

Some examples of places where we need empty state placeholders and some ideas as to what they could look like:

![placeholders](https://user-images.githubusercontent.com/16705946/70871159-063df000-2001-11ea-8e34-2923051239d9.png)

Missing placeholders:

* [ ] Workflow: workflow stopped
* [ ] Workflow: connecting/loading (spinner / foreground icon)
* [ ] Workflow: network error (unable to get data)
* [ ] Workflow: no active view
* [ ] GScan: no workflow
* [ ] GScan: connecting/loading

**Pull requests welcome!**

MetRonnie:
Vuetify has a [`VEmptyState` component](https://vuetifyjs.com/en/components/empty-states/#usage) that should make this easy

oliver-sanders:
I'm not quite sure what `VEmptyState` brings to the party, it's trivial to roll your own:

```
<div v-if="workflow.status === 'stopped'">
  <!-- placeholder -->
</div>
<div v-else>
   <!-- content -->
</div>
```

MetRonnie:
Yes but even trivialier using that component. No need to roll your own styles

PR #493, opened 2020-09-09
async scan updates for the UI

kinow:

Related to https://github.com/cylc/cylc-flow/pull/3724 and https://github.com/cylc/cylc-uiserver/pull/150

This PR contains a patch that allows the UI to show stopped workflows, added in the two issues above.

- [x] Allow the Cylc UI app to work with latest Cylc Async Scan changes, showing stopped workflows
- [x] Change the opacity of workflow items in GScan (`opacity: 0.5`)

**Requirements check-list**

### DIFF
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -16,10 +16,22 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 -->
 
 <template>
-  <div>
-    <div v-if="workflows && workflows.length > 0">
-      <div v-for="workflow in workflows" :key="workflow.id">
-        <v-list-item :to="`/workflows/${ workflow.name }`">
+  <div
+    class="c-gscan"
+  >
+    <div
+      v-if="workflows && workflows.length > 0"
+      class="c-gscan-workflows"
+    >
+      <div
+        v-for="workflow in workflows"
+        :key="workflow.id"
+        class="c-gscan-workflow"
+      >
+        <v-list-item
+          :to="`/workflows/${ workflow.name }`"
+          :class="getWorkflowClass(workflow.status)"
+        >
           <v-list-item-action>
             <v-icon>{{ getWorkflowIcon(workflow.status) }}</v-icon>
           </v-list-item-action>
@@ -37,15 +49,16 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                       <!-- a v-tooltip does not work directly set on Cylc job component, so we use a dummy button to wrap it -->
                       <!-- NB: most of the classes/directives in these button are applied so that the user does not notice it is a button -->
                       <v-btn
-                          v-on="on"
-                          class="mt-1 pa-0"
-                          min-width="0"
-                          min-height="0"
-                          style="font-size: 120%"
-                          :ripple="false"
-                          small
-                          dark
-                          text>
+                        v-on="on"
+                        class="mt-1 pa-0"
+                        min-width="0"
+                        min-height="0"
+                        style="font-size: 120%"
+                        :ripple="false"
+                        small
+                        dark
+                        text
+                      >
                         <job :status="state" />
                       </v-btn>
                     </template>
@@ -179,6 +192,13 @@ export default {
         return 'mdi-pause-octagon'
       default:
         return 'mdi-help-circle'
+      }
+    },
+
+    getWorkflowClass (status) {
+      return {
+        // TODO: replace by constant or enum later (not TaskState as that doesn't have stopped, maybe WorkflowState?)
+        'c-workflow-stopped': status === 'stopped'
       }
     }
   }
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -92,6 +92,7 @@ export default {
   props: {
     /**
      * Vanilla workflows object from GraphQL query
+     * @type {{}|null}
      */
     workflows: {
       type: Array,
@@ -113,8 +114,12 @@ export default {
      */
     workflowsSummaries () {
       const workflowSummaries = new Map()
-      for (const workflow of this.workflows) {
-        workflowSummaries.set(workflow.name, getWorkflowSummary(workflow))
+      // with async scan, the workflows list may be null or undefined
+      // see cylc-uiserver PR#150
+      if (this.workflows) {
+        for (const workflow of this.workflows) {
+          workflowSummaries.set(workflow.name, getWorkflowSummary(workflow))
+        }
       }
       return workflowSummaries
     }
--- a/src/components/cylc/gscan/index.js
+++ b/src/components/cylc/gscan/index.js
@@ -25,23 +25,26 @@
  */
 function getWorkflowSummary (workflow) {
   const states = new Map()
-  for (const taskProxy of workflow.taskProxies) {
-    // a task in waiting, may not have any jobs
-    if (taskProxy.jobs) {
-      for (const job of taskProxy.jobs) {
-        // TODO: temporary fix, as the backend is sending ready jobs, but they will change in cylc flow&uiserver in the future
-        if (job.state === 'ready') {
-          continue
+  // a stopped workflow, may not have any tasks
+  if (workflow.taskProxies) {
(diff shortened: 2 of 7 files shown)


kinow:
Assigning @dwsutherland and @oliver-sanders as both are familiar with the async scan PR's. Found a few strange behaviours in the UI, that can be fixed I think, but thought it would be good to check what we should be doing about them.

1. The GraphQL endpoint returns the initial data burst for a stopped workflow

This breaks the code of the UI as we were validating that the initial data burst would contain information to build the workflow tree.

I think that's intentional? I've commented the `throw new Error`, and replaced by a silent `return` to fix it.

The Lumino tree widget is still added. Should we not add it?

Or maybe we leave as-is, until we are working on #332?

2. GScan subscription does not cope well when a workflow is stopped.

image

With deltas, we have a payload with data about the workflow shutdown. The query of GScan doesn't use deltas, so it never realizes that the workflow has stopped.

I can add a wrapper `<div v-if="node.status !== 'stopped'">...</div>` to prevent the icons being displayed. But we could still have some JS code running for that.

Or we can just wait until GScan is re-written with deltas?

3. Reloading the page for a workflow stopped, shows old data

Here I have `five` running.

image

Here's the same view after I've stopped the workflow.

image

And finally, here's after I refresh the page (F5).

image

We get the last state of the workflow in the initial data burst, and the UI simply builds the tree.

We could add an `if` somewhere to prevent that when the workflow status is `stopped`… but not sure if that's the best option?

kinow:
@dwsutherland the code changes here are small, but you can do just a functional test to review this PR :+1: I will update the unit tests later, after we clear the questions above.

dwsutherland:
> This breaks the code of the UI as we were validating that the initial data burst would contain information to build the workflow tree.
> I think that's intentional? I've commented the throw new Error, and replaced by a silent return to fix it.
> The Lumino tree widget is still added. Should we not add it?

Yes, the burst is intentional, as registered and never run workflows won't have any data to form a tree (and never will, without a workflow DB).. But you still need the burst of whatever info we can gather about it.

So I'd say either, just don't add the widget, or add it with blank data so it can spring to life when the workflow starts running (?)


> I can add a wrapper <div v-if="node.status !== 'stopped'">...</div> to prevent the icons being displayed. But we could still have some JS code running for that.

IMHO - The icons for stopped workflows are fine, as it just shows the last state the workflow was in on shutdown... And can differentiate between registered but never run and run but stopped (or DB loaded) workflows..   The `(?)` icon could probably be more informative, and perhaps the whole row (name and icons) faded for stopped workflows (but that'll be for another PR).


>We could add an if somewhere to prevent that when the workflow status is stopped… but not sure if that's the best option?

Built tree is probably fine? Or is that a problem

kinow:
Thanks for the answers @dwsutherland !

All good for me. On the last one only:

>

>>   We could add an if somewhere to prevent that when the workflow status is stopped… but not sure if that's the best option?
>
>Built tree is probably fine? Or is that a problem

That's a problem I think, because at the moment, without refreshing, you won't see anything in the lumino tab, because we are cleaning the tree when we receive a deltas indicating workflow shutdown.

From your previous answers, I think we could then not remove the items when we receive a shutdown, and instead leave them in the UI and just change the workflow status in GScan?

dwsutherland:
> From your previous answers, I think we could then not remove the items when we receive a shutdown, and instead leave them in the UI and just change the workflow status in GScan?

Yeah, sounds good 👍 

One other thing, does this mean a subscription for each (dead or alive) workflow will be created? Any down sides to that?

kinow:
>One other thing, does this mean a subscription for each (dead or alive) workflow will be created? Any down sides to that?

I don't think so. Only when you click on the workflow in GScan, then a new subscription will be created for that workflow (dead or alive :cowboy_hat_face: ). Otherwise you should have 1 or 2 subscriptions depending on the view you are browsing (GScan `subscription {workflows }` and if you have a tree widget the `subscription { deltas }`).

dwsutherland (approved):
Works great, approved subject to tests and discussion changes 👍

kinow:
Thanks @dwsutherland , will wait for @oliver-sanders to have time to take a look at the questions and code change. @hjoliver if you have time and would like to just give your opinion. Don't need to worry too much about code as @oliver-sanders can do a quick code review, just your opinion if there's anything else I should change in this PR for showing stopped workflows :+1: (all good if you don't have time, we can improve later in other PRs)

oliver-sanders:
> GScan subscription does not cope well when a workflow is stopped.

This is an interesting one, I think you are handling it just fine.

> **Context:**
    In the Cylc7 GScan if a workflow stops with running tasks its task status icons (state total square thinggies) remain (they do get a kinda checkerboard pattern though).

>    If you were to take a look in the suite database those tasks would still be in the pool, they were "active" tasks when the workflow shut down so until it is restarted they must be considered to be "active".

Future work:
* We will need to do some work at the UI Server to query the database so that we get these icons when the UIS starts up.
* I think stopped workflows should show in GScan with some transparency, this will help convey that these "task status icons" are showing the status of a stopped workflow.

oliver-sanders:
> Reloading the page for a workflow stopped, shows old data

Similar answer as for the last question. This is mostly correct, this data could alternatively come from the database.

We need two distinct "modes" for viewing workflows:

* `n={0,1,2,3}` - show all tasks `N` edges from active tasks.
* `n=cycle` - show tasks by cycle point (I think we have called this `N=max` sometimes)

For live workflows it makes sense to default to `n=1` (ala `cylc gui`) for stopped workflows it makes sense to default to `n=cycle` (ala `cylc review`).

From the UI perspective is shouldn't make much of a difference whether a workflow is running or not, its the UI Server which picks up all this horrible complexity and has to work out whether it can get data from a live scheduler or whether it has to go to the database instead. The UI should just display what data it gets given!

so all good here I think.

kinow:
>I think stopped workflows should show in GScan with some transparency, this will help convey that these "task status icons" are showing the status of a stopped workflow.

Great idea @oliver-sanders . Set it to `opacity: 0.5`. It looks good in my monitor/browser, but let me know if this value should be tweaked. It's much easier to see what's running this way (as a complement to icons/colors)

image

hjoliver:
Oooh, that's very nice!

kinow:
>The UI should just display what data it gets given! So all good here I think.

Right. I am only concerned about consistency now.

If I am looking at Cylc UI with a single lumino widget showing the tree view of a workflow, and then I stop it, the lumino widget won't have anything.

But if I open the same URL in another tab, then it will display the latest state of the workflow.

I'm adding two more commits here:

1. when a `deltas.shutdown` is received, instead of clearing the `CylcTree` object (that reactively populates the lumino widget tree), I will simply set the state of the workflow to stopped and leave the UI elements
  - We need to distinguish a workflow running or not, so that if the workflow is restarted, we know that we need to re-create the existing `CylcTree` object (i.e. when the user stopped, it could have been delta number 100, and now that we restarted, the UI elements are still showing delta 100, but the first delta received could be number 400, so it would fail to apply the delta and leave the old data forever in the UI)
2. Fix unit tests

Then it should be good to go :crossed_fingers:

hjoliver:
I guess we need a stopped icon (not question mark) for stopped flows in gscan.

kinow:
From RIOT (hope it's OK to copy that over here @dwsutherland / @hjoliver )

image

oliver-sanders:
All sounds/looks good to me.

For follow-on it would be good to sort GScan so that active suites appear above inactive ones (some people have insane numbers of stopped suites).

oliver-sanders (approved):
Looks good.

Screenshot 2020-09-14 at 13 52 44

Can go in once the tests pass.

kinow:
Going to merge this one as-is, and move the rest of the work for stopped workflows for another issue. That way it gets simpler to review, and also the stopped needs the two PR's that @dwsutherland created yesterday.

kinow:
> Looks good.
> 
> Screenshot 2020-09-14 at 13 52 44
> 
> Can go in once the tests pass.

Thanks @oliver-sanders !

kinow:
Will also include the changelog for showing offline workflows (forgot it, sorry!)

kinow:
b51564ba38a797531260c79b1f9b7c2711bcb1ef adds the change for this PR retroactively

PR #497, opened 2020-09-17
Add global error catcher with Vue's errorCaptured

### LINKED ISSUE #465: Create a global "catcher" for errors in the UI

kinow:
These changes close #465 

Now if an error occurs in the browser console, and if you are not in production mode (we can change this later if others think we should always display the message) it will display an error message.

I hope this will be useful specially when testing the UI with workflows during development. Sometimes the UI may look OK, but you could have hidden errors such as deltas no being applied. Ideally, we should never see that error message. 

This captures errors from children elements of the `Default` layout, that were not captured. Think of it as a `try { renderChildrenOfDefaultLayout } catch { # display error message }`.

image

**Requirements check-list**

### DIFF
--- a/src/layouts/Default.vue
+++ b/src/layouts/Default.vue
@@ -38,6 +38,8 @@ import Drawer from '@/components/cylc/Drawer'
 import Toolbar from '@/components/cylc/Toolbar'
 import ConnectionStatus from '@/components/cylc/ConnectionStatus'
 import { mapState } from 'vuex'
+import store from '@/store/'
+import AlertModel from '@/model/Alert.model'
 
 export default {
   name: 'Default',
@@ -51,6 +53,12 @@ export default {
 
   computed: {
     ...mapState(['offline'])
+  },
+
+  errorCaptured (error, vm, info) {
+    if (process.env.NODE_ENV !== 'production') {
+      store.dispatch('setAlert', new AlertModel(error, null, 'error'))
+    }
   }
 }
 </script>
--- a/tests/e2e/specs/layout.js
+++ b/tests/e2e/specs/layout.js
@@ -1,0 +1,37 @@
+/**
+ * Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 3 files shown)


kinow:
Just need to think how to write a test for this :thinking:

src/layouts/Default.vue
```diff
@@ -51,6 +53,12 @@ export default {
 
   computed: {
     ...mapState(['offline'])
+  },
+
+  errorCaptured (error, vm, info) {
+    if (process.env.NODE_ENV !== 'production') {
```
kinow, line 59: Can be changed later if others prefer to always see the errors.

tests/e2e/specs/layout.js
```diff
@@ -0,0 +1,37 @@
...
+describe('Default layout', () => {
+  it('Should display errors from children elements captured at the Default layout level', () => {
+    // visit any page first, so that we create the window.app reference
+    cy.visit('/#/workflows/one')
+    cy
+      .get('.v-alert')
+      .should('not.be.visible')
+    cy.window().its('app.$workflowService').then(service => {
+      // mock service so that it returns an error
+      cy.stub(service, 'subscribe', () => {
```
kinow, line 27: Note: this `cy.stub`, is actually Sinon's `stub` function. The same we are using for unit tests. Cypress includes Sinon by default, and provides `.stub`, `.spy`, etc Sinon methods so that you don't need to worry importing those libs.

oliver-sanders (approved):
Nice.

hjoliver (approved):
(Amazingly small change, to get this functionality :+1: )

PR #498, opened 2020-09-17
Sort GScan workflows by type (first running, then held, then stopped), and then by name.

kinow:
Follow-up of #493 

https://github.com/cylc/cylc-ui/pull/493#issuecomment-691893455

image

**Requirements check-list**


kinow:
image

image

image

image

kinow:
Will add one or two commits later, for changelog and tests. Then mark as ready for review.

hjoliver:
Very nice! :tada:

oliver-sanders (approved):
Nice, looks good, tested as working.

I think it might be best to mix running, held and stopping flows in together as users who have large numbers of similarly named flows may get confused if they are split up into groups.

kinow:
> Nice, looks good, tested as working.
> 
> I think it might be best to mix running, held and stopping flows in together (i.e. just separate active and stopped flows) as users who have large numbers of similarly named flows may get confused if they are split up into groups.

@oliver-sanders I also think so. In my case, out of all workflows, I have 1 that I always keep an eye on (`five`, no surprise), and found it annoying when I had multiple running/held, and had to look for `five` amongst the workflows that were not transparent.

@hjoliver let me know what you think, and I can quickly update the PR before your review :+1:

hjoliver:
I agree, let's go with the mixed approach - just separate active and stopped.

kinow:
> I agree, let's go with the mixed approach - just separate active and stopped.

Thanks @hjoliver . This PR should be updated (code + tests) by tomorrow morning :+1:

kinow:
Rebased, and pushed one more commit where the "running" and "held" workflows are grouped together when sorting in GScan.

Now it looks like:

image

So the three running workflows were displayed in the following order:

- a/b/c
- families2
- five

Then I held a/b/c and five. The order stayed the same, as in the screenshot :+1: 

I prefer this way, as it's easier to find my five workflow :grimacing: soon with mutations I might be able to quickly start or restart it too :tada: eventually might be able to do 100% with the UI, using command line only to start UIS and look at logs.

kinow:
Ready for review @hjoliver

hjoliver (approved):
Brilliant - thanks @kinow :bouquet:

PR #499, opened 2020-09-21
Add search and filter functionalities to GScan

### LINKED ISSUE #337: GScan: search and filter
Add search and filter functionality to GScan (#139).

gscan-filter

* Search:
  * By partial registration name e.g. "foo' matches "fool/bar" and "bar/fool".
  * Accept globs / regex?
* Filter:
  * Filter by "discrete" data only.
  * May involve adding to the GScan subscription to acquire the required data.

A further enhancement would be searching non-discrete metadata via a different interface. E.G. searching by suite title, description or the Cylc7 "group".

**Pull requests welcome!**

kinow:
These changes close #337

Improves the `GScan` component adding

- [x] search by workflow name
- [x] filter as per design sketch in #337 
  - minus hosts and versions filtering

**Requirements check-list**

### DIFF
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -188,6 +188,7 @@ import { getWorkflowSummary } from '@/components/cylc/gscan/index'
 import { GSCAN_QUERY } from '@/graphql/queries'
 import WorkflowState from '@/model/WorkflowState.model'
 import TaskState from '@/model/TaskState.model'
+import { mdiFilter } from '@mdi/js'
 
 const QUERIES = {
   root: GSCAN_QUERY
@@ -214,6 +215,9 @@ export default {
       subscriptions: {},
       isLoading: true,
       maximumTasksDisplayed: 5,
+      svgPaths: {
+        filter: mdiFilter
+      },
       /**
        * The filtered workflows. This is the result of applying the filters
        * on the workflows prop.
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -24,6 +24,27 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
       type="list-item-three-line"
     >
       <div
+        class="d-flex flex-row mx-4 mb-2"
+      >
+        <v-text-field
+          v-model="searchWorkflows"
+          clearable
+          flat
+          dense
+          hide-details
+          outlined
+          placeholder="Search"
+          class="flex-grow-1 flex-column"
+        />
+        <v-btn
+          link
+          icon
+          class="flex-grow-0 flex-column"
+        >
+          <v-icon>mdi-filter</v-icon>
+        </v-btn>
+      </div>
+      <div
         v-if="!isLoading && workflows && workflows.length > 0"
         class="c-gscan-workflows"
       >
@@ -133,7 +154,8 @@ export default {
       viewID: '',
       subscriptions: {},
       isLoading: true,
-      maximumTasksDisplayed: 5
+      maximumTasksDisplayed: 5,
+      searchWorkflows: ''
     }
   },
   computed: {
@@ -173,6 +195,13 @@ export default {
         }
       }
       return workflowSummaries
+    },
+    filteredWorkflows () {
+      let filtered = [...this.workflows]
+      if (this.searchWorkflows && this.searchWorkflows !== '') {
+        filtered = filtered.filter(workflow => workflow.name.includes(this.searchWorkflows))
+      }
+      return filtered
     }
   },
   created () {
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -36,20 +36,64 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           placeholder="Search"
           class="flex-grow-1 flex-column"
         />
-        <v-btn
-          link
-          icon
-          class="flex-grow-0 flex-column"
+        <v-menu
+          v-model="showFilterTooltip"
+          :close-on-content-click="false"
+          offset-x
         >
-          <v-icon>mdi-filter</v-icon>
-        </v-btn>
+          <template v-slot:activator="{ on, attrs }">
+            <v-btn
+              v-bind="attrs"
+              v-on="on"
+              link
+              icon
+              class="flex-grow-0 flex-column"
+              @click="showFilterTooltip = !showFilterTooltip"
+            >
+              <v-icon>mdi-filter</v-icon>
+            </v-btn>
+          </template>
+          <v-card
+            max-height="250px"
+          >
+            <v-list
+              dense
+            >
+              <div
+                v-for="filter in filters"
+                :key="filter.title"
+              >
+                <v-list-item dense>
+                  <v-list-item-content ma-0>
+                    <v-list-item-title>{{ filter.title }}</v-list-item-title>
+                  </v-list-item-content>
+                </v-list-item>
+                <v-divider />
+                <v-list-item
+                  v-for="item in filter.items"
+                  :key="`${filter.title}-${item.text}`"
+                  dense
+                >
+                  <v-list-item-action>
+                    <v-checkbox
+                      :label="item.text"
+                      v-model="item.model"
+                      dense
+                      height="20"
+                    ></v-checkbox>
+                  </v-list-item-action>
+                </v-list-item>
+              </div>
+            </v-list>
+          </v-card>
+        </v-menu>
       </div>
       <div
-        v-if="!isLoading && workflows && workflows.length > 0"
+        v-if="!isLoading && filteredWorkflows && filteredWorkflows.length > 0"
         class="c-gscan-workflows"
       >
         <div
-          v-for="workflow in sortedWorkflows"
+          v-for="workflow in filteredWorkflows"
           :key="workflow.id"
           class="c-gscan-workflow"
         >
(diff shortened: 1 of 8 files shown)


kinow:
Search is done. I have **not** added debounce, nor implemented the two-step filter as in the tree view as for now we don't have very large lists of workflows. We can fine-tune it during reviews or in follow-up PR's :+1: 

Here's what it looks like now.

GIFrecord_2020-09-21_135705

kinow:
Figuring out how to show so many states in a single tooltip :smile: 

image

kinow:
Tests are broken for now, but it would be better to get a pre-review of the new functionality before I spend some time fixing them.

I have not included workflow host nor Cylc version, as I think I would need to iterate the workflows and collect their hosts and versions. Not sure if there are plans to add this in GraphQL (don't think so? But we can add it later).

image

@cylc/core would anyone be willing to play and see what needs changing, what I missed, any bugs, etc, please? :grimacing: 

GIFrecord_2020-09-22_153118

kinow:
BTW, I tried to add the magnifier icon, as in the design sketch. Unfortunately Material Design puts the icon out of the text field. There might be some way to move the magnify icon inside the text field to match the sketch, but that would probably require some further investigation (i.e. read source code of vuetify in TypeScript, and docs, more than 1/2 hours work at least).

Also tried the `dark` directive, to have the tooltip (that's a menu actually :man_shrugging:, but users won't notice the difference) dark as in the sketch. The `dark` directive uses CSS styles from the vuetify dark theme, which is much darker. And in case we end up using `light`/`dark` from Vuetify, we could have to hack with the colors or theme. So I thought it would be simpler to have the tooltip working first, and then tweak colors, margins, fonts, styles in general :+1: 

But my goal was to be as close to the design sketch as possible.

hjoliver:
Looks good @kinow ; I'll try to take a closer look tomorrow.  I think the other selection criteria can be added later (and will presumably require updating the GraphQL schema and the workflow data sent by the UIS).

oliver-sanders:
> Figuring out how to show so many states in a single tooltip 😄

The good news is that we will kill off a few of those task states pre-release.

kinow:
>Are you filtering for task state or job state? (there are only 5 job states).

workflow state, and task state. The UI displays each state in the enum `TaskState`. Then the filter just checks for the intersection of those states, with the states in `workflow.stateTotals`.

If the intersection is not empty, the workflow stays in the UI.

oliver-sanders:
> states in workflow.stateTotals (I believe these are task states?).

Yep those are task states, however job states are just a subset of task states:

* submitted
* running
* succeeded
* failed
* submit-failed

(should probably stick that in an enum along with task states at some point)

Because GScan is displaying jobs (squares) it would make more sense to filter by job states rather than task states (circles).

hjoliver:
https://cylc.github.io/cylc-admin/proposal-state-names.html

kinow:
Rebased, resolved conflicts, and tested again. Ready for a pre-review. If the UI elements, and the functionality are OK, I will add the unit/e2e tests, and then it should be ready for the final review :+1:

oliver-sanders:
Looks good, only found one niggle which is that stopped workflows don't get filtered out when filtering by task state.

Screenshot 2020-09-29 at 09 37 13

kinow:
>Looks good, only found one niggle which is that stopped workflows don't get filtered out when filtering by task state.

@oliver-sanders ah, good catch! I was intentionally applying task states to only running workflows. Removed that `if` statement in the latest commit. When you first initialize Cylc UI and Cylc 8, you probably won't see any stopped workflows.

That's because GraphQL `workflows.stateTotals` will return empty. But if you start, then stop a workflow, it should work as expected I think :+1: 

image

kinow:
@hjoliver if you have time later, would you mind taking a look at the way the WIP search/filter are working here, please? If you find no issues with how it works, I will add the test/changelog and mark as ready for review :) Thanks!

hjoliver:
I'm not seeing stopped flows on this branch, unless I switch away and come back again. :thinking: 

foo

kinow:
>I'm not seeing stopped flows on this branch, except briefly if I switch away and come back again. thinking

There's an explanation for that. When you come back, the UI takes some ms/seconds to re-apply the filter. The filter is using the task states. By default it shows every workflow that has any task state. Offline/stopped workflows won't have any states, unless you have executed them at least once since the UIS was started.

Thinking about it, I am keen to leave that filter disabled by default, and show everything, every workflow regardless of the task state. I think that makes more sense. If you want to limit by task state, then you have to choose the states and filter.

hjoliver:
Ah, makes sense.

> Thinking about it, I am keen to leave that filter disabled by default, and show everything, every workflow regardless of the task state. I think that makes more sense. If you want to limit by task state, then you have to choose the states and filter.

Agreed.

> Offline/stopped workflows won't have any states, unless you have executed them at least once since the UIS was started.

This might change once the UIS (possibly via `cylc scan` in the stopped flow case) can read flow DBs. (But, that's not the problem of this PR).

hjoliver:
With so many task states (albeit fewer of them by the time we release 8.0) can we have a select-all/deselect-all option?

kinow:
>With so many task states (albeit fewer of them by the time we release 8.0) can we have a select-all/deselect-all option?

By coincidence, I am working on this at this right moment! :grimacing: Just trying to align the checkbox next to the filter name, then will push the new commits.

src/components/cylc/gscan/GScan.vue
```diff
@@ -19,8 +19,87 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+                    dense
+                    height="20"
+                  ></v-checkbox>
+                </v-list-item-action>
+              </v-list-item>
+            </div>
+          </v-list>
+        </v-card>
+      </v-menu>
+    </div>
```
kinow, line 99: These filters are not looking really nice here and in the Tree component or view. Have added a post-it to take a better look, see if we can unify or find a better way to organize them. If so, will create a follow-up issue :+1:

src/components/cylc/gscan/GScan.vue
```diff
@@ -122,8 +202,97 @@ export default {
...
         held: mdiPauseOctagon,
-        unknown: mdiHelpCircle
-      }
+        unknown: mdiHelpCircle,
+        filter: mdiFilter
+      },
+      /**
+       * The filtered workflows. This is the result of applying the filters
+       * on the workflows prop.
+       * @type {[
```
kinow, line 211: This is GH UI. Valid JS & jsdoc.

kinow:
Done @hjoliver 

GIFrecord_2020-10-07_152939

Tested locally. It looked intuitive to me.

First implementation was selecting all only if none was selected. That was a bit strange to use.

Also tried using a checkbox to the right of the text (i.e. "workflow states" [ ]). But aligning that `v-checkbox` component is annoyingly difficult, because of the way Vuetify creates multiple elements with so many classes that it's hard to guess where the extra padding or margin is coming from.

Then realized that it could just be another checkbox, aligned with the other check boxes. It became "easier to read"? Not sure if that makes sense, but I found the final product good enough :grimacing: 

Could you take another look, please, @hjoliver ?

Thanks

hjoliver:
Looks good in your gif - I'll try it myself later, after getting the xtrigger PR tests workings....

hjoliver:
(Apologies for not getting back to this yet @kinow - will return as soon as I can)

kinow:
> (Apologies for not getting back to this yet @kinow - will return as soon as I can)

Not a problem. I am fixing (with some struggles) a functional test in Cylc Flow in the meantime.

Thanks @hjoliver !

hjoliver:
> I am fixing (with some struggles) a functional test in Cylc Flow in the meantime.

Ah, a pain we're unfortunately all too familiar with :rofl:

kinow:
Rebased and fixed conflicts. Should include latest changes, such as Graph component code removal.

oliver-sanders:
Ready for review?

kinow:
> Ready for review?

I need to write tests, but I was waiting for Hilary's feedback first to avoid having to change code + tests.

hjoliver:
> I was waiting for Hilary's feedback first to avoid having to change code + tests.

Apologies @kinow - the usual excuses apply; I need to clone myself or something :confounded:  From another look at your animated gif above it looks great.  Happy for you to un-Draft this on that basis.

kinow:
Haha, not a problem @hjoliver . If it were another smaller change I could ask someone else to take a look. But I think for chamges like this you and Oliver normally pick up things I forget while also keeping the change aligned with the cylc 8 plan.

Will start writing tests and mark it ready for review next week, after the message outputs PR 👍

Thanks!

kinow:
Unit tests done. Now on to the e2e tests.

kinow:
e2e test added. Now wait a bit for CI to confirm it's working, and then for Codecov to produce the coverage report. After that I will confirm I didn't forget to test some important part of the test, and then lastly one more round of manual testing. Then it'll be ready for review :+1:

kinow:
Not sure if Codecov hasn't updated, but it's showing in its UI +0.14% (unit+e2e combined): https://app.codecov.io/gh/cylc/cylc-ui/compare/499

kinow:
Rebased, conflicts fixed.

kinow:
This branch has a problem sorting workflows. The problem has been fixed on #537 , I've added the fix here, but would have to add the updated tests too, so reverted the change.

When reviewing, take a look at the search and filter functionality only :+1: unless that PR is merged first and this is rebased.

oliver-sanders (approved):
:fireworks:

kinow:
Rebased fixing conflicts, but now unit tests failing :disappointed_relieved:

kinow:
Tests fixed. Did a quick manual testing, everything looks to be in order :+1:

hjoliver (approved):
Nice, looks great @kinow :+1: tested as working well.

kinow:
Thanks @hjoliver !!! :champagne:

PR #501, opened 2020-09-22
Support any characters, including slashes, in workflow names

### LINKED ISSUE #500: Browser reads hierarchical workflow name as part of URL structure
On opening the tree view for a hierarchically registered suite;
```
(flow) sutherlander@cortex-vbox:fiz$ cylc register a/fix
REGISTERED a/fix -> /home/sutherlander/workflows/fiz
```
the browser returns "page not found", presumably because it reads the workflow name as part of the URL structure.  

Open the tree view for a hierarchically registered workflow.

**Expected behavior**
Tree view is displayed

**Screenshots**
image
image

**Pull requests welcome!**

kinow:
These changes close #500 

**Requirements check-list**
- [x] Does not need tests (why? It is hard to write a functional or unit test for this :cry: if others would like to have this tested, I will just need some time to think how to address it... maybe with Cypress server in tests, or some mock somewhere…).

### DIFF
--- a/src/router/paths.js
+++ b/src/router/paths.js
@@ -40,7 +40,7 @@ export default [
     }
   },
   {
-    path: '/workflows/:workflowName',
+    path: '/workflows/:workflowName(.*)',
     view: 'Workflow',
     name: 'workflow',
     meta: {
@@ -67,7 +67,7 @@ export default [
     }
   },
   {
-    path: '/tree/:workflowName',
+    path: '/tree/:workflowName(.*)',
     view: 'Tree',
     name: 'tree',
     meta: {
@@ -77,7 +77,7 @@ export default [
     props: true
   },
   {
-    path: '/graph/:workflowName',
+    path: '/graph/:workflowName(.*)',
     view: 'Graph',
     name: 'graph',
     meta: {


kinow:
Tested locally with a running workflow. Tested workflow, tree, and graph views :+1:

dwsutherland (approved):
Works for me 👍 
image

src/router/paths.js
```diff
@@ -40,7 +40,7 @@ export default [
     }
   },
   {
-    path: '/workflows/:workflowName',
+    path: '/workflows/:workflowName(.*)',
```
hjoliver, line 43: Is there any danger of `.*` being too wide-open in this context?  Or is the default "anything except `/`" i.e. `[^/]*`, in which case `.*` is the only way to extend it :grin: )

kinow: Good question. I tried some variations, but will take a better look tomorrow morning.

Vue Router uses [this library](https://github.com/pillarjs/path-to-regexp/tree/v1.7.0). I will take a look at its documentation and see what options we have.

kinow: >Is there any danger of .* being too wide-open in this context?

It looks like that will indeed capture everything after `/workflows/` until the next `/`, from [their docs](https://github.com/pillarjs/path-to-regexp/tree/v1.7.0#named-parameters) "By default, the parameter will match until the following path segment."

The `(.*)` here captures everything until the end of the URL. It is smart enough to stop if it sees a query like `?param=value`, or a fragment like `#anotherfragment`.

So it is greedy, and captures the rest of the URL. Not sure if it has any potential risk of capturing more than we want.

But this regex works too: `/workflows/:workflowName([a-zA-Z0-9_/-])*`

Should we switch to this one? Not sure if I forgot to include some character, but we can just add more if necessary

hjoliver: What I was wondering is, does the default already capture "any character except for `/`" - if it does, then `.*` really only adds one more character (i.e. `\`) and it seems like there would be no point in using the more specific expression (which would presumably exclude some characters that are not currently excluded)... If so, then I think `.*` is fine.  Does that make sense :grimacing: ?

hjoliver: > > Is there any danger of .* being too wide-open in this context?
> It looks like that will indeed capture everything after /workflows/ until the next /, from their docs "By default, the parameter will match until the following path segment."

(Maybe you answered my question there already - but it looks like you're reply to my question about `.*` ... but then talking about "the default" - so I'm a bit confused!)

kinow: Yup! All good. And it was good to dig a bit deeper and understand how `path-to-regexp` works. Handy library, might be useful later.

hjoliver: So we can stick with `.*`?

kinow: The default for named parameters. In Vue Router, we have `/:workflowName`. `path-to-regexp` will parse anything after the slash symbol, until it finds another slash or end of stream, and create the JS object `{ workflowName: "whatever-user-entered" }`.

Vue Router uses that to pass that value as a `prop` to the View.

The `.*` changes that so that it keeps consuming the stream until it is empty, even if it sees a slash. So `/a/b/c` will generate `{ workflowName: "a/b/c" }`.

kinow: > So we can stick with `.*`?

I'm +1 for `.*`. (we are missing each other in the comments here haha, sorry)

hjoliver: Great (and sorry for complicating this small change!)

hjoliver (approved):

PR #504, opened 2020-09-24
mutations in the tree view

oliver-sanders:
Addresses #339 

Add mutations into the tree view.

Caveats:

1) I've only added support for those mutations that don't require any information that we can't get from the context of the object we are mutating (e.g. trigger).
2) At the moment mutations are only available on tasks/families.
     * [x] Cycle points coming soon - #505.
     * [ ] Task only mutations will require some extra logic.
     * [ ] Jobs will require a little argument re-think - https://github.com/cylc/cylc-flow/pull/3880.
3) Support for the remaining mutations and the ability to provide optional arguments to come in a future PR. Until that time I've not removed the old mutations view, though I expect it is thoroughly broken.

This approach can be easily incorporated into other views.

TODO (this PR):

* [x] Tidy.
* [x] Test libs.
* [ ] Test components? (have tested the menu opens, would be good to test mutation generation but can do later)
* [x] Fix commented out tests.

**Requirements check-list**


src/components/cylc/tree/TreeItem.vue
```diff
@@ -32,39 +32,50 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-         :isHeld="node.node.isHeld"
-         :progress=0
-        />
+        <CylcObject :id="node.node.id">
+          <task
+           :status="node.node.state"
+           :isHeld="node.node.isHeld"
+           :progress=0
+          />
+        </CylcObject>
```
oliver-sanders, line 41: Basic premise:

If you've got something you wan't to be interactive, wrap it with a `<CylcObject>` and provide it the Cylc ID for the "thing" your element represents.

This wrapper handles context menus, pop-ups and mutations.

I did think about adding this functionality into the task/job components themselves which would make developing new views a little easier, however, this allows groups of components to be wrapped up together.

I'm not 100% happy with this approach. For example, what if you wanted to wrap an entire row in the tree view? You could do it, however, CylcObject and the Tree compoments would be fighting for interaction events.

Open to ideas.

oliver-sanders:
Ok, getting close with this, need to get started on tests.

kinow:
>What's the best way to delay this until the component is interacted with?

If you only need to calculate the values once, I think it would be to use the `data` attributes` in the UI elements, and have a method under `methods` that is responsible for populating the `data` attributes.

But if you still want reactivity, you can try with a `watch` value, and see if you can control the reactivity. If you don't set `immediate: true` (I think default is `false` anyway?), the watcher is not executed when the component loads.

There are cases when you need to mix `watch` with a `method` too (I think we have that in the new filters?)

oliver-sanders:
Ok, I've added a state variable, sadly this means it's no longer a functional component but at last it's not performing un-necessary computation on mount.

tests/unit/components/cylc/tree/tree.vue.spec.js
```diff
@@ -241,7 +242,8 @@ describe('Tree component', () => {
         // cycle point is displayed (list.filter(() => return filtered=true means OK, filtered=false means remove me)
         expect(taskProxy.filtered).to.equal(false)
         // task proxy is displayed
-        expect(taskProxy.$children[0].filtered).to.equal(false)
+        // TODO: fixme
+        // expect(taskProxy.$children[0].filtered).to.equal(false)
```
oliver-sanders, line 246: I couldn't work out how to fix these, not sure how to debug. I think this is failing because I've added extra components into the hierarchy?

src/components/cylc/tree/TreeItem.vue
oliver-sanders: Also this really won't work with multiple selection etc.

tests/unit/components/cylc/tree/tree.vue.spec.js
kinow: That would make sense. Same happened when I added that skeleton loader wrapping components in Lumino widgets.

kinow (commented):
Haven't tested yet, but read the code in GitHub UI, and it is looking good :+1: Few TODO/notes that I think you are working on, so will leave the review for later :+1:

src/components/cylc/tree/TreeItem.vue
oliver-sanders: Now thinking that this should be available somehow as a function something along the lines of:

```
interact([ids], [posX, posY])
```

This would, for example, allow the Tree view to handle multiple-selection then call the interact function with the IDs of the selected elements and the x,y position of the cursor to create the tooltip at?

In Vue terms this could be:

* A directive, however, I don't know whether a directive can create the context-menu component.
* A mixin for the Cylc "views", however, I don't know whether a mixin can create the context-menu component either.
* A wrapper component that the Cylc "views" are placed inside. The interact method would be passed to the view which would in turn pass it down to whatever components require it.

oliver-sanders:
Rebased and brought in support for mutating cycle points thanks to #505.

oliver-sanders:
Ok, I think I'll open this now, it needs changes, but it'll take me a fair amount of research to work out how to move forward.

kinow (commented):
Tested functionally, and it worked fine! :+1: Going to look at the code now.

src/services/workflow.service.js
```diff
@@ -184,6 +196,19 @@ class WorkflowService extends GQuery {
       }
     })
   }
+
+  loadMutations () {
+    // TODO: this assumes all workflows use the same schema which is and
+    // isn't necessarily true, not quite sure, come back to this later
```
kinow, line 202: :+1:

src/utils/aotf.js
```diff
@@ -0,0 +1,586 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
+ * Functionality for introspecting mutations, associating them with Cylc
+ * objects, filtering and calling mutatons.
```
kinow, line 19: s/mutatons/mutations

src/utils/aotf.js
```diff
@@ -0,0 +1,586 @@
...
+ *     (e.g. taskID).
+ *     Note that this type may be buried inside other GraphQL types
+ *     (e.g. NON_NULL, LIST).
+ *   _multiple:
+ *     true if this field accepts multiple values.
+ *   _required:
+ *     true if this field must be provided for the mutation to be called.
+ *
+ * @param {object} mutation - One Mutation as returned by introspection query.
+ * @param {object} types - Types as returned by introspection query.
```
kinow, line 250: Is it `object` or `Array`? Further down we call `      arg._default = getNullValue(arg.type, types)`. In `getNullValue` there's a `for (const type of types) {`… if an `Object` is given, it would—I think—fail with `Uncaught TypeError: types is not iterable` (not sure how to test it from the UI, sorry)

kinow (commented):
:+1: looks good to me! Found no issues in the code, only one question there about the type of some object. There are some TODO's and commented out tests. Let me know if you need help with any of those.

src/components/cylc/tree/TreeItem.vue
kinow: >Basic premise:
>
>If you've got something you wan't to be interactive, wrap it with a <CylcObject> and provide it the Cylc ID for the "thing" your element represents.
>
>This wrapper handles context menus, pop-ups and mutations.
>
>I did think about adding this functionality into the task/job components themselves which would make developing new views a little easier, however, this allows groups of components to be wrapped up together.

Sounds like a good approach, similar to the Vuetify `v-skeleton` component. It wraps the component(s) and adds it some new feature.

>I'm not 100% happy with this approach. For example, what if you wanted to wrap an entire row in the tree view? You could do it, however, CylcObject and the Tree components would be fighting for interaction events.

Hmmm, well thought. But I think you should be able to capture the `click` or other events in the `CylcObject`, while allowing the children components to receive the event too, if necessary.

>Also this really won't work with multiple selection etc. I think for that to work this functionality would have to be much more centralised. Perhaps this is the stepping stone.

If I understand correctly, the issue would be if we allowed users to select a cycle point and a family proxy at the same time. What would happen when the user click to display the mutations? If so, I think we can just disable the mutations pop up when multiple elements of different types are selected (might be tricky to do that, but doable I think).

>In Vue terms this could be:
>
>    A directive, however, I don't know whether a directive can create the context-menu component.

Hmm, that would be tricky. Not sure how/where I would start to implement this with a directive. I haven't used any custom directive yet, but from what I've used in other components, it's normally used to enable/disable a feature of a component. So the components (`TreeItem` for example) would have to have built-in support to mutations, and that directive would just enable/disable it.

I think the current approach keeps a better separation of concerns, but might be wrong.

>A mixin for the Cylc "views", however, I don't know whether a mixin can create the context-menu component either.

Yeah, I can't think of a clean way, without making each component aware of the context-menu, at least in the template section we would need to have some `if`/`else`s.

>A wrapper component that the Cylc "views" are placed inside. The interact method would be passed to the view which would in turn pass it down to whatever components require it.

That could work, but not sure if that'd work with every sub-component. Worth a try if you think that'd be better. At least I think that's the easiest of the three options you suggested.

But I liked the simple `CylcObject` wrapper :smiley:

kinow:
The Bootstrap.vue Toast component could serve as reference @oliver-sanders for a different approach for the `CylcObject` and the Vuetify menu: https://bootstrap-vue.org/docs/components/toast

You create a "toast" with 

```js
        this.$bvToast.toast(`Toast ${this.counter} body content`, {
          title: `Toaster ${toaster}`,
          toaster: toaster,
          solid: true,
          appendToast: append
        })
```

From any part or your code. That creates a popup with an alert/notification in different parts of the screen. We would just need to make it appear relative to the component that originated the call.

The toast component also uses a plug-in. In Vue we can create a plugin that is executed once when the application installs the plug-in, and loads directives, components, creates global singletons, etc.

e.g.

```js
import CylcObject from '@/components/cylc/cylcObject/CylcObject'

/**
 * Cylc Objects plug-in.
 */
export default {
  /**
   * Called when the Vue application is created, and this plug-in is loaded.
   * @param Vue {object} - Vue application
   * @param options {*} - options passed to the plug-in (if any)
   */
  install (Vue, options) {
    // We could...

    // register the CylcObject component globally (no need to import)
    Vue.component('cylc-object', CylcObject)

    // add a global directive
    Vue.directive('cylc-object', {
      bind (el, binding, vnode, oldVnode) {
        // define that a cylc-object will have an @click and automagically
        // trigger an event?
      }
    })

    // create a singleton in the application to listen for events
    Vue.prototype.$cylcObjectsMenu = {}
  }
}

```

oliver-sanders:
Sounds promising!

kinow:
Before trying anything, I rebased, then `yarn run build --optimize` on this branch. Found no memory leaks, and the garbage collection is working fine. On my computer, the heap memory is going between ~13 and ~18MB, using workflow `five`.

image

image

image

On `master` the performance is similar, but the browser spent less time on Scripting.

image

Nothing major, but having minus one component per `TreeItem` could help to keep the Scripting time down even with `complex`.

kinow:
I've got the conflict fixed locally, and also the directive is partly done. Tomorrow will continue, now working on breaking `CylcObject` into a global `v-menu` that can be added per view/component basis, and another code to react to `@click` on elements with `v-cylc-object` directive.

kinow:
Had an interesting issue. After writing the directive, loading it with a plug-in, and initializing the plug-in in our Vue application, the directive was not executed.

No matter what I tried.

Turns out it apparently doesn't work with functional components :smile:  https://github.com/vuejs/vue-loader/issues/1433

However, if you re-write the functional component template with [JSX/render function](https://vuejs.org/v2/guide/render-function.html#JSX), then it works. Already confirmed by re-writing the `Job` component template (simple, three HTML elements). Now on the `Task` component (more complex).

kinow:
Almost there with the event emitting a custom event `show-mutations-menu`, with a payload containing the mutations returned from `filterAssociations`.

@oliver-sanders for some reason I keep getting `namespace` as the "type" of the tree node of my task proxy.

Debugging, I have:

```
cycle point: "20170730T0000+1300"
namespace: "foo"
user: "kinow"
workflow: "five"
```

But given my ID was `kinow|five|20170730T0000+1300|foo`, I'm thinking this might be related to the new Cylc ID that I've seen in another issue?

Does that make sense to you? Or did I miss some function when updating the JS code locally?

Thanks!

image

oliver-sanders:
> @oliver-sanders for some reason I keep getting namespace as the "type" of the tree node of my task proxy.

Yep, that's on-purpose, the universal ID does not differentiate between tasks and families:

```
# old
user|flow|cycle|namespace|job
# new (not implemented)
~user/flow//cycle/namespace/job
```

So we have to treat them generically as "namespaces", since we make heavy use of globing this abstraction isn't so bad.

kinow:
Thanks! For some reason today I was getting an empty list of mutations (from `filterAssociations` for a task-proxy). Will take another look tomorrow morning to see what I was doing wrong :coffee:

oliver-sanders:
Yikes, though it might not be a user error.

If the object registers as a Task rather than a Namespace (no idea why it would though) then it will get an empty list, that might be it.

During development I used the tooltip to display the ID and detected object type for each CylcObject which sped up debugging somewhat.

kinow:
Rebased, just to make it easier for @oliver-sanders to review my PR to his branch (once I've got it working :+1: )

kinow:
I had been pushing the changes to a test branch at https://github.com/kinow/cylc-ui/tree/pr-504-part2

Now in a more organized branch as a draft PR: https://github.com/oliver-sanders/cylc-ui/pull/3

kinow:
After re-organizing the commits, and reviewing my code @oliver-sanders , everything appears to be working now.

The event is being emitted with the payload that contains the mutations for that element.

image

Just need now to listen to that event and display the menu :sweat_smile:

kinow:
The Vue+Vuetify part is pretty much done @oliver-sanders . The menu should be displayed relatively to the source where the `click` event originated.

There is only one menu in the Tree view, and one in the Workflow view (wasn't sure if it would be better to add it to the parent Layout component, but I guess we can adjust it later).

The event triggers the menu to be positioned, and then displayed.

My brain needs a few minutes of rest, without looking at Vuetify or Vue code/templates :nerd_face: :coffee: but will try to finish today or tomorrow. It needs only testing now. I am not sure if I am passing the correct information to the mutation (things like the ID of the object are wrong I think, as I couldn't hold a task—it complained about the ID not found in the running workflow).

kinow:
If my PR is merged @oliver-sanders , I can take a look at coverage and add tests if necessary too :+1: (bit complicated to look at the reports for my PR to your PR against cylc-ui's master :smile: )

kinow:
Instead of using the browser performance tabs, I've used the Vue Dev Tools as that's more convenient to test components. Here's the number of `Cylc*` components before.

image

And the loading time, clicking on the `one` workflow from the Dashboard view (i.e. loading the `Workflow` view with the mocked workflow).

image

Now after the change, only 1 `Cylc*` component, the global menu.

image

And the loading time, where `TreeItem` takes a while to load, but that's expected, and no other wrapper components. That also reduces the hierarchy in the DOM, which is good.

image

Great improvement suggestion that @oliver-sanders suggested IMO :+1:

kinow:
The way it works with the directive, @oliver-sanders , is to annotate **any** Vue component with the directive and give the directive an ID (which will be tokenized).

```
# for example, you could display the menu for the `cylc-header`, giving it a valid ID
# not that it makes sense, but hopefully it is useful to explain that the directive only depends
# on having a valid cylc ID
<cylc-header v-cylc-object="kinow|five|1|foo" />`
```

And you need to add the `<CylcObjectMenu />` somewhere in your view. That component listens for an event that the directive fires, and when the event is received, displays the menu at the source component X, Y position.

We can probably improve it later, making it easier to add the menu in views via mixins, for instance.

oliver-sanders:
Thanks to some clever work by @kinow centralising the menu this should be good to go.

Some more tests needed, will need to mock some mutations for e2e tests, can do later.

kinow (commented):
CI errors might be due to having to run `yarn install`.

kinow:
@oliver-sanders I just checked in the GH UI, and I think `package.json` is missing from this PR. But it was included in the PR to your repo - https://github.com/oliver-sanders/cylc-ui/pull/3/files.

I think a merge/rebase accident happened somewhere, and the dependency was removed. Probably copying that dependency and re-running `yarn install` will fix it :crossed_fingers:

kinow (approved):
Worked with no issues. I tested with `five`, and everything appears to have worked fine. JS code looks good! :rocket: thanks @oliver-sanders

hjoliver (commented):
Looks great :tada:  I did some pretty extensive testing, found no problems.

hjoliver (approved):

hjoliver:
Sounds like any remaining work is to be down in follow-up PRs, so this is going in :rocket:

PR #505, opened 2020-09-24
Use cycle point ID in the tree view structure

kinow:
These changes are related to #504 #339 

At the moment every node in the tree has an ID like `kinow|five|1|foo`, except the cycle points, which use just `1` (integer-type).

I implemented that way as it was the simplest approach, and because our cycle points are built from family proxies in the GraphQL query (cycle points are not types in the schema I think). @oliver-sanders found out that we didn't have the ID's while working on the mutations & api-on-the-fly integration.

**Requirements check-list**

### DIFF
--- a/src/components/cylc/tree/cylc-tree.js
+++ b/src/components/cylc/tree/cylc-tree.js
@@ -1,7 +1,7 @@
 import { computePercentProgress } from '@/components/cylc'
 import { extractGroupState } from '@/utils/tasks'
 import { merge } from 'lodash'
-import { createFamilyProxyNode } from '@/components/cylc/tree/index'
+import { createFamilyProxyNode, getCyclePointId } from '@/components/cylc/tree/index'
 import TaskState from '@/model/TaskState.model'
 
 export const FAMILY_ROOT = 'root'
@@ -288,12 +288,14 @@ class CylcTree {
     if (!familyProxy.node.state) {
       familyProxy.node.state = ''
     }
+
     // if we got the parent, let's link parent and child
     if (familyProxy.node.firstParent) {
       let parent
       if (familyProxy.node.firstParent.name === FAMILY_ROOT) {
         // if the parent is root, we use the cyclepoint as the parent
-        parent = this.lookup.get(familyProxy.node.cyclePoint)
+        const cyclePointId = getCyclePointId(familyProxy)
+        parent = this.lookup.get(cyclePointId)
       } else if (this.lookup.has(familyProxy.node.firstParent.id)) {
         // if its parent is another family proxy node and must already exist
         parent = this.lookup.get(familyProxy.node.firstParent.id)
@@ -338,8 +340,8 @@ class CylcTree {
     // NOTE: when deleting the root family, we can also remove the entire cycle point
     if (familyProxyId.endsWith('|root')) {
       // 0 has the owner, 1 has the workflow Id, 2 has the cycle point, and 3 the family name
-      const [owner, workflowId, cyclePoint] = familyProxyId.split('|')
-      nodeId = cyclePoint
+      const [owner, workflowId] = familyProxyId.split('|')
+      nodeId = getCyclePointId({ id: familyProxyId })
       node = this.lookup.get(nodeId)
       parentId = `${owner}|${workflowId}`
     } else {
@@ -347,7 +349,7 @@ class CylcTree {
       node = this.lookup.get(nodeId)
       if (node && node.node && node.node.firstParent) {
         if (node.node.firstParent.name === FAMILY_ROOT) {
-          parentId = node.node.cyclePoint
+          parentId = getCyclePointId(node)
         } else {
           parentId = node.node.firstParent.id
         }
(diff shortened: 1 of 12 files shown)


kinow:
The `TreeItem` props contain the tree node. This node now uses the cycle point ID, built from another node's ID.

image

The `CylcTree` object also stores the cycle point nodes received from GraphQL deltas using the complete cycle point ID for consistency. So either using data from deltas, or the available data in the Vue virtual DOM components data, you should have the same cycle point ID with a pattern `user|workflow|cyclepoint`

image

src/components/cylc/tree/cylc-tree.js
```diff
@@ -338,16 +340,17 @@ class CylcTree {
     // NOTE: when deleting the root family, we can also remove the entire cycle point
     if (familyProxyId.endsWith('|root')) {
       // 0 has the owner, 1 has the workflow Id, 2 has the cycle point, and 3 the family name
-      const [owner, workflowId, cyclePoint] = familyProxyId.split('|')
```
kinow, line 341: Removed `cyclePoint` as I don't need that, since I'll build the full cycle point ID

src/components/cylc/tree/cylc-tree.js
```diff
@@ -338,16 +340,17 @@ class CylcTree {
...
+      nodeId = getCyclePointId({ id: familyProxyId })
       node = this.lookup.get(nodeId)
       parentId = `${owner}|${workflowId}`
     } else {
       nodeId = familyProxyId
       node = this.lookup.get(nodeId)
       if (node && node.node && node.node.firstParent) {
         if (node.node.firstParent.name === FAMILY_ROOT) {
-          parentId = node.node.cyclePoint
+          debugger
```
kinow, line 352: grrr

src/graphql/queries.js
```diff
@@ -111,6 +111,11 @@ fragment WorkflowData on Workflow {
   port
 }
 
+fragment CyclePointData on FamilyProxy {
+  id
+  cyclePoint
+}
```
kinow, line 117: A new fragment for cycle points. Notice that it is built from a `FamilyProxy`, so effectively the `id` is a `FamilyProxy` ID

src/components/cylc/tree/index.js
```diff
@@ -37,18 +37,49 @@ function createWorkflowNode (workflow) {
...
+ *   id: string
+ * }} a tree node
+ * @return {string}
+ */
+function getCyclePointId (node) {
+  const tokens = node.id.split('|')
+  if (tokens.length >= 3) {
+    return tokens.splice(0, 3).join('|')
+  }
+  return node.id
```
kinow, line 67: Normally I would `throw new Error`, but errors are now displayed to the user... need to think if we could have a token with less than 3 parts (e.g. workflow, or a token with an invalid ID) in the deltas... :thinking: 

If I can't think of a good case, I will change it to `throw new Error`, as that way at least we will be notified about the error and be able to fix it.

kinow:
Note to self: look at the offline data, to see if we need to re-generate it

src/services/mock/one/flow.cylc
```diff
@@ -100,7 +100,4 @@
             # reset stuff
             cylc unhold \
                 "${CYLC_SUITE_NAME}" "sleepy.${CYLC_TASK_CYCLE_POINT}"
-            cylc reset \
-                "${CYLC_SUITE_NAME}" "retrying.${CYLC_TASK_CYCLE_POINT}" \
-                -s 'succeeded'
```
kinow, line 105: @oliver-sanders could you take a look at the `one` workflow, please? My Cylc-Fu is not good enough. I removed `cylc reset` after @hjoliver confirmed this subcommand was removed with SoD. But not sure how to re-write `one` now. Simply unhold'ing the task still resulted into a stalled workflow for me :cry: 

This is the last part missing, so that I can update the e2e tests, and then it should be ready for review :crossed_fingers:

kinow:
Unit tests green. Almost there. I fixed the workflow `one`, but now got a new problem. The old data was generated before we had deltas. The code used to produce the offline data used `cylc client` to run a `query workflows {}`.

But now, `cylc client` doesn't support a deltas subscription (although I think we can enable that with some parameter?). But even if that worked, we couldn't really use it, because the `generate` shell needs 1 payload returned, instead of multiple messages with the deltas.

So my idea now is to write a new query for the offline data, re-using parts of the deltas fragments. But hit a bump with https://github.com/cylc/cylc-uiserver/issues/159

kinow:
Alright, ready for review. @hjoliver is away this week, so happy if any other @cylc/core volunteers to review after @oliver-sanders , or we can wait until next week :+1:

kinow:
Added one commit that throws an `Error` if the `node.id` doesn't have at least three parts (`part1|part2|part3`). It should never happen, but in case it does, the tree would have incorrect data, and likely miss elements that need to be removed.

Tested with `five`, `families2`, and `complex`, no browser console errors. Then added one more commit with a unit test for that function :+1:

oliver-sanders (approved):
Looks good, tested as working:

Screenshot 2020-09-29 at 09 23 43

hjoliver (approved):
Tested as working :+1: 

Still needs a change log entry @kinow ?

kinow:
> Tested as working +1
> 
> Still needs a change log entry @kinow ?

Ops, just added it. Thanks @hjoliver !

PR #507, opened 2020-09-30
Use the correct node ID when removing the tree item from the cache

### LINKED ISSUE #506: Tree component's treeItemCache is not discarding nodes removed in deltas
**Describe the bug**

@oliver-sanders found that there was a memory leak, obvious after leaving a workflow running overnight. I reproduced it this morning after running `five` for a couple hours with development & product builds, and inspecting memory heap sampling and allocation in Chromium.

image

~`five` keeps only two cycle points in the UI. There are several in this `Map`, which doesn't show the `CylcTree` code, but I am 99.9999% certain that that is `CylcTree`'s lookup map (being the only map I am aware that holds ID's of job/cyclepoint/workflow/etc as key value).~

Turns out it was actually `treeItemCache` in `Tree` component.

image

`master`

Run `five` for >10 minutes, and take a couple snapshots. Compare memory heap size, and compare objects in heap between the two snapshots. Then look for detached. Looking at the retainers panel, you should find the reference to the lookup table.

**Expected behavior**

Proper memory garbage-collected and without growing exponentially over time.

**Screenshots**

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #506 

**Requirements check-list**
- [x] No changelog required (was not part of the last release).

### DIFF
--- a/src/components/cylc/tree/Tree.vue
+++ b/src/components/cylc/tree/Tree.vue
@@ -245,7 +245,7 @@ export default {
     },
     onTreeItemDestroyed (treeItem) {
       // make sure the item is removed from all caches, otherwise we will have a memory leak
-      Vue.delete(this.treeItemCache, treeItem.id)
+      Vue.delete(this.treeItemCache, treeItem.node.id)
       this.expandedCache.delete(treeItem)
       this.activeCache.delete(treeItem)
     },
--- a/src/components/cylc/tree/Tree.vue
+++ b/src/components/cylc/tree/Tree.vue
@@ -245,7 +245,7 @@ export default {
     },
     onTreeItemDestroyed (treeItem) {
       // make sure the item is removed from all caches, otherwise we will have a memory leak
-      Vue.delete(this.treeItemCache, treeItem.node.id)
+      Vue.delete(this.treeItemCache, treeItem.$props.node.id)
       this.expandedCache.delete(treeItem)
       this.activeCache.delete(treeItem)
     },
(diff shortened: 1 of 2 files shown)


kinow:
Found the issue, which looks simple to fix. But pending:

- [ ] test it and confirm the fix works in the morning, after :coffee: to make sure I'm not missing anything else
- [ ] confirm the other caches in that component are being updated correctly
- [ ] check if this was released in the alpha2, if so, add a changelog
- [ ] check if it would be possible to write a unit test to cover the cache operations (add/remove)

kinow:
After running the UI for ~15 minutes, **no detached elements in the UI**! :tada: 

image

kinow:
All done! Reviewing can be done with browser development tools, or by leaving a workflow running for a long period of time :+1: 

This PR addresses memory leaks in caches. The rest of the application looks OK, so we should be good with memory management. Easier to test with the production build, maybe in an incognito browser to avoid issues with extensions.

oliver-sanders (approved):
Learning a bit about profiling 👍.

Comparing snapshots I was able to see detached SVG elements before but not after the change.

I did spot a few other detached elements kicking about:

Screenshot 2020-10-01 at 11 19 59

kinow:
A few element should be fine. These may be collected later. Or if you have Vue Dev Tools and it's loaded, that extension retains some objects to display.

Or it could be a problem in the code. Looking at the list of retainers should help identify which part of the code is involved (for this, it's better to be in development mode, as otherwise the code is minified, and functions and variables are renamed to things like  `$t()`, `$a`, etc.

hjoliver (approved):

PR #510, opened 2020-10-04
Update dependencies

kinow:

Still cannot update `eslint-config-vuetify       ^0.5.0  →   ^0.6.1` due https://github.com/vuetifyjs/eslint-config-vuetify/issues/7

**Requirements check-list**


kinow:
Moving to draft. New linter breaks the build with new Vue rules.

kinow:
Suppressed linter errors in the dashboard.vue component. We are using the template slots as per Vuetify docs: https://vuetifyjs.com/en/components/data-tables/#item

But looks like the linter gets confused because the slots name are "item.column-name1", "item.column-name2", etc.

src/components/cylc/workflow/Lumino.vue
```diff
@@ -151,6 +151,7 @@ export default {
      * }}
      */
     onWidgetActivated (customEvent) {
+      // eslint-disable-next-line vue/custom-event-name-casing
       this.$emit('lumino:activated', customEvent.detail)
```
oliver-sanders, line 155: Funny, the docs seem to suggest that this pattern is correct?

https://eslint.vuejs.org/rules/custom-event-name-casing.html#rule-details

oliver-sanders (approved):

src/components/cylc/workflow/Lumino.vue
kinow: Yup! I guess that's a bug in their linter rule. Another rule failed too, where we had a `v-data-table` with a `v-slot="item.name"` I think. It complained about the slot name, as if it didn't recognize `item.name` and `item.age`, etc, were different slot names :man_shrugging:

oliver-sanders: Yes, was just taking a look through the rule examples and it looks like what we have is compliant.

hjoliver (approved):
LGTM, but would it be worth adding a comment to review the linter disable directives on future Vue updates, in case they fix the rules?

kinow:
> LGTM, but would it be worth adding a comment to review the linter disable directives on future Vue updates, in case they fix the rules?

Good idea. Rebased, then added the `TODO` markers, with `#510` too, so that we can quickly find the issue/PR with context about that note (to save someone from having to go through `git log`)

hjoliver (approved):

PR #512, opened 2020-10-04
Use vuetify 2.x grid classes, support responsive layout too

### LINKED ISSUE #508: Selecting states for tree view filtering changes the combo box size
**Describe the bug**

Noticed it a few days ago, but while testing the UI today I saw it happening again :worried: 

For every state you add in the Tree view filters, the combo box width appears to change a little. At least on Firefox + Ubuntu.

`master`

Open a workflow, and pick a state to filter. Then pick another one, paying attention to the width of the `select` HTML element.

**Expected behavior**

No change in the layout or component.

**Screenshots**

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #508 

No changelog as this is fixing a feature that was not released yet. Also no tests, this is only altering the layout (although it could break e2e tests, but no changes in the code/behavior).

**Requirements check-list**

### DIFF
--- a/src/components/cylc/tree/Tree.vue
+++ b/src/components/cylc/tree/Tree.vue
@@ -21,7 +21,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
       class="mb-1"
     >
       <v-col
-        class="ma-0 py-0 px-3 col-md-5 col-12">
+        class="ma-0 py-0 px-3 col-sm-12 col-md-4">
         <v-text-field
           id="c-tree-filter-task-name"
           clearable
@@ -34,7 +34,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         ></v-text-field>
       </v-col>
       <v-col
-        class="ma-0 pa-md-0 col-md-5 col-12">
+        class="ma-0 pa-md-0 col-sm-12 col-md-4">
         <v-select
           id="c-tree-filter-task-states"
           :items="taskStates"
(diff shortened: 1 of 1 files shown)


kinow:
Fixed width now.

image

Also working on mobile/responsive laytout.

image

There's one problem, when you resize the browser and before it changes from `-md` to `-lg` I think (i.e. layout size goes up from medium to large). There are some `width` values that cause the `v-select` to show more elements than its capacity, causing the element to wrap and use 2 lines.

image

The simplest solution would be to display perhaps 5 items only, and a `…` or "more" indicator. Will move that to another issue.

kinow:
Another thing to note is that Vuetify 2 is replacing Vuetify grid classes. I wasn't aware, as `v-layout` and `v-flex` are both working fine with no warnings.

In Vuetify 2.x we have `v-row` replacing `v-layout`. Immediately under the `v-row`'s you have `v-column`. `v-column` is replacing `v-flex` - https://vuetifyjs.com/en/components/grids/#v-row

kinow:
A simple way to review it, is to test the version from `master` first. Take a look at the width of the select when you pick a task state from the list.

Then check out this branch, and the width should now be fixed, immutable.

hjoliver:
On testing, a minor problem when multiple side-by-side tabs are open:
Screenshot from 2020-10-06 15-23-25

kinow:
Good spotting @hjoliver . Fixed by defining that each column has 4 columns out of the 12 in Vuetify. And that in small devices each column has 12 columns (causing it to display in rows for responsive).

image

image

hjoliver (approved):
Looks good now.

oliver-sanders (approved):

PR #517, opened 2020-10-11
Remove the current graph view and component

### LINKED ISSUE #515: Remove the current graph view and component

It was decided in our last meeting that the current Graph view and component are not working as expected for Cylc 8. Since we are still going to look at options for displaying the graph in the UI, we can safely remove the current version.

That will save us time having to update GraphQL queries, test the Graph view/component when other things are changed, and maintain the dependencies that come with the view/component (e.g. cytoscape & transitive deps like jquery).

You'd need to build & test the app several times, doing something like (not necessarily in this order)

- [x] remove Graph View
- [x] remove Graph Component
- [x] remove any state in Vuex that was used for the Graph
- [x] remove any mixin used for the Graph
- [x] remove any util code used for the Graph
- [x] remove Vue Router routes for the Graph View
- [x] remove the GraphQL queries from the Graph (or leave them commented out, or just leave without use, but check the linter)
- [x] remove unit & e2e tests
- [x] remove dependencies from package.json

**Additional context**

The placeholder for the graph view is #74. Follow the linked issues, or search for "graph" on GitHub UI for more issues related to graph view & components — if interested.

See readme and `package.json` for build commands.

**Pull requests welcome!**

datamel:
These changes close #515
- [x] Does not need tests (why? - removing feature).

So far I have... 
- Removed the Graph dependencies in package.json.
- Removed CSS for Graph components.
- Removed Graph component.
- Removed Graph View.
- Removed import of mdiGraph SVG (icon for Add View).
- Removed from Workflow View.
- Removed reference from paths.js.
- Removed App.json wording (Brazilian Portuguese also). 
- My local build was failing on some JS for ESLint rules so enabled in VS Code and accepted auto-format changes from Prettier ESLint e.g. workflow.js to get this passing.

- When removing the graph view from list in `src/components/cylc/workflow/Toolbar.vue` I removed extra...
`</v-list>
 <v-list class="pa-0">`

- No mixins used exclusively by Graph. 
- No Vuex state used exclusively by Graph.
- No graph QL queries used exclusively by Graph.
- Removed unit & e2e tests

________________________________________________________________________________

I am getting errors when running the tests:

```
GET / 503 11.059 ms - -
[9919:1011/192145.169301:INFO:CONSOLE(0)] "Access to XMLHttpRequest at 'http://userprofile/' from origin 'http://localhost:8080' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.", source: http://localhost:8080/#/workflows/one (0)
```

This error is unrelated to my change as I get the same error on master.
 Looks like `window.location.pathname` is not working in the tests on my environment. 
```js
return axios.get(`${window.location.pathname}/userprofile`).then((response) => {
```
Should that not be `window.location.hostname`? I can eliminate the CORS error with that but then get a `null` back for username. Will have to look closer tomorrow.


package.json
```diff
@@ -1,107 +1,99 @@
...
+        "vue-template-compiler": "^2.6.12",
+        "vuetify-loader": "^1.6.0",
+        "webpack": "^4.44.2"
+    },
+    "resolutions": {
+        "istanbul-instrumenter-loader/**/istanbul-lib-instrument": "4.0.1"
+    },
+    "bugs": {
+        "url": "https://github.com/cylc/cylc-ui/issues"
+    }
```
kinow, line 98: @datamel could you re-format with 2 spaces? No strong opinion on 2 vs 4, but it'd be easier to review with 2 to limit what's changed.

kinow (commented):
:clap: already looking great Mel! Thanks!!!

package.json
datamel: Sorry about that @kinow, it wasn't passing my eslint without this but I've fixed it in vscode now :)

hjoliver:
(I'll do a sanity check review on this once you've done the real review @kinow )

kinow:
> (I'll do a sanity check review on this once you've done the real review @kinow )

Thanks @hjoliver ! Will just wait for @datamel to mark as ready for review.

datamel:
Just to update, after discussing with @oliver-sanders yesterday, I have rebuilt my environment and have still got the following CORS error...
 ```
GET / 503 11.059 ms - -
[9919:1011/192145.169301:INFO:CONSOLE(0)] "Access to XMLHttpRequest at 'http://userprofile/' from origin 'http://localhost:8080' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.", source: http://localhost:8080/#/workflows/one (0) 
```

I can build and run the ui just fine, this error seems to occur on running the e2e tests.

This looks like the cause of the error: ```return axios.get(`${window.location.pathname}/userprofile`).then((response) => {```
Changing from `${window.location.pathname}` to `${window.location.hostname}` does get rid of the CORS error but then it errors with null username. 

I hadn't wanted to make as ready for review until I was able to run the tests successfully locally. 

Wondering @kinow if you have seen this error before when running the e2e tests?

**Comment by oliver-sanders** [comment]
> No 'Access-Control-Allow-Origin'

That sounds like the kind of cross-site-scripting issues you can hit when trying to load JS projects from the `file://` protocol.

**Comment by kinow** [comment]
Hi @datamel !

>This looks like the cause of the error: return axios.get(`${window.location.pathname}/userprofile`).then((response) => {
Changing from ${window.location.pathname} to ${window.location.hostname} does get rid of the CORS error but then it errors with null username.

Ah, I feel like I've seen that error before.

>I hadn't wanted to make as ready for review until I was able to run the tests successfully locally.

Did you change the mode to offline when running the e2e tests locally?

When you run `yarn run test:unit`, I think the application mode will be `test` by default? If you build with `yarn run build:watch`, I think the mode will be `development`. If you build with `yarn run build` the mode will be `production`.

And you can use any other mode if you force it through `NODE_ENV`. In Cylc UI, we use `NODE_ENV=offline`, so that the application can mock the service used to fetch workflow data. Instead of going to the backend, the mocked service will respond with data from a saved JSON.

In that case, I **think** we shouldn't get a CORS error.

Could you try

```
export NODE_ENV=offline
yarn run test:e2e
```

If that works, we can perhaps change the command in `package.json`, so that running `yarn run test:e2e` automatically sets the `NODE_ENV` (as was done in `yarn run serve` by @oliver-sanders )

Hoping that works :crossed_fingers: 
Thanks Mel!

**Comment by datamel** [comment]
> Hoping that works 🤞

Great fix, thanks so much for your help @kinow. I'll now mark this ready for review as I think I'm done :)

**Review by kinow — CHANGES_REQUESTED** [review]
@datamel found no issues with `yarn` or tests! 🎉 

Searched in the IDE for `graph[^q|i]`, and it looks like the GraphQL query for the Graph view hasn't been removed.

Can you remove [the query](https://github.com/cylc/cylc-ui/blob/e0579644abecf5ed4e636da65d9bdb5567474514/src/graphql/queries.js#L169) (also in the `export` at the end of the file; that file is a module, where we can write as much code as we want, like in Python, however, only what is `export`'ed is available to the rest of the JS code to be imported 👍 ).

You can also remove the code in the `Mutation` view that is using that query (otherwise the linter would have alerted for a value exported and never used I think): https://github.com/cylc/cylc-ui/blob/e0579644abecf5ed4e636da65d9bdb5567474514/src/views/Workflow.vue#L83

In the `Workflow` view it will be a bit more challenging. You must remove the `import` statement for the graph query that you deleted.

Then you can remove the `subscribe` method, as that's used for subscription queries without deltas (another method takes care to start subscriptions for queries with deltas).

And can also remove `setActive`, `unsubscribe`, and also the if in the deleted event. Something like this:

**Comment by datamel** [comment]
Ah super, thanks @kinow. I had noticed the `WORKFLOW_GRAPH_QUERY` but had decided to leave it after reading the `-// query to retrieve all workflows` comment. I guess I have quite a bit more to learn about how these queries are used. Thanks for your help and patience!

File: package.json [hunk 2]
```diff
@@ -46,7 +46,8 @@
     "vue-the-mask": "^0.11.1",
     "vuetify": "^2.3.12",
     "vuex": "^3.5.1",
-    "vuex-router-sync": "^5.0.0"
+    "vuex-router-sync": "^5.0.0",
+    "yarn": "^1.22.10"
```
kinow, line 50: Was it added by accident @datamel ? Yarn shouldn't be a runtime dependency for Cylc UI I think

src/services/user.service.js
```diff
@@ -23,7 +23,7 @@ import Alert from '@/model/Alert.model'
 export const UserService = {
 
   getUserProfile () {
-    return axios.get(`${window.location.pathname}/userprofile`).then((response) => {
+    return axios.get(`${window.location.pathname}userprofile`).then((response) => {
```
kinow, line 26: I think this change is not necessary any longer too (it was related to the `NODE_ENV=offline` that we had to add I think?).

I think the `pathname` may contain a file name, not a `/` (slash), in which case we need to add the slash (I think).

- https://www.w3schools.com/js/js_window_location.asp (search pathname)
- https://developer.mozilla.org/en-US/docs/Web/API/Location

kinow (commented):
@datamel check if you need to re-run `yarn install` after removing `yarn` from runtime dependencies. On GitHub actions, the current error says "error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.".

I found [one similar issue](https://github.com/yarnpkg/yarn/issues/4778#issuecomment-522575790) but not sure if we have the same issue here. More likely we just re-run `yarn install` and it goes away (weird issues like this happened more frequently with `npm`).

Thanks!

kinow:
BTW, I've finished reviewing the changed files, and everything looks OK! Also ran the e2e tests locally, and they all passed. So we just need to verify if we need the `yarn`in runtime dependencies, and update `yarn.lock`to see if CI/cypress-run build will pass on GH action 👍

package.json
datamel: yes, I have muddled up a rebase...just fixing now :)

datamel:
Thanks loads for all the help @kinow. I've been off on hols this week and had some frustrating environment issues the previous week resulting in various ui and ui-server errors. Fingers crossed this is now done and wont block the replacement graph view!

kinow (approved):

kinow:
Looks great Mel! Just pending second review. Thanks!!! 👏👏👏🎉

kinow:
@oliver-sanders would you have time to review this one, please? :grimacing:

oliver-sanders (approved):
Tested, no surprises.

PR #518, opened 2020-10-11
Add more tests for Vuex store, and remove unused code

kinow:

Trying to increase our code coverage a bit more. There are pieces of code from the theme we downloaded for the very first version of Cylc UI that are not used and can now be removed.

In our tests, we also have a very bad pattern, where the Vuex store is imported and modified during tests. This means that one unit test may modify states in the store that are used by other tests. Making it difficult to run anything in parallel, as well as to add and update tests. I will try to fix that so that we don't end up with flaky tests :+1: 

**Requirements check-list**

### DIFF
--- a/src/mixins/toolbar.js
+++ b/src/mixins/toolbar.js
@@ -39,7 +39,7 @@ export default {
   },
 
   methods: {
-    ...mapMutations('app', ['setDrawer', 'toggleDrawer']),
+    ...mapMutations('app', ['setDrawer']),
     onClickBtn () {
       this.setDrawer(!this.$store.state.app.drawer)
     },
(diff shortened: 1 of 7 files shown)


kinow:
The work for api-on-the-fly/mutations is changing `WorkflowService` and mutations & graph generator I think, so not adding tests for these files.

And the PR to remove the Graph may also change the coverage. So I think this is the best we can do for now :+1: 

Soon will add more tests for GScan component in #499

kinow:
One review should do, change to test files only.

hjoliver (approved):
Nice :+1: (to my limited understanding!)

PR #519, opened 2020-10-15
Add offline mode to e2e tests

datamel:

This fixes an issue whereby, running e2e tests results in error if NODE_ENV is not set to offline. 

Thanks to @kinow for the fix!

**Requirements check-list**
- [x] Does not need tests (used only for development).
- [x] No change log entry required (why? invisible to users).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -47,7 +47,6 @@ jobs:
           spec: "tests/e2e/**/*"
         env:
           CYPRESS_baseUrl: http://localhost:8080/
-          NODE_ENV: offline
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v1
         with:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,6 +42,7 @@ below.
  - Alex Szabó
  - Ash Newport
  - Garrett Walker
+ - Mel Hall
 
 (All contributors are identifiable with email addresses in the git version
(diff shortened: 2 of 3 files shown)


kinow (commented):
@datamel thanks for doing that! Could you also update GitHub actions, please?

https://github.com/cylc/cylc-ui/blob/master/.github/workflows/main.yml#L50

Thanks!
Bruno

datamel:
Hi @kinow, so I'm a tad confused by this github actions thing. To me it looks as if the NODE_ENV variable is already set to offline, which I guess is why the tests are running correctly on github actions?

kinow:
Hi @datamel !

Sorry for not being clear. I think the `e2e` job step in GH actions is calling `yarn run serve` before running Cypress. It is also defining the `NODE_ENV` in the `env` section of the job. But since `yarn run serve` is a Yarn command to run `NODE_ENV=offline vue-cli-service serve`, we are setting it to offline twice.

Not really related to your change, but if we could merge it here that would leave no traces of the `NODE_ENV` except for the commands defined in `package.json` :grimacing:

kinow:
But you've done everything here already, and this is not related to the change. So I've pushed the extra commit to your branch (hope that's OK!). I tried finding a way to send one of the GitHub edit suggestions, but I think you can only do that with the files included in the PR.

kinow (approved):

kinow:
CI passed! :tada: merging, thanks @datamel !

PR #520, opened 2020-10-20
Remove vuex-router-sync dependency and code.

kinow:

Found this while working on the JupyterLab tests yesterday. While reviewing our dependencies, to see what JupyterLab would need, I spotted this one that I didn't remember why we needed it.

Some `git-log`'ing later, I remembered it came with the Tim Creative theme that we used for the first versions of Cylc UI.

https://github.com/vuejs/vuex-router-sync

The `vuex-router-sync` library adds a `store.state.route` attribute to the Vuex store's state. This `route` attribute represents the current route. We haven't had to use this yet, so safe to remove IMO.

**Requirements check-list**
- [x] Does not need tests (why? removing dependency).

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,8 +53,7 @@
     "vue-spinner": "^1.0.4",
     "vue-the-mask": "^0.11.1",
     "vuetify": "^2.3.12",
-    "vuex": "^3.5.1",
-    "vuex-router-sync": "^5.0.0"
+    "vuex": "^3.5.1"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


kinow:
Tested with https://github.com/cylc/cylc-uiserver/pull/162 due to a bug in cylc-uiserver that I introduced :bug: , and used `five` to navigate the UI and see if I would get any UI error, or browser console error. Looks OK to me :+1:

hjoliver (approved):
One review should do!

PR #521, opened 2020-10-20
Replace string/text by Enums for Workflow, and use lower case by default

kinow:

When working on GScan for JupyterLab, I found some string checking with workflow states. This PR replaces that by an Enum.

I actually could not locate a place in Cylc Flow with the list of states, except for this file from TUI: https://github.com/cylc/cylc-flow/blob/fe598de4a2291335b062541f5a1eafd56b733e1a/cylc/flow/tui/__init__.py#L66

I used the same values for creating the enum.

- [x] added new `WorkflowState` enum
- [x] changed `WorkflowState` and `TaskState` both to use enum names with lower case text—we had `.toLowerCase` in many places, but rarely (never?) used the upper case text; it makes more sense to store the values as we will use it in the majority of time IMO

**Requirements check-list**
- [x] Already covered by existing tests (coverage should increase a little, but it was not the intention, see report here as it didn't comment in this PR: https://codecov.io/gh/cylc/cylc-ui/pull/521)

## Questions

-  Should it be `WorkflowState`, or `WorkflowStatus`? The GraphQL field with the value that we use for the enum is `workflows.status`. But for `TaskState` it's `workflows.taskProxy.state`.
    - Related: [state vs status #3434](https://github.com/cylc/cylc-flow/issues/3434)
- I've changed `Dashboard.vue` to dynamically create the list of states/statuses with their count values. Does it make sense? Or do we want to limit what states/statuses are displayed in the dashboard?
- Should we care about the order of states/statuses in the dashboard?

Screen shot to help with the questions above:

image


kinow:
I was going to add a `JobState`, but I think the intention is to combine Job and Task state. The `Guide.vue` view iterates a list of states and does:

```
<task :status="state" :progress="33" />
<job :status="state" />
```

for the same `state`.

oliver-sanders:
FYI: The task states are in [cylc.flow.task_state](https://github.com/cylc/cylc-flow/blob/fe598de4a2291335b062541f5a1eafd56b733e1a/cylc/flow/task_state.py#L32-L49), there is [a list](https://github.com/cylc/cylc-flow/blob/fe598de4a2291335b062541f5a1eafd56b733e1a/cylc/flow/task_state.py#L74-L84) with them in order which used to be used by the old GTK GUI.

oliver-sanders:
> Should it be WorkflowState, or WorkflowStatus?

We haven't properly standardised (hence https://github.com/cylc/cylc-flow/issues/3434). I think it was kinda supposed to be that the state "contains" the status, i.e. the task state is an entity containing the status, is_held, and some other stuff.

> I've changed Dashboard.vue to dynamically create the list of states/statuses with their count values. Does it make sense?

Fine by me, perhaps omit states which have 0 workflows.

> Should we care about the order of states/statuses in the dashboard?

I guess we should be consistent but there is no established president for order so you're free to choose.

Perhaps running > held > stopped.

oliver-sanders (approved):

hjoliver (approved):
LGTM

PR #526, opened 2020-11-03
Remove retrying and submit-retrying states

### LINKED ISSUE #514: Remove retrying and submit-retrying states
**Describe the bug**

Remove retrying and submit-retrying from `TaskStates` enum. Might be a good idea to look around for any other possible uses too :+1: 

Those states were removed in https://github.com/cylc/cylc-flow/pull/3423

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #514 

**Requirements check-list**

### DIFF
--- a/src/model/TaskState.model.js
+++ b/src/model/TaskState.model.js
@@ -26,8 +26,6 @@ class TaskState extends Enumify {
   static SUBMIT_FAILED = new TaskState('submit_failed')
   static FAILED = new TaskState('failed')
   static EXPIRED = new TaskState('expired')
-  static SUBMIT_RETRYING = new TaskState('submit_retrying')
-  static RETRYING = new TaskState('retrying')
   static RUNNING = new TaskState('running')
   static SUBMITTED = new TaskState('submitted')
   static READY = new TaskState('ready')
--- a/src/utils/tasks.js
+++ b/src/utils/tasks.js
@@ -28,8 +28,6 @@ const isStoppedOrderedStates = [
   TaskState.SUBMITTED,
   TaskState.EXPIRED,
   TaskState.READY,
-  TaskState.SUBMIT_RETRYING,
-  TaskState.RETRYING,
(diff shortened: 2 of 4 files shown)


kinow:
The checkpoint workflow, used to generate the data for the offline mode uses the retrying state too. So I will need a bit more of time understanding how to rewrite it with an xtrigger (I think that's the replacement).

hjoliver:
> The checkpoint workflow, used to generate the data for the offline mode uses the retrying state too. So I will need a bit more of time understanding how to rewrite it with an xtrigger (I think that's the replacement).

Maybe that's not necessary at this point, until we can represent xtriggers in the UI just use waiting as that is the replacement for retrying in terms of state (it is a waiting task, that is waiting on an xtrigger). So a task that fails and retries several times before succeeding (or failing) will look like this: running -> waiting -> running -> waiting -> succeeded (or failed).

kinow:
@hjoliver 

>Maybe that's not necessary at this point, until we can represent xtriggers in the UI just use waiting as that is the replacement for retrying in terms of state (it is a waiting task, that is waiting on an xtrigger). So a task that fails and retries several times before succeeding (or failing) will look like this: running -> waiting -> running -> waiting -> succeeded (or failed).

:+1: I replaced retrying and submit-retrying by waiting in our tests already. I thought that the [workflow used to generate offline data](https://github.com/cylc/cylc-ui/blob/e2726d375575aa0c7c626ed62939831ca28fb60d/src/services/mock/one/flow.cylc#L60-L78) wouldn't work after the latest changes, and I would have to update it.

But I think you are correct!

But I just tried it, and it executes successfully, producing the JavaScript data for the offline mode :tada: So nothing else to be done here. Thanks!

kinow:
@hjoliver assigning you just in case you have time for a quick review today :grimacing: similar to the other one you reviewed yesterday.

hjoliver (approved):
:+1:

PR #527, opened 2020-11-03
Allow nodes without parents in GraphQL deltas payload data

### LINKED ISSUE #523: Ignore/remove no first parent issue appropriately

See @dwsutherland 's comment about the issue in Cylc Flow [here](https://github.com/cylc/cylc-flow/issues/3658#issuecomment-714193658).

We need to ignore if a node in the tree does not have a parent in the deltas.

**Additional context**

Quite sure we would need to

- [x] check for `firstParent` in https://github.com/cylc/cylc-ui/blob/d60c6255f1b0c96c280fed75f32f9b9c7ddc8769/src/components/cylc/tree/cylc-tree.js, when deltas are added, and if not set, skip it
- [ ] update tests

**Pull requests welcome!**

kinow:
These changes close #523

**Requirements check-list**
- [x] No change log entry required (why? e.g. invisible to users, they won't notice that nodes without parents are showing or not, no UI errors fixed as well since the last release).

### DIFF
--- a/src/components/cylc/tree/cylc-tree.js
+++ b/src/components/cylc/tree/cylc-tree.js
@@ -368,6 +368,32 @@ class CylcTree {
   // --- Task proxies
 
   /**
+   * Return a task proxy parent, which may be a family proxy,
+   * or a cycle point (if the parent family is ROOT).
+   *
+   * @private
+   * @param {{
+   *   id: string,
+   *   node: Object,
+   *   children: []
+   * }} taskProxy
+   * @return {null|{
+   *   id: string,
+   *   node: Object,
+   *   children: []
+   * }}
+   */
+  findTaskProxyParent (taskProxy) {
+    if (taskProxy.node.firstParent.name === FAMILY_ROOT) {
+      // if the parent is root, we must instead attach this node to the cyclepoint!
+      const cyclePointId = getCyclePointId(taskProxy)
+      return this.lookup.get(cyclePointId)
+    }
+    // otherwise its parent **MAY** already exist
+    return this.lookup.get(taskProxy.node.firstParent.id)
+  }
+
+  /**
    * @param {{
    *   id: string,
    *   node: Object,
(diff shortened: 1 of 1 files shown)


kinow:
Manually tested it, running `complex` workflow for 20 minutes. No error messages in the browser console, with one tab open with the tree view, and sending GraphQL queries in another tab via GraphiQL (query and subscription, creating new ones every few seconds to see if it would trigger the error).

I just want to play a bit more with the UI, and then look at the tests, to see what/if could be updated, or if new tests are necessary.

kinow:
Tests look to be covering everything. The defensive programming `if` statements don't have the missing parents in the tests, but that shouldn't be a problem IMO. Ready for review.

https://codecov.io/gh/cylc/cylc-ui/commit/34336c640b90d411795d3f0c7a522171115ef19e

src/components/cylc/tree/cylc-tree.js
```diff
@@ -420,14 +422,16 @@ class CylcTree {
...
-        parent = this.lookup.get(taskProxy.node.firstParent.id)
+      if (taskProxy.node.firstParent) {
+        let parent
+        if (taskProxy.node.firstParent.id.endsWith('|root')) {
+          const cyclePointId = getCyclePointId(taskProxy)
+          parent = this.lookup.get(cyclePointId)
+        } else {
+          parent = this.lookup.get(taskProxy.node.firstParent.id)
+        }
+        parent.children.splice(parent.children.indexOf(taskProxy), 1)
```
hjoliver, line 433: This code block is identical to the one above in the CylcTree class, except for the last line - is it worth sharing to avoid potentially getting them out of sync in future?

kinow: Excellent idea @hjoliver ! I've added a `findTaskProxyParent` function that should reduce duplicity. Thanks!!!

kinow: (they were already a bit out of sync, BTW... initially I was doing the `if (taskProxy.node.firstParent.id.endsWith('|root'))` check, until I realized I didn't need to use `.endsWith`, since GraphQL provides the `node.name` attribute with the root family name; well spotted)

hjoliver: Sadly, the first excellent idea I've had regards UI coding :rofl:

hjoliver (approved):
:+1: one review will do

PR #530, opened 2020-11-05
Display message triggers

### LINKED ISSUE #402: display message triggers
[cylc-con-2020]

Display message triggers and [return code](https://github.com/cylc/cylc-flow/issues/3245#issuecomment-530983560) in the UI.

return-code

**Questions:**

* [ ] Happy.
* [ ] Should we display messages against the job which produced them or against the task?

**Pull requests welcome!**

kinow:
These changes close #402 

- [x] display `v-chip`'s with the message "labels"
- [x] display the message text as a tooltip of the `v-chip`
- [x] hide the message if its flag (`satisfied`, `received`, etc) is not `true`
- [x] limit number of `v-chip`'s to 5 most recent (remember to check the backend model for a timestamp...)
- [x] when more than 5 `v-chip`'s, display an icon showing that we have more; when pressed, open job details
- [x] also include all the "label" & "message" in the job details (leaf-node)

**Requirements check-list**

### DIFF
--- a/src/components/cylc/tree/TreeItem.vue
+++ b/src/components/cylc/tree/TreeItem.vue
@@ -89,13 +89,12 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             v-if="node.node.customOutputs.length > 0"
           >
             <v-tooltip
-              v-for="customOutput of node.node.customOutputs.slice(0, 5)"
-              :key="customOutput.id"
+              v-for="customOutput of [...node.node.customOutputs].slice(0, 5)"
+              :key="`${customOutput.id}-chip`"
               bottom
             >
               <template v-slot:activator="{ on, attrs }">
                 <v-chip
-                  :key="`${customOutput.id}-chip`"
                   v-bind="attrs"
                   v-on="on"
                   color="grey"
@@ -140,7 +139,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             >
               <div
                 v-for="customOutput of node.node.customOutputs"
-                :key="customOutput.id"
+                :key="`${customOutput.id}-leaf`"
                 class="leaf-entry"
               >
                 <span class="px-4 leaf-entry-title">{{ customOutput.label }}</span>
--- a/src/components/cylc/tree/TreeItem.vue
+++ b/src/components/cylc/tree/TreeItem.vue
@@ -88,15 +88,6 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             class="grey--text d-flex flex-nowrap flex-row align-center"
             v-if="node.node.customOutputs.length > 0"
           >
-            <v-chip
-              v-if="node.node.customOutputs.length > 5"
-              color="grey"
-              text-color="grey lighten-5"
-              class="ml-2"
-              small
-              link
-              @click="typeClicked"
-            >+{{ node.node.customOutputs.length - 5 }}</v-chip>
             <v-tooltip
               v-for="customOutput of node.node.customOutputs.slice(0, 5)"
               :key="customOutput.id"
@@ -115,6 +106,15 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               </template>
               <span>{{ customOutput.message }}</span>
             </v-tooltip>
+            <v-chip
+              v-if="node.node.customOutputs.length > 5"
+              color="grey"
+              text-color="grey lighten-5"
+              class="ml-2"
+              small
+              link
+              @click="typeClicked"
+            >+{{ node.node.customOutputs.length - 5 }}</v-chip>
           </span>
         </div>
       </slot>
--- a/src/components/cylc/tree/TreeItem.vue
+++ b/src/components/cylc/tree/TreeItem.vue
@@ -81,24 +81,20 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           <span class="mx-1">#{{ node.node.submitNum }}</span>
           <span class="grey--text">{{ node.node.host }}</span>
           <span
-            class="grey--text d-flex flex-nowrap flex-row"
-            v-if="node.node.customMessages.length > 0"
+            class="grey--text d-flex flex-nowrap flex-row align-center"
+            v-if="node.node.customOutputs.length > 0"
           >
             <v-chip
-              v-if="node.node.customMessages.length > 5"
+              v-if="node.node.customOutputs.length > 5"
               color="grey"
               text-color="grey lighten-5"
               class="ml-2"
               small
               link
               @click="typeClicked"
-            >{{ node.node.customMessages.length }}+</v-chip>
-            <!-- eslint-disable-next-line vue/no-unused-vars -->
+            >{{ node.node.customOutputs.length }}+</v-chip>
             <v-tooltip
-              v-for="(customMessage, index) of
-                    node.node.customMessages
-                      .filter(message => message.satisfied)
-                      .slice(0, 5)"
+              v-for="(customOutput, index) of node.node.customOutputs.slice(0, 5)"
               :key="index"
               bottom
             >
@@ -110,9 +106,9 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                   text-color="grey lighten-5"
                   class="ml-2"
                   small
-                >{{ customMessage.label }}</v-chip>
+                >{{ customOutput.label }}</v-chip>
               </template>
-              <span>{{ customMessage.message }}</span>
+              <span>{{ customOutput.message }}</span>
             </v-tooltip>
           </span>
         </div>
@@ -133,14 +129,14 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             <div class="leaf-entry">
               <span class="px-4 leaf-entry-title grey--text text--darken-1">outputs</span>
             </div>
-            <div v-if="node.node.customMessages.length > 0">
+            <div v-if="node.node.customOutputs && node.node.customOutputs.length > 0">
               <div
-                v-for="(customMessage, index) of node.node.customMessages"
+                v-for="(customOutput, index) of node.node.customOutputs"
                 :key="`${node.node.id}-job-custom-message-${index}`"
                 class="leaf-entry"
               >
-                <span class="px-4 leaf-entry-title">{{ customMessage.label }}</span>
-                <span class="grey--text leaf-entry-value">{{ customMessage.message }}</span>
+                <span class="px-4 leaf-entry-title">{{ customOutput.label }}</span>
+                <span class="grey--text leaf-entry-value">{{ customOutput.message }}</span>
               </div>
             </div>
             <div v-else class="leaf-entry">
--- a/src/components/cylc/tree/TreeItem.vue
+++ b/src/components/cylc/tree/TreeItem.vue
@@ -88,6 +88,20 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             class="grey--text d-flex flex-nowrap flex-row align-center"
             v-if="node.node.customOutputs.length > 0"
           >
+            <!--
+              We had a tricky bug in #530 due to the :key here. In summary, the list
+              that is backing this component changes. It contains zero or more entries,
+              up to N (5 at the time of writing).
+              Initially we used `:key=customOutput.id` here. But Vue tried to avoid
+              changing the DOM elements, which caused some elements to be out of order
+              in the final rendered UI (as Vue was trying to optimize and keep the
+              DOM elements in-place whenever possible).
+              That behaviour is not deterministic, so sometimes you would have the list
+              in order. The fix was to use a key that combines a string with the list
+              iteration `index` (the `:key` value must be unique, so we used output-chip
+              prefix).
+              @see https://github.com/cylc/cylc-ui/pull/530#issuecomment-781076619
+            -->
             <v-tooltip
               v-for="(customOutput, index) of [...node.node.customOutputs].slice(0, 5)"
               :key="`output-chip-${index}`"
--- a/src/components/cylc/tree/TreeItem.vue
+++ b/src/components/cylc/tree/TreeItem.vue
@@ -92,7 +92,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               small
               link
               @click="typeClicked"
-            >{{ node.node.customOutputs.length }}+</v-chip>
+            >{{ node.node.customOutputs.length - 5 }}+</v-chip>
             <v-tooltip
               v-for="(customOutput, index) of node.node.customOutputs.slice(0, 5)"
               :key="index"
--- a/src/components/cylc/tree/TreeItem.vue
+++ b/src/components/cylc/tree/TreeItem.vue
@@ -80,6 +80,41 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           />
           <span class="mx-1">#{{ node.node.submitNum }}</span>
           <span class="grey--text">{{ node.node.host }}</span>
+          <span
+            class="grey--text d-flex flex-nowrap flex-row"
+            v-if="node.node.customMessages.length > 0"
+          >
+            <v-chip
+              v-if="node.node.customMessages.length > 5"
+              color="grey"
+              text-color="grey lighten-5"
+              class="ml-2"
+              small
+              link
+              @click="typeClicked"
+            >{{ node.node.customMessages.length }}+</v-chip>
+            <!-- eslint-disable-next-line vue/no-unused-vars -->
+            <v-tooltip
+              v-for="(customMessage, index) of
+                    node.node.customMessages
+                      .filter(message => message.satisfied)
+                      .slice(0, 5)"
+              :key="index"
+              bottom
+            >
+              <template v-slot:activator="{ on, attrs }">
+                <v-chip
+                  v-bind="attrs"
+                  v-on="on"
+                  color="grey"
+                  text-color="grey lighten-5"
+                  class="ml-2"
+                  small
+                >{{ customMessage.label }}</v-chip>
+              </template>
+              <span>{{ customMessage.message }}</span>
+            </v-tooltip>
+          </span>
         </div>
       </slot>
       <slot name="job-details" v-else-if="node.type === 'job-details'">
@@ -93,6 +128,23 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             >
               <span class="px-4 leaf-entry-title">{{ jobDetail.title }}</span>
               <span class="grey--text leaf-entry-value">{{ jobDetail.value }}</span>
+            </div>
+            <v-divider class="ml-3 mr-5" />
+            <div class="leaf-entry">
+              <span class="px-4 leaf-entry-title grey--text text--darken-1">outputs</span>
+            </div>
+            <div v-if="node.node.customMessages.length > 0">
+              <div
+                v-for="(customMessage, index) of node.node.customMessages"
+                :key="`${node.node.id}-job-custom-message-${index}`"
+                class="leaf-entry"
+              >
+                <span class="px-4 leaf-entry-title">{{ customMessage.label }}</span>
+                <span class="grey--text leaf-entry-value">{{ customMessage.message }}</span>
+              </div>
+            </div>
+            <div v-else class="leaf-entry">
+              <span class="px-4 leaf-entry-title grey--text text--darken-1">No custom messages</span>
             </div>
           </div>
         </div>
--- a/src/components/cylc/tree/TreeItem.vue
+++ b/src/components/cylc/tree/TreeItem.vue
@@ -89,8 +89,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             v-if="node.node.customOutputs.length > 0"
           >
             <v-tooltip
-              v-for="customOutput of [...node.node.customOutputs].slice(0, 5)"
-              :key="`${customOutput.id}-chip`"
+              v-for="(customOutput, index) of [...node.node.customOutputs].slice(0, 5)"
+              :key="`output-chip-${index}`"
               bottom
             >
               <template v-slot:activator="{ on, attrs }">
--- a/src/components/cylc/tree/TreeItem.vue
+++ b/src/components/cylc/tree/TreeItem.vue
@@ -35,6 +35,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         <div :class="getNodeDataClass()" @click="nodeClicked">
           <task
             v-cylc-object="node.node.id"
+            :key="node.node.id"
             :status="node.node.state"
             :isHeld="node.node.isHeld"
             :progress=0
@@ -46,6 +47,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         <div :class="getNodeDataClass()" @click="nodeClicked">
           <task
             v-cylc-object="node.node.id"
+            :key="node.node.id"
             :status="node.node.state"
             :isHeld="node.node.isHeld"
             :progress="node.node.progress"
@@ -57,6 +59,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         <div :class="getNodeDataClass()" @click="nodeClicked">
           <task
             v-cylc-object="node.node.id"
+            :key="node.node.id"
             :status="node.node.state"
             :isHeld="node.node.isHeld"
             :progress="node.node.progress"
@@ -76,6 +79,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         <div :class="getNodeDataClass()" @click="nodeClicked">
           <job
             v-cylc-object="node.node.id"
+            :key="node.node.id"
             :status="node.node.state"
           />
           <span class="mx-1">#{{ node.node.submitNum }}</span>
@@ -152,7 +156,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         v-else
       >
         <div :class="getNodeDataClass()">
-          <span @click="nodeClicked" class="mx-1" v-if="node && node.node">{{ node.node.name }}</span>
+          <span
+            v-if="node && node.node"
+            @click="nodeClicked"
+            :key="node.node.id"
+            class="mx-1">{{ node.node.name }}</span>
         </div>
       </slot>
       <slot></slot>
--- a/src/components/cylc/tree/TreeItem.vue
+++ b/src/components/cylc/tree/TreeItem.vue
@@ -96,7 +96,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               small
               link
               @click="typeClicked"
-            >{{ node.node.customOutputs.length - 5 }}+</v-chip>
+            >+{{ node.node.customOutputs.length - 5 }}</v-chip>
             <v-tooltip
               v-for="customOutput of node.node.customOutputs.slice(0, 5)"
               :key="customOutput.id"
--- a/src/components/cylc/tree/TreeItem.vue
+++ b/src/components/cylc/tree/TreeItem.vue
@@ -134,7 +134,10 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             <div class="leaf-entry">
               <span class="px-4 leaf-entry-title grey--text text--darken-1">outputs</span>
             </div>
-            <div v-if="node.node.customOutputs && node.node.customOutputs.length > 0">
+            <div
+              v-if="node.node.customOutputs && node.node.customOutputs.length > 0"
+              class="leaf-outputs-entry"
+            >
               <div
                 v-for="customOutput of node.node.customOutputs"
                 :key="customOutput.id"
--- a/src/components/cylc/tree/TreeItem.vue
+++ b/src/components/cylc/tree/TreeItem.vue
@@ -94,12 +94,13 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               @click="typeClicked"
             >{{ node.node.customOutputs.length - 5 }}+</v-chip>
             <v-tooltip
-              v-for="(customOutput, index) of node.node.customOutputs.slice(0, 5)"
-              :key="index"
+              v-for="customOutput of node.node.customOutputs.slice(0, 5)"
+              :key="customOutput.id"
               bottom
             >
               <template v-slot:activator="{ on, attrs }">
                 <v-chip
+                  :key="`${customOutput.id}-chip`"
                   v-bind="attrs"
                   v-on="on"
                   color="grey"
@@ -118,8 +119,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           <div class="arrow-up" :style="getLeafTriangleStyle()"></div>
           <div class="leaf-data font-weight-light py-4 pl-2">
             <div
-              v-for="(jobDetail, index) in node.node.details"
-              :key="`${node.node.id}-job-detail-${index}`"
+              v-for="jobDetail in node.node.details"
+              :key="jobDetail.id"
               class="leaf-entry"
             >
               <span class="px-4 leaf-entry-title">{{ jobDetail.title }}</span>
@@ -131,8 +132,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             </div>
             <div v-if="node.node.customOutputs && node.node.customOutputs.length > 0">
               <div
-                v-for="(customOutput, index) of node.node.customOutputs"
-                :key="`${node.node.id}-job-custom-message-${index}`"
+                v-for="customOutput of node.node.customOutputs"
+                :key="customOutput.id"
                 class="leaf-entry"
               >
                 <span class="px-4 leaf-entry-title">{{ customOutput.label }}</span>
--- a/src/components/cylc/tree/index.js
+++ b/src/components/cylc/tree/index.js
@@ -18,6 +18,7 @@
 // eslint-disable-next-line no-unused-vars
 import CylcTree from '@/components/cylc/tree/cylc-tree'
 import TaskOutput from '@/model/TaskOutput'
+import Vue from 'vue'
 
 /**
  * Create a workflow node. Uses the same properties (by reference) as the given workflow,
@@ -215,7 +216,7 @@ function createCustomOutputs (job) {
 // TODO: add job-leaf (details) in the hierarchy later for infinite-tree
 function createJobNode (job) {
   const jobDetailsNode = createJobDetailsNode(job)
-  job.customOutputs = createCustomOutputs(job)
+  Vue.set(job, 'customOutputs', createCustomOutputs(job))
   return {
     id: job.id,
     type: 'job',
--- a/src/components/cylc/tree/index.js
+++ b/src/components/cylc/tree/index.js
@@ -115,10 +115,6 @@ function createFamilyProxyNode (familyProxy) {
  */
 // TODO: move expanded state to data later for infinite-tree
 function createTaskProxyNode (taskProxy) {
-  // A TaskProxy could be a ghost node, which doesn't have a state/status yet
-  if (!taskProxy.state) {
-    taskProxy.state = ''
-  }
   return {
     id: taskProxy.id,
     type: 'task-proxy',
@@ -155,6 +151,7 @@ function createJobDetailsNode (job) {
       value: job.finishedTime
     }
   ]
+
   // NOTE: `node.node.customOutputs` is not from the GraphQL query, but added
   //       in createJobNode in this module.
   return {
@@ -174,6 +171,7 @@ function createJobDetailsNode (job) {
  * If no outputs or no custom outputs are provided, then we return an empty list.
  *
  * @param {{
+ *   id: string,
  *   taskProxy: {
  *     outputs: [
  *       {
@@ -197,6 +195,11 @@ function createCustomOutputs (job) {
     .taskProxy
     .outputs
     .filter(output => !TaskOutput.enumValues.map(taskOutput => taskOutput.name).includes(output.label))
+    .map(output => {
+      return Object.assign({}, output, {
+        id: `${job.id}-output-${output.label}`
+      })
+    })
   return customOutputs || []
 }
 
--- a/src/components/cylc/tree/index.js
+++ b/src/components/cylc/tree/index.js
@@ -192,6 +192,10 @@ function createJobDetailsNode (job) {
 function createCustomOutputs (job) {
   // NOTE: the query has the filter satisfied:true already, no need to filter for that here
   //       but we want only custom outputs, so we will filter removing the standard outputs
+  if (!job || !job.taskProxy || !job.taskProxy.outputs) {
+    return []
+  }
+
   const customOutputs = job
     .taskProxy
     .outputs
--- a/src/components/cylc/tree/index.js
+++ b/src/components/cylc/tree/index.js
@@ -159,7 +159,7 @@ function createJobDetailsNode (job) {
     id: `${job.id}-details`,
     type: 'job-details',
     node: {
-      customOutputs: createCustomOutputs(job),
+      customOutputs: job.customOutputs,
       details
     }
   }
@@ -215,8 +215,8 @@ function createCustomOutputs (job) {
 // TODO: re-work the latest message, as this is the task latest message, not the job's...
 // TODO: add job-leaf (details) in the hierarchy later for infinite-tree
 function createJobNode (job) {
-  const jobDetailsNode = createJobDetailsNode(job)
   Vue.set(job, 'customOutputs', createCustomOutputs(job))
+  const jobDetailsNode = createJobDetailsNode(job)
   return {
     id: job.id,
     type: 'job',
--- a/src/components/cylc/tree/index.js
+++ b/src/components/cylc/tree/index.js
@@ -158,7 +158,8 @@ function createJobDetailsNode (job) {
     id: `${job.id}-details`,
     type: 'job-details',
     node: {
-      details
+      details,
+      customMessages: job.customMessages
     }
   }
 }
--- a/src/components/cylc/tree/index.js
+++ b/src/components/cylc/tree/index.js
@@ -17,6 +17,7 @@
 
 // eslint-disable-next-line no-unused-vars
 import CylcTree from '@/components/cylc/tree/cylc-tree'
+import TaskOutput from '@/model/TaskOutput'
 
 /**
  * Create a workflow node. Uses the same properties (by reference) as the given workflow,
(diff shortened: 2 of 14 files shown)


kinow:
Long messages are now abbreviated/truncated. It shows the first message text only, accompanied by a number as "(2)", where 2 is the total number of custom messages.

image

Also added a tooltip that displays each custom message in a paragraph. Still a draft, so will improve the code later to remove duplicated code and avoid iterating the arrays multiple times (if possible; due to `v-if` to check if we will display the component or not...)

hjoliver:
(Big discussion on Element on how to get the output message in the upstream (emitting) task, just needs a small change on the back end).

kinow:
Working on changes to start working on the updates for this PR.

First step is to:

- expand the `TaskProxy` (will edit commit to remove it later), so that I don't have to always expand a task proxy while developing :smile: lazy, I know
- add an array to the Jobs, with custom messages

I'm not worrying about which message is coming from which task/job. Instead, every job will have the same 10 messages, **only for developing/testing**. I've used the following structure

```js
const messages = [
  {
    'label1': 'message1',
    timestamp: 123456
  },
  {
    'label2: 'message2',
    timestamp: 234567
  },
```

That should be easy to use in the UI, and we easy to be modified to whichever data model is implemented in the backend :+1: 

image

kinow:
I'm working on display the messages in the job-details panel. For that I'm also doing something I've been postponing, that's to make job-details a node in the tree, which reduces a bit the logic from the template.

How does this look for now?

- Used a divider to separate job details from job messages
- Also added a message when there are no messages (ignore the actual messages for now :smile: just quick templating hacking for development)

image

- Displaying the messages

image

The questions I have right now are:

1. Should we have a divider? Or should we just display everything in a single place/group?
2. If this is looking OK, should we align the details and messages (see they are not quite the same width, but that can be fixed, just requires more work/time, so I'm asking before doing that to avoid re-work)
3. Is that OK to display all messages for now? Later we can hide and show only 5 too, and have a button or something that displays all the messages…

Thanks!

hjoliver:
I'm not too concerned about alignment and divider, whatever you think looks best is fine for this PR; we can tweak those things later if necessary.

I see you're not displaying most-recent-message first yet in the chip ordering? (if there are lots of messages, seeing the only the first 5 while the next 95 come in slowly wouldn't be ideal).

We should probably list *all* the messages in job details, so we can see what hasn't been received yet as well as what has been received, in which case some way of distinguishing the two caseds is needed - maybe greyed out for not received?

hjoliver:
Also, the `latest message` field will not be needed anymore, since that will be obvious from the list of messages and the chips.  @dwsutherland can you remove that from the schema too?

kinow:
Ah! Forgot it must be the Nth most recent messages. Will fix in the next commit.

And good idea on displaying pending messages in a different style! Thanks Hilary!

oliver-sanders:
> 1. Should we have a divider? Or should we just display everything in a single place/group?

Perhaps a sub-header called "outputs".

> 2. If this is looking OK, should we align the details and messages (see they are not quite the same width, but that can be fixed, just requires more work/time, so I'm asking before doing that to avoid re-work)

Can we use the same alignment as in the previous (job details) section? That looks pretty good.

> 3. Is that OK to display all messages for now? Later we can hide and show only 5 too, and have a button or something that displays all the messages…

We're still in α so anything flies!

We will definitely need to collapse them somehow as (for some niche cases) there will be a large number. 5 sounds like a sensible number.

kinow:
Here's today update:

- implemented all features discussed on Riot
- added a subheader as @oliver-sanders suggested, now it's much easier to locate the outputs section, and also easier for new users to get used to it IMO :+1: 
- reverted the expanded tasks so that we have the same old behaviour (will edit commits to remove this change later)

GIFrecord_2020-11-12_104825

The items are now aligned equally, in job details & output messages

image

I've changed the mocked data to return the list ordered by timestamp, with most recent messages first. That's useful for the chips, but I left it in the job details panel too. So you see `msg-label-14`, `msg-label-13`, etc. Is this order OK?

What should we do about the mobile/smaller viewports?

image

hjoliver:
Looking really good @kinow :+1: 

> I've changed the mocked data to return the list ordered by timestamp, with most recent messages first. That's useful for the chips, but I left it in the job details panel too. So you see msg-label-14, msg-label-13, etc. Is this order OK

If we always displayed the entire list (in job details) it should probably be in the natural order (not reversed like the chips). BUT we intend to truncate that list by default, to say five lines (in this PR, or in a follow-up?), so that would require reverse ordering like the chips to ensure that the most recent messages are prioritized for display.  That being the case, it seems to me we'd want:
always reverse order (with and without truncation); OR natural order when expanded and reverse order when truncated - but that would look weird when toggling between expanded and truncated?   So on that basis I think we should stick with reverse order!

> What should we do about the mobile/smaller viewports?

Can we cut off more chips as the window shrinks, down to a single one, rather than flowing onto the next lines?

Hmm, that suggests we should not have a fixed number like 5, but as many as will fit in the current window size? (Is that easy enough to do?)

kinow:
@dwsutherland shared this query of what it might look like in the GraphQL implementation. It matches the schema used in this draft PR (thanks a lot David!), so finishing up this PR once we are done with the final adjustments should be fairly simple.

```graphql
query {
  workflows (ids: ["*|vix"]) {
    tasks (ids: ["foo"]) {
      proxies (ids: ["1|foo"]) {
        latestMessage
        outputs (satisfied: false, limit: 5, sort: {keys: ["time"], reverse: true}) {
          label
          message
          satisfied
          time
        }
        extras
      }
    }
  }
}
```

kinow:
> Looking really good @kinow +1

Thanks!

> > I've changed the mocked data to return the list ordered by timestamp, with most recent messages first. That's useful for the chips, but I left it in the job details panel too. So you see msg-label-14, msg-label-13, etc. Is this order OK
> 
> If we always displayed the entire list (in job details) it should probably be in the natural order (not reversed like the chips). BUT we intend to truncate that list by default, to say five lines (in this PR, or in a follow-up?), so that would require reverse ordering like the chips to ensure that the most recent messages are prioritized for display. That being the case, it seems to me we'd want:
> always reverse order (with and without truncation); OR natural order when expanded and reverse order when truncated - but that would look weird when toggling between expanded and truncated? So on that basis I think we should stick with reverse order!

All good for me :+1: 

> > What should we do about the mobile/smaller viewports?
> 
> Can we cut off more chips as the window shrinks, down to a single one, rather than flowing onto the next lines?
> 
> Hmm, that suggests we should not have a fixed number like 5, but as many as will fit in the current window size? (Is that easy enough to do?)

Good idea! Let me think a bit how to implement this… :thinking:

kinow:
So far the only thing I could think of, is controlling the amount of chips displayed based on the viewport. So small viewport display 1, medium display 3, and bigger ones display 5.

However, this works only when the host name is not really long. Right now it looks a bit busy with so many messages in two side-by-side tree views.

image

The simplest approach would be to display just 1 chip, with either the number of messages, or just the first message label? Or do that in the small viewport, and then show 5 when the viewport is bigger (tablet & desktop sizes)?

hjoliver:
> The simplest approach would be to display just 1 chip, with either the number of messages, or just the first message label? Or do that in the small viewport, and then show 5 when the viewport is bigger (tablet & desktop sizes)?

I think that would be fine for this PR, and consider improving it later.

kinow:
> > The simplest approach would be to display just 1 chip, with either the number of messages, or just the first message label? Or do that in the small viewport, and then show 5 when the viewport is bigger (tablet & desktop sizes)?
> 
> I think that would be fine for this PR, and consider improving it later.

Do you mean to display 1 chip, or doing that when the viewport is small, and keep the 5 when the viewport is bigger @hjoliver ?

hjoliver:
Actually I'm not sure. I don't think it's great to display only a single chip no matter what the view port.  Ideally we need to show as many messages as possible without wrapping the v-chips onto the next line.  But 5 most recent messages with wrapping in smaller view ports is probably fine for this PR - would you agree @oliver-sanders ?

oliver-sanders:
I would have thought something like displaying the five most recent with `overflow-y: none` (and an ellipsis at the end of the line in the future).

hjoliver:
I was wondering if we couldn't get the ellipsis in the right place if using CSS overflow like that. But I suppose it could be in another div on the right of the truncated list of v-chips?

kinow:
Ellipsis are easier to use with text. With children <div>/<span>/etc it gets trickier.

Tried a few variations today, without luck. Hiding when viewport is small doesn't necessarily solve it. In medium size it may still create extra lines due to label width.

The easiest approach I found was to make it display the job information, including messages, in a single row, without wrapping. Does that look like a possible approach (maybe for now?)?

GIFrecord_2020-11-19_215600

hjoliver:
@kinow  in @oliver-sanders comment:

> I would have thought something like displaying the five most recent with overflow-y: none (and an ellipsis at the end of the line in the future).

I guess by `overflow-y:none` he means `overflow-y:hidden`?  Wouldn't that cause the chips to split onto following lines, but be hidden, with no scroll bar?  Have you tried that?

What about doing that, with max 5 chips, but put the ellipsis (or `+`) **first**, in front on the chips?   Even if ellipsis first isn't the ideal place to put it, that would provide the necessary interface (most recent chip(s) visible, and ellipsis if there's more, without having to change the number of chips as the window shrinks.  That might be the best interim solution, until/if we figure out something better.

kinow:
>I guess by overflow-y:none he means overflow-y:hidden? Wouldn't that cause the chips to split onto following lines, but be hidden, with no scroll bar? Have you tried that?
>
>What about doing that, with max 5 chips, but put the ellipsis (or +) first, in front on the chips? Even if ellipsis first isn't the ideal place to put it, that would provide the necessary interface (most recent chip(s) visible, and ellipsis if there's more, without having to change the number of chips as the window shrinks. That might be the best interim solution, until/if we figure out something better.

I tried that, but couldn't get the ellipsis or indicator of more chips to work. Moving the + sign before the chips could work. I think as long as we are consistent, that should be OK. But if in other elements we have an ellipsis or + after the list, then it would be at least awkward IMO.

How does this look?

GIFrecord_2020-11-20_100758

If you have a long task proxy name (or deeply nested hierarchy for parents), several jobs, or a long host name, the scroll bar might be displayed anyway (repair that the `20000102T0000Z / GOOD / SUCCEEDED / eventually_succeeded`) the scroll bar will be activated anyway. So I thought that limiting the number of chips for that case wasn't worth, that's why last evening I tried to show the 5 chips inline.

image

But not sure which approach is better for Cylc users :-) so happy to go with either way.

hjoliver:
Hmm maybe you're right.

@oliver-sanders - what do you think - shall we just go with 5 chips and a scroll bar for narrow view ports, for now?

kinow:
ping @oliver-sanders :point_up: :grimacing:

oliver-sanders:
The second example is really nice (no idea how it works)!

But would want to find some way to get the ellipsis after rather than before or else make it more informational (e.g. it could give the number of custom outputs).

kinow:
Huh, trying to rebase it, but git is getting utterly confused after we added slots. Trying not to remove anything important, and build/testing it locally after every major rebase/conflict-fixing.

kinow:
Rebased! Now going through some parts of the code that were under development to remind myself what I was doing :smile: after that, just will use the approach we discussed above, to update the PR. Then ask a quick pre-review from @hjoliver & @oliver-sanders to make sure I got it right, and finish it with tests & changelog :muscle:

kinow:
@hjoliver , @oliver-sanders here's the preview of the current version of this PR:

Screenshot from 2021-01-13 14-35-25

If there are more than 5 messages, then we first display a chip with the number of messages and the plus sign, indicating that there are more messages, followed by the first 5 messages. Otherwise, if there are 5 or less messages, we just display the messages.

Clicking on the chip with the number of messages shows the job details. Same behaviour if the user clicks on the expand button in the tree for the job.

Passing the mouse over the chips displays the message text in the tooltip.

What do you think?

hjoliver:
Looks good, I'll aim to have a play with it this evening.

Does it have a scrollbar, if chips are hidden? (my comment above suggests I was asking for overflow:hidden to avoid that, but I'm not sure if that's desirable now...)

kinow:
>Looks good, I'll aim to have a play with it this evening.

Great! Let me know if there's anything you'd like to change. I've just gone through the changes twice, and now I think I remembered everything that was done in this PR.

> Does it have a scrollbar, if chips are hidden? (my comment above suggests I was asking for overflow:hidden to avoid that, but I'm not sure if that's desirable now...)

I think you are talking about the horizontal scrollbar, on the x-axis (`overflow-x`). There will be one, even if chips are hidden.

GIFrecord_2021-01-13_154106

kinow:
I think I have fixed the unit tests. It was an actual bug, where I used the wrong variable name :-) good to have tests.

Now it should update coverage, which will be useful later when I have to write the tests.

src/components/cylc/tree/TreeItem.vue
```diff
@@ -281,6 +334,50 @@ export default {
...
       return classes
+    },
+    /**
+     * Get the summary of custom messages for a node.
+     * @param {{
+     *   node: {
+     *     taskProxy: {
+     *       prerequisites: [
+     *         {
+     *           conditions: [
```
dwsutherland, line 345: Alongside conditions you may want to get the expression (as sometimes there are aliases for conditional statements)

kinow:
Rebased. Will start working to add the tests now.

kinow:
Added changelog (previous commit with new tests passed CI). Rebased. Ready for review! :+1:

src/components/cylc/tree/TreeItem.vue
kinow: Not sure how/when I would use it though?

image

Is it so that we can link the prerequisite with the condition?

kinow:
Moved back to draft because I realized I didn't implement the part where I get the messages from the backend :sweat_smile: 

@dwsutherland / @hjoliver could you help me remembering how we should fetch that from the backend? Last I worked on it, I think we had defined that it should be displayed as `message=value`, and that it was pending some work in the UI.

There's a linked PR that has been merged, but I am not sure if that's the pending issue? Here's the workflow I am using

```
[scheduling]
   cycling mode = integer
   initial cycle point = 1
   [[graph]]
      R1 = """
           foo:out1 => bar
           baz:out1 => bar
      """
[runtime]
   [[foo]]
       script = """
           sleep 20
           cylc message "Aliquam a lectus euismod, vehicula leo vel, ultricies odio."  # <-----
           sleep 20
       """
       [[[outputs]]]
           out1 = "Aliquam a lectus euismod, vehicula leo vel, ultricies odio."
   [[baz]]
      script = """cylc message "Lorem ipsum dolor sit amet, consectetur.""""
      [[[outputs]]]
         out1 = "Lorem ipsum dolor sit amet, consectetur."
```

So should we have in the UI a message `out1="Lorem ipsum dolor sit amet, consectetur."` when the backend returns that prerequisite with `satisfied=true`?

hjoliver:
@kinow I just built this branch to take a look, and I can't open task nodes or job leaves anymore - and same on current master - we might have merged the collapse/expand PR a bit prematurely! (in final review I tested collapse expand functionality without opening jobs, I guess)

kinow:
Strange. But I don't recall opening jobs today either. Will test master first thing in the morning 👍

kinow:
Expanding jobs and tasks appears to be working with `yarn run serve` (offline mode)

GIFrecord_2021-02-16_084131

And with `yarn run build:watch`, both from `master` branch.

GIFrecord_2021-02-16_085123

@hjoliver or do you mean the "expand all" button is not expanding the tasks & jobs?

hjoliver:
Huh, no I meant what you've just demonstrated is working. Must be my bad, sorry!   I did try recreating all my environments.  And, actually, going back to build an earlier version worked :thinking:   I'll try again...

hjoliver:
It's working now.  Sorry for the false alarm, no idea what I did to cause that last night :grimacing:

kinow:
> It's working now. Sorry for the false alarm, no idea what I did to cause that last night grimacing

Not a problem :-) it might have been some browser cache, or yarn/npm/etc unhappy with some dependency, or producing a corrupted JS file.

hjoliver:
Hmm, now I'm getting the aforementioned problem again, on this branch. Console shows:
> TypeError: n.node.node.customMessages is undefined

kinow:
Sorry, I realized yesterday that I hadn't implemented the GraphQL conversion to JS. THe offline mode should provide the right UI components, while I try to convert the GraphQL data into the expected JS data structure.

hjoliver:
Ah, that might explain it! :grin:

hjoliver:
Yep, it's looking good in offline mode.

hjoliver:
> So should we have in the UI a message out1="Lorem ipsum dolor sit amet, consectetur." when the backend returns that prerequisite with satisfied=true?

Just trying to recall the problem here.  I think it is this:
```
[scheduling]
   [[graph]]
       R1 = foo:out1 => bar
[runtime]
   [[foo]]
        [[[outputs]]]
             out1 = "data file ready"
```
Here `foo` generates the output message, and `bar` depends (prerequisite) on it.  In the UI, we want the outputs associated with the generating task, not the dependent task. I.e. when `foo` generates the message, it should appear against `foo` in the UI. But at the moment we are querying the dependent tasks for their prerequisites.

I'll go look at the schema...

kinow:
Thanks @hjoliver ! I'm looking at GraphiQL at this exact moment, trying to figure out how to link prerequisites/conditions and outputs.

I was modifying the fragment for Job.

```graphql
fragment JobData on Job {
  id
  firstParent: taskProxy {
    id
    __typename
  }
  jobRunnerName
  jobId
  host
  startedTime
  submittedTime
  finishedTime
  state
  submitNum
  taskProxy(stripNull: false) {
    prerequisites {
      expression
      satisfied
      conditions {
        exprAlias
        reqState
        satisfied
        message
      }
    }
    outputs {
      label
      message
      satisfied
    }
  }
}
```

@hjoliver I think that would give us the messages for that job, which is what we want? My problem is that at the moment I get some data like:

```js
              "jobs": [
                {
                  "id": "kinow|messagetrigger1|1|foo|1",
                  "firstParent": {
                    "id": "kinow|messagetrigger1|1|foo",
                    "__typename": "TaskProxy"
                  },
                  "jobRunnerName": "background",
                  "jobId": "14106",
                  "host": "localhost",
                  "startedTime": "2021-02-16T11:48:46+13:00",
                  "submittedTime": "2021-02-16T11:48:45+13:00",
                  "finishedTime": "2021-02-16T11:49:26+13:00",
                  "state": "succeeded",
                  "submitNum": 1,
                  "taskProxy": {
                    "prerequisites": [],
                    "outputs": [
                      {
                        "label": "expired",
                        "message": "expired",
                        "satisfied": false
                      },
                      {
                        "label": "submitted",
                        "message": "submitted",
                        "satisfied": true
                      },
                      {
                        "label": "submit-failed",
                        "message": "submit-failed",
                        "satisfied": false
                      },
                      {
                        "label": "started",
                        "message": "started",
                        "satisfied": true
                      },
                      {
                        "label": "succeeded",
                        "message": "succeeded",
                        "satisfied": true
                      },
                      {
                        "label": "failed",
                        "message": "failed",
                        "satisfied": false
                      },
                      {
                        "label": "out1",
                        "message": "Aliquam a lectus euismod, vehicula leo vel, ultricies odio.",
                        "satisfied": true
                      }
                    ],
                    "__typename": "TaskProxy"
                  },
                  "__typename": "Job"
                }
              ]
            }
          ]
```

There are several `output`s, but I **think** in the UI we want to display only the custom ones... they are in `prerequisites`, but I don't know how to link the `prerequisite` with `output` (one has `c0`, the other has `out1`).

hjoliver:
> There are several outputs, but I think in the UI we want to display only the custom ones

Yes

hjoliver:
>  but I don't know how to link the prerequisite with output (one has c0, the other has out1).

Why do we need the prerequisites here at all? For displaying outputs, isn't the outputs query sufficient?

hjoliver:
(I was thinking outputs must be missing from the schema, but they are there - and your query just proved it)

hjoliver:
This was the comment above that I was recalling https://github.com/cylc/cylc-ui/pull/530#issuecomment-722739619

> (Big discussion on Element on how to get the output message in the upstream (emitting) task, just needs a small change on the back end).

I can't remember what the small back-end change required was, or if we've done it yet.  Searching in element...

kinow:
> This was the comment above that I was recalling [#530 (comment)](https://github.com/cylc/cylc-ui/pull/530#issuecomment-722739619)
> 
> > (Big discussion on Element on how to get the output message in the upstream (emitting) task, just needs a small change on the back end).
> 
> I can't remember what the small back-end change required was, or if we've done it yet. Searching in element...

I saw that comment too, and thought it was this PR https://github.com/cylc/cylc-flow/pull/3938 that was linked around the time of that message.

kinow:
> (I was thinking outputs must be missing from the schema, but they are there - and your query just proved it)

Yup, but I can't tell what is a custom output or not. Unless we have an enum somewhere with values to ignore (e.g. submitted, submit-failed), or if we have a place to find the custom outputs (that's why I thought we needed the `prerequisites`, to find the custom outputs...)

hjoliver:
> I saw that comment too, and thought it was this PR cylc/cylc-flow#3938 that was linked around the time of that message.

Yeah I think you're right, in which case we are good to go with outputs from the back end.

> Yup, but I can't tell what is a custom output or not. Unless we have an enum somewhere with values to ignore (e.g. submitted, submit-failed), or if we have a place to find the custom outputs (that's why I thought we needed the prerequisites, to find the custom outputs...)

Ah, OK. In that case it's easy (phew) - no need to use the prerequisites of downstream tasks.

A custom output is anything not in the list of standard outputs, which reflect the standard state changes that all tasks go through.  

Here's the list of all possible standard outputs: https://cylc.github.io/cylc-admin/proposal-state-names.html#outputs 

(But note we decided not to implement the "new" column).

hjoliver:
Standard outputs are the same as standard task states, except for started (message) instead of running (state) ... because a "started" message gets sent when the task starts running.

kinow:
> Standard outputs are the same as standard task states, except for started (message) instead of running (state) ... because a "started" message gets sent when the task starts running.

Ah, excellent. In which case I can write a new enum in Cylc UI. Before that, just to confirm: the values in `TaskState` cannot be re-used for these standard outputs, right? The table you linked doesn't have `running`, for example, which is a task state in Cylc UI. But asking just to make sure :grimacing:

hjoliver:
You might have missed my previous comment just above, on the running state/started output :grin:

hjoliver:
But anyway, yes, outputs and states unfortunately aren't quite the same because of started/running.  They're defined `task_outputs.py` in cylc-flow:
```python
# Standard task output strings, used for triggering.
TASK_OUTPUT_EXPIRED = "expired"
TASK_OUTPUT_SUBMITTED = "submitted"
TASK_OUTPUT_SUBMIT_FAILED = "submit-failed"
TASK_OUTPUT_STARTED = "started"
TASK_OUTPUT_SUCCEEDED = "succeeded"
TASK_OUTPUT_FAILED = "failed"
```

kinow:
Ops, my bad. Re-reading it, all good! Let me have a go at it.

kinow:
I think I got it now :crossed_fingers: 

I'm using the `outputs` of a Job's task proxy (so the outputs of each job). I'm also filtering for only `satisfied: true` outputs, and reverse-sorting by timestamp (I believe that will give us the newest ones first).

The UI components haven't changed, except that I added a class to have `align-items: center` in the flex container for the v-chips. That way, if you expand the job details (easier in offline as it has multiple nodes with job details and custom outputs), you should see a small space versus the previous behaviour where the v-chips were glued to the details panel (easier to see in the screenshot below where it changed)

image

@hjoliver tested with offline, and with real data (messagetrigger1 workflow below).

```
[scheduling]
   cycling mode = integer
   initial cycle point = 1
   [[graph]]
      R1 = """
           foo:out1 => bar
           baz:out1 => bar
      """
[runtime]
   [[foo]]
       script = """
           sleep 20
           cylc message "Aliquam a lectus euismod, vehicula leo vel, ultricies odio."  # <-----
           sleep 20
       """
       [[[outputs]]]
           out1 = "Aliquam a lectus euismod, vehicula leo vel, ultricies odio."
   [[baz]]
      script = """cylc message "Lorem ipsum dolor sit amet, consectetur.""""
      [[[outputs]]]
         out1 = "Lorem ipsum dolor sit amet, consectetur."
~                                                               
```

kinow:
Hmm, looking at the failing e2e test...

kinow:
e2e test passed on ubuntu + chrome. The macos failure is related to #590 . @hjoliver ready for review I think.

hjoliver:
Brilliant, I’m just out picking the kids up from school (rainy day)

hjoliver:
You've basically nailed it :tada: - but some minor problems noted:

Test flow, with tasks slowly generating 10 messages:
```ini
#!Jinja2
{% set N_MSGS = 10 %}

[scheduling]
   [[graph]]
      R1 = """ 
          {% for I in range(0,N_MSGS) %}
             foo:file_{{I}} & bar:blob_{{I}} => baz 
          {% endfor %}
      """ 
[runtime]
   [[foo]]
       script = """ 
          {% for I in range(0,N_MSGS) %}
              cylc message "File {{I}} ready"
              sleep 5
          {% endfor %}
       """
      [[[outputs]]]
          {% for I in range(0,N_MSGS) %}
              file_{{I}} = "File {{I}} ready"
          {% endfor %}
   [[bar]]
       script = """ 
          {% for I in range(0,N_MSGS) %}
              cylc message "Blob {{I}} ready"
              sleep 10
          {% endfor %}
       """
      [[[outputs]]]
          {% for I in range(0,N_MSGS) %}
              blob_{{I}} = "Blob {{I}} ready"
          {% endfor %}
   [[baz]]
      script = "sleep 20"
```

Screenshot from 2021-02-16 16-43-09

- we should sort for most recent chip first, so that the latest message is always visible first
   - (but listed in order of generation in the job leaf, which you have right already)
- chips occasionally appear out of order (see above: `blob_4, _0, _1, _2, _3`)
- can we have the overflow chip (for want of a better word) say `+<n-not-shown>` rather than `<total-number>+`
   - so for 7 messages (say), it would be:  `+2, blob_7, blob_6, blob_5, blob_4, blob_3`

Also:
- the chips aren't appearing reactively - I have to close and open the job node or job leaf to update after new messages have arrived

kinow:
Thanks for the great testing (as always!!!) @hjoliver :tada: 

Will go through your feedback tomorrow in the morning with a fresh pair of eyes :nerd_face: 

>the chips aren't appearing reactively - I have to close and open the job node or job leaf to update after new messages have arrived

Hmmm, will look into it tomorrow, but I wonder if that's the same problem we have in #586 ? I might be able to test it with #591

kinow:
@hjoliver 

>- we should sort for most recent chip first, so that the latest message is always visible first
>   - (but listed in order of generation in the job leaf, which you have right already)

:+1: I used `sort: ["timestamp"], because I think I had used it somewhere else. But without auto-complete, I didn't confirm it existed in the schema. The right key is "time". I'll add an issue to cylc-uiserver about it

>- chips occasionally appear out of order (see above: `blob_4, _0, _1, _2, _3`)

Same issue as above, now they appear to be working OK after the query was updated. I'm not doing any sort in the UI and delegating it to the backend GraphQL query.

>- can we have the overflow chip (for want of a better word) say `+<n-not-shown>` rather than `<total-n>+`
>   - so for 7 messages (say), it would be:  `+2, blob_7, blob_6, blob_5, blob_4, blob_3`

Ah, didn't think about this option :smile: probably more useful to know how many more 

>Also:
>- the chips aren't appearing reactively - I have to close and open the job node or job leaf to update after new messages have arrived

Hmmm, will need more time for this one.

src/components/cylc/tree/index.js
```diff
@@ -155,18 +153,64 @@ function createJobDetailsNode (job) {
     },
     {
       title: 'latest message',
-      property: job.latestMessage
+      value: job.latestMessage
```
kinow, line 156: Going away in #596, but included it to make sure I wasn't mutating any object incorrectly (i.e. maybe somewhere else I could be doing `obj.value=''` and then it would have `title`, `property`, and `value`. Will rebase and remove it if #596 is merged first.

kinow:
Rebased. Picked the changes from #586 , massaged that code & tests for this branch. Updated tests, unit & e2e tests passing locally and on CI. Ignore the macos tests as per #590 :+1: 

@hjoliver I spent some time debugging and testing the UI, but still curious whether you'll find some bug that I missed :smile: ready for review again :rocket:

hjoliver:
This works really well now @kinow :tada: 

Two small questions on the "and N more" chip:
- can you change it from (e.g.) `3+` to `+3`? 
  - ("3+" would normally be read as *"more than 3"* :-1: , whereas "+3" means *"and 3 more"* :+1: )
- can it be moved to from the left side to the right? 
  - with the most recent outputs now on the left, it makes more sense to put "+3" at the end IMO. It doesn't really matter that "+3" will be obscured if the view is too narrow. 
  - (which makes me wonder if we could revisit the idea of hiding the div overflow instead of using a scroll bar ... but not for this PR)

E.g. `c9, c8, c7, c6, c5, +5` instead of `5+, c9, c8, c7, c6, c5`

oliver-sanders:
> can it be moved to from the left side to the right?

Would be great, however, I think there was some technical reason why it is tricky.

Looks good:

outputs

We should probably do something with the expanded job info box incase someone has a task that produces hundreds of outputs, maybe make it scrollable? Follow-on for later.

kinow:
> This works really well now @kinow tada

:tada: 

> Two small questions on the "and N more" chip:
> 
>     * can you change it from (e.g.) `3+` to `+3`?
>       
>       * ("3+" would normally be read as _"more than 3"_ -1 , whereas "+3" means _"and 3 more"_ +1 )

Sure, easy-peasy. Done!

>     * can it be moved to from the left side to the right?
>       
>       * with the most recent outputs now on the left, it makes more sense to put "+3" at the end IMO. It doesn't really matter that "+3" will be obscured if the view is too narrow.
>       * (which makes me wonder if we could revisit the idea of hiding the div overflow instead of using a scroll bar ... but not for this PR)
> 
> 
> E.g. `c9, c8, c7, c6, c5, +5` instead of `5+, c9, c8, c7, c6, c5`

Ah, I remember you suggested it earlier. It's easy to do that, but not sure if that's the best location for that indicator. See screenshot below:

image

It's using your first improvement suggested above BTW, so while my workflow is running, I can see the +5 chip indicating it has other 5 tasks (really good idea to show remaining tasks, not total).

But since I have more than 1 widget, my viewport is not large enough to show the last chips (left widget). Which means if we move the +5 to the right, then it won't be visible, only when the widget is large enough to show all chips. Even 1 single chip with a very long text is enough to hide the +N indicator.

What do you think?

Thanks for the review!

kinow:
Rebased, and latest message is no more :tada: thanks @hjoliver !

kinow:
> > can it be moved to from the left side to the right?
> 
> Would be great, however, I think there was some technical reason why it is tricky.
> 
> Looks good:
> 
> outputs
> 
> We should probably do something with the expanded job info box incase someone has a task that produces hundreds of outputs, maybe make it scrollable? Follow-on for later.

Ah!! :facepalm: I didn't consider what would happen with the job info box. It would be really annoying having to scroll, scroll, scroll… and then maybe more outputs appear, scroll a little bit more…

Let's fix it. There is already a `<div>` wrapping the outputs. I've named it `leaf-outputs-entry`. The other entries are `leaf-entry`. The `leaf-outputs-entry` got a `overflow-x: scroll` and also `max-height: ($line-height * $visible-outputs + 1.0em);` css attributes.

I defined `$visible-outputs` equal `5` for testing, but without thinking too hard about it. Does it sound like a good number? Or should we allow more? Probably I chose 5 just because we have 5 visible up next to the task icon (the chips).

The `$line-height` was added during that infinite tree experiment, setting every node in the tree to have the same height (so that we could have that infinite scroll, which is helpful now too). Since I wanted to show 5, I've simply multiplied 5 times the $line-height. Looks like it's enough space to show ~4.5 items. For some reason the last item is partially hidden (even though I couldn't find any padding/margin).

So I added a `1.0em` buffer, and that appears to work on Chrome and Firefox on Linux. 

image

hjoliver:
>  It's easy to do that, but not sure if that's the best location for that indicator.

We're typically interested in the most recent messages, then (with the new ordering: most recent at left) you would scroll right to see older messages, and finally you would see the "+3" chip that says there are three more even older ones that are not displayed. So I think that's fine.  And I see @oliver-sanders said this "would be great, however, I think there was some technical reason why it is tricky." ... so if it's easy, let's do it :tada:

hjoliver:
5 messages before scrolling in the job leaf is good, IMO :+1:

kinow:
> > It's easy to do that, but not sure if that's the best location for that indicator.
> 
> We're typically interested in the most recent messages, then (with the new ordering: most recent at left) you would scroll right to see older messages, and finally you would see the "+3" chip that says there are three more even older ones that are not displayed. So I think that's fine. And I see @oliver-sanders said this "would be great, however, I think there was some technical reason why it is tricky." ... so if it's easy, let's do it tada

Then let's do it! 

image

kinow:
Alright, ready for review again @hjoliver ! :rocket:

hjoliver:
Looking really good now, except that the chip sorting isn't quite working, as illustrated by task foo in your screenshot just above. (Did you already explain that yesterday?)

kinow:
Oh, that happened yesterday, and I was going to file an issue to Cylc UI Server but I forgot. I created an issue for `sort: ["keydoesnotexist"]`. But yesterday I monitored a GraphQL with `sort: ["time"]` that returned the tasks out of the expected order.

However; the `time` on each task was correct? I'll try to reproduce it here to show what I mean. I can easily fix it by `sort`ing the tasks, but I want to see if I can delegate that to the backend instead.

kinow:
The data in Vue is sorted correctly. Now trying to understand why it's not sorted when rendered. The issue I mentioned above hasn't happened since yesterday, but I have added a post-it to try that later on `master` (after syncing branches). But doesn't seem related to this problem.

kinow:
It's not clear to me what's the lifecycle of objects in our UIS/Flow pools when workflows are restarted, sorry. Could someone help me :grimacing: 

@hjoliver @dwsutherland using GraphiQL ([query](https://gist.github.com/kinow/ed3de3de44c3b06e53129bd2733b9199#file-query-gql)) when I start/stop the [workflow](https://gist.github.com/kinow/ed3de3de44c3b06e53129bd2733b9199#file-workflow-cylc) I get outputs out of order. But I **think** some of the outputs are from a previous run?

image

Variables for query in GraphiQL:

```js
{
  "workflowId": "kinow|530"
}
```

Just start `530`, wait for some `file_` and `blob_` messages, then stop the workflow, wait a couple seconds, start it again. I'm destroying my venv's and syncing branches to have `cylc play` now, but I think this issue might not be fixed after that?

hjoliver:
Good question, I'll have to take a look

hjoliver:
There's a strange box at the bottom of the job leaf outputs div, BTW (for horizontal scroll bar?)
Screenshot from 2021-02-18 10-59-47

hjoliver:
> when I start/stop the workflow

What do you mean by start/stop exactly? `stop --now` then `restart`?  (I'm confused because in the example flow all the output-generating tasks run immediately - no dependencies, and no queues, so a normal stop won't happen until all the output-generating tasks are finished).

kinow:
> > when I start/stop the workflow
> 
> What do you mean by start/stop exactly? `stop --now` then `restart`? (I'm confused because in the example flow all the output-generating tasks run immediately - no dependencies, and no queues, so a normal stop won't happen until all the output-generating tasks are finished).

Sorry, I normally stop a workflow with CTRL + C in the terminal. Here's what my terminal looks like when I stop it:

image

hjoliver:
OK that's essentially the same as `stop --now --now`. It leaves the "orphaned" task jobs running, and the scheduler will re-connect with them if you do a restart before they finish.  When you do that, are you saying that subsequent outputs (generated after the scheduler starts up again) come through to the UI with bad sorting? Or, do you still see the old (pre-stop) outputs too, and they're out of order?

kinow:
> OK that's essentially the same as `stop --now --now`. It leaves the "orphaned" task jobs running, and the scheduler will re-connect with them if you do a restart before they finish.

Ah, interesting!

>When you do that, are you saying that subsequent outputs (generated after the scheduler starts up again) come through to the UI with bad sorting? Or, do you still see the old (pre-stop) outputs too, and they're out of order?

I think the latter, the old outputs and out of order. I've stopped it that way just now, and have these in the stopped workflow (returned via GraphQL).

image

Somehow 5 is coming before 9. And 9 before 4.

image

Now I've edited the workflow, and replaced file and blob by folder and glob, and then `cylc run --no-detach 530`.

image

No `file_` and no `blob_` :thinking: Stopped again, started without editing the workflow.

image

Now I see the old ones.

It appears there could be something wrong with the UI too, as the state and outputs are not always right. But I'm not sure whether it's coming from the backend like that, or if I have a reactivity problem somewhere. The state I have a guess, but no idea what would cause the outputs to be out of order.

kinow:
p.s. I **think** the old outputs of `file_` and `blob_` are in the DB/store, but will never be satisfied, so they never appear in the UI.

hjoliver:
OK, task outputs are stored in the DB without timing info. I haven't checked, but timing is presumably added by the datastore when the output gets generated.  So if the pre-stop outputs are still in the data at all after restart, they won't have correct timing info (i.e. inot the UIs fault).

hjoliver:
Is the order (in the data) always correct if you don't do any stop/start of the scheduler?

kinow:
> Is the order (in the data) always correct if you don't do any stop/start of the scheduler?

Hmmm, not always! Investigating :tired_face:

hjoliver:
In that case, not the problem of this PR!  We need advice from @dwsutherland

kinow:
Ah, no, sorry, I thought you were talking about the data as in the Vue.js data, that is backing each component. GraphQL appears to return the entries in order, if I create a new workflow like `530-1`, copying just the `flow.cylc`.

In this brand new workflow, I still got outputs out of order. It's always next to the job icon. The job info box shows them in order. But the jobs icon shows 1 or 2 outputs out of order.

And more, if I remove the widget and again again, then the outputs are in order. Or if I add a new widget without removing the previous, then the new one will have the outputs also in order. Refreshing the page also fixes it.

There's definitely something wrong with the UI code… but so far I cannot tell what's wrong. The components have keys, the keys are unique. We are adding new elements to the array in a reactive form (i.e. the UI is rendering the newly added items).

At the moment I'm thinking about the possibility that the browser or Vue.js (or both) are optimizing, and when we add a new output, it could be trying to avoid to modify the rendered components, or move them around, causing the elements out of order :thinking:

kinow:
@hjoliver spent a long time reviewing this code today, looking for possible issues. Tried `computed` data, and a `watcher` modifying an array. Debugging and looking at the objects, I realized we had two lists of custom outputs. Fixed in previous commit.

Also that I was setting key's in chips, but only needed in the tooltip. Harmless, but unnecessary — reading Vue docs, I learned it means Vue will start tracking that component… after this I realized it could be something related to the `key` attribute of the components, but couldn't understand what was wrong.

Went to gym, ate something, looked at the code… and :boom: !

- We start with 0 chips
- We get the first delta with (simplified) `outputs: ['glob_0']`, so our Vue component outputs contain `['glob_0']`
- The UI renders a `<div :key='glob_0'>`
- We get the second delta with (simplified) `outputs: ['glob_1']`, so our Vue component outputs now contain `['glob_1', 'glob_0']` (correct order, same as in GraphQL :+1: )
- The UI renders a `<div :key='glob_0'><div :key='glob_1'>` :-1: 

But Vue is not doing anything wrong. It actually tries to re-use DOM elements in components, whenever possible. But since I defined a specific key, Vue will try as much as possible to avoid touching that DOM element, unless it receives an update to the object with that key.

Since we have an update for `glob_1`, Vue may or may not modify the DOM, leaving the DOM elements where they are if it can. It may work and show everything in the right order, or not. The only thing for sure we have, is that Vue will render each component that we defined, and it will be reactive.

Since we are modifying the list, and the order of elements in the list, we actually **want** Vue to modify the elements in the DOM. So, to force Vue to modify the order, I've changed the code to now use the array index as the key.

That way, the last step of the list above would be

- The UI renders a `<div :key='0'><div :key='1'>` :+1:

And the keys match what we have in the array! The element in the list with index `0` (key) is `glob_1`, and the `1` is `glob_0`. So they are displayed in order!

Of course this is my 7PM diagnostic :smile: I hope @hjoliver can test the latest commit and confirm it indeed fixes the issue! Otherwise I'll need to start from scratch, erasing everything from the drawing board :eyes: 

Ready for review @hjoliver !

kinow:
Squashing the commits locally. Then will test again, and push-force to this branch :+1:

kinow:
Done, from 40 commits, down to the 14 that I could group together. 5 main commits, the rest are addressing feedback (move `+` symbol, show extra messages, not total, etc).

Tested locally, everything worked. Ready again for review. :nerd_face:

kinow:
BTW, tested after syncing all projects to `master`. Used `cylc play --no-detach`, but it wouldn't execute the workflow, no matter what. It would just quickly run/play and stop without any error. I didn't want to spend too much time, so I just copied the `flow.cylc` file elsewhere, nuked the directory, re-created it, and then starting with just `530/flow.cylc` it worked :medal_sports:

hjoliver:
Wow, well done @kinow, that sounds like it was a gnarly one to figure out!  (I presume you've put enough comments in to explain how that works for mere mortals?)

kinow:
> Wow, well done @kinow, that sounds like it was a gnarly one to figure out! 

:raised_hands: 

>(I presume you've put enough comments in to explain how that works for mere mortals?)

_Bruno opens the IDE to write a comment about the `:key` used…_ ops! :sweat_smile:

oliver-sanders:
> Used cylc play --no-detach, but it wouldn't execute the workflow, no matter what.

The `cylc play` command restarts flows by default so if the flow has run to completion it will just shut down straight away.

The simple way to force a flow to run again without re-installing it is to:

```
rm ~/cylc-run/<flow>/.service/db
```

We will make this a little easier with `cylc clean` in the near future, something like:

```
# remove the db
$ cylc clean <flow> --rm .service/db

# move the db to <timestamp>-db
$ cylc clean <flow> --mv .service/db
```

hjoliver:
> The simple way to force a flow to run again without re-installing it is to:

I missed @kinow 's comment on `cylc play`, I was responding to his previous comment, we must have been writing at the same time. Anyhow, I just re-install each time now: `cylc install` -> `cylc play foo/run1` ... `cylc install` -> `cylc play foo/run2` ... etc.

hjoliver (approved):
I'm still pretty hazy on the Vue.js but LGTM, and tests as working really well now. :+1:

oliver-sanders:
Here's my facetious output generating workflow with needlessly long message strings:

```
#!Jinja2

{% from "random" import getrandbits %}
{% from "random" import random %}

{% set outputs = [] %}
{% set num_outputs = 100 %}
{% set max_label_len = 30 %}
{% set max_msg_len = 500 %}

{% for _ in range(0, num_outputs) %}
    {% set label = getrandbits((random() * max_label_len) | int + 1) %}
    {% set msg = getrandbits((random() * max_msg_len) | int + 1) %}
    {% do outputs.append((label, msg)) %}
{% endfor %}

[scheduling]
    [[graph]]
        R1 = foo

[runtime]
    [[foo]]
        script = """
{% for label, message in outputs %}
            # {{ label }}
            cylc message -- {{ message }}
            sleep 1
{% endfor %}
        """
        [[[outputs]]]
{% for label, message in outputs %}
            {{ label }} = {{ message }}
{% endfor %}
```

And here's how it looks in the UI:

outputs2

:+1:

oliver-sanders (approved):

kinow:
Will fix conflicts in a bit. I think you started the workflow with --hold, then used the aotf mutation to start it?

A good way to test the UI, especially with workflows that have just 1 cycle.

PR #532, opened 2020-11-09
Tweak CSS flex/overflow values for GScan workflow names, to avoid word-break, and hidden job statuses

### LINKED ISSUE #529: Items not aligned when workflow name is too long, height changed in component (reflow)
**Describe the bug**

While using a workflow called "messagetrigger1" for #402 , I noticed that the UI layout broke, and the job icon was displayed in another row, and the icons were also not aligned.

image

The height of the component changes, which causes reflow in the browser.

We could control the CSS flow when that happens, to avoid the element from growing, and also use word break I think, to perhaps trim the workflow name at certain length, abbreviating it.

If the height is ignored, we still have the issue that the icons were not aligned with the workflow name, appearing a little to the right of the beginning of the workflow name.

`master`

Create a workflow with a long name, display it in GScan.

**Expected behavior**

Either layout stable, with no changes in height/width of the TreeItem, or an extra row added, but elements still vertically aligned.

**Screenshots**

See above.

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #529 

**Requirements check-list**
- [x] Does not need tests (why? cosmetic change, and testing that two elements have some `x`/are aligned, is quite hard ).

### DIFF
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -36,9 +36,9 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             <v-icon>{{ getWorkflowIcon(workflow.status) }}</v-icon>
           </v-list-item-action>
           <v-list-item-title>
-            <v-layout align-center align-content-center wrap>
-              <v-flex grow>{{ workflow.name }}</v-flex>
-              <v-flex shrink ml-4>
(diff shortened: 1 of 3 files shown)


kinow:
With this change, the GScan with workflows with longer names, will look like:

image

image

hjoliver (approved):
Nice; one review should do (small change, definite improvement!) :+1:

PR #533, opened 2020-11-09
Use skeleton loaders in the Cylc UI landing page

### LINKED ISSUE #531: Display an indicator that the workflows are being loaded

Maybe we could display something while the first reply of the GScan subscription has not been received, or is being processed.

I'm testing #499 , and as I have some 20 workflows (due to functional tests), now the UI is taking longer to load the workflows and display them in GScan. So for ~5 seconds the UI shows no workflows on GScan (also on the Dashboard).

It would be better to have an indicator showing that the request is still being processed.

Users won't understand about GraphQL subscriptions initially. So probably better to show that the request we sent (in GScan that's a request to get all the workflows, which is then subscribed/observed to update whenever it changes) is still being processed.

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #531 

**Requirements check-list**

### DIFF
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -19,72 +19,77 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
   <div
     class="c-gscan"
   >
-    <div
-      v-if="workflows && workflows.length > 0"
-      class="c-gscan-workflows"
+    <v-skeleton-loader
+      :loading="isLoading"
+      type="list-item-three-line"
     >
       <div
-        v-for="workflow in sortedWorkflows"
-        :key="workflow.id"
-        class="c-gscan-workflow"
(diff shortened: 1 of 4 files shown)


kinow:
When loading:

image

When loaded:

image

We could possibly standardise it, and have some sort of HOC (High Order Component), or mixins, that when used would have a `data () { return { isLoading: true } }` and also `v-skeleton-loader`'s.

That can be done in a follow-up PR or after Vue 3 (I believe HOC and mixins are better tailored for Vue2).

hjoliver:
Nice :grin:

kinow:
> Nice grin

I almost had a heart attack once when it took a bit longer to load my workflows, as I was in the middle of some development and thought I had broken GScan :smile:

oliver-sanders (approved):

hjoliver (approved):

PR #538, opened 2020-11-24
Fix e2e coverage

kinow:

Thought it was strange that my last PR's the coverage wasn't increasing as much as I expected. Something happened some commits ago, and looks like the e2e reports are empty. We are still not too bad, with about 70% for unit tests coverage :partying_face: 

Let's hope this PR fixes e2e coverage, and that the reported number is good as.

**Requirements check-list**

### DIFF
--- a/.nycrc
+++ b/.nycrc
@@ -15,7 +15,7 @@
   "extension": [".js", ".vue"],
   "reporter": ["html", "lcov", "text-summary"],
   "cache": false,
-  "all": true,
+  "all": false,
   "check-coverage": false,
   "instrument": true,
   "sourceMap": true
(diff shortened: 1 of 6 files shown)


kinow:
From 69.90% to 81.93%. I was more concerned about `Lumino.vue`. I went to look what was missing tests, since I had just written some tests for another PR and it wouldn't hurt to add a few more tests, when I realized the Codecov reports for Lumino were showing methods that are called when you add/delete a widget as not covered.

Since I was sure I had written tests for those methods, started chasing the issue today. Looks like it's fixed, and ready to review.

One review should do, just use either GitHub actions output, or Codecov reports to review :+1: 

Thanks

oliver-sanders (approved):

PR #545, opened 2020-12-01
Load application services in a plugin

### LINKED ISSUE #476: Replace NODE_ENV use by mode/--mode

I had noted several commands support the `--mode` or `mode=`, etc. But since I was used to `NODE_ENV` from Angular and other projects with Node, I never bothered looking at that `mode` option.

Looks like that's being deprecated in webpack, and the `mode` option is the preferred way.

**Additional context**

Was reading [this article](https://ui.dev/webpack/) about Webpack, when I saw this comment for the [reddit post](https://old.reddit.com/r/javascript/comments/hwka8k/webpack_a_gentle_introduction/) where I found this article.

image

And looks like `NODE_ENV` [doesn't automatically](https://webpack.js.org/configuration/mode/) set the `mode`?

image

Probably better to use what is recommended by Webpack. Just needs some experimenting with the env-var vs. the command line option, re-read the webpack docs later, and then prepare a PR to update our configuration.
**Pull requests welcome!**

kinow:
These changes close #476 

What's included in this PR:

- Removes the use of `NODE_ENV`
- Removes the code that was making the `package.json` available in Vuex - import it with ES6 import instead
- Moves the creation of the services (`WorkflowService` and `UserService`) to a plug-in
- Removes the service import-alias (i.e. `workflow-service`, so that we could import it with `import WorkflowService from 'workflow-service'`) - import the implementation or mocked service based on an option/variable
- Makes `UserService` into a class, so that the `MockWorkflowService` extends it

## Why not NODE_ENV?

The `NODE_ENV` variable is used by Babel and by Webpack. But in our case I think we have webpack setting the variable and passing it down to Babel. Webpack has the concept of modes, with valid values `none`, `production`, and `development`. You can use `NODE_ENV` to set the Webpack mode, but offline is not one of the webpack valid values.

The `vue-cli-service` allows any value for `NODE_ENV`. One solution is to use the `--mode` of the `vue-cli-service` in tandem with `dotenv` (from [Vue docs](https://cli.vuejs.org/guide/webpack.html)). This way you tell `vue-cli-service` to load the variables from `.env.offline` (or any other value) with `vue-cli-service $command --mode offline`.

Inside the dotenv file you define one of the valid webpack modes for `NODE_ENV`. This way we will be using the valid modes for webpack, while specifying a custom offline mode (or staging as in the Vue docs example).

## Why not a boolean like `VUE_APP_OFFLINE=true`?

The values parsed from dotenv files are loaded as strings. So you would need to parse the values again. This is a problem as we need to know whether we are using the offline flag in `vue.config.js`. Unfortunately `vue.config.js` is loaded by another vue utility, before the application is loaded.

This means that `import Vue from 'vue'` is not allowed in `vue.config.js`. We could `const vue = require('vue')` I think, but that wouldn't help, as later it would be imported again from scratch and lose any global variable or service.

See other SO answers and issues related to boolean values in environment variables:

- https://github.com/motdotla/dotenv/issues/51
- https://stackoverflow.com/questions/4870328/read-environment-variables-in-node-js

The simplest approach is to have a string value, and compare in JS. That's why now `.env.offline` contains `VUE_APP_SERVICES=offline`.

For security reasons, only `BASE_URL` and `NODE_ENV` are made available in the application, as well as any value that starts with `VUE_APP`. If you set `VUE_APP_SERVICES` to any other value, it will load the normal services.

**Requirements check-list**


kinow:
Just need to take another look with calm, to confirm everything is working OK, and see if there's anything else that needs updating.

src/plugins/offline.js
```diff
@@ -0,0 +1,51 @@
...
+    Vue.prototype.$offline = ((process.env.VUE_APP_OFFLINE_MODE || 'false').trim().toLowerCase() === 'true')
+    let subscriptionClient = null
+    const graphQLUrls = createGraphQLUrls()
+
+    if (!Vue.prototype.$offline) {
+      subscriptionClient = createSubscriptionClient(graphQLUrls.wsUrl)
+    }
+
+    // Set Vue global workflow service.
+    Vue.prototype.$workflowService = new SubscriptionWorkflowService(graphQLUrls.httpUrl, subscriptionClient)
```
kinow, line 47: Hmmm, probably Offline Mode is not the best name for this service. The most useful change here is setting the default `WorkflowService`... :thinking:

src/store/index.js
```diff
@@ -41,10 +41,6 @@ const state = {
    * Application base URL (set by the backend).
    */
   baseUrl: '/',
-  /**
-   * Application environmnet (e.g. offline, development, production), retrieved from NODE_ENV.
-   */
-  environment: process.env.NODE_ENV.toUpperCase(),
```
kinow, line 47: Less one property. We can access the version directly as ES6 imports allow you to import `package.json` directly.

src/components/cylc/Drawer.vue
```diff
@@ -82,22 +82,26 @@ import Header from '@/components/cylc/Header'
 import { mapState } from 'vuex'
 import GScan from '@/components/cylc/gscan/GScan'
 import { mdiHome, mdiGraphql } from '@mdi/js'
+import { version } from '@/../package.json'
```
kinow, line 85: Easier to just import it directly from `package.json` and use it in the component's `data` section, rather than needing to use Vuex (more expensive than 1 import).

vue.config.js
```diff
@@ -65,17 +55,5 @@ module.exports = {
...
-    const isOffline = process.env.NODE_ENV === 'offline'
-    const workflowService = isOffline
-      ? '@/services/mock/workflow.service.mock'
-      : '@/services/workflow.service'
-    config.resolve.alias.set('workflow-service', workflowService)
-
-    const userService = isOffline
-      ? '@/services/mock/user.service.mock'
-      : '@/services/user.service'
-    config.resolve.alias.set('user-service', userService)
```
kinow, line 79: Hopefully this file is simpler now :+1:

src/plugins/services.js
```diff
@@ -0,0 +1,23 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+import Vue from 'vue'
+import ServicesPlugin from '@/services/plugin'
+
+// Load the plug-in
+Vue.use(ServicesPlugin, {
+  offline: process.env.VUE_APP_SERVICES === 'offline'
+})
```
kinow, line 23: This is similar to the plug-in from #504, and to the other plug-ins in the project (following what was defined in the Tim Creative's theme). The plug-in is defined in `src/services/plugin.js`, and this file just loads it.

The `plugins/index.js` imports each installable plug-in file (such as this one).

src/services/plugin.js
```diff
@@ -0,0 +1,72 @@
...
+   * @param Vue {object} - Vue application
+   * @param options {{
+   *   offline: boolean
+   * }} - options passed to the plug-in (if any)
+   */
+  install (Vue, options) {
+    const offline = options.offline || false
+    this._createWorkflowService(offline)
+    this._installUserService(offline)
+  },
```
kinow, line 42: Now we can import this file, write tests or mock it if necessary,a nd only when this `install` function is called the `Vue.prototype` will be modified.

Before it was harder to control as the values were set in the `main.js`, and via aliases. The only way to know whether you'd get the right service was by loading everything in the right order, and either checking the value of the alias, or the type of object of the service object.

kinow:
This is ready for review, but has very low priority. It would also cause conflicts in other more important PR's (such as the mutation one I think). So we can leave it for later. I can fix any conflicts as necessary, and if we don't have time to review it before the next alpha or beta release, it won't hurt waiting for another release to ship it :+1:

oliver-sanders:
Tried this on Linux/Firefox, `build:watch` worked fine, however, `build` resulted in a broken app. I'll try again on MacOS where I have other browsers and a better idea of what's going on, might not be an issue with this branch.

kinow:
Worked for me, but could be something with node/npm/yarn/etc version.

Screenshot from 2020-12-09 08-47-42

My environment (using yarn, not npm)

```
kinow@ranma:~/Development/python/workspace/cylc-ui$ node -v
v12.20.0
kinow@ranma:~/Development/python/workspace/cylc-ui$ npm -v
6.14.8
kinow@ranma:~/Development/python/workspace/cylc-ui$ yarn -v
1.22.5
```

Fixing the conflicts and rebasing now too. Thanks!

hjoliver:
It works for me.  I have an older node version (v10.23.0), same npm and yarn.

oliver-sanders:
Ah, got it working, you need to kill your uiserver when switching between development (`build:watch`) and production (`build`) builds.

oliver-sanders (approved):
Tested `build`, `build:watch` and `serve` builds using the Vue devtools as the indicator of the environment. All good.

oliver-sanders:
Simple change, worked for Hillary, merging.

kinow:
> Ah, got it working, you need to kill your uiserver when switching between development (`build:watch`) and production (`build`) builds.

Unless you use `--debug` or `-debug` in `jupyterhub` I think (can't recall which one), https://github.com/cylc/cylc-uiserver/issues/144#issuecomment-697095070

That starts Tornado in debug mode, and then it doesn't cache the templates.

PR #554, opened 2020-12-08
Fix icons in the Dashboard

kinow:

Fixing issue reported by @hjoliver on Element, where he would get the icons in the dashboard with a really large width/height. I reproduced the issue, but instead of large icons, my dashboard had no icons at all.

From what I could tell, something might have changed in Vuetify, where the parent element VList or VListItem stopped giving the right dimensions, and instead returned large or invalid values, causing the icon—which inherits the height and widget—to be either very large or hidden.

**Requirements check-list**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -55,7 +55,7 @@ in the tree view structure.
 
 ### Fixes
 
-None.
+[#554](https://github.com/cylc/cylc-ui/pull/554) - Fix icons in the Dashboard.
 
 ### Documentation
 
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -76,7 +76,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         <v-list three-line>
           <v-list-item to="/workflows">
             <v-list-item-avatar size="60" style="font-size: 2em;">
-              <v-icon medium>{{ svgPaths.table }}</v-icon>
+              <v-icon large>{{ svgPaths.table }}</v-icon>
             </v-list-item-avatar>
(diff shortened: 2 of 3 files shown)


kinow:
No change in coverage, but one test added. It is now checking `be.visible` of the first icon in the `.c-dashboard` panel. It uses JQuery to check the visibility.

hjoliver:
Is there any problem with hardwire the size to 24 here, as opposed to the mysteriously-broken "medium"?

kinow:
> Is there any problem with hardwiring the size to 24 here, as opposed to the mysteriously-broken "medium"?

Good question. The default value is 24: https://github.com/vuetifyjs/vuetify/blob/5a39a01de42809157b026cff81ad63e2637661a8/packages/vuetify/src/components/VIcon/VIcon.ts#L20

But they could change it too. The medium size was 28. In my IDE, there's no `medium` directive. Only `x-small`, `small`, `large`, and `x-large`. I think it might be related to this bug: https://github.com/vuetifyjs/vuetify/issues/12168, fixed (not sure if released).

So I think we can either use a fixed size like `24`, `28`, or use the `large` value for now? What do you think @hjoliver ? 

Untitled

(image showing the `large` directive replacing `medium`)

hjoliver:
I vote for large - it looks fine

kinow:
Looks good to me too. Will try on mobile/responsive modet o confirm it is still proportional and then update this PR.

kinow:
Working on large and smaller screens. Edited commit, ready for review again :+1:

oliver-sanders (approved):

PR #571, opened 2021-01-15
Sort other states, remove unnecessary Enum.toLowercase, use Enum for Workflow state filters, and add more tests for GScan

### LINKED ISSUE #553: Add more unit and/or e2e tests for GScan sort

In #537 there is a change to sorting that doesn't seem to trigger any tests. It shows that we probably need a few more tests to verify the order of the workflows is working as expected.

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #553 

There were new states added (`STOPPING`, and `ERROR`). So now the code is using `Enumify`'s `enumOrdinal` to sort these values. They are based on the order the enum values are declared. At the moment the order is:

- RUNNING
- HELD
- STOPPING
- STOPPED
- ERROR

Also remove the now redundant `.toLowerCase` as the enums are already using the lower case string, and uses the `WorkflowState` enum to create the list of filters. We were missing stopping and another state that were included after the filters were initially written for GScan.

The unit tests should cover most of GScan. The only part covered exclusively by e2e tests, are the watchers, as that's easier to test with an e2e, rather than changing values and making sure we wait for the watcher to be triggered in an unit test (in e2e we have the browser running the event loop and handling this watcher updates).

**Requirements check-list**

### DIFF
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -418,6 +418,7 @@ export default {
         this.$workflowService.unsubscribe(
           this.subscriptions[queryName]
         )
+        delete this.subscriptions[queryName]
       }
     },
 
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -263,29 +263,19 @@ export default {
       filters: [
         {
           title: 'workflow state',
-          items: [
-            {
-              text: 'running',
-              value: 'running',
-              model: true
-            },
-            {
-              text: 'held',
-              value: 'held',
-              model: true
-            },
-            {
-              text: 'stopped',
-              value: 'stopped',
+          items: WorkflowState.enumValues.map(state => {
+            return {
+              text: state.name,
+              value: state,
               model: true
             }
-          ]
+          })
         },
         {
           title: 'task state',
           items: TaskState.enumValues.map(state => {
             return {
-              text: state.name.toLowerCase(),
+              text: state.name,
               value: state,
               model: false
             }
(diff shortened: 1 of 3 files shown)


kinow:
Right, I think I have added unit tests for all methods and computed values in `GScan`. Now I just need to add the tests for #553 .

I remembered writing the tests and reading `GScan.vue` code what happened in that issue. The line & branch coverage is OK for `GScan.vue`. But I think there are scenarios not tested for sorting. On Monday will go through the current tests, and see what combinations are missing.

Then add new tests :crossed_fingers: and it should be good to be included in 0.3 :+1:

src/components/cylc/gscan/GScan.vue
```diff
@@ -263,29 +263,19 @@ export default {
...
-              text: 'stopped',
-              value: 'stopped',
+          items: WorkflowState.enumValues.map(state => {
+            return {
+              text: state.name,
+              value: state,
               model: true
             }
-          ]
+          })
```
kinow, line 272: We were not displaying a few values (added afterwards in the server side I think?). Using the enum it should be easier to keep everything up to date. We just need to remember to update the enums.

src/components/cylc/gscan/GScan.vue
```diff
@@ -263,29 +263,19 @@ export default {
...
         },
         {
           title: 'task state',
           items: TaskState.enumValues.map(state => {
             return {
-              text: state.name.toLowerCase(),
```
kinow, line 288: Redundant now (thanks to @oliver-sanders !)

src/components/cylc/gscan/GScan.vue
kinow: And also returning the enum as value. We were doing that in the `TaskState` filter, no reason for just repeating the string here. This way we are more consistent, and it's also easier to write tests, and less likely to make mistakes in the code (as I did a few minutes ago, and then had to spend ~15 minutes chasing what was wrong; I had used `.state.name` by accident in the wrong filter)

kinow:
Done! Whew. I thought it would be just a couple tests, but found more methods that were not (intentionally) being tested. Add small unit tests for these methods.

Then also found that there were states not used in sort/filtering. And finally removed some more redundant code, and also added more tests for filter/sorting :+1:

oliver-sanders (approved):

hjoliver (approved):

Issue #471, opened 2020-07-09
View toolbar


oliver-sanders:
Work out what to do with the "view" controls.

## Context

There is a very important feature which appeared in the early-2019 design outline which quietly disappeared in subsequent design documents.

### *The "view" toolbar circa early 2019-design outline:*

![old-toolbar](https://user-images.githubusercontent.com/16705946/87053776-46adc700-c1fa-11ea-83f5-5f291436c98e.png)

### *The UI as of July 2020:*

<img width="454" alt="Screenshot 2020-07-09 at 15 41 04" src="https://user-images.githubusercontent.com/16705946/87054118-a906c780-c1fa-11ea-965b-943d3c41ac08.png">

The initial idea had been to avoid the clutter of the Cylc7 toolbar by implementing something less obtrusive, however, the wonderful Lumino tabs have since superseded that more primitive layout model. Now we need to put them somewhere else.

## Functionality

The toolbar should provide quick access to regularly used options (e.g. group and ungroup).

The main options expected are:

* Group / Ungroup.
* Filtering (e.g. by task state).
* Visual Filtering (e.g. by reflow).
* Window (e.g. `n=1`).
* Transpose (for graphical layouts).
* Subgraph (i.e. group by cycle point - for graphical layouts).
* Freeze (for all layouts?).
* Search (e.g. by task name).
* Selection sharing (i.e. open the selection of one view in another).

## Ideas

Sadly I think a toolbar is inevitable but we can minimise its visual intrusion on the UI. Thus far this is the best I've come up with:

![toolbar](https://user-images.githubusercontent.com/16705946/87055194-ef105b00-c1fb-11ea-81ee-82fdf28f34e1.png)

For mobile we will have to replace the toolbar with a burger bar, perhaps providing limited functionality.

## Questions

* [ ] Any better ideas?
* [ ] Should all tab-related settings be in this toolbar (e.g. the graph view layout options)?
* [ ] Should we hide the toolbar and have it display when the burger bar is clicked?
* [ ] Should we only display a limited set of icons and expand to the full set when an ellipses icon is clicked?

hjoliver:
I've not had time to think about this in any great depth, but at first glance I like your suggestions above as they are.

hjoliver:
(Something for @kinow post infinite tree :grin: )

oliver-sanders:
The search and filter icons would need to toggle the search and filter inputs.

Perhaps these could appear in the row below?

kinow:
[...]

PR #574, opened 2021-01-19
Add collapse all and expand all buttons to Tree component

### LINKED ISSUE #246: Add Collapse All button to the tree view

kinow:
These changes close #246 

GIFrecord_2021-02-15_091918

**Requirements check-list**

### DIFF
--- a/tests/unit/components/cylc/tree/tree.vue.spec.js
+++ b/tests/unit/components/cylc/tree/tree.vue.spec.js
@@ -25,6 +25,7 @@ import { simpleWorkflowTree4Nodes } from './tree.data'
 import TaskState from '@/model/TaskState.model'
 import TreeItem from '@/components/cylc/tree/TreeItem'
 import CylcObjectPlugin from '@/components/cylc/cylcObject/plugin'
+import cloneDeep from 'lodash.clonedeep'
 
 const localVue = createLocalVue()
 localVue.prototype.$eventBus = {
@@ -85,7 +86,7 @@ describe('Tree component', () => {
     expect(wrapper.props().workflows[0].node.__typename).to.equal('CyclePoint')
     expect(wrapper.find('div')).to.not.equal(null)
   })
-  describe('activable', () => {
+  describe('Activable', () => {
     it('should not activate by default', () => {
       const wrapper = mountFunction({
         propsData: {
@@ -119,8 +120,8 @@ describe('Tree component', () => {
       expect(workflowTreeItemNode.classes('node--active')).to.equal(true)
     })
   })
-  describe('filter tree', () => {
-    describe('filter by name', () => {
+  describe('Filter', () => {
+    describe('Filter by name', () => {
       it('should not filter by name by default', () => {
         const wrapper = mountFunction({
           propsData: {
@@ -146,7 +147,7 @@ describe('Tree component', () => {
         expect(wrapper.vm.activeFilters).to.equal(null)
       })
     })
-    describe('filter by state', () => {
+    describe('Filter by state', () => {
       it('should not filter by state by default', () => {
         const wrapper = mountFunction({
           propsData: {
@@ -178,7 +179,7 @@ describe('Tree component', () => {
         expect(wrapper.vm.activeFilters.states).to.equal(states)
       })
     })
-    describe('enable filters', () => {
+    describe('Enable filters', () => {
       it('should not have any filters enabled by default', () => {
         const wrapper = mountFunction({
           propsData: {
@@ -224,7 +225,7 @@ describe('Tree component', () => {
         expect(wrapper.vm.activeFilters.states).to.deep.equal(states)
       })
     })
-    describe('filter tree items', () => {
+    describe('Filter tree items', () => {
       it('should not filter the tree items by default', () => {
         const wrapper = mountFunction({
           propsData: {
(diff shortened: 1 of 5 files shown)


kinow:
First version:

GIFrecord_2021-01-19_153724

Today I just did the template after reading the design issue, tomorrow will do more testing with a real workflow.

@oliver-sanders, @hjoliver I couldn't find where to put the collapse/expand all buttons in the design issue. Any suggestions?

Thanks!

hjoliver:
@kinow I think we'd want expand-all to stop at the task level.  i.e. expand all task families, but don't expand the jobs (the icons show how many jobs there are anyway) and definitely don't expand the job details leaves.

kinow:
> @kinow I think we'd want expand-all to stop at the task level. i.e. expand all task families, but don't expand the jobs (the icons show how many jobs there are anyway) and definitely don't expand the job details leaves.

Done! The function already supported a filter function :-) so just passed a filter that looks for the types we don't want to expand (`job` and `job-details`).

hjoliver:
The button positions seem good to me.

Noticed two issues though, from a quick play:
- expand-all is still expanding the task nodes so that each job is listed on a  new line.  IMO the expansion should probably stop at tasks. That's enough to see every task, and the job icons, without using up a lot more screen lines than necessary (thinking of large flows).
- when a new cycle point is added it appears in the expanded, even if collapse-all was pressed

kinow:
>expand-all is still expanding the task nodes so that each job is listed on a new line. IMO the expansion should probably stop at tasks. That's enough to see every task, and the job icons, without using up a lot more screen lines than necessary (thinking of large flows).

Ah, sorry, I misunderstood it. That'll be easy to fix.

>shrinking the window width, the minus drops under the plus, then they both disappear off the end

:+1: will make it a flex element, and adjust how it behaves (wrap, justify, etc). I'm more concerned about the location, whether the icon is OK (I used the first that came to my mind).

And the feature could be a bit different. We could make it stateful. So if you click expand all, then every subsequent delta will be expanded too. Meaning that the UI keeps the "expand all" state, until you click the expand all icon again. That can be tricky to implement/test, but new deltas would be expanded too.

oliver-sanders:
> @oliver-sanders, @hjoliver I couldn't find where to put the collapse/expand all buttons in the design issue. Any suggestions?

There is an issue for this, although the answer is pretty much "where you've put them". https://github.com/cylc/cylc-ui/issues/471

kinow:
> > @oliver-sanders, @hjoliver I couldn't find where to put the collapse/expand all buttons in the design issue. Any suggestions?
> 
> There is an issue for this, although the answer is pretty much "where you've put them". #471

Thanks @oliver-sanders !

tests/unit/components/cylc/tree/tree.vue.spec.js
```diff
@@ -23,8 +23,27 @@ import Vuetify from 'vuetify/lib'
...
+  },
+  subscribe: function () {
+  },
+  mutations: [
+    {
+      args: []
+    }
+  ]
+}
+localVue.use(CylcObjectPlugin)
```
kinow, line 46: The code above removes several warnings from the console about missing the directive for `cylc-object`.

tests/unit/components/cylc/tree/tree.vue.spec.js
```diff
@@ -39,7 +58,7 @@ describe('Tree component', () => {
   // mount function from Vuetify docs https://vuetifyjs.com/ru/getting-started/unit-testing/
   /**
    * @param options
-   * @returns {Tree}
+   * @returns {Wrapper<Tree>}
```
kinow, line 61: The wrapper has different attributes and functions, so the proper annotation gives the IDE better auto-complete (e.g. `findAllComponents` and displays its signature)

tests/unit/components/cylc/tree/tree.vue.spec.js
```diff
@@ -258,18 +276,15 @@ describe('Tree component', () => {
...
+        const taskProxy = treeItems.at(1)
         // task proxy is displayed
-        expect(taskProxy.$children[0].filtered).to.equal(false)
+        expect(taskProxy.vm.filtered).to.equal(false)
         wrapper.vm.removeAllFilters()
-        // cycle point is displayed (list.filter(() => return filtered=true means OK, filtered=false means remove me)
-        expect(taskProxy.filtered).to.equal(true)
         // task proxy is displayed
-        expect(taskProxy.$children[0].filtered).to.equal(true)
+        expect(taskProxy.vm.filtered).to.equal(true)
```
kinow, line 287: These tests were broken!!!

I thought `children` would return the children components of `Tree`. And it does, but I thought the children were `TreeItem`. Turns out it brings several components, like the `TreeItem`s, but also all the Vuetify components (like `VTextField` and whatnot).

So fixed it by using the `Wrapper` and `WrapperArray` objects functions (and with some debugging :grimacing: ).

They got broken in this PR as I added new children components for the plus and minus icons.

kinow:
Still pending

- [ ] prevent the plus and minus icons from wrapping
- [ ] move the expand/collapse out of the filters area (i.e. out of element with `v-if='filterable')

kinow:
>expand-all is still expanding the task nodes so that each job is listed on a new line. IMO the expansion should probably stop at tasks. That's enough to see every task, and the job icons, without using up a lot more screen lines than necessary (thinking of large flows).

Last commit expands workflows, cycle points, and family proxies. Tasks, jobs, and job details won't be expanded.

kinow:
@hjoliver I've fixed the plus and minus icons wrapping. However, I noticed that the filter elements (task search name, task statuses) are responsive. That means that when you are using a small viewport, they adjust to be displayed in a different way.

But when you have a small widget area, they are not responsive. Here's what I mean:

image

The filter elements are reducing the width, but I think they shouldn't. And there is no way to tell the filter or expand/collapse toggle that they need to be displayed in a different way. In this case, I think making a fixed width would fix it.

But, in the end, I believe this will be fixed when we adjust the filter to match the design toolbar @oliver-sanders mentioned earlier #471 

Once we adopt that design, this issue will be gone. Note too, that this happens before this PR, that's why I stopped messing with layout/responsive/flexbox/etc. Otherwise this change would grow more, making it harder to review/revert/conflicts/etc.

WDYT?

hjoliver:
> WDYT?  

Seems fine to me, as we have a plan that will fix it.

kinow:
Started writing the tests for expand/collapse methods. Even though that's not part of this PR, as it had been implemented since the initial version of the tree component, only now we are using that code. Should be done by tomorrow, then this one will be ready for review.

kinow:
Added unit test and changelog. Had some issues writing the tests because I used an array of test parameters (wish there was something like parametrized tests in mocha+chai), and was calling `.expandAll` and `collapseAll`, and getting confusing results.

Turns out these functions mutate the objects I was passing to the tests, so the first test mutated the data, the second mutated again, etc. Until one failed. :disappointed_relieved:  Fixed with Lodash's mutate (because `Object.assign` is a shallow-copy).

oliver-sanders (approved):
Looks good, tested as working.

One quirk, the tooltip seems to appear in-line with the Lumino/Workflow component rather than the expand/collapse button.

Screenshot 2021-02-11 at 08 37 43

kinow:
>Looks good, tested as working.

Thanks!

>One quirk, the tooltip seems to appear in-line with the Lumino/Workflow component rather than the expand/collapse button.

Ah! Tested locally, and you are correct. The tooltip appears over the Add View link, giving the wrong impression that they are related somehow. I've added the `bottom` directive to the tooltip, and now it should always be under the "Collapse All", "Expand All" links (which should work well for responsive and desktop viewports)

image

kinow:
Also rebased :+1:

kinow:
The CI failure is related to #590 not related to this PR

hjoliver (approved):
Works for me, nice :+1:

PR #575, opened 2021-01-19
Remove queued task state.

### LINKED ISSUE #573: Remove Queued status

Sibling of https://github.com/cylc/cylc-flow/pull/4035

**Additional context**

**Pull requests welcome!**

hjoliver:
Close #573 

**requires https://github.com/cylc/cylc-flow/pull/4035 on the back end**

A quick hack at it :grimacing: - changes will be required; probably broke the tests too... but it seems to work, no console errors. 

I've temporarily copied and tweaked the task held indicator to show `isQueued` with a red dot.  Same flow as demoed in https://github.com/cylc/cylc-flow/pull/4035 : 

Screenshot from 2021-01-19 17-58-57

**Requirements check-list**

### DIFF
--- a/src/components/cylc/Task.vue
+++ b/src/components/cylc/Task.vue
@@ -387,8 +387,7 @@ export default {
     const taskIconSvgCssClasses = ['task']
     if (context.props.isHeld) {
       taskIconSvgCssClasses.push('held')
-    }
-    if (context.props.isQueued) {
+    } else if (context.props.isQueued) {
       taskIconSvgCssClasses.push('queued')
     }
     if (['waiting', 'preparing', 'expired', 'submitted', 'running', 'succeeded', 'failed', 'submit-failed'].includes(context.props.status)) {
--- a/src/components/cylc/Task.vue
+++ b/src/components/cylc/Task.vue
@@ -166,7 +166,7 @@ export default {
       {
         attrs: {
           id: 'held',
-          transform: 'scale(0.45)'
+          transform: 'scale(0.55)'
         }
       },
       [
--- a/src/components/cylc/Task.vue
+++ b/src/components/cylc/Task.vue
@@ -221,13 +221,15 @@ export default {
       },
       [
         createElement(
-          'circle',
+          'rect',
           {
             attrs: {
-              cx: '50',
-              cy: '50',
-              r: '46',
-              'stroke-width': '8'
+              x: 0,
+              y: 0,
+              width: '90',
+              height: '18',
+              rx: '10',
+              ry: '10'
             }
           }
         ),
@@ -235,12 +237,12 @@ export default {
           'rect',
           {
             attrs: {
-              x: 32.5,
-              y: 25,
-              width: '10',
-              height: '50',
-              rx: '5',
-              ry: '5'
+              x: 0,
+              y: 33,
+              width: '90',
+              height: '18',
+              rx: '10',
+              ry: '10'
             }
           }
         ),
@@ -248,12 +250,12 @@ export default {
           'rect',
           {
             attrs: {
-              x: 52.5,
-              y: 25,
-              width: '10',
-              height: '50',
-              rx: '5',
-              ry: '5'
+              x: 0,
+              y: 66,
+              width: '90',
+              height: '18',
+              rx: '10',
+              ry: '10'
             }
           }
         )
--- a/src/components/cylc/Task.vue
+++ b/src/components/cylc/Task.vue
@@ -32,6 +32,10 @@ export default {
       type: Boolean,
       require: true
     },
+    isQueued: {
+      type: Boolean,
+      require: true
+    },
     progress: {
       type: Number,
       required: false,
@@ -206,6 +210,57 @@ export default {
       ]
     )
     taskIconSvgChildren.push(isHeld)
+    // isQueued - the isQueued portion of the task state.
+    const isQueued = createElement(
+      'g',
+      {
+        attrs: {
+          id: 'queued',
+          transform: 'scale(0.45)'
+        }
+      },
+      [
+        createElement(
+          'circle',
+          {
+            attrs: {
+              cx: '50',
+              cy: '50',
+              r: '46',
+              'stroke-width': '8'
+            }
+          }
+        ),
+        createElement(
+          'rect',
+          {
+            attrs: {
+              x: 32.5,
+              y: 25,
+              width: '10',
+              height: '50',
+              rx: '5',
+              ry: '5'
+            }
+          }
+        ),
+        createElement(
+          'rect',
+          {
+            attrs: {
+              x: 52.5,
+              y: 25,
+              width: '10',
+              height: '50',
+              rx: '5',
+              ry: '5'
+            }
+          }
+        )
+      ]
+    )
+    taskIconSvgChildren.push(isQueued)
+
     // Warning icon.
     const warningIcon = createElement(
       'g',
@@ -330,6 +385,9 @@ export default {
     const taskIconSvgCssClasses = ['task']
     if (context.props.isHeld) {
       taskIconSvgCssClasses.push('held')
+    }
+    if (context.props.isQueued) {
+      taskIconSvgCssClasses.push('queued')
     }
     if (['waiting', 'preparing', 'expired', 'submitted', 'running', 'succeeded', 'failed', 'submit-failed'].includes(context.props.status)) {
       taskIconSvgCssClasses.push(context.props.status)
--- a/src/components/cylc/Task.vue
+++ b/src/components/cylc/Task.vue
@@ -185,12 +185,12 @@ export default {
           'rect',
           {
             attrs: {
-              x: 32.5,
-              y: 25,
-              width: '10',
-              height: '50',
-              rx: '5',
-              ry: '5'
+              x: 30,
+              y: 20,
+              width: '15',
+              height: '60',
+              rx: '10',
+              ry: '10'
             }
           }
         ),
@@ -198,12 +198,12 @@ export default {
           'rect',
           {
             attrs: {
-              x: 52.5,
-              y: 25,
-              width: '10',
-              height: '50',
-              rx: '5',
-              ry: '5'
+              x: 55,
+              y: 20,
+              width: '15',
+              height: '60',
+              rx: '10',
+              ry: '10'
             }
           }
         )
--- a/src/components/cylc/Task.vue
+++ b/src/components/cylc/Task.vue
@@ -390,7 +390,7 @@ export default {
     } else if (context.props.isQueued) {
       taskIconSvgCssClasses.push('queued')
     }
-    if (['waiting', 'preparing', 'expired', 'submitted', 'running', 'succeeded', 'failed', 'submit-failed'].includes(context.props.status)) {
+    if (['waiting', 'preparing', 'expired', 'submitted', 'running', 'succeeded', 'failed', 'submit-failed', 'queued'].includes(context.props.status)) {
       taskIconSvgCssClasses.push(context.props.status)
     } else {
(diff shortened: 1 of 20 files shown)


kinow:
That was quick! Thanks!

src/model/TaskState.model.js
```diff
@@ -29,9 +29,7 @@ class TaskState extends Enumify {
   static RUNNING = new TaskState('running')
   static SUBMITTED = new TaskState('submitted')
   static PREPARING = new TaskState('preparing')
-  static QUEUED = new TaskState('queued')
   static WAITING = new TaskState('waiting')
-  static HELD = new TaskState('held')
```
kinow, line 34: Ah, so the held state is gone too? I think some tests will need to be re-written or removed (the ones that are using `=== 'held'`).

tests/unit/components/cylc/task.vue.spec.js
```diff
@@ -29,6 +29,7 @@ describe('Task component', () => {
         props: {
           status: '',
           isHeld: true
+          isQueued: true
```
kinow, line 32: Missing comma above, after `isHeld: true`, is causing the linter to fail in CI

kinow (changes requested):

src/model/TaskState.model.js
hjoliver: Yes, there's now a "held" indicator that gets attached to the actual task state.
As a temporary measure I replaced the value in the tests with the string `'held'`.

kinow (approved):

kinow:
@hjoliver I've fixed the linter issue, and the tests. Also waited for CI to finish to confirm it's building OK (had tested locally), and that there were tests already for this code. It's probably blocked by the issue in Cylc Flow, but approving as it looks good.

Great work updating the guide, the styleguide, everything! I would have missed some parts you updated in this PR :)

hjoliver:
Thanks for fixing the tests @kinow :+1:

kinow:
@hjoliver , the macos build job appears to be a bit flaky. I kicked it in a couple other PR's, and it passed after that :+1:

hjoliver:
Latest commit makes the task icon queued indicator conform to the design sketch. (with gap between "burger" lines slightly bigger for visibility at smaller sizes).

shot

kinow (approved):
Looks good to me. Not sure if the `overflow: visible` shouldn't be fixed by changing the width/height, but we can think about it later.

Did a quick test manually changing the data to show `isHeld` and `isQueued` enabled in the UI (not sure if you can have a task queued/held at the same time :laughing: but just wanted to look at the rendered icons).

image

hjoliver:
Haha, yes I was just looking at that.  The design doc shows queued and held icons in the same location (see my screenshot above). Currently "held" overlaps the task icon.  My latest commit tries moving it to the right to provide some separation:

shot

In Cylc 7 a task couldn't be both queued and held at the same time, because those were mutually exclusive task states. But now that is possible.  A held task in the queue should not be released to run, but other non-held tasks can jump over it. (I'd better check that on the back end).

@oliver-sanders - what do you think about icon locations, in light of this?

hjoliver:
(We'll need to fit other icons in too, of course ... xtriggers)

src/components/cylc/workflow/Toolbar.vue
```diff
@@ -152,7 +153,7 @@ export default {
     ...mapState('app', ['title']),
     ...mapGetters('workflows', ['currentWorkflow']),
     isHeld: function () {
-      return this.currentWorkflow.status === 'held'
+      return this.currentWorkflow.status === WorkflowState.HELD.name
```
kinow, line 156: :+1:

src/components/cylc/Task.vue
```diff
@@ -331,6 +388,9 @@ export default {
     if (context.props.isHeld) {
       taskIconSvgCssClasses.push('held')
     }
+    if (context.props.isQueued) {
```
oliver-sanders, line 391: We can only display one indicator at a time (so held or queued).

I think we agreed that held takes precedence? So this should be

```suggestion
    else if (context.props.isQueued) {
```

hjoliver: > We can only display one indicator at a time (so held or queued).

We can display more than one if we want to, obviously (e.g. my screenshot) ... but maybe we don't want to...

>  I think we agreed that held takes precedence? 

Yeah, I had forgotten that.  I guess if a task is held it doesn't matter if it's queued or not (queuing is essentially an automated means of holding a task, after all).  So OK, I'll tweak the code to do this tomorrow (have to reposition the held indicator as well, back to the left side). :+1:

hjoliver:
Now featuring mutually exclusive held and queued indicators (held takes preference), and a bigger held indicator that looks more like the design sketch and is actually visible a smaller sizes.

Screenshot from 2021-01-28 14-30-09

(The unknown state icon `?`, which should be "waiting" is being addressed elsewhere)

hjoliver:
@kinow could you take a look at the failed unit test here, please? I'm struggling to figure it out.

kinow:
That's strange.

```
kinow@ranma:~/Development/python/workspace/cylc-ui$ git status
On branch pr-575
nothing to commit, working tree clean
kinow@ranma:~/Development/python/workspace/cylc-ui$ git log
commit 3d3b40db943425aad112b0d89b0ed14e92c13dbd (HEAD -> pr-575)
Author: Hilary James Oliver <hilary.j.oliver@gmail.com>
Date:   Thu Jan 28 14:28:43 2021 +1300

    Task indicators: bigger held, move queued left
```
I think I am on the right branch (fetched from your PR). The error in the GH action says

```
should filter by workflow tasks state:

      AssertionError: expected 1 to equal 2
```

But what I see in the unit test locally, is `expect(wrapper.vm.filteredWorkflows.length).to.equal(0)`. Either I am seeing an old copy of the test, or GH has a cached version somewhere?

kinow:
Added your clone as remote, and checked out `remove-queued-state`. Both lint and unit tests are passing for me locally :confused:

hjoliver:
Huh. Yeah, that's what's confusing me.  `expect(1)` is before that test. I'll trying kicking CI then.

kinow:
Ah, reproduced the issue after doing a `git rebase upstream/master` :confused: I wonder if our GH action does the same before running our tests?

```
kinow@ranma:~/Development/python/workspace/cylc-ui$ git log -n 1
commit 3d3b40db943425aad112b0d89b0ed14e92c13dbd (HEAD -> remove-queued-state, hilary/remove-queued-state)
Author: Hilary James Oliver <hilary.j.oliver@gmail.com>
Date:   Thu Jan 28 14:28:43 2021 +1300

    Task indicators: bigger held, move queued left
kinow@ranma:~/Development/python/workspace/cylc-ui$ git rebase upstream/master 
First, rewinding head to replay your work on top of it...
Applying: Remove queued task state.
Applying: Add missing comma, fixing linter error
Applying: Use WorkflowState for toolbars
Applying: Implement task icon queued indicator.
Applying: tweak
Applying: Move task held icon.
Applying: Update src/components/cylc/Task.vue
Applying: Task indicators: bigger held, move queued left
kinow@ranma:~/Development/python/workspace/cylc-ui$ git log -n 1
commit 1fb691bdcbfd7ca4bb8b7b76558b354fc157bf82 (HEAD -> remove-queued-state)
Author: Hilary James Oliver <hilary.j.oliver@gmail.com>
Date:   Thu Jan 28 14:28:43 2021 +1300

    Task indicators: bigger held, move queued left
```

hjoliver:
> Ah, reproduced the issue after doing a git rebase upstream/master confused I wonder if our GH action does the same before running our tests?

Wow, surely not?!

hjoliver:
Still, I'll rebase this.

kinow:
> > Ah, reproduced the issue after doing a git rebase upstream/master confused I wonder if our GH action does the same before running our tests?
> 
> Wow, surely not?!

I think it does. Looking at our raw logs:

```
2021-01-28T02:41:02.1437851Z From https://github.com/cylc/cylc-ui
2021-01-28T02:41:02.1439101Z  * [new ref]         a775bc360544c79e203fedc0946b2851d9b0619a -> pull/575/merge
2021-01-28T02:41:02.1440495Z ##[endgroup]
2021-01-28T02:41:02.1441396Z ##[group]Determining the checkout info
2021-01-28T02:41:02.1442114Z ##[endgroup]
2021-01-28T02:41:02.1442766Z ##[group]Checking out the ref
2021-01-28T02:41:02.1444079Z [command]/usr/bin/git checkout --progress --force refs/remotes/pull/575/merge
2021-01-28T02:41:02.1450292Z Note: switching to 'refs/remotes/pull/575/merge'.
```

And probably done in this issue, with some explanation why that's done I think: https://github.com/actions/checkout/issues/15

hjoliver:
I think you're right. The line number referred to in the gscan unit tests file did not actually exist before I rebased!

hjoliver:
There is a `TaskState.HELD` in the rebased `tests/unit/components/cylc/gscan/gscan.vue.spec.js`, which doesn't (shouldn't) exist anymore.

kinow:
> There is a `TaskState.HELD` in the rebased `tests/unit/components/cylc/gscan/gscan.vue.spec.js`, which doesn't (shouldn't) exist anymore.

Let me see if I can fix that test. I wrote it recently, and I just spotted a part where I should have used the `TaskState` enum, but ended up using `WorkflowState` :facepalm:

hjoliver:
Yeah just noticed that myself - in the `const workflows = ...` assignment?

kinow:
> Yeah just noticed that myself - in the `const workflows = ...` assignment?

Ha! Yes, you found it! :running_man:

hjoliver:
That's got it :tada:

kinow (approved):
Tested again with https://github.com/cylc/cylc-flow/pull/4035

Working as expected!

image

kinow (approved):
Tested again just now, with the sibling cylc-flow PR https://github.com/cylc/cylc-flow/pull/4088

Worked with no issues, re-approving :+1:

oliver-sanders (commented):
All good on the queue functionality, some thinking to do on the SVG.

This PR moves the held/queued parts of the icon outside of the SVG view box, I'm surprised it gets rendered, I would have expected it to be trimmed:

Screenshot 2021-03-02 at 15 12 58

The position of the queued/held icon changes relative to the task icon depending on zoom level (not sure if that was the case before).

hjoliver:
In the CSS, I used:
```css
     overflow: visible;
```

hjoliver:
Struggling to recall now, but I might have moved the badges outward a bit when I was initially trying to accommodate both the held and queued ones at the same time.  Anyhow, I'll try moving it back ...

oliver-sanders:
We can always boost the size of the task icons by setting the font size (they are set to width=height=1em). Thinking that that might be a sensible configuration for accessibility reasons anyhoo.

oliver-sanders:
> This PR moves the held/queued parts of the icon outside of the SVG view box
>> In the CSS, I used:
>>>     overflow: visible;

Derp, no it doesn't, I misread `svg.task` as an element within the SVG :facepalm:.

I thought the icon had exceeded the SVG viewBox which would have resulted in cut-off (which is hasn't) because that would make it invisible (which is isn't).

This is fine, the overflow actually handles the baseline issue nicely. Need to be careful with vertical spacing, that should be easy enough.

```diff
--- a/src/styles/cylc/_task.scss
+++ b/src/styles/cylc/_task.scss
@@ -115,6 +115,7 @@
     /* scale the icon to the font-size */
     width: 1em;
     height: 1em;
+    border: 1px solid red;
     overflow: visible;
     vertical-align: middle;
```

Screenshot from 2021-03-09 15-40-59

PR #577, opened 2021-01-19
Responsive GScan, handle workflows with long names

### LINKED ISSUE #576: gscan: responsive rows
Line wrapping issue in GScan spotted in a screenshot from @hjoliver in https://github.com/cylc/cylc-ui/pull/575.

Screenshot 2021-01-19 at 10 14 44

When the GScan column is not wide enough I think the best resolution is to:

* Let the workflow name `overflow: hidden`.
* Move the status icons to the start of the next row.

I.E:

This has the unfortunate effect of cutting off the full name of the workflow, however, it may not be possible to display a long name in a narrow column. We could display the full name in a tooltip and potentially consider smart abbreviations one day.

**Pull requests welcome!**

kinow:
These changes close #576 

What it looks like after this change:

GIFrecord_2021-01-21_092110

See linked issue for description and screenshots of the before.

**Requirements check-list**

### DIFF
--- a/src/styles/cylc/_tree.scss
+++ b/src/styles/cylc/_tree.scss
@@ -33,12 +33,16 @@ $active-color: #BDD5F7;
 }
 
 .treeitem {
-  display: table;
+  display: block;
   width: 100%;
   .node {
     line-height: 1.8em;
-    display: flex;
-    flex-wrap: nowrap;
+    display: block;
+
+    .node-expand-collapse-button {
+      float: left;
+      margin-right: 4px;
+    }
 
     &--hoverable {
       @include states()
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,9 @@ None or N/A.
 
 ### Enhancements
 
+[#577](https://github.com/cylc/cylc-ui/pull/577) - Improve how GScan behaves
+in responsive mode (smaller viewports).
+
 [#560](https://github.com/cylc/cylc-ui/pull/560) - Make job-details a node
 in the tree, expand/collapse control, align tree-leaf-triangle. Also align
 job icons vertically in the middle of the HTML element.
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -116,13 +116,28 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               :to="workflowLink(scope.node)"
               :class="getWorkflowClass(scope.node.node)"
             >
-              <v-list-item-action v-if="scope.node.type === 'workflow'">
-                <workflow-icon :status="scope.node.node.status" />
-              </v-list-item-action>
               <v-list-item-title>
-                <v-layout align-center align-content-center wrap>
-                  <v-flex grow>{{ scope.node.node.name }}</v-flex>
-                  <v-flex shrink ml-4 v-if="scope.node.type === 'workflow'">
+                <v-layout align-center align-content-center d-flex flex-wrap>
+                  <v-flex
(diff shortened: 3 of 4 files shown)


src/services/mock/checkpoint.js
```diff
@@ -3,7 +3,7 @@ const checkpoint = {
     "workflows": [
         {
             "id": "user|one",
-            "name": "one",
+            "name": "one-two-three-four-five-fix-seven-eight-nine-ten-eleven-twelve-thirteen",
```
kinow, line 6: Will undo this change later before asking for the final review, it just makes it easier for dev/testing for now :+1:

src/styles/cylc/_tree.scss
```diff
@@ -33,12 +33,16 @@ $active-color: #BDD5F7;
...
 
 .treeitem {
-  display: table;
+  display: block;
   width: 100%;
   .node {
     line-height: 1.8em;
-    display: flex;
-    flex-wrap: nowrap;
+    display: block;
```
kinow, line 40: Making these two `block`, limit the width of the treeitem & its node.

This way, the children elements (like v-list-item's) can be flex, and will have the total width limited by the parent's width.

Using that, we then set `flex-wrap` to `nowrap`. And use the styles from the [Flexbox and Truncated Text article from css-tricks](https://css-tricks.com/flexbox-truncated-text/) to truncate the text.

kinow:
I had some couple hours this morning while other PR's were being reviewed, and before our Auckland monthly morning tea, so had a go at this one. Note that it's not exactly the solution proposed by @oliver-sanders .

I started with the proposed solution, then thought it was awkward to have the statuses icons all aligned to the right, except the ones for the workflows with long names :disappointed_relieved: so I thought it would be simpler to scan for workflows this way:

GIFrecord_2021-01-20_124626

I also added the tooltip now. I imagined that users could want to filter the workflows by a part of the text in their name. In which case, if you wanted to see the name, having to click on the workflow first to see what is in the end of the name would be annoying… in the GIF I'm filtering by "thirteen", which appears in the tooltip.

What do you think @oliver-sanders , @hjoliver ? If that looks OK I can update the tests. Or if you prefer to go with the initial solution I can update the PR later too.

hjoliver:
That looks good, and it certainly looks better than on current master, but the downside is you see even less of the long workflow name.  I take your point about icon alignment consistency between long and short name flows though.  Is it possible to have the icons split onto the next line and aligned to the right side instead of the left?

kinow:
Good idea @hjoliver . Let me try and see what it looks like.

kinow:
Next commit will have the icons in the next line (similar to original solution from the issue):

image

kinow:
And now the next commit implements @hjoliver 's suggestion above:

image

We can compare and choose the one that works best now I think :+1:

hjoliver:
Thanks @kinow - I like the last one, let's see if @oliver-sanders agrees...

oliver-sanders:
Much nicer.

kinow:
Squashed commits to use the last approach. Will see if we need tests for this, add changelog, and also add a new screencast showing what it looks like. Then it should be ready for review.

src/styles/cylc/_tree.scss
oliver-sanders: So is it this bit that turns the ellipsis on?

```diff
-<v-layout align-center align-content-center wrap>
+<v-layout align-center align-content-center d-flex flex-wrap>
```

oliver-sanders (approved):

hjoliver (approved):
:tada:

PR #579, opened 2021-01-21
Sort tree nodes as they are added, instead of doing it in the template recursively

### LINKED ISSUE #578: Incorrect cycle point sorting in Tree View
**Describe the bug**

From Element, reported by @hjoliver . Looks like our GraphQL query in the UI, or the JS code, either of these two need updating to sort the cycle points correctly.

**Expected behavior**

Cycle points sorted in natural order, from the most recent to the oldest (if I remember well, that's how it's supposed to be displayed)

**Screenshots**

Screenshot from 2021-01-20 17-34-58

Screenshot from 2021-01-20 17-33-03
**Additional context**

**Pull requests welcome!**

kinow:
These changes close #578

**Requirements check-list**


src/components/cylc/tree/Tree.vue
```diff
@@ -108,13 +108,9 @@ import Vue from 'vue'
 import TaskState from '@/model/TaskState.model'
 import Task from '@/components/cylc/Task'
 import clonedeep from 'lodash.clonedeep'
-import { treeitem } from '@/mixins/treeitem'
 
 export default {
   name: 'Tree',
-  mixins: [
-    treeitem
-  ],
```
kinow, line 117: It was declared for the component, but not used.

src/components/cylc/tree/cylc-tree.js
```diff
@@ -16,7 +16,7 @@
  */
 import { computePercentProgress } from '@/components/cylc'
 import { extractGroupState } from '@/utils/tasks'
-import { merge } from 'lodash'
+import { merge, sortedIndex } from 'lodash'
```
kinow, line 19: A utility function from Lodash that does binary-search to locate the next index to add an element to a sorted array. This is used for every `add....` function of the tree.

The cycle points are the only ones sorted in reverse order (the newest first), so we adjust the index for that case.

src/components/cylc/tree/cylc-tree.js
```diff
@@ -230,7 +230,15 @@ class CylcTree {
...
-      this.root.children.push(cyclePoint)
+      const parent = this.root
+      const names = this.root.children
+        .map(child => child.node.name)
+        .reverse()
+      const insertIndex = sortedIndex(names, cyclePoint.node.name)
+      // cycle points are inserted in the reverse order, so we have to handle that sortedIndex will give you the
+      // wrong index (except when the list is empty). That's why we reverse the list first, to get the index, and
+      // find where it would be in the not-reversed list.
+      parent.children.splice(parent.children.length - insertIndex, 0, cyclePoint)
```
kinow, line 241: In the case where the index is `0`, it won't be negative. Shouldn't matter that we used `.splice(0, 0, elem)` instead of simply `.push(elem)` I believe.

src/components/cylc/tree/cylc-tree.js
```diff
@@ -324,7 +332,9 @@ class CylcTree {
       // the parent-child could end up repeated by accident; it means we must make sure to create this relationship
       // exactly once.
       if (parent.children.length === 0 || !parent.children.find(child => child.id === familyProxy.id)) {
-        parent.children.push(familyProxy)
+        const names = parent.children.map(child => child.node.name)
+        const insertIndex = sortedIndex(names, familyProxy.node.name)
+        parent.children.splice(insertIndex, 0, familyProxy)
```
kinow, line 337: For the other elements, we know we can just add in the next index (i.e. the index where the elements would still be in order).

src/components/cylc/tree/cylc-tree.js
```diff
@@ -86,6 +86,25 @@ function computeCyclePointsStates (cyclePointNodes) {
...
+    return 1
+  }
+  // name
+  return left.node.name.toLowerCase()
+    .localeCompare(
+      right.node.name.toLowerCase(),
+      undefined,
+      { numeric: true, sensitivity: 'base' })
+}
+
```
kinow, line 107: We cannot use the binary-search function, as it will simply search by one value. Tasks and Family Proxies are grouped together, so the location of the next insertion depends on type and value.

kinow:
Not sure if the best solution, and it was written on Friday afternoon, so there could be some bugs or better ways to fix it :)

hjoliver (commented):
(looks good, but I still need to test it)

kinow:
> (looks good, but I still need to test it)

I didn't compare the performance, but I think this should improce time to update the tree.

Currently we have a function in the template called for each tree item recursively, to sort its children.

This PR moves the logic tp a function before the template is rendered. Which should perform better, but also helps writing tests. We can unit test to assert a data structure is ordered, instead of having to write more complex tests that rely on the template being rendered virtually ori n a browser (e2e test).

oliver-sanders (approved):
Tried out with an integer cycling suite and multiple `trigger --reflow` events, the cycle points remained sorted.

They are lexicographically sorted so `10` appears between `1` and `2` though:

Screenshot 2021-01-28 at 17 18 08

I'm not sure if we have a plan for more advanced sorting yet? I can't remember how the Cylc7 GUI ended up either. I think if we split on numbers and sort them as integers that would do nicely, reference Python algorithm:

```python
def sort_key(string):
    """

    Examples:
        >>> sort_key('')
        ('',)
        >>> sort_key('123')
        (123,)
        >>> sort_key('abc')
        ('abc',)
        >>> sort_key('abc123def')
        ('abc', 123, 'def')
        >>> sort_key('123abc456')
        (123, 'abc', 456)
        >>> sort_key('a1b2c3d4')
        ('a', 1, 'b', 2, 'c', 3, 'd', 4)

    """
    buff = ''
    ret = []
    in_string = True
    for char in string:
        try:
            int(char)
        except ValueError:
            if not in_string:
                if buff:
                    ret.append(int(buff))
                in_string = True
                buff = ''
            buff += char
        else:
            if in_string:
                if buff:
                    ret.append(buff)
                in_string = False
                buff = ''
            buff += char
    if in_string:
        ret.append(buff)
    else:
        ret.append(int(buff))
    return tuple(ret) or ''
```

hjoliver (approved):

PR #584, opened 2021-01-27
Prevent progress NaN when meanElapsedTime is undefined

### LINKED ISSUE #583: `NaN` progress in task
**Describe the bug**

While testing the UI for a bug @hjoliver found. In `five`, while inspecting components, noticed one `Task` had its progress to `NaN`. While not a major issue, it could lead to unexpected issues later on.

`master`

More investigation needed.

**Expected behavior**

Progress from `0` to `100`.

**Screenshots**

Screenshot from 2021-01-27 22-46-16

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #583 

**Requirements check-list**


kinow:
I think I broke the tests on `master` when fixing branch conflicts. Waiting for CI to confirm.

kinow:
One review should do I think @hjoliver ? Coverage is reporting a huge increase, but that's only because I broke `master` tests, so the `master` coverage is quite low at the moment.

src/components/cylc/tree/cylc-tree.js
```diff
@@ -14,14 +14,39 @@
...
+ * Compute percent progress.
+ *
+ * @see https://github.com/cylc/cylc-flow/blob/de7d938496e82dbdfb165938145670dd8e801efd/lib/cylc/gui/updater_tree.py#L248-L263
+ * @param {number|undefined} startedTime in milliseconds since 1970-01-01 00:00:00 UTC, e.g. 1568353099874
+ * @param {number|undefined} meanElapsedTime mean elapsed time in seconds
+ * @returns {number} the percent progress, e.g. 25 (meaning 25% progress)
+ * @private
+ */
+function computePercentProgress (startedTime, meanElapsedTime) {
+  // startedTime > now reportedly possible via interaction with `cylc reset`
```
hjoliver, line 34: This comment belongs with the next code block?  In fact, `cylc reset` no longer exists (since SoD) so `startedTime > now` should not be possible anymore and we can remove the comment and the code.

src/components/cylc/tree/cylc-tree.js
```diff
@@ -14,14 +14,39 @@
...
+  if (!startedTime || !meanElapsedTime) {
```
hjoliver, line 35: Maybe a comment to explain why we need this? (mean elapsed time is undefined until the first instance of the task has finished running).

hjoliver (changes requested):

src/components/cylc/tree/cylc-tree.js
kinow: Good catch! Moved it down.

src/components/cylc/tree/cylc-tree.js
kinow: Yup, good idea. Added a comment.

src/components/cylc/tree/cylc-tree.js
hjoliver: (I think you missed the second sentence in my comment here :grin: )

src/components/cylc/tree/cylc-tree.js
hjoliver: I'd rather it gave the reason why undefined is possible, for future reference: task proxies exist before they start running, so startedTime will be undefined until then; and "mean elapsed time" is necessarily undefined until the first instance of a task has completed running (before that, nothing has elapsed to compute the mean of).

src/components/cylc/tree/cylc-tree.js
kinow: Sorry. I agree that can be removed if `cylc reset` has been removed. Not sure if that could happen due to time zone, difference in time or time precision between UI/browser & UI Server (as you pointed in Element).

What do you think @hjoliver ? If we remove this, we just need to move the `meanElapsedTime === 0` check up and return `0` to avoid division by zero.

src/components/cylc/tree/cylc-tree.js
kinow: I've edited my last commit @hjoliver . Hope you don't mind I've copied your text as-is :smile: Just added an extra line about the NaN. Also moved the `|| meanElapsedTime === 0` to this `if` statement. So the other `if` is only for the `startedTime > now`.

src/components/cylc/tree/cylc-tree.js
hjoliver: OK let's remove the comment about reset, but keep the if block in case of clock sync issues (and add a comment for that, so we know there's not some more nefarious reason that we need to be aware of in future).

kinow: Roger, removed the part about `cylc reset`, tried to list a few possible cases that could cause an issue without that check.

hjoliver (approved):
:tada:

hjoliver:
CI failure looks unrelated to the change??

kinow:
> CI failure looks unrelated to the change??

me-kicks-the-build

kinow:
And now the build passed :man_shrugging:

PR #586, opened 2021-01-27
Fix updating task states, preventing state="" replacing existing state

kinow:

**Requirements check-list**

### DIFF
--- a/src/graphql/queries.js
+++ b/src/graphql/queries.js
@@ -39,7 +39,13 @@ subscription OnWorkflowDeltasData ($workflowId: ID) {
 fragment WorkflowTreeDeltas on Deltas {
   id
   shutdown
-  added {
+  # NOTE: We set stripNull here so that values that are null in the data store
+  # are not stripped. Otherwise we could, for example, receive a task-proxy
+  # without the isHeld false/true entry. That happens when there is no
+  # value for isHeld, and then it is stripped. When stripNull: false is used,
+  # we will actually get the default value from Protobuf (in the case of
+  # isHeld, we will get isHeld: false, which is what we want for the UI).
+  added (stripNull: false) {
     ...WorkflowTreeAddedData
   }
   updated {
--- a/tests/unit/components/cylc/tree/index.spec.js
+++ b/tests/unit/components/cylc/tree/index.spec.js
@@ -16,7 +16,6 @@
  */
 
 import {
-  createTaskProxyNode,
   getCyclePointId,
   populateTreeFromGraphQLData
 } from '@/components/cylc/tree/index'
@@ -123,12 +122,6 @@ describe('Tree component functions', () => {
     expect(task.node.progress).to.equal(50)
     // remove the mock
     stub.restore()
-  })
-  it('should initialize the task proxy state if undefined', () => {
-    const taskProxy = sampleWorkflow1.taskProxies[0]
-    delete taskProxy.state
-    const taskProxyNode = createTaskProxyNode(taskProxy)
-    expect(taskProxyNode.node.state).to.equal('')
   })
   it('should not throw an error when the workflow to be populated is invalid', () => {
     const tree = {}
--- a/tests/unit/components/cylc/tree/tree.spec.js
+++ b/tests/unit/components/cylc/tree/tree.spec.js
@@ -501,6 +501,26 @@ describe('CylcTree', () => {
       cylcTree.updateTaskProxy(taskProxy)
       expect(cylcTree.root.children[0].children[0].children[0].node.state).to.equal(TaskState.WAITING.name)
     })
+    it('Should not change the state of an existing task proxy when updating it', () => {
+      const taskProxyId = `${WORKFLOW_ID}|${cyclePoint.id}|foo`
+      const taskProxyState = TaskState.RUNNING.name.toLowerCase()
+      const taskProxy = createTaskProxyNode({
+        id: taskProxyId,
+        firstParent: {
+          id: familyProxy.id
+        },
+        state: taskProxyState
+      })
+      cylcTree.addTaskProxy(taskProxy)
+      expect(cylcTree.root.children[0].children[0].children[0].node.state).to.equal(TaskState.RUNNING.name)
+
+      // state="running", now a delta without state should NOT change the existing state
+      const updateTaskProxy = createTaskProxyNode({
+        id: taskProxyId
+      })
+      cylcTree.updateTaskProxy(updateTaskProxy)
+      expect(cylcTree.root.children[0].children[0].children[0].node.state).to.equal(TaskState.RUNNING.name)
+    })
     it('Should not update an task proxy if it is not in the tree', () => {
       const taskProxyId = `${WORKFLOW_ID}|${cyclePoint.id}|foo`
       const taskProxyState = TaskState.RUNNING.name
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -64,6 +64,8 @@ in the tree view structure.
 
 ### Fixes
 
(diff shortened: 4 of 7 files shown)


kinow:
This should be fixed now. The test failures are fixed in #584

kinow:
@hjoliver we can wait for #584 to be merged before reviewing. Then I'll rebase this one and CI will be OK again.

hjoliver:
Just tested, the unknown task status is gone, but now the held indicator doesn't show up.

kinow:
> Just tested, the unknown task status is gone, but now the held indicator doesn't show up.

Hmmm. @hjoliver what happens if you hold a task, and then expand it in the tree?

hjoliver:
Then it shows up!

kinow:
Rebased :crossed_fingers:

hjoliver:
From @kinow on Element chat: 
> looks like it's just a reactivity issue ...
>
> I've done some progress with the Task component not updating. When I update name or status, the new values are immediately rendered (both are string 🤔).
> 
> But changing progress (number) or isHeld (Boolean) has no effect until something else changes (like status, expanded in another element, etc) and then the component re-renders

kinow:
Rebased

kinow:
@hjoliver I was adding the ID to the element key, so I could more easily manipulate/find it in vue-dev-tools or programmatically while debugging the issue.

But adding `:key="node.node.id"` appears to have made it work. My guess is that Vue was not able to link the change in state to the component. But having the `:key` of the component, when the data changed Vue was able to now locate the component and update it. But I would like to keep digging in the reactivity of the CylcTree and components using it a bit longer after this issue.

But if that fixes the issue for now, at least we can fix this issue and open another one for follow-up.

kinow:
To test, I have Chromium and Firefox open, in two monitors. With `queued` workflow (how I named the workflow you sent me yesterday @hjoliver). After a hard-refresh with `CTRL + F5`, things appear to be working OK. Though some indication that the UI is working after/during a mutation is executed would be helpful too.

hjoliver:
Hmmm, isHeld still isn't reactive for me (updated branch, re-install, re-build, hard refresh). Tried with Firefox and Chrome.

kinow:
I think I found the issue. Appears to be because `isHeld` is not defined in the `deltas.added`, due to `stripNull: true` (when `stripNull: false`, `isHeld` is returned as `false` for the task proxies).

When that happens, the JS object added to the `CylcTree` does not have a `isHeld` attribute, so Vue doesn't create an observer for it. When we receive a delta with the update that contains `isHeld: any-value`, we merge object, setting `isHeld` in the object.

But since there's no observer for `isHeld`, Vue components do not react. I tested changing locally `added(stripNull: false)` and the component was updated correctly. (@hjoliver I will leave my last commit from some days ago, key-ing the components, as that's actually good; and when vue-dev utils is open, that appears to somehow trigger the component update, probably that's why it was working for me but didn't work for you).

Posted a question to the Element chat room to see if `stripNull` is working as expected. A possible solution for this is simply use `added(stripNull: false)`, but will wait for the reply there first.

kinow:
image

kinow:
@hjoliver whenever you have time, could you please take another look and confirm if that fixes the issue for you? The `isHeld` should be applied correctly now after you trigger the mutation :crossed_fingers: Thanks!

src/graphql/queries.js
```diff
@@ -39,7 +39,13 @@ subscription OnWorkflowDeltasData ($workflowId: ID) {
...
+  # NOTE: We set stripNull here so that values that are null in the data store
+  # are not stripped. Otherwise we could, for example, receive a task-proxy
+  # without the isHeld false/true entry. That happens when there is no
+  # value for isHeld, and then it is stripped. When stripNull: false is used,
+  # we will actually get the default value from Protobuf (in the case of
+  # isHeld, we will get isHeld: false, which is what we want for the UI).
+  added (stripNull: false) {
     ...WorkflowTreeAddedData
   }
   updated {
```
oliver-sanders, line 51: Should we use `stripNull` here too?

kinow: Good question. From what I understood, if we add `stripNull: false` (it is `true` at the moment as it's inheriting from an upper node), and there is an object in the deltas pool that does not contain the `isHeld`, for example, then that would mean getting a `isHeld: false`, since that is the default value from Protobuf for booleans.

That would mean that an object from the deltas pool without `isHeld` would cause the UI to show the task as not held... though that may not be what we wanted (perhaps a poor example, sorry!)

oliver-sanders: Sorry, only just realised this is set to `stripNull: false` and remembered that `stripNull` cascades down from higher levels of the query.

Yes this makes sense. I don't think we should ever use `stripNull: true` on any of the `added` fields?

kinow: Don't think so. Hopefully this change fixes our issue without breaking something that we are not able to foresee here :crossed_fingers:

oliver-sanders (approved):
This seems like a logical change to me though I've encountered the issue it's fixing myself.

src/graphql/queries.js
```diff
@@ -39,7 +39,13 @@ subscription OnWorkflowDeltasData ($workflowId: ID) {
...
   id
   shutdown
-  added {
+  # NOTE: We set stripNull here so that values that are null in the data store
+  # are not stripped. Otherwise we could, for example, receive a task-proxy
+  # without the isHeld false/true entry. That happens when there is no
+  # value for isHeld, and then it is stripped. When stripNull: false is used,
+  # we will actually get the default value from Protobuf (in the case of
+  # isHeld, we will get isHeld: false, which is what we want for the UI).
+  added (stripNull: false) {
```
kinow, line 48: Marking as draft. @hjoliver the new merge customizer (from Lodash API) used in #530 appears to remove the need for this `stripNull`. Will test again once we have merged that PR.

tests/unit/components/cylc/tree/tree.spec.js
```diff
@@ -501,6 +501,26 @@ describe('CylcTree', () => {
...
+      cylcTree.addTaskProxy(taskProxy)
+      expect(cylcTree.root.children[0].children[0].children[0].node.state).to.equal(TaskState.RUNNING.name)
+
+      // state="running", now a delta without state should NOT change the existing state
+      const updateTaskProxy = createTaskProxyNode({
+        id: taskProxyId
+      })
+      cylcTree.updateTaskProxy(updateTaskProxy)
+      expect(cylcTree.root.children[0].children[0].children[0].node.state).to.equal(TaskState.RUNNING.name)
+    })
```
kinow, line 523: Added in #530

tests/unit/components/cylc/tree/index.spec.js
```diff
@@ -124,12 +123,6 @@ describe('Tree component functions', () => {
     // remove the mock
     stub.restore()
   })
-  it('should initialize the task proxy state if undefined', () => {
-    const taskProxy = sampleWorkflow1.taskProxies[0]
-    delete taskProxy.state
-    const taskProxyNode = createTaskProxyNode(taskProxy)
-    expect(taskProxyNode.node.state).to.equal('')
-  })
```
kinow, line 132: Removed in #530

kinow (commented):
I've moved it back to draft as I've incorporated these changes into #530 since the state could be interfering with the outputs being updated correctly.

If that PR is merged, we can probably close this one as superseded. I will take a better look after #530 is merged.

kinow:
Confirmed it has been added to #530, with the only difference being that we did not need the `stripNull: true` in the `added` delta. @hjoliver I've tested the UI after updating my `master` branch, everything seems to be working OK.

Calling mutations with the aotf form, then the state is updated correctly in the UI in ~1 second. Without the need of changing the expand/collapse state of the component.

PR #587, opened 2021-01-28
Fix task states.

hjoliver:
Minor tweaks to get all the task icons working properly on master.
- remove vestiges of the obsolete held task state
- corrected several uses of `submit_failed` to `submit-failed`
- use `WorkflowState` instead of `TaskState` in a bunch of places
- make the queued state get recognized properly (it hasn't been removed quite yet)

Result:
- queued tasks (correctly) get the waiting icon, not the unknown state icon
- nonexistent held state removed from state filter drop-down
- submit-failed and queued get the correct icon in the state filter drop-down

Note this overlaps a little with #575 but that might not be merged for a little while.

**Requirements check-list**
- [x] Does not need tests (erm, I think?).
- [x] No change log entry required (? invisible to users).

### DIFF
--- a/src/model/TaskState.model.js
+++ b/src/model/TaskState.model.js
@@ -23,7 +23,7 @@ import { Enumify } from 'enumify'
 class TaskState extends Enumify {
   // NOTE: the order of the enum values is important to calculate the group states in the UI. Items at
   // the top of the list have preference over the items below in the algorithm.
-  static SUBMIT_FAILED = new TaskState('submit_failed')
+  static SUBMIT_FAILED = new TaskState('submit-failed')
   static FAILED = new TaskState('failed')
   static EXPIRED = new TaskState('expired')
   static RUNNING = new TaskState('running')
@@ -31,7 +31,6 @@ class TaskState extends Enumify {
   static PREPARING = new TaskState('preparing')
   static QUEUED = new TaskState('queued')
   static WAITING = new TaskState('waiting')
-  static HELD = new TaskState('held')
   static SUCCEEDED = new TaskState('succeeded')
   static _ = this.closeEnum()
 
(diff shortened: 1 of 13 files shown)


tests/unit/components/cylc/workflow/toolbar.vue.spec.js
```diff
@@ -18,27 +18,27 @@
...
 import store from '@/store/index'
 import Vuetify from 'vuetify/lib'
 
 const mockedWorkflowService = {
   releaseWorkflow: function () {
     return new Promise((resolve) => {
-      if (store.state.workflows.workflows[0].status === TaskState.HELD.name.toLowerCase()) {
-        store.state.workflows.workflows[0].status = TaskState.RUNNING.name.toLowerCase()
+      if (store.state.workflows.workflows[0].status === WorkflowState.HELD.name.toLowerCase()) {
+        store.state.workflows.workflows[0].status = WorkflowState.RUNNING.name.toLowerCase()
```
hjoliver, line 29: @kinow - why do we have these `toLowerCase()` calls everywhere?  All the state names are already lower case!

kinow: They are. But initially they were all upper case :-) more vestiges from the intial versions of the UI

kinow (approved):

tests/unit/components/cylc/workflow/toolbar.vue.spec.js
oliver-sanders: Can be removed now I think.

hjoliver: Done

kinow (approved):
Whew, I thought we had only a few remaining `.toLowerCase()`'s! Thanks @hjoliver !

src/model/TaskState.model.js
```diff
@@ -23,15 +23,14 @@ import { Enumify } from 'enumify'
 class TaskState extends Enumify {
   // NOTE: the order of the enum values is important to calculate the group states in the UI. Items at
   // the top of the list have preference over the items below in the algorithm.
-  static SUBMIT_FAILED = new TaskState('submit_failed')
```
oliver-sanders, line 26: Dammit, should have spotted that!

oliver-sanders (approved):

PR #594, opened 2021-02-12
Automate release

### LINKED ISSUE #475: Automate release with GitHub actions

Inspired by work from @MetRonnie and @oliver-sanders on isodatetime. We should be able to use an action to upload the zip to GitHub. Must be even easier than `pip`, as we already have the GitHub token for the action in the env var.

The process is:

- check out with `git`
- `cd $PROJECT_HOME`
- `yarn install`
- `yarn build --optimize`
- `zip -r ../cylc-ui-$VERSION-dist.zip *`
- `sha256sum ../cylc-ui-$VERSION-dist.zip` and save the output somewhere
- Create a tag and a release in GitHub
- Upload the `cylc-ui-...-dist-zip` file to the release
- Add sha256 sum to the file description (for conda Release Managers)

**Additional context**

To avoid issues like in the previous release - https://github.com/conda-forge/cylc-ui-feedstock/pull/4

**Pull requests welcome!**

kinow:
These changes close #475 

Adds two new workflows, with the same names as in isodatetime (which I used as reference). The first just creates a PR with updated changelog date, and `package.json` version. Merging that PR triggers the second PR which will create the release/tag, upload the generated production build, and add comments.

The `package.json` is updated too to use `0.3.0-0`. That's the equivalent of Python's `0.3.0dev`, or Java's `0.3.0-SNAPSHOT`. A pre-release as in https://semver.org/ (see item 9).

With all that done, we can safely update our Conda Forge recipe with the sha25sum value, and the zip URL. Credits go to @MetRonnie & @oliver-sanders who did most of the work I copied to automate this release.

**Requirements check-list**
- [x] Does not need tests (why? build CI change).

### DIFF
--- a/.github/workflows/auto_publish_release.yml
+++ b/.github/workflows/auto_publish_release.yml
@@ -1,0 +1,83 @@
+name: Release stage 2 - auto publish
+
+on:
+  pull_request:  # types AND paths
+    types: [closed]
+    paths: ['package.json']
+    # NOTE: While this is too generic, we use the `if` condition of the job to narrow it down
+    # NOTE: Don't use `branches` as we might create release on any branch
+
+env:
+  # Best not to include the GH token here, only do it for the steps that need it
+  MERGE_SHA: ${{ github.event.pull_request.merge_commit_sha }}
+  CHANGELOG_FILE: CHANGES.md
+
+jobs:
+  publish:
+    if: >-  # NOTE: Can't use top-level env here unfortunately
+      github.event.pull_request.merged == true &&
+      contains(github.event.pull_request.labels.*.name, 'release')
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+
+      - name: Checkout repo
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ env.MERGE_SHA }}
+
+      - name: Install
+        run: yarn install
+
+      - name: Build
+        run: yarn build --optimize
+
+      - name: Get the version number
+        run: |
+          VERSION=$(node -p -e "require('./package.json').version")
+          echo "VERSION=$VERSION" >> $GITHUB_ENV
+
+      - name: Create ZIP archive
+        run: |
+          cd dist && zip -r ../cylc-ui-${{ env.VERSION }}-dist.zip *
+
+      - name: sha256sum
+        run: |
+          SHA256SUM=$(sha256sum cylc-ui-${{ env.VERSION }}-dist.zip)
+          echo "SHA256SUM=$SHA256SUM" >> $GITHUB_ENV
+
+      - name: Publish GitHub release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          commitish: ${{ env.MERGE_SHA }}
+          tag_name: ${{ env.VERSION }}
+          release_name: cylc-ui ${{ env.VERSION }}
+          prerelease: ${{ env.PRERELEASE }}
+          draft: false
+          body: |
+            See [${{ env.CHANGELOG_FILE }}](https://github.com/${{ github.repository }}/blob/master/${{ env.CHANGELOG_FILE }}) for detail.
+
+            sha256sum ${{ env.SHA256SUM }}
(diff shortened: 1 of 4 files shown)


kinow:
Push-forced this latest commit to my `kinow master` fork branch. Then triggered the workflow part I with `VERSION=0.3.0` (note that I am setting the version in `package.json` to 0.3.0-0 in this PR too):

- https://github.com/kinow/cylc-ui/pull/9#issuecomment-777884654

Merging that PR, it triggered the workflow part II:

- https://github.com/kinow/cylc-ui/runs/1883793604?check_suite_focus=true

That created the 0.3.0 tag and release. The body of the release contains the notes about changelog and checksum. And finally uploaded the produced ZIP file to the release. I tested the sha25sum value, and it checks. Also inspected the contents of the ZIP and it looks good! :tada: 

p.s.: I had forgotten the `isodatetime` from the release-name, but I have already renamed it to `cylc-ui` (see diff in this PR), but apparently GH actions cache it for a while? Not too concerned since the code in this PR is OK, and I can manually rename it — if that happens in Cylc UI, though I'm guessing it's only because I've been triggering this action non-stop since this morning :rofl: 

Thanks @MetRonnie for the great work on the Cylc GH actions!

kinow:
@MetRonnie taking up on your offer to review :grimacing: 

Only thing I am not entirely happy with this PR is the last step, of the auto publish workflow. The one that comments in the pull request, and has the `if: always()`. When it works, the message is OK.

But when the auto publish workflow fails, the message is for PYPI, saying that it failed to create the release etc... Not sure if we should remove that message from this PR, use the other action to comment directly here, or modify our cylc-release GH action to have a custom message for NPM. WDYT?

@wxtim I think you've reviewed some GH actions before? Would you be able to review this one with Ronnie, please? 

Thanks!

wxtim (approved):
- [x] Read - it seems to make sense.
- [x] Run the actions on wxtim/cylc-ui - happy.

.github/workflows/auto_publish_release.yml
```diff
@@ -0,0 +1,83 @@
...
+      - name: Publish GitHub release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          commitish: ${{ env.MERGE_SHA }}
+          tag_name: ${{ env.VERSION }}
+          release_name: cylc-ui ${{ env.VERSION }}
+          prerelease: ${{ env.PRERELEASE }}
```
MetRonnie, line 58: I don't think this will work as `$PRERELEASE` is not set? In the Python project workflows, I've used the Packaging module to determine if the release is pre or not in the "Get version number" step. Might be possible using a node/npm tool. Or can just get rid of this argument/option.

MetRonnie (commented):
> But when the auto publish workflow fails, the message is for PYPI, saying that it failed to create the release etc... Not sure if we should remove that message from this PR, use the other action to comment directly here, or modify our cylc-release GH action to have a custom message for NPM. WDYT?

Yes, there is an issue for that: https://github.com/cylc/release-actions/issues/2. I think we should modify the create-release-pr action, but probably not a priority right now.

___

Otherwise, all looks good apart from one small comment 👍

.github/workflows/auto_publish_release.yml
kinow: Ah, good catch. I had a look at the `npm version` command, and whether we could find out whether the selected version is a prerelease or not. But found nothing ready to use, the closest I found were regexes, but probably better to look into that later and just remove that option for now IMO.

kinow:
> > But when the auto publish workflow fails, the message is for PYPI, saying that it failed to create the release etc... Not sure if we should remove that message from this PR, use the other action to comment directly here, or modify our cylc-release GH action to have a custom message for NPM. WDYT?
> 
> Yes, there is an issue for that: [cylc/release-actions#2](https://github.com/cylc/release-actions/issues/2). I think we should modify the create-release-pr action, but probably not a priority right now.
> 
> Otherwise, all looks good apart from one small comment +1

Thanks @MetRonnie !

Removed the `prerelease` option for now, but left a `TODO` marker there for later.

MetRonnie (approved):

PR #599, opened 2021-02-22
Run JSDoc in GH actions

### LINKED ISSUE #597: Run JSDoc in the CI pipeline

I use the IDE to validate the JS docs, but we should be able to run JSDoc with our GitHub actions.

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #597 

GH Action: https://github.com/marketplace/actions/jsdoc-action
Good intro to JSdoc: https://dev.to/paulasantamaria/document-your-javascript-code-with-jsdoc-2fbf

References for tags: https://jsdoc.app/

In this PR, we have the first use of `typedef` in Cylc UI. This defines the types that can be re-used in other functions. This is useful to reduce the places where you have to define a type.

For example, we have several functions that receive data from GraphQL queries. We had functions like

```js
// module_a.js
/**
 * @param {{id: string, node: Object, ...}} data
 */
function handleDeltaExample(data) { }

// module_b.js
/**
 * @param {{id: string, node: Object, ...}} data
 */
function getCyclePointData(data) {}
````

And in this case, it's easy for the param types to be out of sync. So instead now we have in one of the modules under `src.components.cylc.tree` a module defining

```js
/**
 * @typedef {Object} SomeNode
 * @property {string} id
 * @property {Object} node - node
 * ...
 */
```

After this we use `SomeNode` everywhere where that object is used. The new script `yarn run docs` produces the JSDoc3 HTML in `docs/jsdoc`, which can be used by a developer that is not using an IDE to see the documentation of each type and function.

**Requirements check-list**
- [x] Does not need tests (why? build only, docs).
- [x] Fixes JSDoc issues


kinow:
The build currently fails due to several JSdoc issues.

image

src/graphql/index.js
```diff
@@ -30,7 +30,6 @@ import store from '@/store'
 /**
  * Create the HTTP and WebSocket URLs for an ApolloClient.
  *
- * @private
```
kinow, line 33: Duplicate

src/components/cylc/tree/tree-nodes.js
```diff
@@ -0,0 +1,364 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import TaskOutput from '@/model/TaskOutput'
+import Vue from 'vue'
+
```
kinow, line 20: This module defines functions and types for the tree nodes, also fixing a cyclic dependency between `tree/index` and `tree/cylc-tree`.

oliver-sanders (approved):

hjoliver (approved):
:+1:

PR #603, opened 2021-02-28
Update dependencies (Vuetify, tests, svgo, etc)

kinow:
This is a small change with no associated Issue. One review should do, or maybe 2? I tested locally with a workflow too, using master from flow&UIS.

`eslint` has always something new to complain, but these are normally really easy to fix (there is a loop in `BroadcastSetting` that broke with the updated `eslint`).

Vuetify is the most annoying if we lack too much in updating it. Spent most of this morning trying to understand why some of our unit tests broke when I updated to the latest version of Vuetify. Turns out we have to use `Vue.use(Vuetify)` so that Vuetify components are able to do `this.vm.$vuetify.etc`: https://github.com/vuetifyjs/vuetify/issues/13163#issuecomment-786439815

Fixed here. Also included minor improvements like removing `console.log` from tests, simplifying some tests by introducing a `mountFunction` to reduce code duplication, and removed console warnings that appeared when you executed the tests locally.

**Requirements check-list**


src/components/graphqlFormGenerator/components/BroadcastSetting.vue
```diff
@@ -83,16 +83,16 @@ export default {
...
+          const key = keys[0]
           ptr = ptr[key]
           if (typeof ptr === 'object') {
             ret += `[${key}]`
           } else {
             ret += `${key}=${ptr}`
             ptr = null
           }
-          // only allow one definition per item
-          break
```
kinow, line 95: The linter complained about a loop iterated once :man_shrugging:

src/views/GraphiQL.vue
```diff
@@ -62,7 +62,8 @@ export default {
     createFetcher () {
       const subscriptionClient = this.$workflowService.subscriptionClient
       return subscriptionClient !== null
-        ? graphQLFetcher(subscriptionClient, fallbackGraphQLFetcher, this) : fallbackGraphQLFetcher
+        ? graphQLFetcher(subscriptionClient, fallbackGraphQLFetcher, this)
+        : fallbackGraphQLFetcher
```
kinow, line 66: This, and the file above, both fixed automatically by the linter :man_shrugging:

tests/unit/components/cylc/tree/tree.spec.js
```diff
@@ -289,8 +289,6 @@ describe('CylcTree', () => {
       })
       cylcTree.addFamilyProxy(familyProxy1)
       cylcTree.addFamilyProxy(familyProxy1)
-      console.log(cyclePoint2.id)
-      console.log(cylcTree.root.children[0])
```
kinow, line 293: My bad, not necessary for this test :+1:  (we do have some `console.log` or `console.error` being tested, to confirm warnings are raised, etc… but here was just debugging I forgot to remove).

tests/unit/components/graphqlFormGenerator/formgenerator.vue.spec.js
```diff
@@ -166,11 +168,25 @@ const NESTED_TYPES = [
...
+   * @returns {Wrapper<FormGenerator>}
+   */
+  const mountFunction = options => {
+    const vuetify = new Vuetify()
+    return mount(FormGenerator, {
       localVue,
+      vuetify,
+      ...options
+    })
+  }
```
kinow, line 187: Added this function as in other tests, to avoid having to `const vuetify = new Vuetify()` etc inside each test function.

kinow:
Had to keep the version of webpack (due to vue-cli) and sass-loader (due to https://stackoverflow.com/a/66087132).

kinow:
Couldn't fix the warning "Download the Vue Devtools extension for a better development experience:", but that will be for another PR.

src/components/graphqlFormGenerator/components/BroadcastSetting.vue
```diff
@@ -83,16 +83,16 @@ export default {
       let ptr = object_
       let ret = ''
       while (ptr) {
-        for (const key in ptr) {
+        const keys = [...ptr.keys()]
+        if (keys && keys.length !== 0) {
+          const key = keys[0]
```
oliver-sanders, line 88: I managed to get this traceback opening the "mutations" view which I think is related to this:

> Error in render: "TypeError: ptr.keys is not a function. (In 'ptr.keys()', 'ptr.keys' is undefined)"

kinow: Argh. Object.keys. good catch, I forgot to test mutations.

Will fix it and see if I can write a short test to look for this error too.

kinow: Fixed!

kinow:
Added a new e2e test that loads the default Workflow view with the tree view, then clicks to add a new Mutation view/component. It will check that "Sample Mutation" is displayed, and fail if it is not, or if there are any alerts displayed in the UI (such as ptr.keys is not a function :+1: ). Confirmed it failed with previous code, and passes after the latest changes.

Screenshot from 2021-03-02 08-56-00

Screenshot from 2021-03-02 09-15-32

hjoliver (approved):
Looks good :+1:

PR #606, opened 2021-03-04
tree: swap task name and job order

oliver-sanders:
completes point one of #333 

Swap the job and task name for readability.

Screenshot from 2021-03-04 13-16-53

**Requirements check-list**

### DIFF
--- a/src/components/cylc/tree/TreeItem.vue
+++ b/src/components/cylc/tree/TreeItem.vue
@@ -57,6 +57,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
       </slot>
       <slot name="task-proxy" v-else-if="node.type === 'task-proxy'">
         <div :class="getNodeDataClass()" @click="nodeClicked">
+          <!-- Task summary -->
           <task
             v-cylc-object="node.node.id"
             :key="node.node.id"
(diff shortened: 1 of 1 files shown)


src/components/cylc/tree/TreeItem.vue
```diff
@@ -64,15 +64,17 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-          <span class="mx-1">{{ node.node.name }}</span>
           <div v-if="!isExpanded" class="node-summary">
             <!-- Task summary -->
             <job
               v-for="(job, index) in node.children"
               v-cylc-object="job.id"
               :key="`${job.id}-summary-${index}`"
-              :status="job.node.state" />
+              :status="job.node.state"
+              style="margin-left: 0.25em;"
```
kinow, line 74: We are using the Vuetify margin-right (`mr`) and margin-left (`ml`) spaces elsewhere. Would that work here too, for consistency? e.g. `ml-1` (`4px`), `ml-2` (`8px`).

oliver-sanders: If we use px as a measurement then this margin won't scale with the font size when it is changed from the user dashboard.

Really surprised vuetify doesn't use relative units like em's rem's, vw, etc, it's pretty essential for responsive and accessible design.

kinow: They have a whole docs section on accessibility, but I haven't checked if these px values vary for other viewports.

So far I think the base layout and most components (including the tree component) have been using vuetify classes.

If you prefer this em value maybe it'd be worth adding a css class + scss var in our styles? Up to you 👍

kinow (approved):
Much better! Easier to scan the tree

kinow (approved):

PR #607, opened 2021-03-04
task: update progress with css animation

### LINKED ISSUE #563: task progress bar update
The radial progress bar computation is correct, but it only gets called when the data of the particular task gets updated. Consequently if a task doesn't emit any messages in between start and finish, the progress bar goes from ~zero to 100 when the task finishes.

So for the following flow, the progress bar only sees intermediate values if the messages are included:
```
[scheduling]
  cycling mode = integer
  initial cycle point = 1
  [[graph]]
     P1 = foo
[runtime]
  [[foo]]
     script = """
sleep 2
cylc message "hello 1"
sleep 2
cylc message "hello 2"
sleep 2
cylc message "hello 3"
sleep 2
cylc message "hello 4"
sleep 2
cylc message "hello 5"
sleep 2
     """
```

Can we have some kind of heartbeat to regularly update the progress bars of active tasks??

oliver-sanders:
closes #563

Also:

* Upgrade the Guide page from Vuetify1->2
* Use the task state enum in the Guide to keep it in sync with task state changes.
* Change the submitted task colour
  * see also https://github.com/cylc/cylc-flow/pull/3831
  * note the terminal colours need to be more vibrant so don't match

Start a CSS3 animation for each `<task :status="running" />` component that will keep the progress clock-face up to date for the life of the running job.

This saves having to routinely compute the progress and should, hopefully, be pretty efficient.

Example screen grab, the page is reloaded a few times to show that the animations are started with the correct offset for tasks that are already running before the page is loaded.

progress

**Requirements check-list**
- [ ] Tests :( (see comment)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,9 @@ None or N/A.
 
 ### Enhancements
 
+[#607](https://github.com/cylc/cylc-ui/pull/607) - Animate task progress
+using average run times.
+
 [#598](https://github.com/cylc/cylc-ui/pull/598) - Add mutation button
 
 [#530](https://github.com/cylc/cylc-ui/pull/530) - Display message triggers.
--- a/src/components/cylc/Mutation.vue
+++ b/src/components/cylc/Mutation.vue
@@ -57,10 +57,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
        style="font-size:1.5em;"
        v-if="status != 'waiting'"
       >
-        <Task
-         :status="status"
-         :progress="0"
-        />
+        <Task :status="status"/>
         {{ status }}
       </p>
       <pre v-if="status === 'failed'">{{ response }}</pre>
--- a/src/components/cylc/Task.vue
+++ b/src/components/cylc/Task.vue
@@ -20,6 +20,35 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
          which can sit in text.
 -->
 <script>
+import TaskState from '@/model/TaskState.model'
+
+/* Compute the style attribute that makes the progress animation work. */
+function getRunningStyle (context) {
+  if (
+    context.props.status === TaskState.RUNNING.name &&
+    context.props.startTime &&
+    context.props.estimatedDuration
+  ) {
+    const startTime = Date.parse(context.props.startTime)
+    const now = Date.now()
+    const elapsedTime = ((now - startTime) / 1000)
+    return `
+      animation-name: c-task-progress-animation;
+      /* clock position should be proportional to elapsed time */
+      animation-timing-function: linear;
+      /* run this animation once and only once /*
+      animation-iteration-count: 1;
+      /* the duration of the animation (the estimated task run time) */
+      animation-duration: ${context.props.estimatedDuration}s;
+      /* start the animation at the right point (note negative duration) */
+      animation-delay: -${elapsedTime}s;
+      /* stay at 100% once the animation has finished */
+      animation-fill-mode: forwards;
+    `
+  }
+  return ''
+}
+
 export default {
   name: 'Task',
   functional: true,
@@ -36,7 +65,12 @@ export default {
       type: Boolean,
       require: true
     },
-    progress: {
+    startTime: {
+      type: String,
+      required: false,
+      default: ''
+    },
+    estimatedDuration: {
       type: Number,
       required: false,
       default: 0
@@ -69,7 +103,7 @@ export default {
               r: '25',
               'stroke-width': '50',
               'stroke-dasharray': '157',
-              style: `stroke-dashoffset: ${157 - (157 * (context.props.progress / 100))}`
+              style: getRunningStyle(context)
             }
           }
         ),
--- a/src/components/cylc/Task.vue
+++ b/src/components/cylc/Task.vue
@@ -34,8 +34,8 @@ function getRunningStyle (context) {
     const elapsedTime = ((now - startTime) / 1000)
     return `
       animation-name: c-task-progress-animation;
-      /* clock position should be proportional to elapsed time */
-      animation-timing-function: linear;
+      /* restrict the number of frames in the animation for performance */
+      animation-timing-function: steps(50);
       /* run this animation once and only once /*
       animation-iteration-count: 1;
       /* the duration of the animation (the estimated task run time) */
--- a/src/components/cylc/tree/Tree.vue
+++ b/src/components/cylc/tree/Tree.vue
@@ -66,12 +66,12 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               v-model="tasksFilter.states"
             >
               <template v-slot:item="slotProps">
-                <Task :status="slotProps.item.value" :progress=0 />
+                <Task :status="slotProps.item.value"/>
                 <span class="ml-2">{{ slotProps.item.value }}</span>
               </template>
               <template v-slot:selection="slotProps">
                 <div class="mr-2" v-if="slotProps.index >= 0 && slotProps.index < maximumTasks">
-                  <Task :status="slotProps.item.value" :progress=0 />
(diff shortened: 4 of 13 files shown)


kinow:
Animation is looking good already!

oliver-sanders:
Would like to test the progress, however, can't find a way to do this quickly and am running out of time pre-beta.

Should be able to inspect the CSS attrs in the live document (timing dependent tests, yuck).

Couldn't get cypress/vue working but then I haven't a clue how to configure these things, opened an issue for another day https://github.com/cylc/cylc-ui/issues/610

Marking as ready for review with manual testing.

kinow (commented):
Found no issues in the new code, excellent work updating existing styles, tests, and code :1st_place_medal: :100: going to do some more testing with a workflow now.

kinow (approved):
Tested on Chrome and FIrefox. The framerate went up a lot, mainly because the paint/rendering of the Cylc UI was very "quiet" before, without much action, unless deltas changed.

I tested the `five` workflow first.

Screenshot from 2021-03-11 09-36-27

From `27ms` it went up to `1241ms`, mainly because of the animations occupying a thread now.

Screenshot from 2021-03-11 09-06-05

Complex started with `63ms`.

Screenshot from 2021-03-11 09-39-08

Going up to `4005ms`.

Screenshot from 2021-03-11 09-44-42

Memory is the same in both cases. In the last picture, the green "boxes" in the graph are the painting. Changed mainly due to animation. There are some red marks on top. These are dropped frames (we don't want a lot of those unnecessary frames).

For `five` and `complex`, I didn't see any [jank](https://developer.mozilla.org/en-US/docs/Glossary/Jank)iness, or any issue with the animations.

Then I wrote this workflow to test with more tasks. I wanted to have multiple tasks animated simultaneously. @hjoliver here's what I managed to write without bothering anyone :tada (a very simple workflow, but I was proud for writing it without having to consult the guide for every line, and for using `cylc validate .`, `cylc view -p .` :grimacing: ) 

```
#!jinja2

{% set letters='abcdefghijklmnopqrstuvwxyz' %}

[scheduler]
UTC mode = true

[scheduling]
initial cycle point = 1
final cycle point = 100
cycling mode = integer

[[queues]]
[[[default]]]
limit = 26


[[graph]]

P1 = """
{% for letter in letters %}
{{ letter }}[-P1] => {{ letter }}
{% endfor %}
"""

[runtime]
[[root]]
script = sleep 15

{% for letter in letters %}
[[ {{letter}} ]]
{% endfor %}
```

First tests, `348ms` painting.
Screenshot from 2021-03-11 10-14-04

Up to `5652ms` with animations. No jank on Chrome. On Firefox I had to restart the experiment a few times, and it was still performing badly. So I closed developer tools, restarted everything and it got a bit better, but I had the felling it wasn't performing great.

Screenshot from 2021-03-11 10-20-16

Besides CPU being a bit more busy now, the memory appears to have some stress, though not noticeable for me on my computer (not a lot of things running, just browser, music in the background, IDE).

Screenshot from 2021-03-11 10-15-09

Screenshot from 2021-03-11 10-20-40

I used Firefox to check the average frame rate (can't find it on Chrome). We have about 45-49 [frames per second on average](https://developer.mozilla.org/en-US/docs/Tools/Performance/Scenarios/Animating_CSS_properties).

110710918-820a0500-8263-11eb-8be4-48d74fe864d1

Energy consumption went up a bit, but that's also normal when you have animations I believe. Before:

Screenshot from 2021-03-11 11-23-06

After (I didn't see High without animations, but the value is not much different to be honest).

Screenshot from 2021-03-11 11-20-30

This workflow shows the UI much more busy than normal due to the number of tasks, but maybe I exaggerated on the number of tasks.

Screenshot from 2021-03-11 10-45-05

In theory, users on mobile may have issues if there are a lot of animations in the screen. But I guess it's safe to merge this and just keep an eye to see if anyone has a workflow with lots of tasks running simultaneously (I don't think so? probably most users with such workflows will need queue limits?).

Great work @oliver-sanders ! Approving.

Bruno

p.s. I think the animations may cause jank as Vue.js uses [requestAnimationFrame](https://github.com/vuejs/vue/blob/b51430f598b354ed60851bb62885539bd25de3d8/dist/vue.runtime.js#L7337-L7347); the animation we have uses `stroke-dashoffset`, which cannot go to a GPU (I think only those `translate` -> `translate3d`, `transform` -> `transform3d` are supported?), meaning they go into the same loop to be updated/run.

p.p.s.: we could add [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) to our scss file to disable animations if the user has enabled it in the operating system level; this could be used too for troubleshooting UI performance issues, but this is for later IMO

kinow:
Not related to the change here, but with the workflow I wrote for this review, the UI appears to have a lot of console errors, due to missing `children` attributes, and also got one of those `node with no parent` errors :confused: 

At one point my Firefox got really slow, so I restarted everything, and noticed this in the UIS log.

```bash
2021-03-11 10:08:23,267 tornado.application ERROR    Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x7f8f58063d90>>, <Task finished name='Task-276210' coro=<WorkflowsManager.update() done, defined at /home/kinow/Development/python/workspace/cylc-uiserver/cylc/uiserver/workflows_mgr.py:226> exception=KeyError('kinow|ui-607/run1')>)
Traceback (most recent call last):
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.8/site-packages/tornado/ioloop.py", line 741, in _run_callback
    ret = callback()
  File "/home/kinow/Development/python/workspace/cylc-uiserver/venv/lib/python3.8/site-packages/tornado/ioloop.py", line 765, in _discard_future_result
    future.result()
  File "/home/kinow/Development/python/workspace/cylc-uiserver/cylc/uiserver/workflows_mgr.py", line 267, in update
    self.active.pop(wid)
KeyError: 'kinow|ui-607/run1'
```

hjoliver:
@kinow - :100: on the PR testing :grin: 

> here's what I managed to write without bothering anyone...

For the amusement value, you have hit upon our old way of generating "parameterised tasks" with Jinja2 loops.  We can do that now more cleanly with built-in task parameters: 
```ini
[task parameters]
   letter = a, b, c, d, e, f, g, h, i, j, k, l, \
            m, n, o, p, q, r, s, t, u, v, w, x, y, z
[scheduler]
   UTC mode = true
[scheduling]
   initial cycle point = 1
   final cycle point = 100
   cycling mode = integer
   [[queues]]
      [[[default]]]
         limit = 26
   [[graph]]
      P1 = "foo<letter>[-P1] => foo<letter>"
[runtime]
   [[root]]
      script = sleep 15
   [[foo<letter>]]
```
To exactly reproduce your task names (`a`, `b`, etc. instead of `foo_a`, `foo_b`, ...):
```ini
[task parameters]
   [[templates]]
       letter = %(letter)s
```

oliver-sanders:
> @kinow - 100 on the PR testing grin

Brilliant testing!

I was a little worried about whether these animations would take a toll, looks like they do.

I've just found CSS3 has something called "steps" which I think can be used to define the number of frames an animation uses which should make a big difference, taking a look in...

oliver-sanders:
Ok that turned out to be really easy, I've now restricted the animation to use 50 frames. It's not quite as smooth but is smooth enough:

prog50

We can easily configure this to whatever number we want.

hjoliver:
> I've just found CSS3 has something called "steps" which I  ...

Yeah I intended to have a play with steps myself late this afternoon, but I ran into horrendous `yarn install` issues on my neglected office computer, and blew my remaining time on figuring that out and upgrading nodejs (which did fix the problem, phew).

kinow:
#TodayILearned about these steps. Will read more about it tomorrow.Thanks!!!

hjoliver (approved):
Nice

PR #611, opened 2021-03-12
Handle errors for each delta data, avoid bailing out of whole deltas …

kinow:

**Requirements check-list**

### DIFF
--- a/src/components/cylc/tree/deltas.js
+++ b/src/components/cylc/tree/deltas.js
@@ -49,7 +49,12 @@ function applyDeltasPruned (pruned, tree) {
   Object.keys(PRUNED).forEach(prunedKey => {
     if (pruned[prunedKey]) {
       for (const id of pruned[prunedKey]) {
-        tree[PRUNED[prunedKey]](id)
+        try {
+          tree[PRUNED[prunedKey]](id)
+        } catch (error) {
+          // eslint-disable-next-line no-console
+          console.error('Error applying pruned-delta, will continue processing the remaining data', error, id)
+        }
       }
     }
   })
@@ -84,10 +89,15 @@ function applyDeltasAdded (added, tree) {
   Object.keys(ADDED).forEach(addedKey => {
     if (added[addedKey]) {
       added[addedKey].forEach(addedData => {
-        const createNodeFunction = ADDED[addedKey][0]
-        const treeFunction = ADDED[addedKey][1]
-        const node = createNodeFunction(addedData)
-        tree[treeFunction](node)
+        try {
+          const createNodeFunction = ADDED[addedKey][0]
+          const treeFunction = ADDED[addedKey][1]
+          const node = createNodeFunction(addedData)
+          tree[treeFunction](node)
+        } catch (error) {
+          // eslint-disable-next-line no-console
+          console.error('Error applying added-delta, will continue processing the remaining data', error, addedData)
+        }
       })
     }
   })
@@ -119,10 +129,15 @@ function applyDeltasUpdated (updated, tree) {
   Object.keys(UPDATED).forEach(updatedKey => {
     if (updated[updatedKey]) {
       updated[updatedKey].forEach(updatedData => {
-        const updateNodeFunction = UPDATED[updatedKey][0]
-        const treeFunction = UPDATED[updatedKey][1]
-        const node = updateNodeFunction(updatedData)
-        tree[treeFunction](node)
+        try {
+          const updateNodeFunction = UPDATED[updatedKey][0]
+          const treeFunction = UPDATED[updatedKey][1]
+          const node = updateNodeFunction(updatedData)
+          tree[treeFunction](node)
+        } catch (error) {
+          // eslint-disable-next-line no-console
+          console.error('Error applying updated-delta, will continue processing the remaining data', error, updatedData)
+        }
       })
     }
   })


kinow:
With this change, every delta should be processed, regardless of errors with other deltas. In case of an error, it will log an error about not being able to process the delta. Will include the JS error stack. As well as the delta received for dev/debugging. Ideally this error should never occur.

hjoliver:
This is an improvement, but I think if any part of a delta update cannot be applied we have to re-fetch the full workflow state and start again, to be sure that the UI represents the true workflow state.  (And, of course, figure out why the delta could not be applied and prevent it from happening again).

kinow:
I've reduced the tasks to 5 (i.e., letters `a->e`), and limited the queue to `1`. The tree had no issues, nor any console errors. The deltas were processed correctly as much as I could tell.

Here's the diff I am using in the UI for testing it.

```diff
diff --git a/src/components/cylc/tree/cylc-tree.js b/src/components/cylc/tree/cylc-tree.js
index 5e5df4dd..15e1f806 100644
--- a/src/components/cylc/tree/cylc-tree.js
+++ b/src/components/cylc/tree/cylc-tree.js
@@ -42,6 +42,13 @@ function mergeWithCustomizer (objValue, srcValue, key, object, source) {
     if (objValue === undefined) {
       Vue.set(object, `${key}`, srcValue)
     }
+    if (key === 'node' &&
+      Object.hasOwnProperty.call(object, 'node') &&
+      Object.hasOwnProperty.call(object.node, 'isQueued') &&
+      Object.hasOwnProperty.call(srcValue, 'isQueued')
+    ) {
+      console.log(`WAS id ${object.node.id}, isQueued: ${object.node.isQueued} - NOW: ${srcValue.isQueued}`)
+    }
     // 2. object[key], or objValue, is defined but without reactivity
     //    this means somehow the object got a new property that is not reactive
     //    so let's now make it reactive with the new value!
diff --git a/src/components/cylc/tree/deltas.js b/src/components/cylc/tree/deltas.js
index 51e98cf4..63ed858c 100644
--- a/src/components/cylc/tree/deltas.js
+++ b/src/components/cylc/tree/deltas.js
@@ -179,6 +179,10 @@ function handleInitialDataBurst (deltas, tree) {
  * @param {CylcTree} tree - Tree object backed by an array and a Map
  */
 function handleDeltas (deltas, tree) {
+  const str = JSON.stringify(deltas)
+  if (str.includes('isQueued')) {
+    console.log(deltas)
+  }
   if (deltas.pruned) {
     applyDeltasPruned(deltas.pruned, tree)
   }
```

There's a bit of a hack, to filter the received delta string for `isQueued` by serializing it as JSON :sweat_smile: Going to slowly increase the queue limit, and the number of tasks to see when I get the errors.

kinow:
Still using 3 cyclepoints, 5 letters + 2 queue limit, no issue. 5 letters + 5 queue limit, no issue. 5 letters + 15 queue limit, no issue.

kinow:
All the letters, 5 queue limit, and 3 cyclepoints. No problem.

kinow:
@dwsutherland you mentioned on Element chat that you'd take a look. If you'd like the workflow @hjoliver and I are using: https://github.com/cylc/cylc-ui/pull/607#issuecomment-796363996 :+1: the issue happens after a while, trying to understand what leads to the error. If it's number of tasks, number of cycle points (i.e. running workflow for a long time?), etc.

kinow:
All the letters, 26 queue limit, 3 cycle points, no issue.

Increased from 3 cycle points to 6 (i.e. `final cycle point = 6`), and then the error finally happened. Both JS console errors, and the UI showing `isQueued` icon on tasks that have started.

When the error happened, I search for the task in the logs, but couldn't find it. Either it caused an error in the UI somewhere else that I am not aware and is not logging anything, or there're some deltas missing.

kinow:
Final test. All letters, 6 cycle points, but limited to 5 queue limit. No errors.

kinow:
Will trigger GH actions later. So please hold reviewing until CI passes. This addresses - partially - the issue we had with deltas last week.

For now I think there is something in the backend that is causing some deltas to not be sent. But regardless of that, I think we can use this change to avoid dropping deltas (and later figure out how to ask for the ones that could not be processed).

hjoliver:
@kinow  where is this PR at now, with regard to my comment above https://github.com/cylc/cylc-ui/pull/611#issuecomment-797144601

kinow:
> @kinow where is this PR at now, with regard to my comment above [#611 (comment)](https://github.com/cylc/cylc-ui/pull/611#issuecomment-797144601)

I added this part thinking about your comment

>(and later figure out how to ask for the ones that could not be processed).

This is not fixing that part @hjoliver. I'm not sure if we have a way to request one specific delta from the backend. The current behaviour is drop the delta and stop processing remaining deltas. This PR is only changing it to drop the current delta and keep processing remaining deltas.

hjoliver:
OK, looking at the code, it still does as the description says.  The trouble is, if any delta or part of it goes unprocessed the tree view will no longer be in sync with the state of the back end.

kinow:
> OK, looking at the code, it still does as the description says. The trouble is, if any delta or part of it goes unprocessed the tree view will no longer be in sync with the state of the back end.

Correct, which is an issue we have had for a while. There are cases where I am forced to refresh the browser when issues in JS (like missing parent node) occur.

I just searched for deltas in the UI issues, and couldn't find one for re-requesting deltas. Should I create one here to follow up on your comment? Or should it be on UIS first, or maybe flow?

hjoliver:
I suppose this can go in as an improvement on what happens in master, but we'll still need to address the more serious problem of what to do if a delta is missed or is bad.  Seems to me we have to start again with a new complete state. I don't think the back end can handle re-sending a particular delta (and if there's something wrong with the delta it would presumably fail again anyway).

hjoliver:
I'll put an issue up...

kinow:
> (and if there's something wrong with the delta it would presumably fail again anyway).

I think so too. A delta for a family proxy, for example, where its parent cyclepoint or family proxy is not in the tree has no way to be added to the tree. Though, similarly, if we drop the delta that removes the family proxy from the tree, then that node is never removed from the UI widget.

I think your suggestion of starting again is the simplest. Perhaps we could warn the user and let him/her choose what to do, in case there's something important in the current UI.

hjoliver:
> A delta for a family proxy, for example, where its parent cyclepoint or family proxy is not in the tree has no way to be added to the tree. 

That's essentially a bug in the datastore that we need to fix.  We should be able to eliminate problems like this, and just fall back to starting again from scratch in case of unexpected problems.

kinow:
GH actions now passing :+1:

hjoliver (approved):
Good :+1: - with follow-up #614

src/components/cylc/tree/deltas.js
```diff
@@ -49,7 +49,12 @@ function applyDeltasPruned (pruned, tree) {
   Object.keys(PRUNED).forEach(prunedKey => {
     if (pruned[prunedKey]) {
       for (const id of pruned[prunedKey]) {
-        tree[PRUNED[prunedKey]](id)
+        try {
+          tree[PRUNED[prunedKey]](id)
+        } catch (error) {
+          // eslint-disable-next-line no-console
+          console.error('Error applying pruned-delta, will continue processing the remaining data', error, id)
```
oliver-sanders, line 56: I wouldn't want people to miss an important error like this. Perhaps we should make this a user facing error en that dismissible status bar thinggy?

oliver-sanders: Bumped to a comment in #614

oliver-sanders (approved):

PR #617, opened 2021-03-18
Use latest task states in GScan

### LINKED ISSUE #401: Check if we can use states and stateTotals in the GScan instead of jobs

ATM, the `GScan` component uses the `workflow.jobs` in the GraphQL query. During the Cylc Meetup 2020, it was pointed we could instead use `workflow.states` and `workflow.stateTotals`.

It appears to be missing the task name, so we may need to review how Cylc 7 was doing it.

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #401 

- [x] update recent message to use state totals count

From @hjoliver on Element

>We need state totals as well, because that represents the current active tasks.
>
>E.g. if there are 20 running tasks, we need the tooltip to say "20 running; the 5 that ran most recently are...".   If there are 3 running tasks, we need it to say "3 running; the 5 that ran most recently are ...".  And if there are no running tasks, "0 running; the 5 that ran most recently are ...".

- [ ] update `one` offline workflow's query
- [ ] write e2e tests for the new states (removed the workflow summary JS unit tests, easier to test now with the browser; probably for a follow-up after #624)

**Requirements check-list**


kinow:
After this change, GScan looks like:

image

kinow:
Now with tooltip including the number of current tasks in each state from query GraphQL query's `stateTotals`.

GIFrecord_2021-03-19_090234

kinow:
Rebased. Will fix tests after we have decided on which icons to use :+1:

hjoliver:
Before I forget, what we decided in the last meeting was gscan should:
> show n=0 (active + unhandled failed) in icon list, 5-most-recent in tooltips

Note n=0 currently also includes waiting-but-limited tasks (queued, runhead, xtriggered, held). But we could decide to exclude those.

IMO the 5-most-recent state X tool-tip *should* be available even when there are no current tasks with state X.  Which argues for graying out rather than removing those icons.

hjoliver:
@kinow - as in interim measure, for beta-0 release, would it be super-easy for you to just show active and failed job icons as now on master, and the tool-tips for those, and don't show the icon at all if there are no current tasks of that state.  For the moment we couldn't see most recent failed tasks when there are no current failed tasks (etc.) but that's fine for now.  (If that would be time consuming we can punt to beta-1)

kinow:
Sounds pretty easy @hjoliver !

kinow:
First part done @hjoliver 

image

Note that I fixed the width of workflow name and workflow states columns. This way, with the default font size, the workflow name does not push the workflow states column out causing the overflow (which is not a flex-wrap, nor an HTML element wrap, so I couldn't modify with CSS :disappointed_relieved: ). The solution for this is setting the `flex-basis`…

If you Zoom with the browser, things should still work. However, if you visit Settings, and increase the font size, then that may cause overflow. Maybe with CSS grid it would be easier to define the fixed grids for GScan, but probably for a follow-up PR.

Pending: 

- [ ] I've modified the GraphQL query @hjoliver , adding @dwsutherland 's latest state tasks , but re-generating the workflow breaks tests (#624), so I won't update it nor add e2e tests
- [ ] fix unit tests due to changes in GScan

@hjoliver could you take a look to see if it's looking OK, please? If so, I will start chasing what's wrong in each failing unit test.

Thanks!

src/graphql/queries.js
```diff
@@ -200,6 +200,13 @@ subscription {
...
     port
     stateTotals
+    latestStateTasks(states: [
+      "failed",
+      "preparing",
+      "submit-failed",
+      "submitted",
+      "running"
+    ])
     taskProxies(sort: { keys: ["cyclePoint"] }) {
```
dwsutherland, line 210: Can we get rid of this `taskProxies` section now?

Also there might be some delay because of the builtin default for `subscription.workflows`:
```
    workflows = List(
        Workflow,
        description=Workflow._meta.description,
        ids=List(ID, default_value=[]),
        exids=List(ID, default_value=[]),
        # TODO: Change these defaults post #3500 in coordination with WUI
        strip_null=Boolean(default_value=False),
        delta_store=Boolean(default_value=False),
        delta_type=String(default_value=DELTA_ADDED),
        initial_burst=Boolean(default_value=True),
        ignore_interval=Float(default_value=2.5),
        resolver=delta_subs
    )
```
So setting the argument `ignoreInterval` to zero would mean all deltas are sent.

hjoliver:
> Can we get rid of this taskProxies section now?

So this will make gscan pretty efficient even though it is not (yet) delta-driven, right?    We were requesting all the tasks in order to compute the state totals in the UI, but now we can get that from the back end.

src/graphql/queries.js
kinow: >Can we get rid of this taskProxies section now?

Yes! It would be pretty bad to merge this one without removing that as the query would still request it :clap:  great catch @dwsutherland !

kinow: >So setting the argument ignoreInterval to zero would mean all deltas are sent.

Should I change it here, or do you think it could have some performance impact on the backend? If that's the case we can leave it for a follow-up.

kinow:
> > Can we get rid of this taskProxies section now?
> 
> So this will make gscan pretty efficient even though it is not (yet) delta-driven, right? We were requesting all the tasks in order to compute the state totals in the UI, but now we can get that from the back end.

Kinda I think. It is still changing all the data structure. Then Vue needs to go and check if there's any change and if there is, update each component.

Shouldn't matter if you have 10 or so workflows. If you have 50+ workflows that may cause some slowness.

src/graphql/queries.js
kinow: Removed `taskproxies` from the query.

dwsutherland: I had it in there as some way of easing the frontend and backend (as any change would trigger delta construction with potentially nested field resolving (i.e. `taskProxies`))

hjoliver:
Perhaps I should have said "this should make the UI more efficient" (not just gscan) because on master it is requesting all the tasks from all the workflows.  That could be a massive number of tasks.  I was more worried about that than rebuilding the gscan tree every time (50+ workflows should be very unusual!)

dwsutherland:
> > Can we get rid of this taskProxies section now?
> 
> So this will make gscan pretty efficient even though it is not (yet) delta-driven, right? We were requesting all the tasks in order to compute the state totals in the UI, but now we can get that from the back end.

Well `subscription.workflows` is delta driven, but the content is a static dump of whatever you query the data-store for.
However, `stateTotals` and `latestStateTasks` are static anyway so it'll be fine as is... 
Only thing is `subscription.workflows` is not workflow specific at the moment (i.e. `ids: ["..."]` isn't used in this subscription (though it could be)), so all workflows might show up

src/graphql/queries.js
kinow: Happy to change it to `0` here if it'd make the UI more responsive with little impact on the server. @hjoliver any thoughts on this?

hjoliver:
> Only thing is subscription.workflows is not workflow specific at the moment (i.e. ids: ["..."] isn't used in this subscription (though it could be)), so all workflows might show up

For gscan, we want all workflows to show up, but we don't want it to request all the tasks from all the workflows. (Or have I misunderstood?)

dwsutherland:
The other option is using `subscription.deltas.updated.workflow` ... But it's essentially the same thing, just single target workflow.

dwsutherland:
> > Only thing is subscription.workflows is not workflow specific at the moment (i.e. ids: ["..."] isn't used in this subscription (though it could be)), so all workflows might show up
> 
> For gscan, we want all workflows to show up, but we don't want it to request all the tasks from all the workflows. (Or have I misunderstood?)

Yes, we can get rid of the `taskProxies` part ... But like you said; at the moment it's all tasks all workflows .. I was addressing the latter.

dwsutherland:
Hold on, will double check; `subscription.workflows` might be able to target all workflows but only send the workflow that changed on delta prompt (install of all)..

dwsutherland:
So we can do this with two subscriptions, one for discovery:
```
subscription {
  workflows (deltaType: "added", deltaStore: true, initialBurst: true, ignoreInterval: 0){
    id
    isHeldTotal
    isQueuedTotal
    latestStateTasks (
      states: [
        "submitted",
        "running",
        "submit-failed",
        "failed",
        "succeeded"
      ]
    )
    stateTotals
  }
}
```
And one for updates:
```
subscription {
  workflows (deltaType: "updated", deltaStore: true, ignoreInterval: 0, stripNull: true){
    id
    isHeldTotal
    isQueuedTotal
    latestStateTasks (
      states: [
        "submitted",
        "running",
        "submit-failed",
        "failed",
        "succeeded"
      ]
    )
    stateTotals
  }
}
```
The id still shows up by itself... So that will need to be looked at in the future (as you know)

hjoliver:
Tested - I rather like that, with the greyed out icons so we can still have the tool-tips on those  (even though I was actually suggesting not to do that until we collectively decide!  ... but let's leave it like you've done it to see what people think).

hjoliver:
@dwsutherland the state-totals seems to be wrong.
```ini
[scheduler]                                                                                                                                       
   allow implicit tasks = True                                                                                                                    
[scheduling]                                                                                                                                      
    cycling mode = integer                                                                                                                        
    initial cycle point = 1                                                                                                                       
       [[graph]]                                                                                                                                      
     P1 = """                                                                                                                                     
         foo => bar & baz => qux                                                                                                                  
     """                                                                                                                                          
[runtime]                                                                                                                                         
   [[root]]                                                                                                                                       
       script = sleep 10                                                                                                                          
   [[qux]]                                                                                                                                        
       script = false  
```

All the unhandled-failed `qux` get kept in n=0. The tree view correctly shows that, but gscan thinks total failed tasks is zero.

dwsutherland:
> @dwsutherland the state-totals seems to be wrong.
> 

Yeah, I need to look into this ... Noticed the same when working on the active state totals

src/graphql/queries.js
dwsutherland: Not a lot of impact on the backend for small subscriptions (without querying all nodes), so see how it goes set to `0`

kinow:
> Tested - I rather like that, with the greyed out icons so we can still have the tool-tips on those (even though I was actually suggesting not to do that until we collectively decide! ... but let's leave it like you've done it to see what people think).

Sorry! Missed that part, but I liked it too :)

kinow:
Pending the backend issue @hjoliver found yesterday, I believe @dwsutherland is looking into it. But now CI passes. Once we fix the offline data workflow, I'll look into adding more tests :+1:

kinow:
For the two subscriptions I would need, I think, to:

- Start the first subscription when the GScan component is created, just like we are doing with the single sub now
- Modify our existing callback, which only updates the component loading state, and instead start a second subscription
- The second subscription would then call the old callback that changes the GScan isLoading variable, and additionally also stop the first subscription
- Both subscriptions would write to the Vuex store

This requires modifying the queries.js file, GScan.vue, hopefully not changing gquery.js nor workflow.service.js. But also updating unit/e2e tests.

I can start it today, and finish either tomorrow noon/avro, or in the beginning of the next week if I don't have any issues implementing the plan above.

@hjoliver should we try this one? I would also add a note to the GScan + deltas PR, as I believe that PR would basically have to delete the two subscriptions and use deltas instead.

hjoliver:
(Going to Element to discuss...)

dwsutherland (approved):
Looks good to me:
image

Needs the `ignoreInterval` lowered or removed though 👍

kinow:
From Element: 

1. Created production build of the UI with `yarn run build` (better performance, less memory leaks due to browser extension/etc).
2. Copied `five` as `five-running` (my `five` is installed via symlink).
3. Set the `workflows (ignoreInterval: 0) {` in the GScan query
4. Started hub with `cylc hub` (already have 10 workflows I normally use for testing, all stopped)
5. Launched browser, UIS happily running, with <12MB of memory, note is OK with no CPU fan, only firefox/chromium open
6. Started the 20 workflows with `for x in {1..20}; do cylc install && cylc play five-running/run${x}; done`
7. UI running OK, memory under 40MB. It went up to 50MB when I clicked to open the tree view. But it's going back to 30, 34,23 MB (i.e. garbage collection is working OK, no memory leaks).

The CPU fan appears to be working though not as bad as before. I can open the other tabs in both browsers, nothing wrong with my computer, so I assume it's the ignoreInterval set to 0 should be OK?

dwsutherland:
> The CPU fan appears to be working though not as bad as before. I can open the other tabs in both browsers, nothing wrong with my computer, so I assume it's the ignoreInterval set to 0 should be OK?

As long as the workload on your machine is not attributed to the UIS or UI then this is fine
(and the workload on the UIS is no noticeably worse than before the the UI change)

src/components/cylc/gscan/GScan.vue
```diff
@@ -528,6 +513,43 @@ export default {
         return `/workflows/${ node.node.name }`
       }
       return ''
+    },
+
+    /**
+     * Count how many tasks we have in a given state. The state will have
+     * been retrieved from `latestStateTasks`. And the list of tasks in
+     * each states is from the `stateTotals`. This gives us the number of
```
hjoliver, line 521:
```suggestion
     * been retrieved from `latestStateTasks`. And the number of tasks in
     * each state is from the `stateTotals`. This gives us the number of
```

hjoliver (approved):
Looks great. Frequent updates with busy workflows, but nice low-volume payloads. :+1: 

I like the greyed out state-total=0 icons, but either way its easy tweak that later if @oliver-sanders has a better design idea.

src/components/cylc/gscan/GScan.vue
```diff
@@ -528,6 +513,43 @@ export default {
...
+     * active tasks in a given state (`latestStateTasks` includes old tasks).
```
hjoliver, line 522:
```suggestion
     * Get number of tasks we have in a given state. The states are retrieved 
     * from `latestStateTasks`, and the number of tasks in each state is from
     * the `stateTotals`. (`latestStateTasks` includes old tasks).
```

oliver-sanders:
> https://github.com/oliver-sanders/cylc-doc/pull/new/install-error

I like it too :+1:, a nice innovation from the task-icon experiment!

PR #627, opened 2021-03-24
Use json-server to start a JupyterHub REST API endpoint, testing Axios now

kinow:

Just staging some changes I had in my local workspace. I was experimenting with the idea of removing the mocked services. Instead, let the real services do the work and be tested by e2e specs.

Then just start the endpoints in `dev` mode (i.e. `test`, `offline`, etc, `!== production`), and proxy it with the devServer. It should be transparent to the user, and also to other devs. Only difference is that there will be a new process in the background listening for some port with `json-server`. Not sure if we can use that for GraphQL too, or if I'd need to start something like an express http server.

**Requirements check-list**


package.json
```diff
@@ -13,7 +13,9 @@
     "dev": "yarn run serve",
     "docs": "jsdoc -c jsdoc.conf.json",
     "lint": "vue-cli-service lint",
-    "serve": "vue-cli-service serve --mode offline",
+    "serve": "yarn run serve:jupyterhub & yarn run serve:vue",
```
kinow, line 16: I wish NPM and Yarn supported multiple commands & command chaining. This isn't so nice as if `yarn run serve:jupyterhub` fails, then the other command continues running and you cannot stop it easily with `CTRL + C`. The alternatives are installing packages like `npm-run-all`, `npm-or-yarn`, or starting both commands separately :man_shrugging:

src/services/mock/db.json
```diff
@@ -0,0 +1,12 @@
+{
+  "userprofile": {
```
kinow, line 2: This creates `/userprofile`. In theory we can create any endpoint like `/users`, `/graphql`, etc.

src/services/plugin.js
```diff
@@ -58,15 +58,14 @@ export default {
...
    * The service is available as `Vue.$userService`.
    *
-   * @param {boolean} offline - whether the application is in offline mode or not
    * @private
    */
-  _installUserService (offline) {
-    Vue.prototype.$userService = offline ? new MockUserService() : new UserService()
+  _installUserService () {
+    const baseUrl = getBaseUrl()
+    Vue.prototype.$userService = new UserService(baseUrl)
```
kinow, line 69: Real service now, no more mocks :+1:

src/utils/urls.js
```diff
@@ -0,0 +1,49 @@
...
+  // TODO: revisit this and evaluate other ways to build the GraphQL URL - not safe to rely on window.location (?)
+  const baseUrl = getBaseUrl()
+  const httpUrl = `${baseUrl}graphql`
+  const websocketsProtocol = baseUrl.startsWith('https') ? 'wss:' : 'ws:'
+  const wsUrl = `${websocketsProtocol}//${baseUrl}subscriptions`
+  return {
+    httpUrl: httpUrl,
+    wsUrl: wsUrl
+  }
+}
```
kinow, line 49: Just to re-use the base URL.

vue.config.js
```diff
@@ -31,6 +31,14 @@ module.exports = {
...
     }
   },
+  devServer: {
+    proxy: {
+      '^/userprofile': {
+        target: 'http://localhost:3000/',
+        changeOrigin: true
+      }
+    }
+  },
```
kinow, line 41: Development server proxy'ing requests to the other process.

kinow:
The e2e errors are very interesting. They are failing, apparently, due to a promise in vue router navigation guard taking longer now, as it does an HTTP request, instead of the mock request. The route is evaluated more slowly, so the test sets the error, then the navigation guard replaces it by empty (every time you navigate to a view, it clears the alerts).

package.json
kinow: Using https://github.com/kimmobrunfeldt/concurrently#readme, after reading about it in several posts. Simpler than `npm-run-all` or `yarn-run-all`, but also doesn't appear to lock us with `npm` nor `yarn`. Also liked the option to have a prefix with some color to identify where the log is coming from (similar to `docker-compose`).

kinow:
If CI is OK, then it's ready for review now.

kinow:
Rebasing, and fixing e2e tests.

kinow:
Now the e2e tests should pass for both browsers :crossed_fingers:

kinow:
e2e tests still failing :sighs: sorry for the noise

kinow:
See last message of the Codecov bot, where the test coverage increases, and `axios.js` is now being tested. Next step is to see if we can kill the mocked `WorkflowService` too, and instead just have the `json-server` JupyterHub fake API returning the GraphQL recorded JSON responses.

kinow:
rebased

kinow:
Ha! It was failing with the same error found in the #653 PR, due to the null coalesce operator missing being transpiled. I've added the same code changes here, and will rebase/fix whichever is conflicted. Ready for review.

src/services/user.service.js
```diff
@@ -21,12 +21,19 @@ import store from '@/store/'
...
+  constructor (baseUrl) {
+    this.baseUrl = baseUrl
+  }
+
   /**
    * Gets the user profile from the backend server.
    * @returns {Promise<*>} - a promise that dispatches Vuex action
    */
   getUserProfile () {
-    return axios.get(`${window.location.pathname}/userprofile`).then((response) => {
```
oliver-sanders, line 29: I had noticed this was causing a `//` in the URL (which the UIS/hub seemed to handle ok), might be worth using a URL join function in the future.

vue.config.js
```diff
@@ -23,14 +23,22 @@ module.exports = {
...
   productionSourceMap: process.env.NODE_ENV !== 'production',
   pluginOptions: {
     apollo: {
       lintGQL: false
     }
   },
+  devServer: {
+    proxy: {
+      '^/userprofile': {
+        target: 'http://localhost:3000/',
```
oliver-sanders, line 37: Is 3000 the default of the json-server?

oliver-sanders (commented):
Looks good.

oliver-sanders (approved):
It works.

The `yarn run serve` output is a bit messy now because it relies on rewriting text lines in the terminal but I'm not sure there is much we can do about that.

vue.config.js
kinow: Yes.

image

But doesn't hurt to be explicit about the port in our `scripts` section (also helpful if `3000` is in use and we want to quickly switch to a different port locally). Pushed a new commit with the port in the `scripts` section.

image

hjoliver (approved):
:+1:

PR #628, opened 2021-03-25
actions: fix release workflow

oliver-sanders:
Run 1 of the release action stage 1 failed:

https://github.com/cylc/cylc-ui/runs/2191797691?check_suite_focus=true

Is this used?

https://github.com/cylc/cylc-ui/blob/11b8f87d506fdf6fec1150f81d3892db966d8ffa/.github/workflows/create_release_pr.yml#L9-L12

### DIFF
--- a/.github/workflows/create_release_pr.yml
+++ b/.github/workflows/create_release_pr.yml
@@ -30,7 +30,7 @@ jobs:
 
       - name: Set the package version
         run: |
-          npm version $VERSION
+          npm version ${{ github.event.inputs.version }}
 
       - name: Update "released on" date in changelog
         continue-on-error: true
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -516,7 +516,7 @@ export default {
     },
 
     /**
-     * Get number of tasks we have in a given state. The states are retrieved 
+     * Get number of tasks we have in a given state. The states are retrieved
      * from `latestStateTasks`, and the number of tasks in each state is from
      * the `stateTotals`. (`latestStateTasks` includes old tasks).
      *


MetRonnie:
I had a go running `npm version` locally, and it wouldn't work for `npm version 0.3`, but it would work for `npm version 0.3.0` 🙃

oliver-sanders:
Hmmm ok, well we might as well move to maj.min.mtn numbering anyway.

oliver-sanders:
Renamed milestones.

.github/workflows/create_release_pr.yml
```diff
@@ -30,7 +30,7 @@ jobs:
 
       - name: Set the package version
         run: |
-          npm version $VERSION
+          npm version ${{ github.event.inputs.version }}
```
MetRonnie, line 33: I don't think this is any different (apart from the env variable is supposedly sanitised)

Maybe a mention of the need to specify `major.minor.patch` in the input description field would be useful?

oliver-sanders: Where is $VERSION set?

MetRonnie: https://github.com/cylc/cylc-ui/blob/11b8f87d506fdf6fec1150f81d3892db966d8ffa/.github/workflows/create_release_pr.yml#L20-L21

oliver-sanders: Magic variable?

MetRonnie: (Looking back at that action, not sure it exactly works. But it does set the `$VERSION` and `$BRANCH` env vars, so kind of needed for now)

PR #638, opened 2021-04-01
toolbar: fix line break for narrow displays

oliver-sanders:
Fix a small responsive display bug using a float trick.

Also had a go at centralising the view logic so we don't have to manually write out each view. One day views should be pluggable things that users could contribute.

Before:

before

After:

after

**Requirements check-list**

### DIFF
--- a/src/components/cylc/workflow/Toolbar.vue
+++ b/src/components/cylc/workflow/Toolbar.vue
@@ -89,17 +89,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         </template>
         <v-list class="pa-0">
           <v-list-item
-            v-bind:class="[
-              'py-0',
-              'px-8',
-              'ma-0',
-              'c-add-view',
-              `c-add-view-${ view.name }`
-            ]"
             :id="`toolbar-add-${ view.name }-view`"
-            @click="$listeners['add']"
             v-for="view in views"
-            v-bind:key="view.title"
+            :key="view.title"
+            @click="$listeners['add'](view.name)"
+            class="py-0 px-8 ma-0 c-add-view"
           >
             <v-list-item-icon>
               <v-icon>{{ view.icon }}</v-icon>
--- a/src/components/cylc/workflow/Toolbar.vue
+++ b/src/components/cylc/workflow/Toolbar.vue
@@ -37,7 +37,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
     </v-btn>
     <!-- title -->
     <v-toolbar-title
-      class="tertiary--text font-weight-light"
+      class="font-weight-light"
     >
       <span class="c-toolbar-title">{{ title }}</span>
     </v-toolbar-title>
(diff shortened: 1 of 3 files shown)


src/components/cylc/workflow/Toolbar.vue
```diff
@@ -37,7 +37,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
     </v-btn>
     <!-- title -->
     <v-toolbar-title
-      class="tertiary--text font-weight-light"
```
oliver-sanders, line 40: This doesn't seem to do anything?

src/components/cylc/workflow/Toolbar.vue
```diff
@@ -150,20 +157,32 @@ export default {
...
       stop: null
-    }
+    },
+    views: [
+      {
+        name: 'tree',
+        title: 'Tree',
+        icon: mdiFileTree
+
+      },
```
oliver-sanders, line 178: Hopefully these values can be provided by the views themselves one day.

src/styles/cylc/_toolbar.scss
```diff
@@ -26,20 +26,39 @@
 .c-toolbar {
   background-color: map-get($grey, 'lighten-4') !important;
 
+  .v-toolbar__title {
+    // clear default padding (too much space on the left)
+    padding: 0!important;
```
oliver-sanders, line 31: This gets rid of the extra large space between the collapsed "drawer" icon and the flow name.

src/components/cylc/workflow/Toolbar.vue
kinow: I think it may have changed in Vuetify. It used to set the font size if I remember correctly.

src/styles/cylc/_toolbar.scss
kinow: Do we need a space between `0` and `!important`?

src/components/cylc/workflow/Toolbar.vue
```diff
@@ -81,23 +81,30 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+              'py-0',
+              'px-8',
+              'ma-0',
+              'c-add-view',
+              `c-add-view-${ view.name }`
+            ]"
+            :id="`toolbar-add-${ view.name }-view`"
+            @click="$listeners['add']"
+            v-for="view in views"
+            v-bind:key="view.title"
```
kinow, line 102: I think this could be simpler.

If I understood it correctly, we have a list in the Vue component with the views. For each view we iterate it and create a CSS class with the view name.

Then we fire an event on click. The method bound to the event will parse the CSS class and parse the view name. I think we should be able to call the listener method passing the view name. That eliminates the need for a dynamic css class, and the need to parse the CSS class to retrieve the view name.

src/components/cylc/workflow/Toolbar.vue
kinow: Ah! It was for color. 

image

I've removed theming from Vuetify as we have another issue for that, but at some point I believe we had something similar to

```js
vuetify: {
  theme: {
    dark: true,
    themes: {
      dark: {
        primary: '#4caf50',
        secondary: '#ff8c00',
        tertiary: '#4682bf',
        accent: '#9c27b0'
      }
    }
  }
},
```

All good to remove it as we won't use the themes from Vuetify I think

kinow:
Here's what I mean by simplifying that function

```diff
diff --git a/src/components/cylc/workflow/Toolbar.vue b/src/components/cylc/workflow/Toolbar.vue
index 341c7e5c..428d2186 100644
--- a/src/components/cylc/workflow/Toolbar.vue
+++ b/src/components/cylc/workflow/Toolbar.vue
@@ -89,17 +89,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         </template>
         <v-list class="pa-0">
           <v-list-item
-            v-bind:class="[
-              'py-0',
-              'px-8',
-              'ma-0',
-              'c-add-view',
-              `c-add-view-${ view.name }`
-            ]"
             :id="`toolbar-add-${ view.name }-view`"
-            @click="$listeners['add']"
             v-for="view in views"
-            v-bind:key="view.title"
+            :key="view.title"
+            @click="$listeners['add'](view.name)"
+            class="py-0 px-8 ma-0 c-add-view"
           >
             <v-list-item-icon>
               <v-icon>{{ view.icon }}</v-icon>
diff --git a/src/views/Workflow.vue b/src/views/Workflow.vue
index b1ae7095..fad8a393 100644
--- a/src/views/Workflow.vue
+++ b/src/views/Workflow.vue
@@ -19,7 +19,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
   <div>
     <CylcObjectMenu />
     <toolbar
-      v-on:add="this.clickAddView"
+      v-on:add="this.addView"
     ></toolbar>
     <div class="workflow-panel fill-height">
       <lumino
@@ -178,27 +178,6 @@ export default {
       this.deltaSubscriptions.push(id)
       return id
     },
-    /**
-     * Handle click events on elements designed to add new views.
-     *
-     * Clickable element must have the ID "c-add-view-<name-of-view>".
-     */
-    clickAddView (e) {
-      let ele = e.target
-      while (ele && !ele.classList.contains('c-add-view')) {
-        // go up the element tree until we find the c-add-view list item
-        ele = ele.parentElement
-      }
-      if (!ele) {
-        return
-      }
-      // const cls = ele.classList.filter(
-      const cls = Array.from(ele.classList.values()).filter(
-        (x) => { return x.startsWith('c-add-view-') }
-      )[0]
-      // extract the name of the view from the element id
-      this.addView(cls.replace('c-add-view-', ''))
-    },
     /**
      * Add a new view widget.
      *
```

Should give same behaviour with less one method/function I think.

src/styles/cylc/_toolbar.scss
kinow: Question answered, looks like webpack or the browser add the space
image

kinow (approved):

hjoliver (approved):
Good.

PR #646, opened 2021-04-16
Terminology: suite to workflow.

hjoliver:

UI sibling of https://github.com/cylc/cylc-flow/pull/4174

**Requirements check-list**


src/views/Dashboard.vue
```diff
@@ -132,11 +132,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           </v-list-item>
           <v-list-item href="https://cylc.github.io/doc/built-sphinx/suite-design-guide/suite-design-guide-master.html">
```
kinow, line 133: Will we change this URL :point_up: later too? It's 404 at the moment, I think the new URL is `https://cylc.github.io/cylc-doc/stable/html/suite-design-guide/suite-design-guide-master.html`… but will we rename the docs too to `workflow-design-guide/etc`?

hjoliver: Yes I expect we will.  I didn't actually check if the current URL is 404

kinow (approved):

PR #649, opened 2021-04-18
Avoid re-sorting a sorted array, instead adding element in its sorted index (respecting locale, and numeric on)

### LINKED ISSUE #644: Check if the tree view is sorting items in natural order
**Describe the bug**

We had an issue some time ago where I think the cycle points, family proxies, task proxies, or jobs, were not sorted in natural order.

So we got `[1, 10, 2]` instead of `[1, 2, 10]`.

This issue is a follow-up so that we verify this behaviour has indeed been fixed. Might be worth to check if this could happen in GScan too.

Apologies for taking so long to create this issue.

`master`

Create a workflow that generates cycle points, tasks, jobs, families, with numbers for names that include 1, 2, 10, for instance.

**Expected behavior**

Items sorted as in Cylc 7, in natural order.

**Screenshots**

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #644 

Workflow used:

```cylc
[scheduling]
  cycling mode = integer
  initial cycle point = 9
  final cycle point = 12
  runahead limit = P12
  [[graph]]
     P1 = "10 => 1 => 2 => 100 => 1000"
[runtime]
  [[10]]
     script = "sleep 5"
  [[1]]
     script = "true"
  [[2]]
     script = """
       if [[ "${CYLC_TASK_SUBMIT_NUMBER}" -gt 3 ]]; then
         true
       else
         false
       fi
"""
    [[[job]]]
      execution retry delays = 3*PT1S
  [[100]]
     script = "true"
  [[1000]]
     script = "sleep 5"
```

Then just `CTRL + C` after Task `2` gets a couple jobs to check it's adding items in-order.

image

(see how `9` is coming **after** `10` :point_up:  ­— at the moment we have the oldest cycle points at the bottom)

**Requirements check-list**

### DIFF
--- a/src/components/cylc/tree/cylc-tree.js
+++ b/src/components/cylc/tree/cylc-tree.js
@@ -156,7 +156,7 @@ function sortTaskProxyOrFamilyProxy (leftObject, leftValue, rightObject, rightVa
  * @param comparator {SortedIndexByComparator=} - function used to compare the newValue with otherValues in the list
  */
 function sortedIndexBy (array, value, iteratee, comparator) {
-  if (!array || array.length === 0 || !value) {
+  if (array.length === 0) {
     return 0
   }
   // If given a function, use it. Otherwise, simply use identity function.
--- a/src/components/cylc/tree/cylc-tree.js
+++ b/src/components/cylc/tree/cylc-tree.js
@@ -72,28 +72,60 @@ function computeCyclePointsStates (cyclePointNodes) {
 }
 
 /**
- * @private
- * @param {FamilyProxyNode} left
- * @param {FamilyProxyNode} right
+ * The default comparator used to compare strings for cycle points, family proxies names,
+ * task proxies names, and jobs.
+ *
+ * @param left {string}
+ * @param right {string}
  * @returns {number}
+ * @constructor
  */
-function sortTaskProxyOrFamilyProxy (left, right) {
+const DEFAULT_COMPARATOR = (left, right) => {
+  return left.toLowerCase()
+    .localeCompare(
+      right.toLowerCase(),
+      undefined,
+      {
+        numeric: true,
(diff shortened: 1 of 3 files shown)


kinow:
Need to 

- [ ] fix tasks sorting (it's showing `[1, 100, 2]` at the moment)
- [ ] fix unit tests & e2e tests
- [ ] write more tests for the new function
- [ ] jsdoc for the comparator function

kinow:
Tested locally, and after writing the tests did a quick debugging session to confirm the `CylcTree` object got families and tasks sorted as expected (i.e. families before tasks if under the same cycle point, using numeric sort)

image

hjoliver (approved):
Looks good, tested as working :+1:

src/components/cylc/tree/cylc-tree.js
```diff
@@ -72,27 +72,113 @@ function computeCyclePointsStates (cyclePointNodes) {
...
+ * Declare function used in sortedIndexBy as a comparator.
+ *
+ * @private
+ * @callback SortedIndexByComparator
+ * @param {object} leftObject - left parameter object
+ * @param {string} leftValue - left parameter value
+ * @param {object} rightObject - right parameter object
+ * @param {string} rightValue - right parameter value
+ * @returns {boolean} - true if leftValue is higher than rightValue
+ */
```
oliver-sanders, line 105: Is this declaring an interface?

oliver-sanders (approved):
Screenshot 2021-04-23 at 16 23 03

🍾

src/components/cylc/tree/cylc-tree.js
kinow: Yes, kind of. A type that can be re-used in the docs. But won't affect the code, i.e. only for docs, linters, etc

PR #651, opened 2021-04-28
Use Vuepress

### LINKED ISSUE #640: Use jsdoc (or another tool?) to document .vue files

At the moment jsdoc is parsing our `.js` files, but not the `.vue` files. I noticed this when I was writing some docs some days ago for @giulianoserrao . I wanted to update our Wiki here with auto-generated docs, instead of writing manually about each component.

I thought since we had `jsdoc` it would be just a copy and paste job until we had some GH action updating `gh-pages` or the Wiki, but alas it's not working as expected.

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #640 

Vuepress was working fine, until I got stuck with Webpack issues. Going down the rabbit hole, looks like updating to Yarn v2 could solve my current issue. It would also help for Cypress I think, so will give it a go tomorrow.

I tried to update it now following their migration guide, but the build failed. So there are probably things that need to be fixed in our build first (e.g. those many warnings I've been sweeping under the rag until now :smile: )

**Requirements check-list**


kinow:
Once #653 is merged I should be able to edit the last commit and make the build pass :+1:

kinow:
Building OK now. But Vuepress still not quite working. First issue was with istanbul-instrument-loader, which can be replaced. I removed it temporarily to test it.

Then `vue-markdown` failed. This is more problematic, and we could have the same issue when upgrading to Vue 3 too. The issue appears to be that `vue-markdown` is bringing `core-js` 0.x or 1.x, while we use v3. It's not able to locate imported functions from `core-js` and just crashes.

I had a look at `vue-markdown`'s code, and it's super simple. The issue is that it requires changes to their build process, but the project has had no activity in 4 years :disappointed_relieved: using the fork `@adapttive/vue-markdown` fixed Vuepress build, but now the Vuepress page is loaded blank, and there are errors in the console.

image

kinow:
The `core-js` issue of Vue Markdown appears to be that Vue Markdown uses babel-runtime that is v6, while we are using Babel 7.

kinow:
The current issue now is in `markdown-it-toc-and-anchor`. A dependency of `vue-markdown`, but with last release in 2019 too :disappointed_relieved:

kinow:
Found one workaround for the global issue, by defining it globally and importing with `require`: https://github.com/vuejs/vuepress/issues/1434#issuecomment-816268942

kinow:
Building passing, Vuepress displaying components correctly. Whew :sleepy: 

image

kinow:
Just need to add more components to the doc, show how it will look like in a GH pages, then it should be ready for review. Probably will finish next week or later, will move on to the GScan deltas now :+1:

kinow:
I have removed my branch, and closed the pull request, but moved the branch to this repo: https://github.com/cylc/cylc-ui/tree/vuepress

The reason is that it would take a bit longer to get everything working. And the everything is actually a lot of duplicated code.

The idea of Vuepress is really great. It compiles and runs your components in a documentation, similar to Sphinx in Python.

The main issue that I found, is that Vuepress does not load the project's configurations, especially not the webpack configuration of Cylc UI. Because of that, there are many configuration steps in the `.docs/enhanceApp` where I am basically duplicating our code from `vue.config.js` and `babel.config.js`.

This might become a problem for future developers since it is not necessarily an exact copy of the existing configuration. See the last commit, for example. I loaded Vuex with `Vue.use(store)`, so that the `Alert` component could be displayed in the Vuepress docs.

However, that was not enough, since the `this.$store` is not loaded automatically as in Cylc UI. I had to troubleshoot a little, and then learn that `Vue.mixin({ store: store })` was missing too. That's not mentioned in their docs, but you can find similar tips from forums & issues :disappointed_relieved: 

Definitely a great tool if other devs are willing to put up time and effort to maintain Vuepress in-sync and working. Main benefit is having the docs reflect exactly what we have in the source code (or fail if it is out of sync).

PR #654, opened 2021-04-28
Remove postcss config js file as we already have the yaml config (and fix e2e tests)

kinow:
This is a small change with no associated Issue.

While working on another PR for Yarn V2, I realized we have 2 configuration files for postcss. From their [docs](https://github.com/postcss/postcss-load-config/blob/master/README.md)

> It's recommended to use an extension (e.g .postcssrc.json or .postcssrc.yml) instead of .postcssrc

Given we already have `postcss.config.js`, this PR removes the extra `.js` file.

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Already covered by existing tests.
- [x] No change log entry required (why? e.g. invisible to users).
- [x] No documentation update required.

PR #657, opened 2021-04-29
Display multiple jobs with a shadow under the icon

kinow:
These changes partially address #333 

**Requirements check-list**

### DIFF
--- a/src/components/cylc/tree/TreeItem.vue
+++ b/src/components/cylc/tree/TreeItem.vue
@@ -74,7 +74,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               v-cylc-object="job.id"
               :key="`${job.id}-summary-${index}`"
               :status="job.node.state"
-              :multiple="node.children.length > 1"
+              :previous-state="node.children.length > 1 ? node.children[1].node.state : ''"
               style="margin-left: 0.25em;"
             />
           </div>
--- a/src/components/cylc/tree/TreeItem.vue
+++ b/src/components/cylc/tree/TreeItem.vue
@@ -74,6 +74,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               v-cylc-object="job.id"
               :key="`${job.id}-summary-${index}`"
               :status="job.node.state"
+              :multiple="node.children.length > 1"
               style="margin-left: 0.25em;"
             />
           </div>
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,11 @@ release.
 -------------------------------------------------------------------------------
 ## __cylc-ui-0.5 (<span actions:bind='release-date'>Released 2021-??-??</span>)__
 
+### Enhancements
+
+[#657](https://github.com/cylc/cylc-ui/pull/657) - Display a different icon, with
+a shadow underneath, for the Job component.
+
 ### Fixes
 
 [#649](https://github.com/cylc/cylc-ui/pull/649) - Avoid re-sorting a
--- a/src/components/cylc/Job.vue
+++ b/src/components/cylc/Job.vue
@@ -20,6 +20,8 @@
         which can sit in text.
 -->
 <script>
+import isEmpty from 'lodash/isEmpty'
+
 export default {
   name: 'Job',
   functional: true,
@@ -28,10 +30,9 @@ export default {
       type: String,
       required: true
     },
-    multiple: {
-      type: Boolean,
-      required: false,
-      default: false
+    previousState: {
+      type: String,
+      required: false
     }
   },
   render: function (createElement, context) {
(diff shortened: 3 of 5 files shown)


kinow:
Some screenshots with `one` offline workflow. See `eventually_succeeded` task.

Screenshot from 2021-04-30 10-53-40

Screenshot from 2021-04-30 10-54-04

Screenshot from 2021-04-30 10-55-28

Screenshot from 2021-04-30 10-56-15

kinow:
An e2e may fail for firefox, fix is in #654

src/components/cylc/tree/TreeItem.vue
```diff
@@ -74,6 +74,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               v-cylc-object="job.id"
               :key="`${job.id}-summary-${index}`"
               :status="job.node.state"
+              :multiple="node.children.length > 1"
```
oliver-sanders, line 77: Works nicely :+1:, quick thought, could we use the status of the previous job for the colour of the "shadow" job?

```suggestion
              :previousStatus="node.children.slice(-2, -1).status"
```

kinow: Ah! Good idea. This way we avoid having to define yet another color in the theme. And TIL. I didn't know we could go negative in `.slice`, thanks! I think it will be `previousState=node.children[1]` though (the query is sorting reverse, so we display the first item, so the previous is the second I think)

kinow: Will leave it for @hjoliver and others to review, and apply all suggestions to this PR on Monday 👍 Thanks again!

oliver-sanders: > the query is sorting reverse

Of course 🤦.

kinow:
@hjoliver any opinion on this change? I like @oliver-sanders ' suggestion to use the previous state instead of the shadow. See comments above. Happy to take any other feedback before I change the code.

hjoliver:
> I like @oliver-sanders ' suggestion to use the previous state instead of the shadow. See comments above. 

If the previous color is the same (several failed jobs, typically) the icon will just look like a misshapen blob. So we might need the previous color to be de-saturated a bit?

How about this:
- most recent job, full color
- previous job, de-saturated color
- if there are 3 or more jobs, a final grey shadow

kinow:
>If the previous color is the same (several failed jobs, typically) the icon will just look like a misshapen blob.

Oh, good point!

>So we might need the previous color to be de-saturated a bit?

An interesting idea, I think we can do that (I think just changing the CSS `opacity` should do, or maybe `filter`, but doable).

Your suggestion sounds good to me. Now back to @oliver-sanders :+1:

oliver-sanders:
* most recent job, full color
* previous job, ~de-saturated color~ ~50% opacity
  * Similar to GScan.
* ~if there are 3 or more jobs, a final grey shadow~
  * too complicated, good enough to just show the previous job, user can expand to see full job history.

kinow:
@oliver-sanders would definitely be simpler to implement (less one `method` or `computed` I think). Maybe we go with this approach first @hjoliver and see if we want or need the "if there are 3 or more jobs, a final grey shadow" feature?

hjoliver:
Personally I think it would be good to show 1 job, vs 2 jobs, vs 3 or more jobs ... because for certain kinds of job failures you might expect a single retry to fix it, so it could be useful to see at a glance if that didn't happen. But, no biggie, possible future enhancement at best.

kinow:
What it looks like now.

image

I'm using the exact same CSS classes we had before. The only differences from a normal Job icon are that when you pass a `previousState` value, the component will reduce the dimensions of the `rect`, and add 2 `rect`'s. One at the top left with full opacity, and the other at the bottom right with `50%` opacity.

kinow:
Rebased too.

hjoliver:
> for certain kinds of job failures you might expect a single retry to fix it, so it could be useful to see at a glance if that didn't happen. 

Replying to myself - I guess two fails shows that well enough, as opposed to fail then succeed.   So on that basis, two layers is good enough :rofl:

hjoliver:
> What it looks like now.

That looks good :tada:

oliver-sanders:
I think the job order has gone funny:

Screenshot 2021-05-05 at 15 00 52
Screenshot 2021-05-05 at 15 00 56

oliver-sanders:
Minor style suggestion, can we make the icons a little larger.

```diff
diff --git a/src/components/cylc/Job.vue b/src/components/cylc/Job.vue
index 7a79ebf..5926674 100644
--- a/src/components/cylc/Job.vue
+++ b/src/components/cylc/Job.vue
@@ -40,9 +40,9 @@ export default {
     //   * let width = 100 - x - stroke-width
     //   * let height = 100 - y - stroke-width
     const DEFAULT_SIZE = '80'
-    const PREVIOUS_STATE_ITEMS_SIZE = '60'
+    const PREVIOUS_STATE_ITEMS_SIZE = '65'
     const DEFAULT_XY = '10'
-    const PREVIOUS_STATE_ITEMS_XY = '30'
+    const PREVIOUS_STATE_ITEMS_XY = '25'
     const width = !isEmpty(context.props.previousState) ? PREVIOUS_STATE_ITEMS_SIZE : DEFAULT_SIZE
     const height = !isEmpty(context.props.previousState) ? PREVIOUS_STATE_ITEMS_SIZE : DEFAULT_SIZE
     const jobStatusIcon = createElement('rect', {
@@ -52,8 +52,8 @@ export default {
         y: DEFAULT_XY,
         width: width,
         height: height,
-        rx: '20',
-        ry: '20',
+        rx: '15',
+        ry: '15',
         'stroke-width': '10'
       }
     })
@@ -70,8 +70,8 @@ export default {
           y: PREVIOUS_STATE_ITEMS_XY,
           width: width,
           height: height,
-          rx: '20',
-          ry: '20',
+          rx: '15',
+          ry: '15',
           opacity: '50%',
           'stroke-width': '10'
         }
```

kinow:
>Minor style suggestion, can we make the icons a little larger.

Thanks for the diff patch. Applied it in the latest commit.

kinow:
>I think the job order has gone funny:

Oh! My bad! I fixed the sorting with natural order, but forgot to maintain the jobs in reverse order I think. That should be happening on `master` too (no changes here that alter that).

I'm fixing that and will push a separate PR :+1:

oliver-sanders:
Nice:

Screenshot 2021-05-06 at 10 03 21

Sadly spotted a new bug in the review, it might be a rare one, keep an eye out incase you run into it - https://github.com/cylc/cylc-ui/issues/666

oliver-sanders (approved):

hjoliver (approved):
Nice :+1:

PR #658, opened 2021-04-30
Allow user to set the order of cycle points

### LINKED ISSUE #333: Tree: subtle refactor proposal

kinow:
These changes close #333 

A draft for discussion I think. @oliver-sanders , @hjoliver I've added a setting under the User Profile page. If you change that setting, then reload the tree view (or just navigate after you've set the value), then it should be displayed with the correct order.

**Requirements check-list**

### DIFF
--- a/src/components/cylc/tree/cylc-tree.js
+++ b/src/components/cylc/tree/cylc-tree.js
@@ -263,8 +263,13 @@ class CylcTree {
    * a workflow in Cylc.
    *
    * @param {?WorkflowNode} workflow
+   * @param {*} options
    */
-  constructor (workflow) {
+  constructor (workflow, options) {
+    const defaults = {
+      cyclePointsOrderDesc: true
+    }
+    this.options = Object.assign(defaults, options)
     this.lookup = new Map()
     if (!workflow) {
       this.root = {
@@ -335,10 +340,14 @@ class CylcTree {
         cyclePoint,
         (c) => c.node.name
       )
-      // cycle points are inserted in the reverse order, so we have to handle that sortedIndex will give you the
-      // wrong index (except when the list is empty). That's why we reverse the list first, to get the index, and
-      // find where it would be in the not-reversed list.
-      parent.children.splice(parent.children.length - insertIndex, 0, cyclePoint)
+      if (this.options.cyclePointsOrderDesc) {
+        // cycle points are inserted in the reverse order, so we have to handle that sortedIndex will give you the
+        // wrong index (except when the list is empty). That's why we reverse the list first, to get the index, and
+        // find where it would be in the not-reversed list.
+        parent.children.splice(parent.children.length - insertIndex, 0, cyclePoint)
+      } else {
+        parent.children.splice(insertIndex, 0, cyclePoint)
+      }
     }
   }
 
--- a/src/components/cylc/tree/cylc-tree.js
+++ b/src/components/cylc/tree/cylc-tree.js
@@ -258,6 +258,7 @@ function sortedIndexBy (array, value, iteratee, comparator) {
  * @class
  */
 class CylcTree {
+  static DEFAULT_CYCLE_POINTS_ORDER_DESC = true
   /**
    * Create a tree with an initial root node, representing
    * a workflow in Cylc.
@@ -267,7 +268,7 @@ class CylcTree {
    */
   constructor (workflow, options) {
     const defaults = {
-      cyclePointsOrderDesc: false
+      cyclePointsOrderDesc: CylcTree.DEFAULT_CYCLE_POINTS_ORDER_DESC
     }
     this.options = Object.assign(defaults, options)
     this.lookup = new Map()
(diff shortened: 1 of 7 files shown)


src/views/UserProfile.vue
```diff
@@ -129,50 +129,59 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
             <v-layout row align-center wrap>
               <v-flex xs3>
-                <span>Colour Theme</span>
+                <span>Color Theme</span>
```
hjoliver, line 132: (Controversial decision :rofl: )

hjoliver:
Works well.

Seems to me this is fine as a user profile preference. It's likely to be a static choice, not something people want to change at different times or for different workflows.  Also, if we keep it there, we don't have to worry about reversing existing open tree views, because there won't be any. 

On naming, maybe *"Latest cycle point at top"*, or if we want top to the be default, *"Latest cycle point at bottom"*.

src/views/UserProfile.vue
kinow: Sorry, I think I just did that automatically in one of the commits. Let me undo this one :+1: (we will probably have en_GB and en_US in translations at some point I think).

kinow:
Reverted change to spelling of colour, and set the user profile settings name to "Latest cycle point at top". Will let @oliver-sanders have time to look at this draft before adding tests/changelog and marking it ready for review.

hjoliver:
Still Draft?  (I guess you want @oliver-sanders opinion on the questions above?)

kinow:
> Still Draft? (I guess you want @oliver-sanders opinion on the questions above?)

That's right. If he agrees with the current implementation I will write the tests and changelog.

oliver-sanders:
> That's right. If he agrees with the current implementation I will write the tests and changelog.

Sorry, missed this.

> where should the setting for the tree view be set? In the user preference

Sounds good to me, I can't think why users would want this to be different in one view to another.

> or we can leave it so that users have to refresh browser

That sounds fine, a notice saying "reload this page for changes to take effect" (or something like that) should be enough. We can always come back to this later if we accumulate other similar settings.

kinow:
Adding test and preparing to mark this PR as ready for review today/tomorrow :+1:

kinow:
Rebased first.

tests/unit/components/cylc/tree/tree.spec.js
```diff
@@ -186,6 +186,43 @@ describe('CylcTree', () => {
       expect(cylcTree.root.children[0].id).to.equal(`${WORKFLOW_ID}|2`)
       expect(cylcTree.root.children[1].id).to.equal(`${WORKFLOW_ID}|1`)
     })
+    it('Should add cycle points sorted DESC by default', () => {
```
kinow, line 189: These two tests assume we have the DESC order by default for cycle points. Should we change it later, we just update the constant in `CylcTree`, and these two tests, plus the remaining tests that are expecting elements in order.

hjoliver (changes requested):
The reverse order (latest point at bottom) doesn't seem to be working properly.

kinow:
Ah, debugging it for a while, finally found what's wrong! Before, I reversed the array to find the next position to add the element. But in ascending mode (1 to 2, to 3, etc) I don't need to reverse it (otherwise we get a very confusing result as you found out @hjoliver :sweat_smile: ). Trying to come up with a unit test to verify this scenario.

Thanks @hjoliver ! :100:

kinow:
Fixed! :crossed_fingers:

hjoliver:
> Ah, debugging it for a while, finally found what's wrong! Before, I reversed the array to find the next position to add the element. But in ascending mode (1 to 2, to 3, etc) I don't need to reverse it (otherwise we get a very confusing result as you found out @hjoliver sweat_smile ). Trying to come up with a unit test to verify this scenario.
> 
> Thanks @hjoliver ! 100

Yes I figured that out too and was trying to fix it for a PR to your branch ... but my JS is too slow so you bet me to it!

src/components/cylc/tree/cylc-tree.js
```diff
@@ -328,17 +334,18 @@ class CylcTree {
     if (!this.lookup.has(cyclePoint.id)) {
       this.lookup.set(cyclePoint.id, cyclePoint)
       const parent = this.root
-      // reverse to put cyclepoints in ascending order (i.e. 1, 2, 3)
-      const cyclePoints = [...parent.children].reverse()
+      // when DESC mode, reverse to put cyclepoints in ascending order (i.e. 1, 2, 3)
+      const cyclePoints = this.options.cyclePointsOrderDesc ? [...parent.children].reverse() : parent.children
```
hjoliver, line 338: Nice, I was using another `if ... else` block, which was annoying hell out of me.

hjoliver (approved):
That's got it :+1:

oliver-sanders (approved):

PR #669, opened 2021-05-11
Avoid content shift by using a skeleton loader in each cell with the item-count

### LINKED ISSUE #660: Dashboard skeleton loader: page jump
A minor but annoying design problem:

I often go to the Dashboard page with the sole intention of clicking on one of the static links (e.g. preferences or hub), then accidentally hit the wrong link when the real Workflows data suddenly replaces the skeleton, which moves the static links downward.

shot

I wonder if we should:
1. make the loader the exact same height as the data table
   -  (but this might not work depending on what gets displayed under Events at the top right side)
1. or put the static links at the top of the page

Thoughts anyone?

kinow:
These changes close #660 

**Requirements check-list**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,9 @@ cycle point icon state. After this, the Task icon used to summarize the cycle
 point state should be updated to reflect the overall state based on the states
 of its children nodes.
 
+[#669](https://github.com/cylc/cylc-ui/pull/669) - Define the height
+of each table row, and specify the Skeleton height to avoid content shift.
+
 -------------------------------------------------------------------------------
 ## __cylc-ui-0.4 (Released 2021-04-16)__
 
--- a/src/styles/cylc/_dashboard.scss
+++ b/src/styles/cylc/_dashboard.scss
@@ -47,4 +47,12 @@
   .v-list__tile {
     padding: 10px 0;
   }
+
+  table {
+    tbody {
+      tr {
+        height: 50px;
+      }
+    }
+  }
 }
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -24,21 +24,25 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
     <v-layout wrap>
       <v-flex xs12 md6 lg6>
         <p class="display-1">Workflows</p>
-        <v-skeleton-loader
-          :loading="isLoading"
-          :height="`${workflowsTable.length * 50}px`"
-          type="table-row@3"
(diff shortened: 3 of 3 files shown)


hjoliver:
Looks good, but I've noticed this method doesn't work properly if you crank the font size up high enough in our Settings page (why do we need that in addition to browser zoom, BTW?)

kinow:
> 
> 
> Looks good, but I've noticed this method doesn't work properly if you crank the font size up high enough in our Settings page (why do we need that in addition to browser zoom, BTW?)

Oh, good catch. I forgot to test after we increase the font size. Back to the whiteboard 😬

kinow:
>Looks good, but I've noticed this method doesn't work properly if you crank the font size up high enough in our Settings page 

@hjoliver I see what you mean now. But that also breaks other parts of our layout, such as the User Profile page itself, GScan, the Toolbar... Let me think how to work around this issue.

>(why do we need that in addition to browser zoom, BTW?)

If you change the zoom level, that affects the whole page, and triggers responsive layout, or mobile version. That may or may not be what the user wants. In my case, I use that feature in some sites as I have (as my wife says) an old man's eyesight. 

That allows users to change the font size without affecting the rest of the layout. But we can disable/remove that feature too if others don't use it :+1: (which unblocks this issue too :rofl: )

hjoliver:
>  But that also breaks other parts of our layout, ...

Yeah, that's why I'm wondering if we really need the ability to change font size independently.

It sounds useful in principle, but it can make the page look ugly, and it's not good if it actually breaks parts of the layout.  Let's see what others think...

oliver-sanders:
> Oh, good catch. I forgot to test after we increase the font size. Back to the whiteboard 😬

Should work with `em`s.

hjoliver:
(Can we do it with `em`s?)

kinow:
On it :+1: just finish some tests with Cylc Flow first, then will start working on this one again.

kinow:
Good ref: https://css-tricks.com/accessible-font-sizing-explained/

kinow:
Fixed with a different approach. Instead of worrying about the height of the element (which could change depending on the `font-size` [which we can use `px` or `em`]) I am using a loader within each row :grimacing: 

That way we have no content shifting, regardless of the font size. Also enabled the loader for the Dashboard table as an extra indicator of the query-loading status :+1:

hjoliver (approved):
Seems to work well, even at pretty extreme font sizes :tada:

oliver-sanders (approved):

Issue #631, opened 2021-03-25
Workflow Table View component


hjoliver:
The Cylc 7 tree/table view has been replaced in Cylc 8 by a tree view where job details collapse into leaf nodes (that's cleaner and it works better with smaller viewports, including side-by-side tabs).  However, some users still need to see the detail of all jobs at once, so we need a Cylc 8 table view.  

## basic requirements

- a [cylc-ui workflow component widget](https://github.com/cylc/cylc-ui/wiki/Workflow-Component)
- delta-driven, via subscription to the common data store
- use the cylc-ui job icons (so linear progress bars not needed)
- configurable (which columns to show)
- task name and task state filtering
  - by name, state, and cycle point.
- column sorting
- (does not need a collapsible family tree at the left - unlike Cylc 7)
- needs a cyclepoint column
- one task/job per table row:
  - by default show the latest job of each task, i.e. one table row per task
  - but click on tasks to expand and show one table row per job

hjoliver:
For reference, the Cylc 7 tree/table view:
![image](https://user-images.githubusercontent.com/2362137/112556110-60e21080-8e2e-11eb-985a-cd9c93902065.png)

hjoliver:
Suggested approach, for @giulianoserrao (@kinow might have better ideas!):

1. make a new [workflow component widget](https://github.com/cylc/cylc-ui/wiki/Workflow-Component) that can be opened in our UI tabs like the treeview, but which initially requests no workflow data and just displays "hello world!" or whatever
2.  learn how to register a data subscription (for the workflow data we want to display in the table view) with the shared UI data store, but initially just display the data as raw text. Play with GraphiQL in cylc-ui to see how data subscriptions work.
3. look for existing high-performance Vue.js or JS table view components that might work for us, possibly after modification
4. display the workflow data in the table. Note the table has to be updated reactively in response to incoming data deltas (i.e. just what changed in the workflow state).
5. mutations integration (click on tasks in the table and execute commands on them)
5. the details: look-and-feel, filtering, sorting

kinow:
>mutations integration (click on tasks in the table and execute commands on them)

For this item, see `cylc-object` directive in the existing components :+1:

hjoliver:
Closed by #672 and #861. (Updating description for auto-close)

Further improvements can be follow-ups.

Issue #667, opened 2021-05-06
Investigate what happens when a Cylc workflow is reloaded


kinow:
**Describe exactly what you would like to see in an upcoming release**

Quite sure the UI will just gets broken if the workflow structure changed too much as we are not using the `reloaded` flag 

**Additional context**

From our group meeting on May 5th.

**Pull requests welcome!**

hjoliver:
From my meeting notes:

> Check on scheduler reload, we get a new "initial data burst" - check for the reload flag in the delta

As I recall this means the scheduler sends a new initial burst (i.e. full scheduler state) after a reload.  But an initial burst is actually just a delta (containing the difference between a null state and the full state, instead of the difference since the previous delta) so the UI will have to *know* it is a burst in order to interpret it correctly.

kinow:
@hjoliver I think if at any point, if we get a Workflow node added in the deltas.added, that contains any property with `host` or any other property besides ID & status (and the ones that are arrays such as taskProxies/jobs until we remove them from deltas.added.workflow) it means that that is an initial data burst?

Or maybe there's a better way to tell that that delta is an initial data burst? Will have a look at the query later, I have forgotten if there is any flag that indicates that is the first delta.

hjoliver:
> I think if at any point, if we get a Workflow node added in the deltas.added, that contains any property with host or any other property besides ID & status (and the ones that are arrays such as taskProxies/jobs until we remove them from deltas.added.workflow) it means that that is an initial data burst?

That might be right, I'd have to take a look. 

Even so, for clarity it might be better to have a boolean flag to indicate initial burst or not.

kinow:
>Even so, for clarity it might be better to have a boolean flag to indicate initial burst or not.

:+1:

dwsutherland:
> Even so, for clarity it might be better to have a boolean flag to indicate initial burst or not.

There already is one, set by the scheduler on reload and passed into data-store re-instantiation. (`workflow.reloaded` I believe, should be in the added delta (I think))

hjoliver:
So `workflow.reloaded` appears in the initial burst even if the workflow was not actually reloaded?  If so, that's useful but the name could be better. "full-state" or something?

dwsutherland:
> So `workflow.reloaded` appears in the initial burst even if the workflow was not actually reloaded? If so, that's useful but the name could be better. "full-state" or something?

[...]

Issue #661, opened 2021-05-04
Avoid data duplication in added deltas


kinow:
**Describe exactly what you would like to see in an upcoming release**

We have added cyclepoints, family proxies, task proxies, and jobs being received twice, once in `deltas.added.workflow.$KEY` (where key is each of these node types), and again in `deltas.added.$KEY`.

**Additional context**

From Element today, issue raised by @hjoliver 

**Pull requests welcome!**

hjoliver:
I think `deltas.added.workflow.$` was added to handle the initial burst, because it only gets populated then (and is then a duplicate of `deltas.added.$`)

It seems to me we should be able to ditch the whole `deltas.added.workflow` part because the initial burst is still just a delta, isn't it? (It's still the difference between the new state and the previous state, but in this case previous state is null). 

Presumably if the tree is empty when a delta is received, then:
- if `deltas.added` is present and `deltas.updated` and `deltas.pruned` are not, it must be an initial burst, and we should construct the tree
- otherwise error

Or have I misunderstood something?

kinow:
What you said makes sense to me, but I am not sure if the initial data burst is just a delta? Say I subscribe to a query now, and at this moment the pool has (sorry if being vague, I have only high level understanding of the pool):

- 1 task proxy
- 2 job of that task proxy
  * job `1` failed
  * retry policy triggered second job `2` that is in `submitted` state

I think the subscription in GraphQL normally gives your entries of some data stream (like a database, log file, messaging system, etc) that changed after you susbcribed to the query. So if I subscribe to a query now, I don't need to receive anything.

Then the second job succeeds. Now I think I should receive the first response in my GraphQL subscription, something like

```json
{
  deltas: {
    updated: {
      id: "kinow|five|1|foo|2,
      state: "succeeded",
      ...
  }
}
```

Here the UI doesn't know about the cycle point, or the task proxy. So the job cannot be added to the tree, unless I create each node from the updated delta.

I think that's why we introduced the initial data burst. And I think that's why Apollo Client has the `subscribeToMore`, where you fetch data with a query (essentially doing the same as our initial data burst), and updates your data with a subscription that receives whatever changed after your subscription was created.

PR #672, opened 2021-05-14
Tableview

giulianoserrao:
These changes partially address #631 (@hjoliver #631 will stay open after this PR is merged; we will just create more PR's against that issue as we improve the table view, or should we create separate issues?)

**Requirements check-list**


src/layouts/Default.vue
```diff
@@ -58,12 +58,13 @@ export default {
        * want to display the default Toolbar—the Workflow view
        * has its own Toolbar that communicates with the Workflow
        * component (e.g. the Workflow Toolbar owns a button that
-       * triggers the action to add a new Tree View, so the events
+       * triggers the action to add a new Tree or Table View, so the events
```
kinow, line 61: This was just an example, I think we don't need to add it here.

src/mixins/tableview.js
```diff
@@ -0,0 +1,69 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { mapActions, mapState } from 'vuex'
+/* eslint-disable no-unused-vars */
+import User from '@/model/User.model'
+/* eslint-enable no-unused-vars */
```
kinow, line 21: We can remove these `no-unused-vars` :point_up:

src/styles/cylc/_table.scss
```diff
@@ -0,0 +1,135 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+@import '../../../node_modules/vuetify/src/styles/styles';
```
kinow, line 18: I think we didn't customize anything in the stylesheet for the Table. This file can be removed.

CONTRIBUTING.md
```diff
@@ -45,6 +45,7 @@ below.
  - Edgardo Ramírez
  - Mel Hall
  - Ronnie Dutta
+ - Giuliano Serrao
```
kinow, line 48: :tada:

src/views/Workflow.vue
```diff
@@ -39,6 +39,18 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+            v-for="widgetId of tableWidgets"
+            :key="widgetId"
+            :id="widgetId"
+            :loading="isLoading"
+            type="list-item-three-line"
+            tab-title="table"
+        >
+        <!-- <table-component-->
+        <!--    :workflows="table.root.children"-->
+        <!--  />-->
```
kinow, line 52: We can remove the code commented out above.

kinow (changes requested):
First quick review :+1:

src/styles/index.scss
```diff
@@ -28,6 +28,7 @@
 @import "cylc/task";
 @import "cylc/toolbar";
 @import "cylc/tree";
+@import "cylc/table";
```
kinow, line 31: When the `_table.scss` file is removed, this file needs to be updated too.

src/components/cylc/table/Table.vue
```diff
@@ -0,0 +1,247 @@
...
+export default {
+  name: 'TableComponent',
+  props: {
+    tasks: {
+      type: Array,
+      required: true
+    },
+    hoverable: Boolean,
+    activable: Boolean,
+    multipleActive: Boolean,
```
kinow, line 149: We can remove `hoverable`, `activable`, and `multipleActive` props here, and in the caller code (i.e. `views/Table.vue`).

src/components/cylc/table/Table.vue
```diff
@@ -0,0 +1,247 @@
...
+    filterable: {
+      type: Boolean,
+      default: true
+    }
+  },
+  components: {
+    Task
```
kinow, line 156: You already have the `Task` component here. You should be able to import the `Job` component in the same manner, and declare it here (so that you can use it in the `template` section).

Then, in the `template` section, use these components to display the icon per task proxy, something like `<Task :status="taskProxy.state" />`

src/components/cylc/table/TableItem.vue
```diff
@@ -0,0 +1,380 @@
+<!--
```
kinow, line 1: This file is not used and can be deleted.

src/components/cylc/table/table-nodes.js
```diff
@@ -0,0 +1,366 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import TaskOutput from '@/model/TaskOutput'
+import Vue from 'vue'
```
kinow, line 19: This file can be removed.

src/views/Table.vue
```diff
@@ -0,0 +1,187 @@
...
+<template>
+  <div>
+    <CylcObjectMenu />
+    <div class="c-table">
+      <table-component
+        :tasks="tasks"
+        :hoverable="false"
+        :activable="false"
+        :multiple-active="false"
+        :min-depth="1"
```
kinow, line 27: We can remove `min-depth`, `multiple-active`, `activable`, `hoverable`. Leaving just the `tasks` prop.

src/views/Table.vue
```diff
@@ -0,0 +1,187 @@
...
+/**
+* @param {DeltasAdded} data
+* @param {Array} array
+*/
+const applyTableDeltas = (data, array) => {
+  const added = data.added
+  const pruned = data.pruned
+  const updated = data.updated
+  if (added) {
+    if (added.taskProxies) {
```
kinow, line 54: These `if`'s could be combined. Not important, but you can have `if (added && added.taskProxies)` instead of 2 `if`'s. Same for other two `if`'s below.

src/views/Table.vue
```diff
@@ -0,0 +1,187 @@
...
+
+  props: {
+    workflowName: {
+      type: String,
+      required: true
+    }
+  },
+
+  components: {
+    CylcObjectMenu,
```
kinow, line 110: Nice, this is the component used for the mutations menu. As it's already imported here, I think we have 50% done. Just need to use the directive and confirm it works with the table too.

src/views/Table.vue
```diff
@@ -0,0 +1,187 @@
...
+  },
+
+  data: () => ({
+    /**
+     * This is the CylcTable, which contains the hierarchical table data structure.
+     * It is created from the GraphQL data, with the only difference that this one
+     * contains hierarchy, while the GraphQL is flat-ish.
+     * @type {Array}
+     */
+    tasks: []
```
kinow, line 127: The comment of this data property is incorrect now. Either remove or edit it :point_up:

src/views/Table.vue
```diff
@@ -0,0 +1,187 @@
...
+
+  /**
+   * Called when the user enters the view. This is executed before the component is fully
+   * created. So there is no direct access to things like `.data` or `.computed` properties.
+   * The component also hasn't been bound to the DOM (i.e. before `mounted()`).
+   *
+   * Here is where we create the subscription that populates the `CylcTable` object, and
+   * also applies the deltas whenever data is received from the backend.
+   */
+  beforeRouteEnter (to, from, next) {
```
kinow, line 138: Comment above needs amending too.

src/mixins/tableview.js
```diff
@@ -0,0 +1,69 @@
...
+        workflowId: this.workflowId
+      }
+    },
+    /**
+     * Return the workflow hierarchical table.
+     * @return {Array} workflow table
+     */
+    workflowTable () {
+      return this.table ? this.table.root.children : []
+    }
```
kinow, line 56: This `computed` variable is never used, and can be removed.

src/views/Table.vue
```diff
@@ -0,0 +1,187 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+-->
+
+<template>
+  <div>
+    <CylcObjectMenu />
```
kinow, line 20: This is the mutations menu. As it's already in the Table view, we should — in theory — have the menu working with the directive :crossed_fingers: 

Will send the diff later after some of these comments are reviewed/resolved.

kinow:
@giulianoserrao can you apply this diff to get the offline mode working, please?

```diff
diff --git a/src/services/mock/workflow.service.mock.js b/src/services/mock/workflow.service.mock.js
index de00c2cc..d2910850 100644
--- a/src/services/mock/workflow.service.mock.js
+++ b/src/services/mock/workflow.service.mock.js
@@ -201,6 +201,7 @@ class MockWorkflowService extends GQuery {
   recompute () {}
 
   startDeltasSubscription (query, variables, subscriptionOptions) {
+    console.log(query.loc.source.body)
     if (!query.loc.source.body.includes('deltas')) {
       this.query = query
     }
@@ -217,7 +218,8 @@ class MockWorkflowService extends GQuery {
       data: {
         deltas: {
           added: {
-            workflow: checkpoint.workflows[0]
+            workflow: checkpoint.workflows[0],
+            taskProxies: checkpoint.workflows[0].taskProxies
           }
         },
         workflows: checkpoint.workflows
```

src/components/cylc/table/Table.vue
giulianoserrao: `Job` component imported and declared.
Icons not yet displaying as will do this later.

src/views/Table.vue
giulianoserrao: My file looks quite different:
```
      <table-component
        :tasks="tasks"
       props :min-depth="1"
        ref="table0"
        key="table0"
      ></table-component>
      ```
      There's no `hoverable` or similar props there

File: src/views/Table.vue (hunk 12 above)
Comment (giulianoserrao) [reply src/views/Table.vue:54]: Great suggestion, done!

File: src/views/Table.vue (hunk 11 above)
Comment (kinow) [reply src/views/Table.vue:27]: In which case I think your local branch is ahead or behind this remote branch? 😕

**Comment by datamel** [comment]
@giulianoserrao, my vue is not really up to scratch enough to review, just just to say looks like a substantial piece of work, congrats!

File: src/views/Table.vue (hunk 11 above)
Comment (hjoliver) [reply src/views/Table.vue:27]: (Or you had a different branch checked out?  What this PR shows *is* your tableview branch (in your GitHub fork of cylc-ui) so your file can't look different, so long as you have the right branch checked out and haven't made any local changes since pushing it to GitHub).

Comment (giulianoserrao) [reply src/views/Table.vue:27]: Hmm quite puzzled by this one to be honest. I haven't really changed anything locally and this issue was already happeni g on Friday...

**Comment by giulianoserrao** [comment]
> @giulianoserrao, my vue is not really up to scratch enough to review, just just to say looks like a substantial piece of work, congrats!

Thanks @datamel - it's been really challenging, yet fun. I must say, without Bruno and Hilary I would NOT have gone this far! :)

File: src/components/cylc/table/Table.vue (hunk 8 above)
Comment (giulianoserrao) [reply src/components/cylc/table/Table.vue:156]: @kinow | re: icons,  Maybe I have this wrong, but isn't the `state` of a task, whether it is `running`, `waiting`, `preparing` etc?
I tried both `state` and `status` and at best I can get the state of a task to be replicated in the task name column. Looked at the tree component and the only icons I could find where associated with `+` and `-`, presumably having to do with the filtering section up the top. Where am I going wrong?

Comment (kinow) [reply src/components/cylc/table/Table.vue:156]: >@kinow | re: icons, Maybe I have this wrong, but isn't the state of a task, whether it is running, waiting, preparing etc?

That's correct. If you have a `task` object, retrieved from our GraphQL API, then you will have a `task.state` with the state as you described above.

>I tried both state and status and at best I can get the state of a task to be replicated in the task name column. Looked at the tree component and the only icons I could find where associated with + and -, presumably having to do with the filtering section up the top. Where am I going wrong?

Here I'm talking about the [`prop`](https://github.com/cylc/cylc-ui/blob/2042834d65f00ca9a77db1836066c56397221d1d/src/components/cylc/Task.vue#L56) that `Task` and `Job` expect you to pass. Unfortunately it's not very consistent at the moment.

We have `state` in most places, but we still have `status` in other places (to be fixed soon-ish). So in this case, you should add to the Table the Task and the Job icons. Something like `<Task :status="taskObject.state">`.

Does this make sense?

File: src/views/Table.vue (hunk 11 above)
Comment (kinow) [reply src/views/Table.vue:27]: It'll be quite hard to diagnose this one without looking at your working copy. We can have a video chat tomorrow to see if we can figure out what's going on.

Otherwise, I think `git fetch --all` and then `git status` should display if your branch is up to date or not to the remote branch (you can also set the remote branch as `git branch --set-upstream-to=upstream/master` or something like that).

**Comment by kinow** [comment]
I think this branch has something that is causing the Tree view and deltas to fail.

1. start Cylc UI on this branch with build:watch
2. start UIS & a workflow
3. visit Cylc UI
4. click on the workflow name in GScan
5. close the tree view that is created by default

image

**Comment by kinow** [comment]
I think the latest merge/push-force broke the table view in the workflow view.

image

File: src/views/Workflow.vue [hunk 18]
```diff
@@ -54,14 +66,17 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 <script>
 import { mixin } from '@/mixins'
 import { datatree } from '@/mixins/treeview'
+import { datatable } from '@/mixins/tableview'
```
kinow, line 69: This can be removed.

src/views/Table.vue
```diff
@@ -0,0 +1,171 @@
...
+import { datatable } from '@/mixins/tableview'
+import TableComponent from '@/components/cylc/table/Table'
+import { WORKFLOW_TABLE_DELTAS_SUBSCRIPTION } from '@/graphql/queries'
+import Alert from '@/model/Alert.model'
+import CylcObjectMenu from '@/components/cylc/cylcObject/Menu'
+import { mergeWith } from 'lodash'
+import Vue from 'vue'
+
+/**
+* @param {DeltasAdded} data
```
kinow, line 43: Correct type is `Deltas`. That fixes the issue in the IDE where it shows that `.added`, `.pruned`, and `.updated` do not exist in `data`.

image

src/views/Workflow.vue
```diff
@@ -39,6 +39,18 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+            :key="widgetId"
+            :id="widgetId"
+            :loading="isLoading"
+            type="list-item-three-line"
+            tab-title="table"
+        >
+         <table-component
+             :tasks="table"
+         />
+        </v-skeleton-loader>
```
kinow, line 53: Can you re-format the code above to use 2 spaces instead of 4 like the rest of the code, please?

src/views/Workflow.vue
```diff
@@ -136,9 +161,9 @@ export default {
     this.isLoading = true
     // clear the tree with current workflow data
     this.tree.clear()
+    this.table.clear()
```
kinow, line 164: `this.table` does not have the `.clear()` function, since `this.table` is an [Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array).

Probably this can be used instead: `this.table.length = 0` (empties an array in JS... or `this.table = []`, thought would need to confirm Vue keeps reactivity that way too)

src/views/Workflow.vue
```diff
@@ -152,6 +177,7 @@ export default {
   beforeRouteLeave (to, from, next) {
     this.$workflowService.stopDeltasSubscription()
     this.tree.clear()
+    this.table.clear()
```
kinow, line 180: See comment above.

src/views/Workflow.vue
```diff
@@ -54,14 +66,17 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 <script>
 import { mixin } from '@/mixins'
 import { datatree } from '@/mixins/treeview'
+import { datatable } from '@/mixins/tableview'
 import { mapState } from 'vuex'
 import Lumino from '@/components/cylc/workflow/Lumino'
-import { WORKFLOW_TREE_DELTAS_SUBSCRIPTION } from '@/graphql/queries'
+import { WORKFLOW_TREE_DELTAS_SUBSCRIPTION, WORKFLOW_TABLE_DELTAS_SUBSCRIPTION } from '@/graphql/queries'
```
kinow, line 72: Both subscriptions are identical. We should be able to use just the tree one, without having to add a new subscription.

image

The only reason why that's not possible, at the moment, seems to be due to #661 

You don't have to worry about this one @giulianoserrao

src/views/Workflow.vue
```diff
@@ -175,9 +201,57 @@ export default {
...
+            next: function next (response) {
+              applyDeltas(response.data.deltas, vm.tree)
+              vm.isLoading = false
+            },
+            error: function error (err) {
+              vm.setAlert(new Alert(err.message, null, 'error'))
+              vm.isLoading = false
+            }
+          })
+      }
```
kinow, line 217: The code above appears to be creating a new delta subscription for the tree. Below is the code for the new delta subscription for the table. This is one of the bugs I mentioned in our Element chat today.

We already have a delta subscription for the tree. So no need to create it again.

src/views/Workflow.vue
```diff
@@ -175,9 +201,57 @@ export default {
...
+            next: function next (response) {
+              applyTableDeltas(response.data.deltas, vm.table)
+              vm.isLoading = false
+            },
+            error: function error (err) {
+              vm.setAlert(new Alert(err.message, null, 'error'))
+              vm.isLoading = false
+            }
+          })
+      }
```
kinow, line 251: The code above also has a bug. The `if` blocks are identical, probably happened due to copy-and-paste, or during git/rebase/merge/etc

src/views/Workflow.vue
```diff
@@ -227,10 +304,17 @@ export default {
         if (this.deltaSubscriptions.length === 0) {
           this.$workflowService.stopDeltasSubscription()
           this.tree.clear()
+        } else if (vm.deltaTableSubscriptions.includes(subscriptionId)) {
+          // if this is a table widget with a deltas subscription, then stop it if the last widget using it
+          vm.deltaSubscriptions.splice(this.deltaSubscriptions.indexOf(subscriptionId), 1)
+          if (this.deltaTableSubscriptions.length === 0) {
+            this.$workflowService.stopTableDeltasSubscription()
```
kinow, line 311: `.stopTableDeltasSubscription` does not exist yet. So when we we removed the table widget, the subscription never stops.

You can confirm that by looking at the WebSockets messages that never cease even when you don't have any widgets (note: subscription `id: 1` is the GScan component... in this case, `2` was the default tree, and now `3` contains the table widget that I added, and then removed... `2` received a stop and is over, while `3` never stopped).

src/views/Workflow.vue
```diff
@@ -227,10 +304,17 @@ export default {
...
+            this.table.clear()
+          }
+        }
+        if (Object.entries(this.widgets).length === 0) {
```
kinow, line 315: Not sure why this `if` had to be moved here? See rest of the `diff` below.

src/views/Workflow.vue
```diff
@@ -227,10 +304,17 @@ export default {
         if (this.deltaSubscriptions.length === 0) {
           this.$workflowService.stopDeltasSubscription()
           this.tree.clear()
+        } else if (vm.deltaTableSubscriptions.includes(subscriptionId)) {
```
kinow, line 307: Hmmm, this is never executed I'm afraid.

Here on GH UI, click on the button icon above to see a few more lines of the code above. You should see that this code is wrapped within a `if (vm.deltaSubscriptions.includes(subscriptionId)) {`. But if this `subscriptionId` exists in `deltaTableSubscriptions`…

kinow (changes requested):

src/components/cylc/table/Table.vue
giulianoserrao: Task import and icons now workingf

src/views/Table.vue
kinow: @giulianoserrao has already added the mutations menu :tada: marking as resolved.

src/views/Table.vue
giulianoserrao: Table Component no longer has the props mentioned above, but only has `tasks` prop so I believe this is also solved.

src/views/Table.vue
giulianoserrao: Mutations working now.

src/views/Table.vue
giulianoserrao: Directive now being used and working along with Mutations menu

giulianoserrao:
> @giulianoserrao can you apply this diff to get the offline mode working, please?
> 
> ```diff
> diff --git a/src/services/mock/workflow.service.mock.js b/src/services/mock/workflow.service.mock.js
> index de00c2cc..b4d05500 100644
> --- a/src/services/mock/workflow.service.mock.js
> +++ b/src/services/mock/workflow.service.mock.js
> @@ -217,7 +217,8 @@ class MockWorkflowService extends GQuery {
>        data: {
>          deltas: {
>            added: {
> -            workflow: checkpoint.workflows[0]
> +            workflow: checkpoint.workflows[0],
> +            taskProxies: checkpoint.workflows[0].taskProxies
>            }
>          },
>          workflows: checkpoint.workflows
> ```

Done

giulianoserrao:
> @giulianoserrao can you apply this diff to get the offline mode working, please?
> 
> ```diff
> diff --git a/src/services/mock/workflow.service.mock.js b/src/services/mock/workflow.service.mock.js
> index de00c2cc..b4d05500 100644
> --- a/src/services/mock/workflow.service.mock.js
> +++ b/src/services/mock/workflow.service.mock.js
> @@ -217,7 +217,8 @@ class MockWorkflowService extends GQuery {
>        data: {
>          deltas: {
>            added: {
> -            workflow: checkpoint.workflows[0]
> +            workflow: checkpoint.workflows[0],
> +            taskProxies: checkpoint.workflows[0].taskProxies
>            }
>          },
>          workflows: checkpoint.workflows
> ```

Done

src/views/Workflow.vue
giulianoserrao: Done

src/views/Table.vue
giulianoserrao: Vue imported.
Replaced `data` with `deltas` in the function args and `const`

src/views/Workflow.vue
giulianoserrao: Spaces removed.

src/views/Workflow.vue
giulianoserrao: I also had to remove `datatable` from `views/Workflow.vue` (line 88 - in `export default` as once I removed the `import` statement, `datatable` became undefined.

src/views/Workflow.vue
giulianoserrao: Line replaced with `this.table = []` and table appears to be functioning fine following the change.

src/views/Workflow.vue
giulianoserrao: Line replaced with this.table = [] and table appears to be functioning fine following the change.

src/views/Workflow.vue
giulianoserrao: Thank you - duplicate `if` code block removed.

kinow:
Pushed one commit to fix the Workflow view in the offline mode (mode which e2e tests run in GitHub actions). The e2e tests were failing as the workflow widgets were empty.

kinow:
Rebased fixing conflicts.

package.json
```diff
@@ -82,6 +82,7 @@
     "istanbul-instrumenter-loader": "^3.0.1",
     "istanbul-lib-coverage": "^3.0.0",
     "json-server": "^0.16.3",
+    "lodash.clonedeep": "^4.5.0",
```
kinow, line 85: Will see about removing this dependency, although I think I started doing it in a PR pending review...

src/components/cylc/table/Table.vue
```diff
@@ -0,0 +1,287 @@
...
+        this.activeFilters = cloneDeep(this.tasksFilter)
+      } else {
+        this.activeFilters = null
+      }
+    },
+    taskStartTime,
+    taskEstimatedDuration,
+    latestJob,
+    getTaskJobProps
+  }
```
kinow, line 285: We need unit tests for the methods above.

src/components/cylc/table/deltas.js
```diff
@@ -0,0 +1,83 @@
...
+    }
+  }
+  if (pruned && pruned.taskProxies) {
+    for (const taskProxyId of pruned.taskProxies) {
+      if (tasks[taskProxyId]) {
+        Vue.delete(tasks, taskProxyId)
+      }
+    }
+  }
+}
```
kinow, line 83: Missing tests.

src/graphql/queries.js
```diff
@@ -102,72 +183,15 @@ fragment WorkflowTreePrunedData on Pruned {
...
-  state
-  submitNum
-  taskProxy {
-    outputs (satisfied: true, sort: { keys: ["time"], reverse: true}) {
-      label
-      message
-    }
-  }
-}
+${JOB_DATA}
```
kinow, line 194: This same change exists in the GScan deltas PR.

src/styles/cylc/_table.scss
```diff
@@ -0,0 +1,22 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+.c-table {
+  .c-task, .c-job {
+    font-size: 1.2em;
```
kinow, line 20: Trying to match the font-size in the tree view.

src/views/Table.vue
```diff
@@ -0,0 +1,135 @@
...
+   * Called when the user leaves the view. We stop the active subscription and clear
+   * data structures used locally.
+   */
+  beforeRouteLeave (to, from, next) {
+    this.$workflowService.stopDeltasSubscription()
+    this.tasks = {}
+    next()
+  }
+}
+</script>
```
kinow, line 135: Need an e2e test for this view.

src/views/Workflow.vue
```diff
@@ -101,28 +121,39 @@ export default {
...
     isLoading: true,
     // the widgets added to the view
     /**
      * @type {
      *   Object.<string, string>
      * }
      */
-    widgets: {}
+    widgets: {},
+    deltasCallbacks: {}
```
kinow, line 133: Not happy about adding one more property here. I think @oliver-sanders ' initial design had a single subscriptions array/dict. I think that would be better than having `.widgets`, `subscriptions`, `deltasCallbacks`, etc. Will think about it over next days.

src/views/Workflow.vue
kinow: Actually might not be necessary at all.

kinow:
Will rebase it after finishing to update the GScan PR (using it to review @dwsutherland 's PR in UIS)

kinow:
Last commit makes the Table view to use the new mixins. Also creates a `table/deltas.js` to handle deltas as was done with GScan and Table components. It's functional, but needs more testing, and some review before it's ready to be reviewed by others. Moving it to draft for now.

kinow:
Build should pass now, so will have an idea of how this PR affects coverage, and use that to finish writing the tests later on.

cypress.json
```diff
@@ -1,7 +1,7 @@
 {
   "pluginsFile": "tests/e2e/plugins/index.js",
   "video": false,
-  "defaultCommandTimeout": 30000,
+  "defaultCommandTimeout": 10000,
```
kinow, line 4: Our tests appear to be more resilient, so changing it back to 10 seconds to improve build time.

src/components/cylc/table/deltas.js
kinow: Last ones now!

src/components/cylc/table/deltas.js
```diff
@@ -0,0 +1,196 @@
...
+  }
+  if (pruned.taskProxies) {
+    pruned.taskProxies.forEach(taskProxyId => {
+      Vue.delete(table, taskProxyId)
+    })
+  }
+  if (pruned.jobs) {
+    pruned.jobs.forEach(jobId => {
+      // TODO: should we use an internal lookup for table too? To replace this loop by a quick O(1) operation
+      //       to fetch the existing job with its ID, and then its existing parent task?
```
kinow, line 114: Looks like we will have to change the table component later, to support showing all the jobs (maybe collapsing too), sorting, and other improvements. So the structure used in Vuex may change... not sure whether we should use something like `table.lookup` as we are doing with the Tree view structure, to quickly access the job & task (NB: the global `lookup` doesn't have the job anymore as it has been pruned in another delta-callback already).

kinow:
There is one part of the code that I want to add tests, but for next week. It's the error handling in the Vuex actions. I'll write the tests for the Table view Vuex actions, and also for the Tree view (missing as well).

With that, this PR will be done, and ready for review or to be postponed after the next alpha release :tada:

src/components/cylc/table/deltas.js
oliver-sanders: > the global lookup doesn't have the job anymore as it has been pruned in another delta-callback already

Does that mean the global flat data store doesn't contain jobs?

> something like table.lookup

Could we run the table off of the regular flat or tree stores?

kinow: >Does that mean the global flat data store doesn't contain jobs?

Not anymore. It had the data, until the callback to apply the new `pruned` delta was applied to the global flat data.

I've been thinking about this issue, which happened with the tree view too.

Ideally I think we would want to

1. apply all the added & update in order (i.e. global, then each component like tree, then view, then graph, etc)
2. then apply the pruned deltas in reverse order (i.e. the components first, so that they can use the global lookup, and lastly the global callback to prune data)

That would solve the issue here I think (not sure about the tree view, since there we want to access the tree-node entry, not the global data). Will need to think more about this one.

>Could we run the table off of the regular flat or tree stores?

With some extra work, perhaps. The structure used for the table is

- task proxies
  * latest job (for now, @hjoliver I think mentioned that we want to display all jobs later, with some collapse/expand functionality)

The flat store has the data. But we would have to scan every entry whenever it's updated, and update a `computed` variable somewhere, with the structure for the component.

The tree store has the structure, but the task proxies are buried deeper in the tree, under cycle-points/family-proxies. So I would have to filter the tree store whenever it's updated, to re-build the data for the table (a `computed` property).

kinow (approved):
GH UI keeps showing me as reviewer, so approving it (even though I'm helping with the code :laughing: so don't consider my approval as one of the two required approvals :+1: )

src/components/cylc/table/deltas.js
```diff
@@ -0,0 +1,215 @@
...
+/**
+ * @param {GraphQLResponseData} data
+ * @param {Table} table
+ * @param {Lookup} lookup
+ */
+export default function (data, table, lookup) {
+  const deltas = data.deltas
+  // first we check whether it is a new start
+  if (deltas && deltas.added && deltas.added.workflow) {
+    if (deltas.added.workflow.status === WorkflowState.RUNNING.name) {
```
kinow, line 181: :exclamation:  Note to self: this  is wrong, but will update the PR only after we have decided which direction to go with the table view.

See https://github.com/cylc/cylc-ui/pull/729 and https://github.com/cylc/cylc-ui/issues/667 for more.

src/components/cylc/table/Table.vue
```diff
@@ -0,0 +1,271 @@
...
+          return true
+        })
+    }
+  },
+  methods: {
+    filterByTaskName () {
+      return this.activeFilters &&
+        this.activeFilters.name !== undefined &&
+        this.activeFilters.name !== null &&
+        this.activeFilters.name.trim() !== ''
```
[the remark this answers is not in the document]
kinow: Good spot! I've modified the template that was using `v-model="tasksFilter.name"` (in a `v-text-field`) to `v-model.trim="tasksFilter.name"` :+1:

kinow:
@MetRonnie I tried fixing the `.mailmap` file. It's now complaining of "Juliano Serrao", but if I run git shortlog locally:

```bash
git shortlog -s "348a05062534c16511b3c0e714e54a47fed08907"
     2  Alex Szabó
     2  Ash Newport
  1285  Bruno Kinoshita
     1  David Sutherland
     4  Edgardo Ramírez
     1  Garrett Walker
     4  Giuliano Serrao
   215  Hilary Oliver
    13  Martin Ryan
    11  Matt Shin
     9  Mel Hall
   201  Oliver Sanders
     6  Ronnie Dutta
     7  Sadie Bartholomew
    13  Tim Pillinger
     4  dependabot[bot]
     7  github-actions[bot]
```

That's the first part of the shell used in the Action I think. Any idea what I'm missing here? Thanks!

kinow:
@oliver-sanders I think your understanding of Vue's reactivity is correct! :tada: 

>    I think our flat data is stored as an Array?

Correct, for the global data that is shared by views: https://github.com/cylc/cylc-ui/blob/cda7551659d3672697062608f6617d88b4bcf67d/src/store/workflows.module.js#L51-L56

Some other objects may use Object too that are not reactive by default in Vue 2 (hence our `Vue.set` use in some parts of the code). But Vue 3 fixes it.

>    I think Vue is reactive for Array item changes - https://vuejs.org/v2/guide/reactivity.html#For-Arrays?

Correct.

Your example is also correct. I think my `TODO` was not clear here, sorry. Here's a screenshot from my IDE.

image

My TODO comment was towards the `Object.values(table).find`. For every job that was pruned, I'm scanning the complete table to locate the parent task.

In the Tree view, we use a lookup in a similar case, to avoid scanning the complete tree (or even its flat sibling structure).

kinow:
>I'm getting this error in my browser at the moment:
>
>TypeError: modules[moduleId] is undefined
>
>Are we missing a dependency somewhere.

Maybe it was a Yarn glitch, due to the hierarchical gscan PR? I checked out the branch, rebased and fixed conflicts. CI appears to be OK building & running the tests. I tried `yarn run serve` and it worked fine.

.mailmap
```diff
@@ -3,6 +3,9 @@ Bruno Kinoshita            <kinow@users.noreply.github.com> Bruno P. Kinoshita
 David Sutherland           <davidwollow@gmail.com>
 Edgardo Ramírez            <soldiercrp@gmail.com>
 Garrett Walker             <garbwalk@ut.utm.edu>
+Giuliano Serrao            <giulianoserrao@gmail.com>
+Giuliano Serrao            <giulianoserrao@gmail.com> giulianoserrao
+Giuliano Serrao            <giulianoserrao@gmail.com> Juliano Serrao
```
MetRonnie, line 8: > @MetRonnie I tried fixing the .mailmap file. It's now complaining of "Juliano Serrao", but if I run git shortlog locally

Strange, I also don't have the extra name showing up locally. Have you tried just having the one entry, as the email address is the same
```suggestion
Giuliano Serrao            <giulianoserrao@gmail.com>
```

oliver-sanders: BTW we don't need a `.mailmap` entry for each user, it is only necessary if the name in the contributor list doesn't match the committer's name.

hjoliver:
I'm not getting any dependency issues on this branch.

However the table and tree views don't seem to work together. If I open one (tree or table) it works, but if then open the other it stays blank.

kinow:
> 
> 
> I'm not getting any dependency issues on this branch.
> 
> However the table and tree views don't seem to work together. If I open one (tree or table) it works, but if then open the other it stays blank.

Huh, I probably broke something when rebasing it, or forgot to update the code after GScan hierarchical PR. Let me check...

.mailmap
kinow: @MetRonnie 

>Strange, I also don't have the extra name showing up locally. Have you tried just having the one entry, seeing as the email address is the same

Thanks for looking into this. I haven't, but it worked for others (Sadie and Hilary I think had different e-mails)

@oliver-sanders 

>BTW we don't need a .mailmap entry for each user, it is only necessary if the name in the contributor list doesn't match the committer's name.

Thanks, I will try that too, but I'm terrible with this mailmap syntax. I tried having less lines, but I was probably doing something wrong since the GH action kept complaining about duplicate users. I found one combination that worked but didn't confirm if it was the best combination 😅

kinow:
From Element chat with @hjoliver this morning:

>Ideally, build the new view from the data store if it does not trigger a subscription restart.

There's no initial data burst when merging table + tree queries at the moment. Spent some time debugging the code this morning, more notes to self for when working on this:

- [ ] are you sure removing the last widget of a type is recomputing the query and removing the excess (if any) fields?
- [ ] could we use the component.$options.name instead of the ._uid? Any reason for keeping each instance of each widget, instead of only one subscription per widget-type (i.e. one for tree, one for table, one for mutations, etc)?
- [ ] there is a wrong JSdoc in workflow.services.js, see typing error in the `variables` field; create a typedef for that object, or use `Object.<String, String>`
- [ ] you had an error for parent-less task, missing a family I believe, when adding/removing/adding/removing widgets successively in short intervals (if confirmed, ping @dwsutherland and see if this could be an issue or just need to ignore parent-less nodes - we had that issue some time ago).

oliver-sanders:
> @oliver-sanders I think your understanding of Vue's reactivity is correct! :tada:

:sweat: :tada: 

> But I didn't want to introduce yet-another map for this case, since I also suspected we could change the table structure (to support multiple jobs, or to have more data, or hierarchy, etc). Does it make sense?

I think searching through the job nodes to find the most recent job should be fine since:

* We probably want to show multiple jobs anyway.
  * Even if we don't go for full multi-job expansion right away we can still use the multi-job icon to make users aware there were multiple job submissions.
* We wouldn't expect many job submissions under normal conditions.

kinow:
Besides the linked issue, and comments above, this docs might be helpful to the assignee of this issue: https://github.com/cylc/cylc-ui/wiki/Architecture#ui-views :+1:

kinow:
@AaronDCole I will try to rebase and fix conflicts next week if I have time. Then, I think we could modify `workflow.service.js` to merge callbacks, even when queries are identical. I guess that would solve the issue of having to reinitialize the subscription, without having to dig too deep into how the components are being created & using data from the store (related to #803 & to what we discussed today)

kinow:
Not sure if the tests will pass. But the offline mode is working.

I have:

- rebased, and fixed conflicts
- updated the code to use callbacks instead of actions
- fixed an issue found when merging queries (tree sorts tasks by cycle points, this was sorting by name; not sure how that would result when merged, current code throws an error, so now both are sorting by cycle points [table can re-sort if necessary by name])

The offline mode works fine, the `/#/table/one` also works OK. What's pending now is:

- Figure out how to have both tree + table in the workflow #803 
- Update tests & build :grimacing: 

:vulcan_salute:

hjoliver:
Nice job @AaronDCole 

Only thing I've noticed so far, is the table view task icons aren't showing their "modifier" badges for "queued", "held", and "runahead limited":

shot

See Cylc UI Dashboard -> Quickstart

Here's my test workflow:
```
[task parameters]
    m = 0..5
[scheduling]
    runahead limit = P1
    cycling mode = integer
    initial cycle point = 1
    [[queues]]
        [[[default]]]
             limit = 1
    [[graph]]
        P1 = "foo<m>"
[runtime]
    [[root]]
        script = sleep 20
    [[foo<m>]]
```

- `runahead limit = P1` causes the next-cycle tasks to be held back even though they are ready to run (dot badge)
- `[queues][default] limit = 1` means only one task can be active at a time, others that are ready will be "queued" (burger badge)
- click on a task in the UI and select "Hold" to hold (pause) the task shown for `foo_m5.1` in the screen shot (pause badge) ... or run `cylc hold <workflow-name> foo_m5.1`

AaronDCole:
> Nice job @AaronDCole
> 
> Only thing I've noticed so far, is the table view task icons aren't showing their "modifier" badges for "queued", "held", and "runahead limited":
> 
> shot
> 
> See Cylc UI Dashboard -> Quickstart
> 
> Here's my test workflow:
> 
> ```
> [task parameters]
>     m = 0..5
> [scheduling]
>     runahead limit = P1
>     cycling mode = integer
>     initial cycle point = 1
>     [[queues]]
>         [[[default]]]
>              limit = 1
>     [[graph]]
>         P1 = "foo<m>"
> [runtime]
>     [[foo<m>]]
>         script = sleep 20
> ```
> 
>  * `runahead limit = P1` causes the next-cycle tasks to be held back even though they are ready to run (dot badge)
> * `[queues][default] limit = 1` means only one task can be active at a time, others that are ready will be "queued" (burger badge)
> * click on a task in the UI and select "Hold" to hold (pause) the task shown for `foo_m5.1` in the screen shot (pause badge) ... or run `cylc hold <workflow-name> foo_m5.1`

Cheers, Im investigating today, no double something to do with the way row nodes are passed to the icon renderers.

hjoliver:
@AaronDCole - confirmed the task icon modifier badges are working now in the table view, but something has gone wrong with the initial rendering of a new view.  If I close the lumino tree tab, then the open a new tab (tree or table) it works.  But if I open a second tab of the other kind (whether tree it table) it stays blank.

hjoliver:
Also @AaronDCole - since @giulianoserrao is no longer working on the project, it might be wise to close this PR after pushing the branch to your own cylc-ui fork, then open a new PR from there. And that would also get rid of the above clutter from all the early work.  

(@giulianoserrao - if you are watching, this move will preserve the identity of your commits).

giulianoserrao:
Thanks Hilary,
Much appreciated 😊
I have been following the table view's progress. Sounds like it's getting
better and better!

hjoliver (approved):
Had another good play with this, tests now passing.  All good :tada:  Thanks @AaronDCole for finishing the branch off (esp.: views driven by shared datastore only). As discussed at last project meeting, merging this as-is; further changes to the tableview can be done as follow-ups.

PR #676, opened 2021-05-16
Update the cycle point icon state

### LINKED ISSUE #666: tree: inconsistent cycle point / family / task state
In this screenshot the group status for the cycle point is not reflecting the status of the tasks within it.

Screenshot 2021-05-06 at 10 03 21

* When I refresh the page the cycle point status is corrected.
* Cylc Tui displays it correctly (note: at the moment Tui uses a global update, not deltas)
* cylc-flow@f5802012d1bdd03af52afae205fed2a1b19f8597 (master)

The problem only occurs as the workflow state evolves (i.e. it does not occur when the page first loads) which suggests a delta has been missed or not applied.

kinow:
These changes close #666

@oliver-sanders could you test and see if this solves the issue for your workflow? I tested on `five`, and noticed that when the job started running, the cycle point icon never changed. Then, after this change, it was correctly updated. Thanks!

**Requirements check-list**

### DIFF
--- a/src/components/cylc/tree/cylc-tree.js
+++ b/src/components/cylc/tree/cylc-tree.js
@@ -67,7 +67,8 @@ function computeCyclePointsStates (cyclePointNodes) {
     for (const child of cyclePointNode.children) {
       childStates.push(child.node.state)
     }
-    cyclePointNode.node.state = extractGroupState(childStates, false)
+    const cyclePointState = extractGroupState(childStates, false)
+    Vue.set(cyclePointNode.node, 'state', cyclePointState)
   }
 }
 
--- a/src/components/cylc/tree/cylc-tree.js
+++ b/src/components/cylc/tree/cylc-tree.js
@@ -68,6 +68,8 @@ function computeCyclePointsStates (cyclePointNodes) {
       childStates.push(child.node.state)
     }
     const cyclePointState = extractGroupState(childStates, false)
+    // Initially the .node object retrieved from the GraphQL endpoint does
+    // not have the .state property. So we need to ask Vue to make it reactive.
     Vue.set(cyclePointNode.node, 'state', cyclePointState)
   }
 }
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,11 @@ a shadow underneath, for the Job component.
 sorted array, instead adding a new element in its sorted index (respecting
 locale, and numeric on).
 
+[#676](https://github.com/cylc/cylc-ui/pull/676) - Fix the reactivity of the
+cycle point icon state. After this, the Task icon used to summarize the cycle
+point state should be updated to reflect the overall state based on the states
+of its children nodes.
+
 -------------------------------------------------------------------------------
 ## __cylc-ui-0.4 (Released 2021-04-16)__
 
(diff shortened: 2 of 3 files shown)


kinow:
Looking how hard it would be to write a test for this :+1: it should be covered by existing tests, but not necessarily verifying whether the icon is updated when a state changes after the page & component have been loaded. Right now I'm thinking about simply `setTimeout` to change a state in some JS object after ~5 secs, and then mocha or Cypress asserting it changed in the DOM element as well (`nextTick`).

src/components/cylc/tree/cylc-tree.js
```diff
@@ -67,7 +67,8 @@ function computeCyclePointsStates (cyclePointNodes) {
     for (const child of cyclePointNode.children) {
       childStates.push(child.node.state)
     }
-    cyclePointNode.node.state = extractGroupState(childStates, false)
+    const cyclePointState = extractGroupState(childStates, false)
+    Vue.set(cyclePointNode.node, 'state', cyclePointState)
```
hjoliver, line 71: (I had been wondering about `Vue.set`; nice doc here: https://vuejs.org/v2/api/#Vue-set )

oliver-sanders:
Yes this fixes it :+1:.

kinow:
Thanks for testing it! I will try to come up with a test to prevent it from happening again (in Vue 3 you just create an object like `const cyclePoints = ref(someObject)` or `... = reactive(someObject)` and in theory reactivity just-works™).

oliver-sanders:
Why was this particular property non-reactive?

kinow:
> Why was this particular property non-reactive?

Ah, good question! The answer is in the docs @hjoliver linked

>Adds a property to a reactive object, ensuring the new property is also reactive, so triggers view updates. This must be used to add new properties to reactive objects, as Vue cannot detect normal property additions (e.g. this.myObject.newProperty = 'hi').

We have a `cyclePointNode.node` object. The `.node` part, if I recall correctly, contains the GraphQL object as it was returned from the GraphQL endpoint. That object is reactive I think, but the Cycle Point (a [family proxy](https://github.com/cylc/cylc-ui/blob/2042834d65f00ca9a77db1836066c56397221d1d/src/graphql/queries.js#L56-L58) actually in our query) does not contain the `.state` property.

The old code was simply adding `.state = 'bla'` in the reactive object.

But as the Vue docs text above states, that does not mean that that attribute is reactive; so when used in templates, only the initial value is displayed. Updates to the value do not trigger an update to the template.

kinow:
Added a comment to the code quickly explaining why we have `Vue.set`. Added a unit test which verifies that the JS object is an observable (by confirming it has a JS [getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) which is defined by Vue), and added a changelog entry.

I started writing an e2e test… but wasn't sure if that was the best way to test such a small part of the code. Reading the Vue docs and some of its code, I realized it would be simpler to test that the `node.state` was reactive, and not a simple JS property.

This means we can have a small and simple unit test, instead of an e2e test (e2e tests are harder to maintain and debug, but more useful when you want to test DOM, a real page, site navigation, etc).

Ready for review!

oliver-sanders (approved):

hjoliver (approved):
:+1:

PR #677, opened 2021-05-18
Simplify how we create URL's in the UI

### LINKED ISSUE #675: Review how we concatenate strings to create URLs

Ref https://github.com/cylc/cylc-ui/pull/627#discussion_r632532438

We have URLs in Cylc UI that are created dynamically, like the JupyterHub user service URL used to start a WebSocket connection.

In some cases, the URL is getting an extra `/`. Libraries and browsers are able to handle the double slashes, but we should fix it in case it causes troubles with other libraries or in our code.

Ideally a library or existing code to join it. The simpler the merrier.

**Additional context**

- [x] Does it work when deploying the UI/UIS to a path? https://github.com/cylc/cylc-ui/issues/258
- [x] Do we need the Python provided BASE URL? https://github.com/cylc/cylc-ui/issues/549
  * I don't think we need the BASE URL being passed from Python anymore, and it should be OK to trust that the URL used in the browser will be the same one used to access GraphQL via Apollo Client and Hub Via Axios

**Pull requests welcome!**

kinow:
Closes #675 

Remove existing code that relied on the Python backend code to provide
the Base URL. Now this value is always whatever the browser URL contains.

There is now a single createUrl function that is used in the application.
This removes the need to use Vuex.

**Requirements check-list**


kinow:
Just need some good testing before it can be marked as ready for review :+1:

kinow:
Tested with the UI+UIS running locally, all good. Then enabled the path `c.JupyterHub.bind_url = 'http://127.0.0.1:8000/cylc'` and tested again (had to log-in again) and everything worked fine too.

kinow:
The Base URL provided by Python doesn't seem to be needed. That was my first fix for an issue that was reported. But in hindsight I think it was never necessary, and simply using the URL from the browser would be sufficient to build the GraphQL & Hub API URL's.

kinow:
If approved & merged, I will raise an issue in UIS to [remove the base URL](https://github.com/cylc/cylc-uiserver/blob/4a32321f8199ba5d9d5102fccc91a8b7664ed432/cylc/uiserver/handlers.py#L114-L115) from the rendered template :+1:

hjoliver (approved):
Looks fine to me :+1:

kinow:
Rebased.

src/utils/urls.js
```diff
@@ -16,34 +16,54 @@
...
-  const baseUrl = `${window.location.protocol}//${window.location.hostname}${port}${window.location.pathname}`
-  if (baseUrl.endsWith('//')) {
-    return baseUrl.substring(0, baseUrl.length - 1)
-  } else if (baseUrl.endsWith('/')) {
-    return baseUrl
-  }
-  return `${baseUrl}/`
+function unslash (url) {
+  return url.replace(/([^:]\/)\/+/g, '$1')
 }
```
oliver-sanders, line 28: I'm not sure what this is doing?

```node
> function unslash (url) {
...    return url.replace(/([^:]\/)\/+/g, '$1')
... }
undefined
> unslash('http://abc.def/g/h/ijk/l')
'http://abc.def/g/h/ijk/l'
```

kinow: Ah, sorry, missing docs. Will add some tomorrow, but happy to rename it too. Wrote the function name without thinking too much about it, sorry.

```js
unslash('http://abc.def///g////h/ijk/l')
"http://abc.def/g/h/ijk/l"
```

oliver-sanders: Ah, thanks, that makes sense, could call it `noramlise` or something like that.

oliver-sanders (approved):
Thanks, much simpler.

PR #543, opened 2020-11-29
Use deltas in GScan

kinow:
These changes close a lot of issues, listed below. Do not worry about assigning to me, setting milestone, etc. After the final review, and when this issue is closed, then I will go through each issue linked below, will assign and set the milestone :+1: 

- close #542 (main issue)
- close #234
  * Now there is a JS object with the variables in `SubscriptionQuery` class, that gets filled from the VueRouter + VuexStore info via the `graphql.js` mixin :+1: 
  * **code review**
- close #209 
  * Now we declare the query and callbacks. Previously, there was some code in WorkflowService & GQuery that was calling setWorkflows(data.workflow), but now the callbacks receive the data as is, and can do whatever they need with it, be it nodes, workflows, deltas, etc.
  * **code review**
- close #160
  * Achieved that by using Mixins. Should be easy to port to Vue 3 too (thought about that when trying a few different approaches).
  * **code review**
- close #356 
  * **code review**
- partially address #376 (@oliver-sanders I've got an idea to improve it)
  * We are now adding views directly to the widgets. So any view (with the right mixins) can be loaded in the Workflow view & component.
  * **code review**
- close #388 
  * By coordinating in mixins when a subscription is merged/created, and only starting it in the VueRouter navigation guard (`subscriptionView` mixin) we guarantee that we don't start subscriptions that are immediately going to be stopped.
  * **review with browser network & other developer tools**
- close #416
  * We are not mixing the GScan and Tree queries any longer, they are kept as separated GraphQL web sockets subscriptions.
  * **code review or use browser network & other developer tools**
- close #453 
  * e2e tests added, also simplified with the mixins.
  * **code review**
- close #661 
  * The workflows data (GScan, Dashboard, WorkflowsTable) is in Vuex store's `workflows.workflows`. The workflow data is stored in `workflows.lookup` (workflow subscription lookup with `id` as key, and GraphQL data as value), and Tree view data is stored in `workflows.workflow`. The Tree view is re-using the data in `workflows.lookup` when it creates `workflows.workflow`, avoiding data duplication.
  * **code review and/or vue-dev-utils browser extension, inspecting the Vuex store when browsing different views**
- close #684 
  * See above.
- close #625
  * Noticed that after dismissing the alert, I was not getting notified about the new alerts (that I knew were supposed to happen). Remembered about this issue, and added a call to `setAlert(null)` when the user dismisses the alert. Note: if the user does not dismiss the alert, once a new alert appears, it will replace the existing alert :+1: 
  * **code review**
[...]

PR #680, opened 2021-05-25
Fix GraphiQL URL function, use json-server instead of MockWorkflowService, and other minor fixes

kinow:
Sorry, I think I broke GraphiQL. Trying to fix it here, while also adding a test.

Removes the `MockWorkflowService` class that was used in the offline mode. Now, instead, we reply with `json-server` for GraphQL requests sent to the `/graphql` endpoint, and to the `/subscriptions` (this one via WebSockets).

Fixes other build warnings and minor issues.

**Requirements check-list**


src/components/cylc/tree/Tree.vue
```diff
@@ -160,7 +160,7 @@ import TreeItem from '@/components/cylc/tree/TreeItem'
 import Vue from 'vue'
 import TaskState from '@/model/TaskState.model'
 import Task from '@/components/cylc/Task'
-import clonedeep from 'lodash.clonedeep'
+import cloneDeep from 'lodash/cloneDeep'
```
kinow, line 163: I think this is fixed in the table view PR, and probably also in another PR. Whoever gets in first wins, and the other PRs can then be updated :smile:

src/graphql/graphiql.js
```diff
@@ -18,7 +18,7 @@
 // Code related to GraphiQL
 
 import { parse } from 'graphql'
-import { createGraphQLUrls } from '@/utils/urls'
+import { createGraphQLUrls } from '@/graphql/index'
```
kinow, line 21: Here's the bug :bug:

kinow:
Managed to get `json-server` serving GraphQL requests today. That would remove the need to extend `WorkflowService`, and instead return GraphQL responses over HTTP. Only issue — that I didn't anticipate — is that `json-server` has no WebSockets :grimacing: looking into that now.

kinow:
If CI tests pass, final step is to — finally — add a test so I don't break GraphiQL again :laughing:

Now with `json-server` serving real data via HTTP, we should be able to run GraphiQL in the offline mode :tada:

package.json
```diff
@@ -15,7 +15,7 @@
     "lint": "vue-cli-service lint",
     "serve": "concurrently --names 'SERVER,CLIENT' --prefix-colors 'yellow,blue' --success \"first\" --kill-others 'yarn run serve:jupyterhub' 'yarn run serve:vue'",
     "serve:vue": "vue-cli-service serve --mode offline",
-    "serve:jupyterhub": "json-server --port 3000 ./src/services/mock/db.json",
+    "serve:jupyterhub": "nodemon src/services/mock/json-server.js",
```
kinow, line 18: We are not running `json-server` any longer, but instead running `nodemon` to monitor for changes, and starting `json-server` programmatically. It's the documented way in `json-server` when you want to have dynamic responses — as is our case with the `/graphql` and the `/subscriptions` endpoints, which may return Tree data, or GScan data, or Dashboard, etc.

package.json
```diff
@@ -79,9 +79,11 @@
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-standard": "^5.0.0",
     "eslint-plugin-vue": "^7.9.0",
+    "express-ws": "^4.0.0",
```
kinow, line 82: `json-server` uses Express, but for WebSockets we need `express-ws` as the built-in `ws` module doesn't seem to work well with Express :man_shrugging: https://stackoverflow.com/a/32705838

package.json
```diff
@@ -79,9 +79,11 @@
...
     "istanbul-instrumenter-loader": "^3.0.1",
     "istanbul-lib-coverage": "^3.0.0",
     "json-server": "^0.16.3",
+    "nodemon": "^2.0.7",
```
kinow, line 86: The alternative to `nodemon` is to write something like

```js
const fs = require('fs')

const someFile = __filename
fs.watch(someFile, () => { ... })
```

Which gets trickier when you want to monitor multiple files (e.g. `index.json` for `json-server`, plus any of the `JSON` files, or even Cylc UI JS code...)

src/graphql/queries.js
```diff
@@ -216,15 +216,15 @@ subscription {
...
-    }
+subscription WorkflowsTableQuery {
+  workflows (ignoreInterval: 0) {
+    id
+    name
+    owner
+    host
+    port
   }
+}
```
kinow, line 227: Fixed indentation here, and added a "name", or "operation name" to each of our queries, so that it's easier to identify the queries in the `json-server` handler.

src/services/mock/generate
```diff
@@ -20,7 +20,7 @@ if ! sed --version 2>/dev/null | grep -q 'GNU' 2>/dev/null; then
 fi
 
 cd "$(dirname "$0")" || exit 1
-CHECK_FILE='checkpoint.js'
+CHECK_FILE='json/OnWorkflowTreeDeltasData.json'
```
kinow, line 23: Didn't test, but now it should be simpler. Just dump whatever is returned from GraphQL into a JSON file, and let `json-server` serve it. We could even have more requests, use a URL with params for tests like `/graphql?throwException` or `/graphql?returnInvaildDelta`, to return a different JSON.

src/services/mock/graphql.js
```diff
@@ -0,0 +1,94 @@
...
+  OnWorkflowTreeDeltasData: {
+    deltas: {
+      added: {
+        workflow: data.OnWorkflowTreeDeltasData.workflows[0]
+      }
+    }
+  },
+  DashboardSubscriptionQuery: data.DashboardSubscriptionQuery,
+  GscanSubscriptionQuery: data.GscanSubscriptionQuery,
+  WorkflowsTableQuery: data.WorkflowsTableQuery
```
kinow, line 38: The key/left is what should be in the client GraphQL query (why I changed `queries.js`). Right is the JSON object returned.

src/services/mock/graphql.js
```diff
@@ -0,0 +1,94 @@
...
+ * @param {string} query - GraphQL query
+ * @returns {string} operation name, or empty string if not found
+ */
+function getOperationName (query) {
+  const regexp = /\s*(subscription|query|mutation)\s+([a-zA-Z0-9_]+)/
+  const match = regexp.exec(query)
+  if (match && match.length >= 3) {
+    return match[2]
+  }
+  return ''
```
kinow, line 61: Sometimes we have the `req.body.operationName`. I think ApolloClient adds it for Query requests. But GraphiQL doesn't, and WebSockets also don't include the operation name, so we have to search for it in the text query.

src/services/mock/json-server.js
```diff
@@ -0,0 +1,70 @@
...
+      responseData = graphql.handleGraphQLRequest(req)
+    }
+  }
+  // NB: json-server returns 404 for requests that are not in db.json, but
+  //     we have dynamic requests for the /graphql endpoint (i.e. no single
+  //     data response) hence the status(200) below.
+  res.status(200).jsonp(responseData)
+}
+
+server.use(router)
```
kinow, line 67: Now we are using `server` and `router`, objects provided by `json-server`, but that are actually part of [ExpressJS](http://expressjs.com/). We can add middlewares too, so we can have the server returning an HTTP header like `X-Token`, or rewriting URL's, or marking a request as not authorized, etc. Just for for thought for later.

src/services/mock/json/GscanSubscriptionQuery.json
```diff
@@ -0,0 +1,27 @@
...
+        "succeeded": 0
+      },
+      "latestStateTasks": {
+        "running": [
+          "running-task"
+        ]
+      }
+    }
+  ]
+}
```
kinow, line 27: To change with #543 , where I will have to return the JSON as an initial burst delta (as we are doing here for the tree data).

tests/e2e/specs/aotf.js
```diff
@@ -19,22 +19,110 @@ import {
...
+        name: 'job',
+        type: {
+          name: 'JobID',
+          kind: null
+        }
+      }
+    ]
+  }
+]
+
```
kinow, line 110: This was in the `MockWorkflowService`, but was used only for this e2e test. We can perhaps make it permanent in the offline mode, so that we can play with mutations. However, it would require more changes (that I started doing and gave up as it would bloat this PR) to the WebSockets code added here first.

tests/e2e/specs/aotf.js
```diff
@@ -56,10 +144,28 @@ function mockWorkflowService () {
...
+          __schema: {
+            mutationType: {
+              fields: MUTATIONS
+            },
+            types: TYPES
+          }
+        }
+      })
+      .as('IntrospectQuery')
+  })
```
kinow, line 160: Cypress allows us to intercept any HTTP requests in the e2e tests. This way, we can avoid mocking the `WorkflowService`, and instead just pretend the server returned our mutations instead of the default ones.

tests/e2e/specs/mutation.js
```diff
@@ -28,11 +28,14 @@ describe('Mutations component', () => {
...
     cy
       .get('.c-mutation-menu-list')
+      .should('be.visible')
       .find('.v-list-item__action')
+      .first()
       .click()
   }
   const submitMutationForms = () => {
     cy.visit('/#/workflows/one')
+    cy.get('.treeitem').should('be.visible')
```
kinow, line 38: Brittle test that started failing as the assertions were executed before the page was fully loaded.

tests/e2e/specs/tree.js
```diff
@@ -108,8 +108,12 @@ describe('Tree component', () => {
           .get('.node-data-cyclepoint')
           .should('be.visible')
         cy.window().its('app.$workflowService.subscriptions').then(subscriptions => {
-          // FIXME: likely wrong, but to be fixed later in a follow-up PR to housekeep subscriptions
```
kinow, line 111: There were several tests like this one. Where we were checking for the wrong number of subscriptions. That's because the `MockWorkflowService` still didn't handle starting/stopping subscriptions. Since we are not using it any longer, now we are able to confirm the subscriptions are created and destroyed as expected :tada: (it was unintentionally fixed here :sweat_smile: )

vue.config.js
```diff
@@ -33,9 +33,14 @@ module.exports = {
   },
   devServer: {
     proxy: {
-      '^/userprofile': {
+      '(^/userprofile|^/graphql)': {
```
kinow, line 36: The URL's listed here will be served by `json-server`.

vue.config.js
```diff
@@ -33,9 +33,14 @@ module.exports = {
...
         target: 'http://localhost:3000/',
         changeOrigin: true
+      },
+      '^/subscriptions': {
+        target: 'http://localhost:3000/',
+        changeOrigin: true,
+        ws: true
```
kinow, line 43: This URL needs a new block for the `ws` flag of the `webpack-dev-server` - https://github.com/webpack/webpack-dev-server/issues/283#issuecomment-162163329

kinow:
Had to learn code & docs of `json-server`, `express-ws`, and Cypress :sweat_smile: started boring, but slowly became an interesting day. I think the tests should pass now. If so, Monday morning, after some :coffee: I'll go through the code once more to see if I can simplify or add more docs, and will finally write the e2e test for GraphiQL :+1:

kinow:
>Merging #680 (a17410a) into master (5c8c2e5) will increase coverage by 4.47%

This also was not intentional. But without the mocked service, it appears we are exercising more lines of our code :slightly_smiling_face: Maybe we will pass the 90% coverage for the first time in the UI projects :crossed_fingers:

kinow:
Added the unit test, and confirmed the new test would have caught the bug where I broke GraphiQL :+1: 

image

So at least that should help preventing me from adding this same bug again (cannot promise not adding new bugs though :smile: )

kinow:
New test is brittle (passes locally sometimes). Will fix on Monday :wave:

kinow:
Ah, finally reproduced the issue locally. I wrote the tests on Firefox, but CI runs Electron. The issue happens on Electron and Chrome. Wrote an entry in our Wiki for troubleshooting for future developers :nerd_face: :neckbeard: https://github.com/cylc/cylc-ui/wiki/Troubleshooting#e2e-tests

kinow:
The error in the GH actions log is:

```
  1) GraphiQL
       should execute a GraphQL query and get a valid response:
     AssertionError: expected 'Syntax Error: Unexpected <EOF>.' to include 'user|one'
      at Context.eval (http://localhost:8080/__cypress/tests?p=tests/e2e/specs/graphiql.js:1743:51)
```

Couldn't find in the code what would cause it. But searching the Internet: https://github.com/dotansimha/graphql-code-generator/issues/4426#issue-664977147

Turns out sending the commented out text causes this issue.

image

So now I know that the code to set the query value is not working on Electron/Chrome when in headless mode (it works with the GUI).

kinow:
The real endpoint, instead of the mocked class, made two tests brittle, the new one `graphiql.js`, but also the existing `mutation.js`.

Took me a while to realize what was going on there.

Screenshot from 2021-05-31 20-58-15

Running the tests locally, something like 1 out of 10 times it would fail with the same error seen on GH actions. The issue was that the Skeleton Loader could still be displayed when the test started. What would cause the test to fail to locate the tree node element.

Pushed last commit with a fix for that (I think :crossed_fingers: )

kinow:
Whew, finally ready for review. Coverage increased a bit, but not really important. The most important parts of this PR are fixing the GraphiQL URL, and being able to test GraphiQL and other parts of the UI without needing to mock code :+1:

cypress.json
```diff
@@ -1,10 +1,10 @@
 {
   "pluginsFile": "tests/e2e/plugins/index.js",
   "video": false,
-  "defaultCommandTimeout": 10000,
+  "defaultCommandTimeout": 30000,
   "execTimeout": 60000,
   "taskTimeout": 60000,
   "pageLoadTimeout": 60000,
-  "requestTimeout": 10000,
+  "requestTimeout": 30000,
```
kinow, line 8: Increased because I noticed some tests on GH actions were taking 6, 7 seconds. And `mutations.js` failed a couple times, first for taking longer than 10 seconds, then because it had the skeleton loader still being displayed. Locally this timeout shouldn't matter, but 30 seconds seems to be good enough for GH actions.

package.json
```diff
@@ -71,17 +71,19 @@
     "concurrently": "^6.0.2",
     "cross-env": "^7.0.3",
     "cross-fetch": "^3.1.4",
-    "cypress": "^7.3.0",
+    "cypress": "^7.4.0",
```
kinow, line 74: Cypress GUI was showing an update available so updated it :man_shrugging: why not right?

tests/e2e/specs/graphiql.js
```diff
@@ -0,0 +1,66 @@
...
+        editors[0].CodeMirror.setValue(query)
+        // This appears to force the CodeMirror command above to be executed, or at least
+        // waited for.
+        expect(editors[0].CodeMirror.getValue()).to.equal(query)
+        expect(editors[1].CodeMirror.getValue()).to.equal('')
+        expect(editors[2].CodeMirror.getValue()).to.equal('')
+      })
+    // TODO: CodeMirror seems to have a delay to actually set the value to the underlying
+    //       textarea. Which can cause the test below to fail as the query submitted is
+    //       the default commented-out text, instead of the given query above.
```
kinow, line 51: This was tricky. And I couldn't find a cleaner way to send the query only after CodeMirror has updated the text area.

What you see in GraphiQL is a CodeMirror element. When you type something, CodeMirror is parsing your input, and eventually filling a hidden textarea. It takes a few ns/ms to update it, but occasionally we can get this test failing because CodeMirror still has the old value (the comment you see when you first start GraphiQL).

Waiting any time appears to be enough to give the browser time to update the element. Probably even a timeout of `0` (zero) could work (common in JS, I think we had it in the infinite tree before, see [this SO for more](https://stackoverflow.com/a/779785)), but `500` looks safer IMO.

kinow:
Rebased.

hjoliver:
One `tree.js` test failing?

kinow:
> One `tree.js` test failing?

Ah, I replaced the mocked workflow JSON with the new one that contained the isRunahead data, but forgot that I needed to have the task outputs as well. Added them back. Tests should all pass now :crossed_fingers: Thanks!

hjoliver (approved):
LGTM :tada:  One review will do for now, as it's offline functionality.

kinow:
Thanks @hjoliver !

PR #681, opened 2021-06-01
Remove flex-basis and flex-wrap from gscan

kinow:

Another iteration on responsiveness for GScan. Noticed this while testing GScan for recent PR's. The flex-basis ratio of `40/60` worked well in the past. But after we got more states displayed in GScan, it didn't work that well.

image

The reason is that 5 states, with the icon now bigger, passes the flex-basis limit we set. That, combined with flex-wrap, cause the GScan workflow row to grow.

Tested with a really long workflow name, and hacking the code to allow several states. Then adding more than 5 states. Screenshots below show what it looks like after this PR.

Screenshot from 2021-06-01 11-48-33
Screenshot from 2021-06-01 11-52-08
Screenshot from 2021-06-01 11-52-39
Screenshot from 2021-06-01 11-56-35

**Requirements check-list**

### DIFF
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -118,7 +118,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               :class="getWorkflowClass(scope.node.node)"
             >
               <v-list-item-title>
-                <v-layout align-center align-content-center d-flex flex-wrap>
+                <v-layout align-center align-content-center d-flex flex-nowrap>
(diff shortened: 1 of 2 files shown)


hjoliver (approved):
IMO this is an improvement, given that you get the full workflow name as a tooltip. Easy to revisit later if others disagree. One review will do.

kinow:
> IMO this is an improvement, given that you get the full workflow name as a tooltip. Easy to revisit later if others disagree. One review will do.

Agreed, thanks!!!

PR #682, opened 2021-06-02
Fix dashboard hub URL.

hjoliver:
#677 broke the Dashboard's Hub URL, which isn't supposed to have the user bit in it.

Not sure if this is the best way to fix it?

**Requirements check-list**
- [x] No change log entry required (fixes recent dev break).

### DIFF
--- a/src/utils/urls.js
+++ b/src/utils/urls.js
@@ -43,9 +43,7 @@ function getBaseUrl (websockets = false, baseonly = false) {
     ? window.location.protocol.startsWith('https') ? 'wss:' : 'ws:'
     : window.location.protocol
   const host = window.location.host
-  console.log('HOST', host)
   const baseUrl = `${protocol}//${host}`
-  console.log(baseonly)
   if (baseonly) {
     return normalize(new URL(baseUrl).href)
   } else {
@@ -67,7 +65,6 @@ function createUrl (path, websockets = false, baseonly = false) {
   const url = [baseUrl, path]
     .map(part => part.trim())
     .join('/')
-  console.log(url)
   return normalize(url)
 }
 
--- a/src/utils/urls.js
+++ b/src/utils/urls.js
@@ -34,18 +34,21 @@ function normalize (url) {
  *
  * The returned URL does not include the query search params (i.e. excludes ?redirectTo=/hub/login).
  *
+ * The parameter `baseOnly` can be used to define whether the final URL will incl
+ *
  * @private
  * @param {boolean} websockets - whether the URL will be used for websockets or not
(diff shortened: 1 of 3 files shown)


kinow:
Ah, good catch! I think the `console.log` lines need to be removed, but looks like a good way to solve it. Will review tomorrow morning after :coffee: and see if we can add a test somewhere so it doesn't happen again. Sorry for yet another :bug: (2 in less than 1 month :laughing: )

src/utils/urls.js
```diff
@@ -38,14 +38,20 @@ function normalize (url) {
  * @param {boolean} websockets - whether the URL will be used for websockets or not
  * @returns {string} - the application base URL, containing protocol, hostname, port, and pathname
  */
-function getBaseUrl (websockets = false) {
+function getBaseUrl (websockets = false, baseonly = false) {
   const protocol = websockets
     ? window.location.protocol.startsWith('https') ? 'wss:' : 'ws:'
     : window.location.protocol
   const host = window.location.host
+  console.log('HOST', host)
```
kinow, line 46: Removing it should fix part of the build (maybe the whole CI build?)

hjoliver: Oops!

src/utils/urls.js
```diff
@@ -34,18 +34,25 @@ function normalize (url) {
  *
  * The returned URL does not include the query search params (i.e. excludes ?redirectTo=/hub/login).
  *
+ * The parameter `baseOnly` can be used to define whether the final URL will incl
```
kinow, line 37: Is it OK to use camel case @hjoliver ? Sorry my OCD :laughing:

tests/unit/utils/urls.spec.js
```diff
@@ -35,6 +35,13 @@ describe('urls', () => {
         expected: `${PROTOCOL}//${HOST}/${PATHNAME}`,
         location: DEFAULT_LOCATION
       },
+      {
+        path: '',
+        websockets: false,
+        baseOnly: true,
+        expected: `${PROTOCOL}//${HOST}/`,
+        location: DEFAULT_LOCATION
+      },
```
kinow, line 44: I pushed a commit with this test, and the docs & camel-cased parameter name. I've approved it @hjoliver . If you are OK with the changes feel free to merge it, otherwise we can drop my commit :+1:

kinow (approved):

src/utils/urls.js
hjoliver: Yes, much as I hate it, I should conform to the practice here!

tests/unit/utils/urls.spec.js
hjoliver: Thanks for doing the test!

PR #699, opened 2021-07-05
Simplify how we add views to the Workflow view as widgets

### LINKED ISSUE #376: expose the cylc views standalone
This has been discussed before but @hjoliver made me aware we hadn't documented it anywhere.

The Cylc workflow views (tree, dot, graph, etc) and Cylc GScan should all be accessible via their own URL endpoints. This is currently true of the Graph view and should be fairly straight forward.

This is to allow the components to be used standalone for embedding and other purposes.

The main use case I'm aware of is that some users (namely operators) *might* want to have GScan open full-screen on one monitor.

Cylc views can be built to be responsive to make use of the available screen space.

**Pull requests welcome!**

kinow:
These changes partially address #376

Coverage change result of code that was removed, reducing the total coverage. But the new code appears to be covered/exercised by e2e and unit tests - https://app.codecov.io/gh/cylc/cylc-ui/compare/699

**Requirements check-list**

### DIFF
--- a/src/components/cylc/workflow/Toolbar.vue
+++ b/src/components/cylc/workflow/Toolbar.vue
@@ -96,7 +96,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             class="py-0 px-8 ma-0 c-add-view"
           >
             <v-list-item-icon>
-              <v-icon>{{ view.icon }}</v-icon>
+              <v-icon>{{ view.data().widget.icon }}</v-icon>
             </v-list-item-icon>
             <v-list-item-title>{{ view.name }}</v-list-item-title>
           </v-list-item>
@@ -124,8 +124,6 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 <script>
 import { mapGetters, mapState } from 'vuex'
 import {
-  mdiAppleKeyboardCommand,
-  mdiFileTree,
   mdiMicrosoftXboxControllerMenu,
   mdiPause,
   mdiPlay,
@@ -135,8 +133,6 @@ import {
 } from '@mdi/js'
 import toolbar from '@/mixins/toolbar'
 import WorkflowState from '@/model/WorkflowState.model'
-import TreeView from '@/views/Tree'
-import MutationsView from '@/views/Mutations'
 
 import {
   mutationStatus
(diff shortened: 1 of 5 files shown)


kinow:
After this pull request, any view that provides the following data property

```js
# file: Table.vue <- for the table view, for instance
import { mdiTable } from '@mdi/js'

export default {
...
  data () {
    return {
      widget: {
        title: 'table',
        icon: mdiTable
      }
    }
  }
...
}
```

can be added as a widget. The Toolbar component will use the title and icon to automatically populate the dropdown (e.g. "Add Table"). And the workflow component will use the title in the widget tab name.

@hjoliver after this is reviewed & merged, I will undo many of my changes to @giulianoserrao 's pull request, and then update his view to use the new mixins, and to provide this `widget` property. After that it should be ready to be reviewed and merged too :tada: 

@oliver-sanders let me know if you think there's more that needs to be done before we can close #376 or if this new way to create widgets from view, combined with our new mixins, is enough :+1: 

Thanks!
Bruno

oliver-sanders (approved):
:+1:

oliver-sanders:
> @oliver-sanders let me know if you think there's more that needs to be done before we can close #376

I think there are three things left to do, we can bump them to their own issues

* I think we should auto-generate the routing if/where possible so new views get their own pages automatically.
  (we might want to change the URL for the views to `#/view/<view>[/<flow>]`)
* The standalone tree view has the GScan sidebar which shouldn't be there.
* The GScan sidebar should be available standalone too.
  (we might want to give GScan the URL `#/workflows` and move the workflow table to something like `#/workflow-table`.

kinow:
>I think we should auto-generate the routing if/where possible so new views get their own pages automatically.
(we might want to change the URL for the views to #/view/<view>[/<flow>])

Ohh. Interesting. Not sure if doable, but should be fun to try!

>The standalone "tree" and "workflows" views have the GScan sidebar which shouldn't be there
(should only be one view per standalone page)

Not having GScan in every view would be helpful since it is a component consuming data like a view. If instead GScan were a view, things would be more uniform.

I think @hjoliver and others might want to have a look at these ideas and comment before we implement them. Let me change the PR from "closes" to just mentioning the issue number then :+1: Later we can manually close it and open new issues, or keep the issue open, happy either way.

Thanks!

hjoliver:
Yep, I agree with @oliver-sanders' suggestions, and that they can be done as follow-ups.  The auto-gen routing one is the least important, but would be helpful if possible.

hjoliver (approved):
:-1:

kinow:
> Yep, I agree with @oliver-sanders' suggestions, and that they can be done as follow-ups. The auto-gen routing one is the least important, but would be helpful if possible.

Perfect. I've copied @oliver-sanders ' comment into the issue, and this PR is not closing the issue, so will keep working on that later. Thanks!

PR #700, opened 2021-07-05
Use WorkflowState to control Tree view state, ignoring the shutdown property

### LINKED ISSUE #556: UI sees old and new state of a re-run flow
**Describe the bug**

If a stop a flow, the tree view remembers its final state.

If I then run the flow again after reconfiguring it to look different, the tree view still shows the old stopped state along with the new restarted state

Current master

```ini
[scheduling]                                                                    
    initial cycle point = now                                                  
    [[graph]]                                                                   
        R1 = foo => bar & baz & qux                                             
[runtime]                                                                       
    [[foo]]                                                                     
       script = "sleep 20"        
```
Run the above flow, stop it while the first task is running, comment out `initial cycle point = now` to make it an integer flow instead of date-time, then run it again.  

**Expected behavior**

Clear out the old state, once the flow starts running again.

**Screenshots**

Screenshot from 2020-12-09 12-24-10

**Pull requests welcome!**

### LINKED ISSUE #689: Received a delta before the workflow initial data burst
**Describe the bug**

Error message in the console and displayed using the error banner:

> Received a delta before the workflow initial data burst. Please reload your browser tab to retrieve the full flow state

1. Open the tree view at a running workflow
2. Stop the workflow (wait for UI to update)
3. Play the workfow.

**Additional context**

This diff shows that the error is caused by an empty delta:

```diff
diff --git a/src/components/cylc/tree/deltas.js b/src/components/cylc/tree/deltas.js
index b953f9ce..a14a16a8 100644
--- a/src/components/cylc/tree/deltas.js
+++ b/src/components/cylc/tree/deltas.js
@@ -220,6 +220,7 @@ export function applyDeltas (deltas, tree) {
       // page. So we fail silently and wait for a request with the initial data.
       if (!deltas.added || !deltas.added.workflow) {
         // eslint-disable-next-line no-console
+        console.error(deltas)
         console.error('Received a delta before the workflow initial data burst')
         store.dispatch('setAlert', new AlertModel('Received a delta before the workflow initial data burst. Please reload your browser tab to retrieve the full flow state', null, 'error'))
         return
```

```
{"id": "2", "type": "data", "payload": {"data": {"deltas": {"id": "oliver|one", "shutdown": true, "added": {"__typename": "Added"}, "updated": {"__typename": "Updated"}, "pruned": {"__typename": "Pruned"}, "__typename": "Deltas"}}}}	1624270854.4726202
```

I can't tell if the delta being sent relates to the previous run or the new one so the error might be legitimate.

As the subscription uses `stripNull=true` I think this becomes a UIS issue?

**Pull requests welcome!**

kinow:
These changes:

- close #689 
- close #556 

**Requirements check-list**

### DIFF
--- a/src/components/cylc/tree/deltas.js
+++ b/src/components/cylc/tree/deltas.js
@@ -24,6 +24,7 @@ import {
   createTaskProxyNode
 } from '@/components/cylc/tree/nodes'
 import * as CylcTree from '@/components/cylc/tree/index'
+import WorkflowState from '@/model/WorkflowState.model'
 
 /**
  * Helper object used to iterate added deltas data.
@@ -213,11 +214,13 @@ function handleDeltas (deltas, workflow, lookup, options) {
  */
 export default function (data, workflow, lookup, options) {
   const deltas = data.deltas
-  // first we check whether it is a shutdown response
-  if (deltas.shutdown) {
-    CylcTree.clear(workflow)
-    return {
-      errors: []
+  // first we check whether it is a new start
+  if (deltas && deltas.added && deltas.added.workflow) {
+    if (deltas.added.workflow.status === WorkflowState.RUNNING.name) {
+      // The workflow could be stopped. In this case when restarted (cold or hot)
+      // it would be hard to apply the new deltas keeping the UI state valid. So
+      // for now we clear the tree, and start from scratch rebuilding it.
+      CylcTree.clear(workflow)
     }
   }
   // Safe check in case the tree is empty.
@@ -248,7 +251,6 @@ export default function (data, workflow, lookup, options) {
   // 1. data added in deltas.added
   // 2. data updated in deltas.updated
   // 3. data pruned in deltas.pruned
-  // 4. a delta with some data, and the .shutdown flag telling us the workflow has stopped
   try {
     return handleDeltas(deltas, workflow, lookup, options)
   } catch (error) {
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -137,6 +137,13 @@ in the tree view structure.
 
 [#554](https://github.com/cylc/cylc-ui/pull/554) - Fix icons in the Dashboard.
 
+[#556](https://github.com/cylc/cylc-ui/issues/556) - UI sees old and new state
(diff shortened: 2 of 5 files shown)


kinow:
@hjoliver I tested starting `five`, then stopping it and starting it from scratch without reloading the page. Also paused then restarted `five`. In both cases the view cleared the Tree before adding new nodes.

There might be a follow-up discussion for later, whether we can do something better than removing everything and starting from scratch when the workflow is stopped/paused, but that will require more thinking. This PR only fixes the issue where we had old data, and an inconsistent state too (i.e. the tree could be cleared, while the UIS data pool still contained data; so reloading the page would show the last state of the workflow).

src/components/cylc/tree/deltas.js
```diff
@@ -213,11 +214,13 @@ function handleDeltas (deltas, workflow, lookup, options) {
...
-    CylcTree.clear(workflow)
-    return {
-      errors: []
+  // first we check whether it is a new start
+  if (deltas && deltas.added && deltas.added.workflow) {
+    if (deltas.added.workflow.status === WorkflowState.RUNNING.name) {
+      // The workflow could be stopped. In this case when restarted (cold or hot)
+      // it would be hard to apply the new deltas keeping the UI state valid. So
+      // for now we clear the tree, and start from scratch rebuilding it.
+      CylcTree.clear(workflow)
```
kinow, line 223: Note too that at the moment the tree view has no reaction when the workflow status change. We could gray-out the whole area, disable the tree, set an icon somewhere, etc. But I think that will be done in a follow-up, or maybe part of #332

oliver-sanders: I think the background colour could change if the workflow is stopped just to make it clear that the data is offline. But yes, work for later.

(experience with Cylc Review has show us that people look past the workflow status in the top left and often interpret a running task to mean a running workflow)

oliver-sanders (approved):
tested:

* fixes #556
* think this fixes #689 (haven't managed to replicate on this branch)

hjoliver (approved):

PR #703, opened 2021-07-06
Fix sorting in GScan due to strange enumify behaviour

### LINKED ISSUE #702: gscan: workflow state sorting bug
GScan can get confused when workflows change state.

Here's a screenshot showing a running workflow shown at the bottom of the GScan list (should be at the top):

Screenshot from 2021-07-06 12-34-39

When the page is refreshed it appears at the top:

Screenshot from 2021-07-06 12-34-59

Steps to reproduce:

* Install two workflows.
* Navigate to the second workflow (in the sort order) in GScan.
* Start it.

**Pull requests welcome!**

kinow:
These changes close #702 

**Requirements check-list**
- [x] Already covered by existing tests. (it should be covered by tests, but the bug appears to happen in runtime, while behaving as expected in the IDE or unit tests :confused: )
- [x] No change log entry required (why? e.g. invisible to users as we haven't released a version with this bug).

### DIFF
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -316,12 +316,20 @@ export default {
      * (natural sort).
      */
     sortedWorkflows () {
-      const workflowStateOrder = WorkflowStateOrder.map(state => state.name)
+      // Create a new object with state.name => order.
+      const workflowStateOrder = Object.fromEntries(
+        WorkflowStateOrder
+          .map(([state, order]) => {
+            return [state.name, order]
+          })
+      )
       return [...this.filteredWorkflows].sort((left, right) => {
-        if (left.status !== right.status) {
-          const leftState = left.status
-          const rightState = right.status
-          return workflowStateOrder.indexOf(leftState) - workflowStateOrder.indexOf(rightState)
+        const leftStateOrder = workflowStateOrder[left.status]
+        const rightStateOrder = workflowStateOrder[right.status]
+        // Here we compare the order of states, not the states since some states
+        // are grouped together, like RUNNING, PAUSED, and STOPPING.
+        if (leftStateOrder !== rightStateOrder) {
+          return leftStateOrder - rightStateOrder
         }
         return left.name
           .localeCompare(
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -195,7 +195,7 @@ import subscriptionComponentMixin from '@/mixins/subscriptionComponent'
 import JobState from '@/model/JobState.model'
 import TaskState from '@/model/TaskState.model'
 import SubscriptionQuery from '@/model/SubscriptionQuery.model'
-import WorkflowState from '@/model/WorkflowState.model'
+import { WorkflowState, WorkflowStateOrder } from '@/model/WorkflowState.model'
 import Job from '@/components/cylc/Job'
 import Tree from '@/components/cylc/tree/Tree'
 import WorkflowIcon from '@/components/cylc/gscan/WorkflowIcon'
@@ -316,11 +316,12 @@ export default {
      * (natural sort).
      */
     sortedWorkflows () {
+      const workflowStateOrder = WorkflowStateOrder.map(state => state.name)
       return [...this.filteredWorkflows].sort((left, right) => {
         if (left.status !== right.status) {
-          const leftState = WorkflowState.enumValueOf(left.status.toUpperCase())
-          const rightState = WorkflowState.enumValueOf(right.status.toUpperCase())
-          return leftState.enumOrdinal - rightState.enumOrdinal
+          const leftState = left.status
+          const rightState = right.status
+          return workflowStateOrder.indexOf(leftState) - workflowStateOrder.indexOf(rightState)
         }
         return left.name
           .localeCompare(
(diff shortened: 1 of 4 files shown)


src/model/TaskState.model.js
```diff
@@ -21,8 +21,6 @@ import { Enumify } from 'enumify'
  * Cylc valid task states.
  */
 export class TaskState extends Enumify {
-  // NOTE: the order of the enum values is important to calculate the group states in the UI. Items at
-  // the top of the list have preference over the items below in the algorithm.
```
kinow, line 25: There's an array used to sort things in the UI. Expand the file here or click View File, then scroll down. Which means this comment is outdated.

src/model/WorkflowState.model.js
```diff
@@ -51,4 +50,16 @@ class WorkflowState extends Enumify {
...
+ * Workflow states ordered for display purposes.
+ */
+export const WorkflowStateOrder = [
+  WorkflowState.RUNNING,
+  WorkflowState.PAUSED,
+  WorkflowState.STOPPING,
+  WorkflowState.STOPPED,
+  WorkflowState.INSTALLED,
+  WorkflowState.ERROR
+]
```
kinow, line 63: Matching what we have in `TaskState.model.js`, where the enum order is managed in an array, instead of the order of elements declared in the enum.

oliver-sanders (approved):
I've not been able to replicate the bug, looks good.

I think we used to mix running and paused workflows?

Really I think we would like to mix running, paused and stopping workflows together (just to be awkward :cry:) as it might get a bit confusing for users looking for workflows by name and not being able to find them. It makes sense to separate stopped workflows and they are displayed differently which makes the distinction clear.

Perhaps we could use a sort order mapping?

```js
WorkflowSortOrder = {
    WorkflowState.RUNNING: 0,
    WorkflowState.PAUSED: 0,
    WorkflowState.STOPPING: 0,
    WorkflowState.STOPPED: 1,
}
```

We could tackle this one later.

kinow:
> I've not been able to replicate the bug, looks good.

:tada: 

> I think we used to mix running and paused workflows?

Oh, now that you mention... how did it happen that we stopped doing that? I probably deleted it in some pull request, sorry.

> Really I think we would like to mix running, paused and stopping workflows together (just to be awkward cry) as it might get a bit confusing for users looking for workflows by name and not being able to find them. It makes sense to separate stopped workflows and they are displayed differently which makes the distinction clear.
> 
> Perhaps we could use a sort order mapping?
> 
> ```js
> WorkflowSortOrder = {
>     WorkflowState.RUNNING: 0,
>     WorkflowState.PAUSED: 0,
>     WorkflowState.STOPPING: 0,
>     WorkflowState.STOPPED: 1,
> }
> ```
> 
> We could tackle this one later.

Great idea, and took me just a few minutes to implement, but near 1 hour then to fix 1 of the existing test cases. Which was good as that spot a bug in my few-minutes-implementation :facepalm: 

Updated code and unit tests, and manually tested the change. Added @hjoliver as second reviewer since now it changes a little how GScan works (comparing to what's on master).

Screenshot from 2021-07-09 09-26-34

Thanks @oliver-sanders !

src/components/cylc/gscan/GScan.vue
```diff
@@ -316,11 +316,20 @@ export default {
...
-        if (left.status !== right.status) {
-          const leftState = WorkflowState.enumValueOf(left.status.toUpperCase())
-          const rightState = WorkflowState.enumValueOf(right.status.toUpperCase())
-          return leftState.enumOrdinal - rightState.enumOrdinal
+        const leftStateOrder = workflowStateOrder[left.status]
+        const rightStateOrder = workflowStateOrder[right.status]
+        // Here we compare the order of states, not the states since some states
+        // are grouped together, like RUNNING, PAUSED, and STOPPING.
+        if (leftStateOrder !== rightStateOrder) {
+          return leftStateOrder - rightStateOrder
```
kinow, line 332: One case in a unit test was failing because of this. It ended up with the incorrect order since it was finding workflow `e` as `PAUSED` and workflow `a` as `RUNNING` and returning `0` since the states were different. When actually it was supposed to skip the states comparison since they have the same order now, and then sort alphabetically. :neckbeard:

oliver-sanders (approved):
:+1:

kinow:
Working on #707 I realized I was using an array, then converting to an object with `Object.fromEntries`. While that works, simply using an Object would be much easier to maintain :sweat_smile: 

Updated to match what's used in #707. Tests have not been changed, and also tested manually locally to confirm it's still working :+1:

oliver-sanders:
Having some fun reading about object property order in JS.

https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order/38218582#38218582

I can't tell whether order is guaranteed to be preserved (until ES2020).

kinow:
> Having some fun reading about object property order in JS.
> 
> https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order/38218582#38218582
> 
> I can't tell whether order is guaranteed to be preserved (until ES2020).

Whew, didn't know about this thanks. JS is fun, but it's also extremely scary when you think all the possible things that can go wrong (compare variables of different types, access properties expecting some order, ...).

I think we should be fine as long as we have a very flat object, and where keys and values have a single type (`string` for key, `number` for value)?

kinow:
Although using a `Map` is super easy... will change that! :+1: this way we are definitely sure nothing can go wrong (famous last words).

kinow:
Updated to use a `Map`. Ran unit tests locally and also tested manually with a couple workflows. No issues found :+1:

oliver-sanders (approved):

kinow:
Rebased fixing conflicts.

Issue #679, opened 2021-05-19
workflow subscriptions when the hub user != uis user


oliver-sanders:
The user from the store is used to construct the workflow ID for the tree view subscription:

https://github.com/cylc/cylc-ui/blob/2042834d65f00ca9a77db1836066c56397221d1d/src/mixins/treeview.js#L39

This is fine so long as the user viewing the UI is the same as the user running the workflow.

However, the user in the store is the authenticated user which may be different. We need to capture and store the workflow owner somewhere so it can be used for this purpose later.

Note that the Jupyter Server plan (https://github.com/cylc/cylc-uiserver/issues/219) means that we cannot get the workflow owner from the URL as that would only work when the UIS is run via the hub and not when it is run standalone.

Ideas:

1. We could add a `User` field to the GraphQL schema.
2. We could add an `owner` field to the userprofile.

I may try (2) as part of https://github.com/cylc/cylc-uiserver/issues/219 as it should be quick.

**Pull requests welcome!**
This is an Open Source project - please consider contributing a bug fix
yourself (please read `CONTRIBUTING.md` before starting any work though).

kinow:
@oliver-sanders realized just now that https://github.com/cylc/cylc-ui/pull/724 appears to be related to this issue? Let me know if that PR is unnecessary with this new owner field in the UIS. Or if that approach could still be used (happy to close that PR if it'd be better to use the owner attribute in some way).

oliver-sanders:
Sorry, I'm falling behind with reviewing at the moment, haven't got around to #724 yet.

The tricky thing with the Jupyter Server conversion is dealing with standalone UI servers that use token authentication rather than JupyterHub.

PR #711, opened 2021-07-29
Jupyter server

oliver-sanders:
jupyter_server: support Cylc UIS Jupyter Server extension
    
* https://github.com/cylc/cylc-uiserver/pull/230
* The UIS is now a Jupyter Server extension.
* The endpoint is now `cylc/#/` rather than `#/`
* Authentication via token/cookie is now supported.
* Sets XSRF header to support new hubless single-user mode.
* closes https://github.com/cylc/cylc-ui/issues/679

**Requirements check-list**
- [x] Docs: https://github.com/cylc/cylc-doc/pull/285

### DIFF
--- a/src/model/User.model.js
+++ b/src/model/User.model.js
@@ -24,11 +24,17 @@
  * @property {string} server - server URL
  */
 export default class User {
-  constructor (username, groups, created, admin, server) {
+  constructor (username, groups, created, admin, server, owner) {
+    // the authenticated user
+    // (full info only available when authenticated via the hub)
     this.username = username
     this.groups = groups
     this.created = created
     this.admin = admin
-    this.server = server
+    this.server = server || '?' // server can be unset
+    // the UIS owner
+    // (i.e. the user who's workflows we are looking at)
+    // (this might not be the authenticated user for multi-user setups)
+    this.owner = owner
   }
 }
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,10 @@ release.
 
 [#502](https://github.com/cylc/cylc-ui/pull/502) - Hierarchical GScan.
 
+[#711](https://github.com/cylc/cylc-ui/pull/711) -
+Support Jupyter Server conversion. (the CylcUIServer has been converted
+to a Jupyter Server extension).
+
 ### Fixes
 
 -------------------------------------------------------------------------------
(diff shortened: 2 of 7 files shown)


oliver-sanders:
Rebased onto 0.6.0

kinow:
Ready for review @oliver-sanders ?

oliver-sanders:
Sorry, getting behind, yep I think so.

src/model/User.model.js
```diff
@@ -24,11 +24,17 @@
...
-  constructor (username, groups, created, admin, server) {
+  constructor (username, groups, created, admin, server, owner) {
+    // the authenticated user
+    // (full info only available when authenticated via the hub)
     this.username = username
     this.groups = groups
     this.created = created
     this.admin = admin
-    this.server = server
+    this.server = server || '?' // server can be unset
```
kinow, line 34: Could we just leave it blank? Or is the `?` necessary or used elsewhere?

kinow (approved):
Looks good @oliver-sanders ! Couple comments, but no blockers.

>The endpoint is now cylc/#/ rather than #/

Just checking, but I think it didn't require any change in Cylc UI since it (VueRouter) is only responsible for the part after `/#/`? Thanks! 👍

src/model/User.model.js
oliver-sanders: I think I made this change because I was getting undefined errors from somewhere in the code.

oliver-sanders:
> Just checking, but I think it didn't require any change in Cylc UI since it (VueRouter) is only responsible for the part after /#/? Thanks! +1

I started this branch a few months back and had to rebase it recently, I think it's smaller now thanks to other UI improvements (like the URL creation thinggy) :+1: so this is probably outdated now.

oliver-sanders:
Rebased and fixed GraphiQL.

src/model/User.model.js
kinow: Quite sure this can be removed @oliver-sanders . Maybe it was the unit test that failed when it was undefined?

From what I recall, I used this `server` in the past to craft the GraphQL endpoint URL. But then we ran into problems after a user reported s/he couldn't get `/extra-path/users/blabla/#/` to work, and then I stopped using `server` and used the `window.location` to create the URL.

Only reason the value is still saved, I think, is because it's returned from the JupyterHub REST service, so no harm in leaving it.

datamel (approved):
Works smoothly for me.

src/model/User.model.js
oliver-sanders: Ah that might have been it, will take a look...

PR #712, opened 2021-07-30
Yarn 3.0.0 released, disable telemetry

kinow:

**Requirements check-list**


.yarnrc.yml
```diff
@@ -1,12 +1,15 @@
-nodeLinker: node-modules
-
-yarnPath: .yarn/releases/yarn-3.0.0-rc.2.cjs
-
 enableNetwork: false
+
 enableStrictSsl: true
 
+enableTelemetry: 0
```
kinow, line 5: Didn't know it was opt-in :angry: apologies. Disabled now.

.yarnrc.yml
```diff
@@ -1,12 +1,15 @@
...
+
 networkSettings:
-  "registry.yarnpkg.com":
+  registry.npmjs.org:
     enableNetwork: true
-  "registry.npmjs.org":
+  registry.yarnpkg.com:
     enableNetwork: true
```
kinow, line 11: Note: when you upgrade Yarn, it will fail because it will try to use GitHub.com. I've enabled github.com, installed Yarn, and then disabled it. This way we can install Yarn, but don't allow other libraries installed directly from github.com (which we don't need at the moment, and prevents malicious packages hosted at GH of being installed)

oliver-sanders (approved):

PR #722, opened 2021-08-04
Check contributing with GH actions

### LINKED ISSUE #710: actions: check contributor list
Automatically test `git shortlog` against the list of contributors in order to ensure they match.

Required functionality implemented in cylc-release-actions, see the cylc-flow PR: https://github.com/cylc/cylc-flow/pull/4205

kinow:
These changes close #710 

**Requirements check-list**
- [x] Does not need tests (why? build only).

### DIFF
--- a/.github/workflows/create_release_pr.yml
+++ b/.github/workflows/create_release_pr.yml
@@ -20,6 +20,9 @@ jobs:
       - name: Sanitise workflow inputs # Should be 1st step
         uses: cylc/release-actions/stage-1/sanitize-inputs@v1
 
+      - name: Check CONTRIBUTING.md
+        uses: cylc/release-actions/check-shortlog@v1
+
       - name: Checkout repo
         uses: actions/checkout@v2
         with:
--- a/.github/workflows/shortlog.yml
+++ b/.github/workflows/shortlog.yml
@@ -1,0 +1,20 @@
+name: check contributor list
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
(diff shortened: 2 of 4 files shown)


.github/workflows/create_release_pr.yml
```diff
@@ -20,6 +20,9 @@ jobs:
       - name: Sanitise workflow inputs # Should be 1st step
         uses: cylc/release-actions/stage-1/sanitize-inputs@v1
 
+      - name: Check CONTRIBUTING.md
+        uses: cylc/release-actions/check-shortlog@v1
```
kinow, line 24: Here and in the shortlog.yml, I think I'll have to use either a new tag or a commit after this PR: https://github.com/cylc/release-actions/pull/18 @MetRonnie

MetRonnie: cylc/release-actions#18 is more of a bug fix imo, can keep the v1 tag*

---

*Actually it's a branch rather than tag - makes it easier to update. The idea is the branch always points to the latest commit that is still backwards compatible with the original state of the actions repo when v1 was ready to use. (We could also use tags for `major.minor.patch` releases, but it's not really needed.)

MetRonnie (approved):

.github/workflows/create_release_pr.yml
kinow: Got it, let me rebase it now.

oliver-sanders (approved):
Thanks

PR #724, opened 2021-08-05
Use window.location to retrieve the username for GraphQL variables

kinow:

**Requirements check-list**

### DIFF
--- a/src/mixins/graphql.js
+++ b/src/mixins/graphql.js
@@ -19,6 +19,7 @@ import { mapState } from 'vuex'
 
 // eslint-disable-next-line no-unused-vars
 import User from '@/model/User.model'
+import { getUserNameFromUrl } from '@/utils/urls'
 
 /**
  * A mixin that contains data used for a GraphQL subscription, such as the
@@ -51,7 +52,8 @@ export default {
      * @return {string} - the Workflow ID used in this view
      */
     workflowId () {
-      return `${this.user.username}|${this.workflowName}`
+      const username = getUserNameFromUrl(this.user.username)
+      return `${username}|${this.workflowName}`
     },
     /**
      * GraphQL query variables.
--- a/src/utils/urls.js
+++ b/src/utils/urls.js
@@ -72,6 +72,28 @@ function createUrl (path, websockets = false, baseOnly = false) {
   return normalize(url)
 }
 
+/**
+ * Get the user name from the current browser URL. If the `window.location.pathname`
+ * object is not available (e.g. running without a browser?) then the
+ * fallback given value is returned.
+ *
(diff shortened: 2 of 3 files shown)


src/mixins/graphql.js
```diff
@@ -51,7 +52,8 @@ export default {
      * @return {string} - the Workflow ID used in this view
      */
     workflowId () {
-      return `${this.user.username}|${this.workflowName}`
+      const username = getUserNameFromUrl(this.user.username)
+      return `${username}|${this.workflowName}`
```
kinow, line 56: The Vuex user is still there, but we will try to use the user found in the URL. If no user found, then we fallback to the Vuex value.

src/utils/urls.js
```diff
@@ -72,6 +72,28 @@ function createUrl (path, websockets = false, baseOnly = false) {
...
+ *
+ * The expected URL format is <protocol>://<hostname>:<port>/user/<username>/. The returned
+ * value will be either the <username> or the given fallback username.
+ *
+ * @param {string} fallbackValue - fallback value
+ * @returns {string} - the
+ */
+function getUserNameFromUrl (fallbackValue) {
+  return window.location.pathname
+    ? window.location.pathname
```
kinow, line 88: TodayILearned it's better/safer to use `window.location` than `document.location`, even though I get the same with both on Firefox/Chromium/etc: https://stackoverflow.com/a/2431375

kinow:
Just need to prepare the set-up to test @datamel 's PR again, to validate that this PR indeed allows us to access the workflows of other users.... then it will be ready for review.

kinow:
It works hohoho!

image

Left monitor with `cylc` user, but looking at `kinow|five`. Right monitor with `kinow` user looking at `kinow|five`.

Using the latest commit from @datamel 's https://github.com/cylc/cylc-uiserver/pull/204

Also tested changing 

```py
c.UIServer.user_authorization = {
  "cylc": ["ALL"]                                 
}
```

to

```py
c.UIServer.user_authorization = {
  "cylc": ["!ALL"]                                 
}
```

in `~/.cylc/hub/config.py` and got the 403 error as expected :+1: so ready for review, shouldn't break the UI if merged now, but can be used to help reviewing the UIS PR.

hjoliver (approved):

kinow:
Note: @oliver-sanders ' #711 PR uses the owner property returned by the UIS. This owner would fix the issue here too, so we might hold back on merging this one until @datamel finishes reviewing that PR. If merged, I'll test it again, and close this PR and the associated issue if the issue doesn't occur anymore 👍

kinow:
Superseded by #711

PR #725, opened 2021-08-08
Add button to toggle drawer

### LINKED ISSUE #389: sidebar: collapse should always be an option

kinow:
These changes close #389

**Requirements check-list**

### DIFF
--- a/src/styles/cylc/_drawer.scss
+++ b/src/styles/cylc/_drawer.scss
@@ -19,6 +19,17 @@
 
 /* this is not in our styles/material-dashboard, so we need to force-override */
 .v-navigation-drawer {
+  // overflow so we see the button to toggle the drawer
+  overflow: visible !important;
+  // toolbar has z-index 5; since the toggle-button appears over the toolbar, z-index 6 it is
+  z-index: 6;
+  #drawer-toggle-btn {
+    // toolbar height is 48px, using 24px puts the button at the center/middle of the toolbar
+    // (since we translated(x, -50%) the button below
+    top: 24px;
+    // translate(x, y) in 2d space so we move it to the right and up (relative to width/height)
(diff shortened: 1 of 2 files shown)


src/styles/cylc/_drawer.scss
```diff
@@ -40,3 +51,7 @@
     padding-right: 15px;
   }
 }
+
+.v-navigation-drawer--close {
+  visibility: unset !important;
```
kinow, line 56: The main panel has higher z-index, so it's displayed if the drawer is still visible.

kinow:
@hjoliver @oliver-sanders here's a possible solution; the drawer normally is toggled in two ways in Vuetify. One way is to have a toggle button outside of the drawer. Normally it's in the main content area, or in the toolbar. Cylc UI, however, does not always show a toolbar. For example, the dashboard has no toolbar visible, so it wouldn't be possible to put a toggle (or burger) button there.

The other way is to have a button in the drawer. The only issue is that when this approach is used, then the toolbar is not hidden. Instead a variant is activated when toggled, showing a "mini" toolbar (showing only icons, no text). That wouldn't work in mobile though. So I used this second option, with a variation with a floating button. It seems to work OK. If this looks like a good approach, I'll write the e2e test :+1: 

GIFrecord_2021-08-09_083415

GIFrecord_2021-08-09_083656

(looks like Dashboard is broken for mobile, will take care of that in another PR)

hjoliver:
Nice, that looks good to me.

wxtim:
stale

hjoliver:
> stale

(We can now resize the sidebar away)

PR #731, opened 2021-08-12
Remove margin/padding from Tree component

### LINKED ISSUE #730: Reduce or remove border for GScan items when hovered

Follow-up for GScan hierarchical from https://github.com/cylc/cylc-ui/pull/502#pullrequestreview-728556936

**Additional context**

Screenshot from 2021-08-12 13-45-12

**Pull requests welcome!**

kinow:
These changes close #730 

The Tree component now has no borders, padding, margin, etc. This way the caller code can choose whether to add them or not.

For the Tree **view** we want padding to match the previous appearance. But for GScan we don't want any padding or spaces so that the hover effect occupies as much space as possible without distracting users.

**Requirements check-list**
- [x] Does not need tests (why? hard to test, small change).
- [x] No change log entry required (why? e.g. invisible to users (since the gscan-hierarchical code was not released yet)).

### DIFF
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -109,7 +109,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           :filterable="false"
           :expand-collapse-toggle="false"
           :workflows="filteredWorkflows"
-          class="c-gscan-workflow ma-0 pl-2"
+          class="c-gscan-workflow ma-0 pa-0"
         >
           <template v-slot:node="scope">
             <v-list-item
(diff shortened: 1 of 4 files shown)


kinow:
What is looks like in offline mode: 

image

And with real workflows:

GIFrecord_2021-08-13_150031

@oliver-sanders I think one review should do. Note that I added one more commit because I was (and am still not) happy with the alignment of GScan items. But it's hard to get everything align since the row with a Workflow has one structure (WorkflowIcon + Span with title + Task States) and the Workflow name part row has another one (Expand or collapse button, followed by span/text).

But at least now there should be no more padding-right for the items, and hovering the items they should be left-aligned when you moved your mouse over them (see final seconds of the GIF above).

oliver-sanders:
> I was (and am still not) happy with the alignment of GScan items.

Yeah, that looks tricky, much easier to draw than to implement.

oliver-sanders (approved):
Thanks, that looks better.

I don't know whether the left padding is something we could easily remove, I'm guessing the way the indentation for hierarchies is implemented makes that tricky?

Happy with one review, will leave you to merge or comment.

kinow:
>I don't know whether the left padding is something we could easily remove, I'm guessing the way the indentation for hierarchies is implemented makes that tricky?

Exactly, and I didn't parametrize the indentation. It's hard-coded to 1.5 rem. That, and the different structure makes it trickier to make GScan children elements aligned and uniform. But there is a lot of improvements pending from the design issue for GScan that I will start to go through and create issues once we have more time :+1: Thanks!

PR #733, opened 2021-08-17
Propagate state gscan

### LINKED ISSUE #721: Propagate summary state up the tree in GScan

>As a follow-up, we should have a summary state propagate up the tree (so you can see if anything is running in a collapsed branch...)

**Additional context**

Follow-up from #502 

**Pull requests welcome!**

kinow:
These changes close #721 

**Requirements check-list**
- [x] No change log entry required (why? e.g. invisible to users because hierarchical GScan has not been released yet).

### DIFF
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -142,7 +142,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                   </v-flex>
                   <!-- We check the latestStateTasks below as offline workflows won't have a latestStateTasks property -->
                   <v-flex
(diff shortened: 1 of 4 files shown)


kinow:
Was triaging issues, and updating the requirements for the #721 , thought what would be required for this to happen, and decided to create this draft since I had it fresh on my mind.

It adds the `node.latestStateTasks` property to workflow-name-part nodes (workflow nodes are the leaves in the GScan tree), though empty for now.

Also moves the code used to display the state summary into its own component. This way we should be able to simply call the component for both workflow and workflow-name-part passing the `node` as prop. Just need to populate the `node.latestStateTasks` of the workflow-name-part nodes, write tests, and voilà.

PR #736, opened 2021-08-23
GScan improvements, propagate state

### LINKED ISSUE #721: Propagate summary state up the tree in GScan

kinow:
These changes close #721 and supersede #733 (not yet, but will in a few more commits :crossed_fingers: )

**Requirements check-list**


kinow:
While working on #733 for #721, I noticed the `computed` variable that would re-trigger more elements in the UI to be updated than desired.

In this pull request I will move the GScan data out of the component, into Vuex, and update it incrementally based on deltas. The goal is to have the least amount of elements being updated as possible.

Here's a screenshot from this morning where I hacked the code to print whenever a `TreeItem` was updated.

Screenshot from 2021-08-23 10-50-06

Screenshot from 2021-08-23 10-50-32

Even though `five` was being updated, the TreeItem's for `jh1`, `rose/run1`, etc, were being updated as well :grimacing: which results in a major performance issue.

src/store/tree.module.js
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+import { clear } from '@/components/cylc/tree'
+import applyDeltasLookup from '@/components/cylc/common/deltas'
+import Alert from '@/model/Alert.model'
+import applyDeltasTree from '@/components/cylc/tree/deltas'
+
+const state = {
```
kinow, line 22: I've split the `workflows.module` since it had data used by GScan, and also by the Tree view. The tree view data is relatively more complex at this point, so moving it out simplifies the `workflows.module`.

src/components/cylc/common/deltas.js
```diff
@@ -79,3 +84,133 @@
...
+ * @typedef {Object} Result
+ * @property {Array<Object>} errors
+ */
+
+/**
+ * @param {DeltasAdded} added
+ * @param {Object.<String, Object>} lookup
+ * @return {Result}
+ */
+function applyDeltasAdded (added, lookup) {
```
kinow, line 98: The code added here was in `workflows/deltas.js`, but was used only by the Tree view. Instead of moving to the tree view, I've moved it here to common since if we ever need to use a delta that creates a lookup, anyone can simply call this function :point_up: 

I **think** I won't need a lookup for GScan, at least for now. Or at least I am avoiding adding one unless really needed. But if I need to add one, then I'll use this function.

kinow:
I've implemented most of what I had in mind for GScan. But this afternoon just before a meeting I decided to look if the `TreeItem`'s were still being updated, and strangely everything was still updating.

I created a smaller test scenario, with `yarn run serve` and only 1 workflow `one`. Then I push a delta manually to have another worklfow `a`. This workflow is sibling of `one` in the GScan tree, but `one` is being updated too, which is not really desirable.

I'm investigating now why that's happening, and so far these are the two best leads: https://github.com/vuejs/vue/issues/7257 & https://github.com/vuejs/vue/issues/10115#issuecomment-518503180 since they match what `why-did-you-render` is reporting in the browser console... more to come :+1: :bed:

kinow:
All right, coming back to this one after lunch, and strong :coffee: to help me fixing conflicts, and remember where was I.

kinow:
Hmm, I think I figured how to store the data in each node in a way that I can update with added/updated/pruned deltas.

The leaf nodes in the GScan tree are workflows. They will keep the same structure (latestStates + stateTotals). The intermediary nodes (workflow-name-part nodes, as I called them :monocle_face: ) will have a computed variable with same name. This way the component's templates won't need to be changed.

However, the real data will hold:

- the ID of every intermediate nodes (WIP: thinking)

kinow:
Done for the day, next week will start sorting the `parent.children` of a node when it was updated. Noticed this error when I started a random run and it didn't move up the tree.

Previously, we were always re-generating the whole tree :fearful: so it looked like the tree element was correctly re-positioned… but it was actually due to the new three being created from scratch.

kinow:
Think I found how to improve GScan a bit more, preventing it from re-rendering all the nodes in the GScan tree, refs:

* https://dev.to/veebuv/5-extremely-easy-ways-to-drastically-improve-your-vuejs-app-s-speed-5k0
* https://betterprogramming.pub/6-ways-to-speed-up-your-vue-js-application-2673a6f1cde4

kinow:
Will edit the last commit tomorrow, but wanted to save what I wrote today. Pending more manual tests with working workflows. Here's a preview of what it's looking like in the offline mode right now (modified `one` workflow a little, will undo tomorrow since it's used by e2e tests).

GIFrecord_2021-09-20_164611

wxtim:
Stale.

Issue #701, opened 2021-07-06
task: differentate running and submitted icons


oliver-sanders:
The submitted task icon is the same as the running task icon (just with no progress).

This is by design, however, it would help to differentiate them a little.

* [x] When a task has started running we should advance the "progress" animation forward a tick or two so that the clock face is visible.
* [ ] https://github.com/cylc/cylc-ui/issues/608
* [ ] #922

![Screenshot from 2021-07-06 10-03-46](https://user-images.githubusercontent.com/16705946/124573473-7ad5bb00-de41-11eb-9acd-698c04aee3b5.png)

**Pull requests welcome!**

oliver-sanders:
@wxtim why does this have the blocked label?

MetRonnie:
Removing the blocked label as there doesn't seem to be a reason for it

wxtim:
I have no idea - September was 5 months ago!

oliver-sanders:
Converted the last point into an issue.

Closing this as superseeded by the issues that replace it, #608, #922

PR #740, opened 2021-08-27
start progress 'clock' at ~20% to differentiate submitted from running.

wxtim:
These changes partially address #701 

Edit - this really only answers the first bullet point of the ticket, and may be more than a "tick or two". 

The main outcome from my point of view is that I've had a play with Vue and got my hands dirty.

N.B. I was thinking about what it might look like to the visually impaired, and was pleased with the results of pressing ctrl+ multiple times.

**Requirements check-list**
- [x] Does not need tests (manual testing? Visual feature).
- [x] No change log entry required (beta).

### DIFF
--- a/src/styles/cylc/_task.scss
+++ b/src/styles/cylc/_task.scss
@@ -18,7 +18,7 @@
 @keyframes c-task-progress-animation {
   from {
     /* 0% progress */
-    stroke-dashoffset: 157;
+    stroke-dashoffset: 150;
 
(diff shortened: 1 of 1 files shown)


kinow (approved):
Looks good to me @wxtim ! Later we can change the icon in the filter as per the issue linked too. Thanks!

datamel (approved):
Looks good, thanks @wxtim

PR #749, opened 2021-09-02
Create delta callbacks, control execution order of added, updated, pruned

### LINKED ISSUE #747: UI reporting incorrectly that a delta is not in the lookup
**Describe the bug**

From Element, reported today by @dwsutherland . In the screenshot below, it's possible to see an alert error message for a delta missing from the lookup. But since we have the updated and pruned deltas, both in the same delta, it should be already in the lookup map.

Unless... it never arrived in an added delta? But in this case **maybe** we are adding nodes in updated too? Will have to confirm.

`master`

**Expected behavior**

**Screenshots**

image

**Additional context**

**Pull requests welcome!**

kinow:
These changes close #747 

Used the workflow provided by @hjoliver to reproduce and fix the issue. See screenshot of the UI after fixing the issue, no browser console, no alerts :ok_man: 

GIFrecord_2021-09-03_150458

**Requirements check-list**

### DIFF
--- a/src/services/workflow.service.js
+++ b/src/services/workflow.service.js
@@ -16,7 +16,6 @@
  */
 
 import isEqual from 'lodash/isEqual'
-import unionBy from 'lodash/unionBy'
 import ViewState from '@/model/ViewState.model'
 import Subscription from '@/model/Subscription.model'
 import {
@@ -332,7 +331,11 @@ class WorkflowService {
       baseSubscriber.query.query = mergeQueries(baseSubscriber.query.query, subscriber.query.query)
       // Combine the arrays of callbacks, creating an array of unique callbacks.
       // The callbacks are compared by their class/constructor name.
-      subscription.callbacks = unionBy(subscription.callbacks, subscriber.query.callbacks, (callback) => callback.constructor.name)
+      for (const callback of subscriber.query.callbacks) {
+        if (!subscription.callbacks.find(element => element.constructor.name === callback.constructor.name)) {
+          subscription.callbacks.push(callback)
+        }
+      }
     }
     const finalQuery = print(baseSubscriber.query.query)
     // TODO: consider using a better approach than print(a) === print(b)


kinow:
I finished the core changes, pending to fix the tests. But before that I will think about it over the weekend, and see if there are other ways to solve the #747 bug.

We have 2 callbacks for the Tree view. The first callback populates a workflow lookup, used by Tree view (and Graph view, and Table view, in the future). The second callback creates the workflow tree structure, where the nodes of the tree contain pointers to the workflow lookup created by the first callback.

The #747 bug happens when we have processed the first callback, applying some pruned deltas that removed nodes from the global lookup. And then the second callback has the same nodes in the updated delta, but now it fails to locate the nodes that have been just removed.

This pull request splits the execution of deltas, creating a common interface for each delta callback, where we have, in order:

- `before()`
- `onAdded()`
- `onUpdated()`
- `onPruned()`
- `after()`
- `commit()`
- `tearDown()`

For each delta, then, we have in pseudo-code:

```
for (callback of callbacks) {
  callback.before(deltasData)
  callback.onAdded(deltasData)
  callback.onUpdated(deltasData)
  callback.commit()
}
for (callback of [...callbacks].reverse()) {
  callback.onPruned()
  callback.commit()
}
```

And when the component is destroyed, `tearDown` is called to clear the Vuex store.

The second for is fixing this bug, by applying pruned deltas in reverse order… we add & update everything, everywhere. Then we can safely remove the pruned data. This way we always apply every delta, with no risk of trying to access data that was already removed.

kinow:
Compilation errors fixed, now just need to fix these 32 unit tests that are broken... errr.. :coffee: :dizzy_face:

hjoliver:
Only 32! try messing with the core of cylc-flow ... :sob:

src/services/workflow.service.js
```diff
@@ -292,18 +317,15 @@ class WorkflowService {
...
-        subscription.actionNames = union(subscription.actionNames, subscriber.query.actionNames)
-        subscription.tearDownActionNames = union(subscription.tearDownActionNames, subscriber.query.tearDownActionNames)
-      })
+    for (const subscriber of subscribers.slice(1)) {
+      if (!isEqual(subscriber.query.variables, baseSubscriber.query.variables)) {
+        throw new Error('Error recomputing subscription: Query variables do not match.')
+      }
+      this.mergeQueries(baseSubscriber.query.query, subscriber.query.query)
+      subscription.callbacks = union(subscription.callbacks, subscriber.query.callbacks)
+    }
```
kinow, line 328: for-loops are easier to debug :grimacing:

src/views/Dashboard.vue
```diff
@@ -199,8 +200,9 @@ export default {
         DASHBOARD_DELTAS_SUBSCRIPTION,
         {},
         'root',
-        ['workflows/applyWorkflowsDeltas'],
-        []
+        [
+          new GScanCallback()
+        ]
```
kinow, line 205: I think this is clearer. Before it was using the same Vuex action that is used by GScan. But now it's even clearer that for now the Dashboard and WorkflowsTable are re-using the query & callbacks of GScan.

src/services/workflow.service.js
```diff
@@ -292,18 +317,15 @@ class WorkflowService {
...
-        this.mergeQueries(baseSubscriber.query.query, subscriber.query.query)
-        subscription.actionNames = union(subscription.actionNames, subscriber.query.actionNames)
-        subscription.tearDownActionNames = union(subscription.tearDownActionNames, subscriber.query.tearDownActionNames)
-      })
+    for (const subscriber of subscribers.slice(1)) {
+      if (!isEqual(subscriber.query.variables, baseSubscriber.query.variables)) {
+        throw new Error('Error recomputing subscription: Query variables do not match.')
+      }
+      this.mergeQueries(baseSubscriber.query.query, subscriber.query.query)
+      subscription.callbacks = union(subscription.callbacks, subscriber.query.callbacks)
```
kinow, line 327: Note to self: manual test, and confirm this is working OK.

kinow:
Just pending some manual testing, review if I didn't miss any parts of the code to be tested, and then it should be ready for review :sleepy: but will probably leave the manual test for tomorrow morning.

src/services/workflow.service.js
kinow: Ha! I knew it. Quickly wrote a unit test to verify it.

image

Let me fix it, and make the test permanent.

kinow:
Tested locally with some unit tests, found one bug, fixed. Tested locally again with `cylc hub` and 4 workflows, no issues found :tada: Had 4 workflows running, with and w/o hierarchy, and with and w/o families. I believe I found a bug, but in cylc-flow (will confirm & report later).

Ready for review! :tada:

kinow:
Rebased.

hjoliver:
(Testing this now, next major review priority)

hjoliver:
Hmm, opening a second tree view on this branch causes the tree to freeze :thinking:

hjoliver:
(although that doesn't seem to be the case in your screen recording above!)

kinow:
I've been keeping this PR up to date with master, maybe something changed and caused the freeze. Will try to reproduce in a bit (or in the morning, on a Windows computer trying to get a nvidia computer board to boot 🤓 )

kinow:
But if you are investigating it, what I would look at are

- anything in the browser console?
- does it matter if you try on a different browser?
- or with a different workflow?
- any chance there's cache?
- nothing in the UIS or flow logs?

Then I'd start looking at vue dev utils to see what's the state of the Tree view (literally, look for the Tree component/view, which must contain a TreeItem which is the top of the tree). And also look at the Network tab to confirm the subscription is going OK. Then from there it'd be the IDE to see what could be wrong with the code.

kinow:
Working OK for me @hjoliver 

1. checked out branch (double-checked it was up to date)
2. `yarn install`
3. `yarn run build:watch`
4. Started UIS & `five`

GIFrecord_2021-09-21_091324

hjoliver:
Looks good. Does it carry on to the next cycle points though, with multiple tabs open?

kinow:
> Looks good. Does it carry on to the next cycle points though, with multiple tabs open?

I think so. Two trees, CP 180.

image

Some seconds later, CP 182.

image

I found why it's restarting the subscription, but that's a bug in master too. Will create an issue for that one and start preparing a fix and an e2e test to prevent that regression :+1:

kinow:
If you would like I can try with the same workflow you are using and see if it works?

hjoliver:
Huh, strange (but good).  I've tried with several workflows. Let me just nuke everything and start from scratch again...

kinow:
In case it helps, heres `five`

```cylc
[scheduling]
  cycling mode = integer
  initial cycle point = 1
  [[queues]]
     [[[default]]]
       limit = 1
  [[graph]]
    R1 = "prep => foo"
    P1 = "foo[-P1] => foo => bar"

[runtime]
  [[root]]
      script="sleep 5"
  [[prep]]
  [[foo]]
  [[bar]]

```

kinow:
@hjoliver shared his screen showing what this PR looks like.

1. `yarn run build`
2. start UIS + `five`
3. add a second Tree View

Now the tree views are both frozen, not being updated. The network tab shows 2 GraphQL subscriptions, `1` being GScan as always, and `3` the tree views. Normally the ID for the tree view would be `2`, so I guess the subscription was reloaded but then the UI got confused somehow.

We tested on Chrome and Firefox on Linux.

hjoliver:
(Result: the bug was only triggered in production builds, something to do with lodash's `UnionBy()`)

hjoliver:
I can confirm the latest commit fixes it :+1:

kinow:
@hjoliver sent me a workflow that shows errors due to missing parents. It happens when a task cannot find the parent family proxy, and also when a job cannot find the parent task.

I've created a gist with the deltas collected, the `diff` used to fire the debugger and to log more info, and also to remove GScan to have one less subscription to filter: https://gist.github.com/kinow/817205bbbef7684722fce5cce1939764

One parent task only appears as updated, skipping the added status. But I'm still investigating it that's the cause of the errors with that workflow.

kinow:
(rebased)

oliver-sanders (approved):
> I'm still investigating

Not sure what the status of this is.

The code looks good, I ran some local tests and everything worked fine.

kinow:
>Not sure what the status of this is.

Just pending review.

>The code looks good, I ran some local tests and everything worked fine.

Thanks! @hjoliver or @dwsutherland can sign it off then (I think they were the first ones to report the issue)

PR #802, opened 2021-10-11
Update GScan incrementally, using tree/lookup

kinow:

The PR to propagate states in GScan has two changes. One where we propagate states, and another where we replace the `computed` property by a Tree+Lookup approach to avoid re-creating the `computed` property any time something changes in GScan, instead building the tree incrementally.

To simplify the other PR, I have moved the latter over to this PR, so we can compare before/after and merge it first, especially since I didn't get good performance with the propagate states part (which means it could take a bit longer for that change to be ready for review.)

**Requirements check-list**

### DIFF
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -241,9 +241,9 @@ export default {
     }
   },
   computed: {
-    ...mapState('workflows', ['workflows']),
+    ...mapState('workflows', ['lookup']),
     workflowsTable () {
-      const count = Object.values(this.workflows)
+      const count = Object.values(this.lookup)
         .map(workflow => workflow.status)
         .reduce((acc, state) => {
           acc[state] = (acc[state] || 0) + 1
--- a/src/views/WorkflowsTable.vue
+++ b/src/views/WorkflowsTable.vue
@@ -132,9 +132,9 @@ export default {
     }
   }),
   computed: {
-    ...mapState('workflows', ['workflows']),
+    ...mapState('workflows', ['lookup']),
     workflowsTable () {
-      return Object.values(this.workflows)
+      return Object.values(this.lookup)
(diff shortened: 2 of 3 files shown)


kinow:
Yesterday: 18 failed tests
Today: 10 failed tests :sweat_smile:

kinow:
Unit tests fixed locally. Now figure out how to fix these conflicts :cold_sweat:

kinow:
One brittle test. When we visit URL's, and then click on buttons, it's taking too long to load the subscription (sometimes) and then the test is failing as it did for firefox :point_up: will investigate tomorrow.

MetRonnie:
@oliver-sanders I'm guessing this will be closed now?

oliver-sanders:
I think we'll have to work out how to pull the propagate states logic out onto the redesigned store. I think we could probably put that logic into a computed property on the GScan component?

oliver-sanders:
Since the data store refactor we now have a universal tree containing both the GScan (workflow) and Tree (task) parts. This is updated incrementally using the global-callback.

So the workflow tree is being computed incrementally, but GScan is still using a computed property rather than iterating over this directly because of sorting and filtering.

In order to sort/filter incrementally it is still necessary to write a callback for GScan using the data from the lookup as done on this branch. The callbacks here are still valid, `lookup` has become `$index` but otherwise it's the same. Ideally we would avoid callbacks as these duplicate data processing and require loads of maintenance because they are bespoke to each view.

Because GScan is the only component (so far) that is worried about workflow iteration order, we could consider changing the sort order of the central store to what GScan expects, then use `v-if` to handle filtering. That would cut out the computed property completely.

Alternatively, we could consider maintaining a separate index in the central store just for GScan (e.g. `$gscan`).

PR #805, opened 2021-10-11
removed the installed state

oliver-sanders:
PR trio: https://github.com/cylc/cylc-flow/pull/4459, https://github.com/cylc/cylc-uiserver/pull/256, https://github.com/cylc/cylc-ui/pull/805

* The installed state was the same as "stopped" but with the additional
  information that the workflow had not yet been run.
* Added this information to the workflow status message in cylc-uiserver
  (as it is offline information).
* Improved the workflow status message provided by cylc-flow.
* Put the status message into the UI toolbar.
    workflow-status-message

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [ ] Appropriate tests are included (unit and/or functional).
- [x] No change log entry required (why? e.g. invisible to users).
- [x] No documentation update required.

PR #806, opened 2021-10-13
Toolbar play from stopped

oliver-sanders:
built on #805

This completes the play-pause/resume-stop sequence in the workflow toolbar.

**Requirements check-list**
- [ ] Appropriate tests are included (unit and/or functional) - hard to test with mocked data

### DIFF
--- a/src/components/cylc/workflow/Toolbar.vue
+++ b/src/components/cylc/workflow/Toolbar.vue
@@ -53,9 +53,20 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
       </v-icon>
 
       <v-icon
+        id="workflow-play-button"
+        color="#5E5E5E"
+        :disabled="!enabled.playToggle"
+        v-if="!isRunning"
+        @click="onClickPlay"
+      >
+        {{ svgPaths.run }}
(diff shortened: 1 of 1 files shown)


oliver-sanders:
Testing this sort of thing without live data is going to be a bit tricky/limited, can have a go, will take a look next week.

Long term I think a system integration test suite for cylc-flow, cylc-uiserver and cylc-ui with Cypress might be a good way to go.

Would be good to slide this into the beta if possible so may need to go on manual testing for now and bump to an issue.

oliver-sanders:
Rebased, added changelog.

hjoliver (approved):
LGTM

datamel (approved):
I have checked out the branch and tested the change. Looks great, thanks @oliver-sanders. 
The change is not covered by a test, like you said it is a hard one to test and I'm not quite sure how I would go about it so I am merging having manually tested, I could open an issue for the testing if you think it would be fitting?

oliver-sanders:
I think we have an issue on cylc-admin for system testing.

My best idea ATM is to use Cypress to test the UI but using live data provided by a UIS.

PR #824, opened 2021-10-29
Handle the preparing job state properly.

hjoliver:

Togehter with the datastore tweak in https://github.com/cylc/cylc-flow/pull/4482 this makes the UI handle the `preparing` task and job states properly.

shot

**Requirements check-list**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,9 @@ to a Jupyter Server extension).
 
 ### Fixes
 
+[#824](https://github.com/cylc/cylc-ui/pull/844) - Fix handling of task/job
+"preparing" state.
+
 -------------------------------------------------------------------------------
 ## __cylc-ui-0.5 (<span actions:bind='release-date'>Released 2021-07-28</span>)__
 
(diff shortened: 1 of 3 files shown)


hjoliver:
> Preparing isn't supposed to be a job state right?
> 
> I don't think this makes sense - moved to element chat.

See response here: https://github.com/cylc/cylc-flow/pull/4482#issuecomment-955766865

hjoliver:
@oliver-sanders is correct that "preparing" is not supposed to be a job state, but (with reference to my reponse linked to above):

This PR un-breaks the current UI, because "preparing" currently IS a job state to the datastore that feeds the UIs.

And, this is a trivial change:
- one line of code (add a job state to match the received data)
- a few lines of CSS to style that job state

So I suggest we merge this, and I'll post an Issue for a proper fix via the datastore.

oliver-sanders:
Could become more difficult to unpick these changes at a later date, we should check that a simple `if job.status == TASK_STATUS_PREPARING` wouldn't be an easier fix?

hjoliver:
> Could become more difficult to unpick these changes at a later date 

This particular molehill seems to have turned into a mountain somehow :grin:  It's literally one line of code and a bit of CSS, and it unbreaks the UI for the current datastore, which seemed a slightly more complicated thing to fix in a hurry. All I wanted was to get the GUI and TUI to correctly show the data they were receiving in order to be able to compare with task state changes in the scheduler exactly. But at this point I don't mind if we do fix the datastore straight off and supersede this, so long as it's quick.

hjoliver:
Closing as superseded by cylc/cylc-flow#4485

PR #842, opened 2021-11-10
actions: fix release stage one (again)

oliver-sanders:
Hopefully this'll do it?!


MetRonnie (approved):
Just had a look and the same thing will be needed for cylc-uiserver

oliver-sanders:
And metomi-rose!

PR #861, opened 2021-11-29
Updates to View filtering, and table display and functionality

### LINKED ISSUE #631: Workflow Table View component

### LINKED ISSUE #853: Instant filtering (by task name/state)

Instead of having to click a button to implement the filter, why not instantly filter? (For the task name, which is typed out, this would have to be debounced)

Marked as question in case there is a reason not to do this.

**Pull requests welcome!**

AaronDCole:
This merge request updates the design and functionality of the table and tree views. It includes;
- removed the requirement to manually 'submit' filtering options in the filter bar
- fixed the filter bar to the top of the Lumino container in the table view
- added a new column with a dropdown toggle to the task rows (if they have current jobs)
- modified the table data store to include information about the jobs relating to the current task (and add, update and remove as instructed via delta changes)
- modified the table to list the task jobs' when the jobs dropdown toggle is applied
- fixed the display of the 'layered' task status icon
- added the job status icon and submission number in the job rows
- updated the respective tests

These changes partially address #471
These changes:
  - close #853
  - close #631

**Requirements check-list**

### DIFF
--- a/src/components/cylc/table/Table.vue
+++ b/src/components/cylc/table/Table.vue
@@ -315,9 +315,9 @@ export default {
             valueB = typeof taskB.node.name !== 'undefined' ? taskB.node.name : ''
             return sortDesc ? valueB.localeCompare(valueA) : valueA.localeCompare(valueB)
           case 'Cycle Point':
-            valueA = taskA.node.cyclePoint !== '' && typeof taskA.node.cyclePoint !== 'undefined' ? (new Date(taskA.node.cyclePoint)).getTime() : 0
-            valueB = taskB.node.cyclePoint !== '' && typeof taskB.node.cyclePoint !== 'undefined' ? (new Date(taskB.node.cyclePoint)).getTime() : 0
-            return sortDesc ? valueB - valueA : valueA - valueB
+            valueA = taskA.node.cyclePoint !== '' && typeof taskA.node.cyclePoint !== 'undefined' ? String(taskA.node.cyclePoint) : ''
+            valueB = taskB.node.cyclePoint !== '' && typeof taskB.node.cyclePoint !== 'undefined' ? String(taskB.node.cyclePoint) : ''
+            return sortDesc ? valueB.localeCompare(valueA) : valueA.localeCompare(valueB)
           case 'Jobs':
             valueA = typeof taskA.jobs !== 'undefined' ? taskA.jobs.length : 0
             valueB = typeof taskB.jobs !== 'undefined' ? taskB.jobs.length : 0
--- a/src/components/cylc/table/Table.vue
+++ b/src/components/cylc/table/Table.vue
@@ -17,24 +17,21 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 <template>
   <v-container
-    class="c-table ma-0 pa-2"
+    class="c-table ma-0 pa-2 h-100 flex-column d-flex"
   >
     <!-- Toolbar -->
     <v-row
-      no-gutters
-        class="d-flex flex-wrap"
+        class="d-flex flex-wrap table-option-bar no-gutters flex-grow-0"
     >
       <!-- Filters -->
       <v-col
         v-if="filterable"
-        class="grow"
+        class=""
       >
-        <v-row
-          no-gutters
-        >
+        <v-row class="no-gutters">
           <v-col
             cols="12"
-            md="5"
+            md="6"
             class="pr-md-2 mb-2 mb-md-0"
           >
             <v-text-field
@@ -46,12 +43,13 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               outlined
               placeholder="Filter by task name"
               v-model.trim="tasksFilter.name"
+              @change="filterTasks"
             ></v-text-field>
           </v-col>
           <v-col
             cols="12"
-            md="5"
-            class="pr-md-2 mb-2 mb-md-0"
+            md="6"
+            class="mb-2 mb-md-0"
           >
             <v-select
               id="c-table-filter-task-states"
@@ -64,6 +62,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               outlined
               placeholder="Filter by task state"
               v-model="tasksFilter.states"
+              @change="filterTasks"
             >
               <template v-slot:item="slotProps">
                 <Task :status="slotProps.item.value"/>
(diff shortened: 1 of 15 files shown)


AaronDCole:
rebuild

hjoliver:
Looks good @AaronDCole.

Since you make the filter bar sticky on the table view, and got rid of the Filter button on both views, could you also make it sticky on the tree view?

Also, column sort (via clicking on the heading) would be an uncontroversial thing to get working before the (yet-to-be-arranged) meeting on the table view.

hjoliver:
Also, job expansion would probably be better on the left, near the task icons, rather than off to the far right.

src/components/cylc/table/Table.vue
```diff
@@ -239,9 +304,73 @@ export default {
...
+            valueB = typeof taskB.latestJob.platform !== 'undefined' ? taskB.latestJob.platform : ''
+            return sortDesc ? valueB.localeCompare(valueA) : valueA.localeCompare(valueB)
+          case 'Job System':
+            valueA = typeof taskA.latestJob.jobRunnerName !== 'undefined' ? taskA.latestJob.jobRunnerName : ''
+            valueB = typeof taskB.latestJob.jobRunnerName !== 'undefined' ? taskB.latestJob.jobRunnerName : ''
+            return sortDesc ? valueB.localeCompare(valueA) : valueA.localeCompare(valueB)
+          case 'Job ID':
+            valueA = typeof taskA.latestJob.jobId !== 'undefined' ? taskA.latestJob.jobId : 0
+            valueB = typeof taskB.latestJob.jobId !== 'undefined' ? taskB.latestJob.jobId : 0
+            return sortDesc ? valueB - valueA : valueA - valueB
```
hjoliver, line 336: As I recall, job IDs are not necessarily numerical for batch systems like PBS and Slurm (gotta check though...)

src/components/cylc/table/Table.vue
```diff
@@ -239,9 +304,73 @@ export default {
...
+          const sortDesc = this.sortDesc[this.sortBy.indexOf(sortByProperty)]
+          switch (sortByProperty) {
+          case 'Task':
+            valueA = typeof taskA.node.name !== 'undefined' ? taskA.node.name : ''
+            valueB = typeof taskB.node.name !== 'undefined' ? taskB.node.name : ''
+            return sortDesc ? valueB.localeCompare(valueA) : valueA.localeCompare(valueB)
+          case 'Cycle Point':
+            valueA = taskA.node.cyclePoint !== '' && typeof taskA.node.cyclePoint !== 'undefined' ? (new Date(taskA.node.cyclePoint)).getTime() : 0
+            valueB = taskB.node.cyclePoint !== '' && typeof taskB.node.cyclePoint !== 'undefined' ? (new Date(taskB.node.cyclePoint)).getTime() : 0
+            return sortDesc ? valueB - valueA : valueA - valueB
```
hjoliver, line 320: Date-time cycle points are ISO8601 format, e.g. `20240502T00+1200` - can `Date().getTime()` cope with that?

hjoliver (commented):
Partially reviewed. Looks good, a couple of minor comments so far.

src/components/cylc/table/Table.vue
hjoliver: Alphanumeric sort is probably fine, because the format is the same for all tasks in a given workflow.

hjoliver: Just checked, and cyclepoint sorting seems not to be working for a datetime cycling workflow.

hjoliver (approved):
Looks good to me!

src/components/cylc/table/Table.vue
wxtim: Definately correct about Slurm:

```
sacct -a  -S 09:00 -o "job" | python -c '''
import sys
for line in sys.stdin.readlines():
    try:
        int(line)
    except:
        print(line)
'''
```
gives a load of answers such as
```
123456.bat+   # ch?
123456.0
```

Slurm job array would also probably produce job ids in the form `123456_42`

MetRonnie:
I've noticed when clicking the "x" in the text field to clear the filtering by task name, the view doesn't update

MetRonnie:
A couple more issues:
- Vertical scrollbars have gone missing from the Tree and Table views
- Clicking on the column defaults to sorting in descending order, which is counter-intuitive. Also I think an arrow icon makes more sense than a chevron for indicating sorting (chevron is better suited for expand/collapse imho)
  
  image

But overall a nice improvement 👍

src/components/cylc/table/Table.vue
wxtim: You can ignore the last comment - I don't think that we support slurm job array right now. Trying to use it may well confuse Cylc.

wxtim:
I've removed myself as a reviewer, I got sidetracked into playing with Slurm Job Arrays.

AaronDCole:
> I've noticed when clicking the "x" in the text field to clear the filtering by task name, nothing happens / the view doesn't update.
> 
> 
> 
> ~Also the job colours aren't always correct in the task row - I'm seeing turquoise and orange for some reason~ Actually this is only the case for the mock data when you do `yarn serve`
> 
> 

I cant actually reproduce this, and I 'think' the orange color might be intentional. The other issues were all valid and Ill push fixes to them in a few minutes.

src/components/cylc/table/Table.vue
```diff
@@ -82,70 +84,128 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-            class="subheading font-weight-light text-success text--darken-3"
-            v-text="header.text"
-          />
-      </template>
-      <template
-          slot="item"
-          slot-scope="{ item }"
+      <v-col
+        cols="12"
+        class="overflow-y-scroll mh-100 position-relative"
```
MetRonnie, line 96: I don't think this is needed. I tried removing it and the scrollbars appear automatically when the content overflows
```suggestion
        class="mh-100 position-relative"
```

src/components/cylc/tree/Tree.vue
```diff
@@ -132,25 +123,34 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-        :hoverable="hoverable"
-        :initialExpanded="expanded"
-        v-on:tree-item-created="onTreeItemCreated"
-        v-on:tree-item-destroyed="onTreeItemDestroyed"
-        v-on:tree-item-expanded="onTreeItemExpanded"
-        v-on:tree-item-collapsed="onTreeItemCollapsed"
-        v-on:tree-item-clicked="onTreeItemClicked"
+      <v-col
+        cols="12"
+        class="overflow-y-scroll mh-100 position-relative"
```
MetRonnie, line 132:
```suggestion
        class="mh-100 position-relative"
```

src/styles/index.scss
```diff
@@ -59,3 +59,39 @@
 .c-interactive:hover {
     cursor: pointer;
 }
+
+.overflow-y-scroll {
+  overflow-y: scroll;
+}
+
```
MetRonnie, line 66: See above
```suggestion
```

MetRonnie (approved):

MetRonnie:
Strange thing happened when testing this out. I was running a simple workflow and triggering `*/foo` a few times using the UI when

> Received a Table delta before the workflow initial data burst

image

MetRonnie:
Ok so
- the "Received a Table delta before the workflow initial data burst" error I also get on the table view for master
- the workflow shutting down seems to be a bug in cylc-flow

PR #928, opened 2022-02-16
Add mutations to Gscan workflow icons.

### LINKED ISSUE #867: make mutations available in all places
"Cylc Objects" are things that have Cylc ID's such as workflows, cycles, tasks & jobs.

All Cylc objects can be made "mutable" by adding the `cylc-object` directive to the relevant component.

This adds the on-click menu which is automatically propagated with the appropriate mutations.

The tree view currently implements mutations for all "Cylc Objects", however, I don't think other views do.

We need to go through and add the directive where appropriate, taking care to avoid any conflict with other on-click interactivity & tooltips.

**Pull requests welcome!**

wxtim:
These changes close #867 

Massive thanks to @MetRonnie for helping me get my head around this one.

**Requirements check-list**

### DIFF
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -113,7 +113,10 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           class="c-gscan-workflow ma-0 pa-0"
         >
           <template v-slot:node="scope">
-            <span class="mr-2">
+            <span
+              class="mr-2"
+              v-if="scope.node.type === 'workflow'"
+            >
               <workflow-icon
                 :status="scope.node.node.status"
                 :statusMsg="scope.node.node.statusMsg"
@@ -135,7 +138,6 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                     v-else-if="scope.node.type === 'workflow'"
                     class="c-gscan-workflow-name"
                   >
-
                     <v-tooltip top>
                       <template v-slot:activator="{ on }">
                         <span v-on="on">{{ scope.node.name }}</span>
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -113,9 +113,15 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           class="c-gscan-workflow ma-0 pa-0"
         >
           <template v-slot:node="scope">
+            <span class="mr-2">
+              <workflow-icon
+                :status="scope.node.node.status"
+                :statusMsg="scope.node.node.statusMsg"
+                v-cylc-object="scope.node.node"
+              />
+            </span>
             <v-list-item
               :to="workflowLink(scope.node)"
-              :class="getWorkflowClass(scope.node.node)"
             >
               <v-list-item-title>
                 <v-layout align-center align-content-center d-flex flex-nowrap>
@@ -129,13 +135,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                     v-else-if="scope.node.type === 'workflow'"
                     class="c-gscan-workflow-name"
                   >
-                    <span class="mr-2">
-                      <workflow-icon
-                        :status="scope.node.node.status"
-                        :statusMsg="scope.node.node.statusMsg"
-                        v-cylc-object="scope.node.node"
-                      />
-                    </span>
+
                     <v-tooltip top>
                       <template v-slot:activator="{ on }">
                         <span v-on="on">{{ scope.node.name }}</span>
@@ -374,12 +374,6 @@ export default {
     }
   },
   methods: {
-    getWorkflowClass (node) {
-      return {
-        'c-workflow-stopped': node && node.status && node.status === WorkflowState.STOPPED.name
-      }
-    },
-
     filterHierarchically,
 
     /**
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -133,6 +133,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                       <workflow-icon
                         :status="scope.node.node.status"
                         :statusMsg="scope.node.node.statusMsg"
+                        v-cylc-object="scope.node.node"
                       />
                     </span>
                     <v-tooltip top>
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -113,16 +113,12 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           class="c-gscan-workflow ma-0 pa-0"
         >
           <template v-slot:node="scope">
-            <span
+            <workflow-icon
               class="mr-2"
               v-if="scope.node.type === 'workflow'"
-            >
-              <workflow-icon
-                :status="scope.node.node.status"
-                :statusMsg="scope.node.node.statusMsg"
-                v-cylc-object="scope.node.node"
-              />
-            </span>
+              :status="scope.node.node.status"
+              v-cylc-object="scope.node.node"
+            />
             <v-list-item
               :to="workflowLink(scope.node)"
             >
--- a/tests/e2e/specs/gscan.js
+++ b/tests/e2e/specs/gscan.js
@@ -22,15 +22,14 @@ describe('GScan component', () => {
       .get('.c-gscan-workflow')
       .should('have.length', 1)
   })
+
   it('should filter by workflow name', () => {
-    cy.visit('/#/')
-    cy
-      .get('#c-gscan-search-workflows')
+    cy.get('#c-gscan-search-workflows')
       .type('abc')
-    cy
       .get('.c-gscan-workflow')
       .should('have.length', 0)
   })
+
   it('should filter by workflow state', () => {
     cy.visit('/#/')
     cy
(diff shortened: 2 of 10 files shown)


src/components/cylc/gscan/GScan.vue
```diff
@@ -133,6 +133,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                       <workflow-icon
                         :status="scope.node.node.status"
                         :statusMsg="scope.node.node.statusMsg"
+                        :info="scope.node.node"
```
oliver-sanders, line 136: Guessing there's a reason we can't add the directive on the `workflow-icon` component directly?

wxtim: Not sure, will check if that's possible.

oliver-sanders: Might be the tooltip, we should remove that now, when you open the menu does it show the workflow status at the top of the menu?

wxtim: It could be, but the menu doesn't include the workflow status.

wxtim: Have a look at https://github.com/cylc/cylc-ui/pull/928/commits/d2dfbd52eb5f8d665defe0a7cc7b2d7f388a10e2 -

MetRonnie:
Can you rebase onto master/merge master into this, then I will open a PR against this with an e2e test

wxtim:
> Can you rebase onto master/merge master into this, then I will open a PR against this with an e2e test

Done

MetRonnie (changes requested):
Opened a PR against this at https://github.com/wxtim/cylc-ui/pull/2

The new test has uncovered a bug: if you run it you will see that the menu doesn't appear if you click on a workflow icon that isn't the current workflow

tests/e2e/specs/gscan.js
```diff
@@ -22,15 +22,14 @@ describe('GScan component', () => {
       .get('.c-gscan-workflow')
       .should('have.length', 1)
   })
+
   it('should filter by workflow name', () => {
-    cy.visit('/#/')
```
oliver-sanders, line 26: @MetRonnie could you explain this diff.

MetRonnie: As this test can be carried out on the state of the UI after the success of the previous test, it is quicker not to reload the page

oliver-sanders: So Cypress maintains its state from one test to the next? Is this just for "it" declarations or also "describe"? Will this test still work if it is change to "it.only"?

MetRonnie: It seems to me it maintains the state during a whole test file. I think you are only allowed 1 top-level `describe`, but any sub-`describe`s probably have the same state. Unless you do `cy.visit()`. I suspect it won't work if you do `it.only`, but if you're doing that you could just as easily add `cy.visit()` as well?

MetRonnie: Unfortunately I don't think you're allowed `cy.visit()` outside an `it()`. There are some tests where it is useful to have inter-dependent tests because the behaviour relies on past interactions, particularly with the mutations menu (e.g. checking it is collapsed when you open it after having closed it while expanded), for those I think it would be somewhat a pain to have to duplicate getting it into a particular state when the earlier tests already got it into that state.

I don't think `--watch` is a problem though? When I run the e2e tests it hot reloads when I save the file just fine

wxtim: > The new test has uncovered a bug: if you run it you will see that the menu doesn't appear if you click on a workflow icon that isn't the current workflow

This bug is more specific than that - It only occurs if you try to press a workflow icon in the dashboard. You can in fact control other workflows as long as the current view isn't the dashboard!

oliver-sanders:
Javascript events "bubble" up the DOM tree which can result on multiple actions being called as a result of a single click.

The way to handle this is to manually tell JS to stop the propagation at the appropriate level.

Try clicking both hello and world in this example (results in the console) then comment out the `event.stopPropagation()` line.

```html
<html>
  <head></head>
  <body>
    <a onclick="anc()">
      hello
      <span onclick="span(event)">world</span>
    </a>
    <script type="text/javascript">
      function anc() {
        console.log('anc')
      }
      function span(event) {
        console.log('span')
        event.stopPropagation()
      }
    </script>
  </body>
</html>
```

Easy to implement in a toy example, hard to figure out when you're using complex Vuetify components.

Easy way to sidestep might be to move the icon out of the title and into an ["avatar"](https://dev.vuetifyjs.com/en/api/v-list-item-avatar/).

wxtim:
> Easy way to sidestep might be to move the icon out of the title and into an ["avatar"](https://dev.vuetifyjs.com/en/api/v-list-item-avatar/).

Did something like that in my local branch. Note that the avatar doesn't have a `:to` prop...

oliver-sanders:
> Note that the avatar doesn't have a :to prop...

The `:to` attribute comes from the `router-link` component which we are free to use standalone.

https://router.vuejs.org/api/

wxtim:
None of the arguments to be sent to the mutations menu by
```
    vnode.context.$eventBus.emit('show-mutations-menu', {
      id: cylcId,
      type: type,
      types: types,
      tokens: tokens,
      mutations: componentMutations,
      node: binding.value,
      event: e
    })
 ```
 
 Are actually set

**Comment by wxtim** [comment]
I've come to the same conclusion as ronnie.

**Comment by wxtim** [comment]
> What level is the menu component currently at? I wonder if we need to bump it up the tree a bit?

File tree? DOM tree?

**Comment by MetRonnie** [comment]
Possible lead:

Doing
```js
document.querySelectorAll('.v-menu')
```
while on a workflow view returns 4 divs, whereas on the dashboard it only returns 1 div. It could be that the `vnode` that the `show-mutations-menu` event is being emitted on doesn't exist when on the dashboard view?

Note: the divs with class `v-menu` are empty because 
https://github.com/cylc/cylc-ui/blob/73958fb951d3c183683e64fb36ae7bb296f943a0/src/components/cylc/cylcObject/Menu.vue#L34-L37

MetRonnie:
I'm starting to think the `CylcObjectMenu` component is not mounted when in the dashboard view, so there is nothing to receive the `show-mutations-menu` event

wxtim:
> I'm starting to think the `CylcObjectMenu` component is not mounted when in the dashboard view, so there is nothing to receive the `show-mutations-menu` event

Yes, trying...

MetRonnie:
I've cracked it!

oliver-sanders:
Presumably we need to lift the CylcObjectMenu component up a level in the tree?

wxtim:
IMO this is now a Tim and Ronnie show, and we should ask @AaronDCole to be 2nd Reviewer

src/views/Dashboard.vue
```diff
@@ -21,6 +21,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
     grid-list
     class="c-dashboard mt-4"
   >
+    <CylcObjectMenu/>
```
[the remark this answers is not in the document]
wxtim: @MetRonnie - I'd like to have a go at doing this if you don't mind.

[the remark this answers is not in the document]
MetRonnie: Ah wait, got it

src/utils/aotf.js
```diff
@@ -68,7 +69,8 @@ export const mutationIcons = {
   resume: mdiPlay,
   setOutputs: mdiGraph,
   stop: mdiStop,
-  trigger: mdiCursorPointer
+  trigger: mdiCursorPointer,
+  clean: mdiBroom
```
MetRonnie, line 73: Crept in to the wrong PR?

Besides, I think we might want to consider changing the icon for remove to e.g. `mdiMinusCircleOutline` and using the trash can icon `mdiDelete` or `mdiDeleteSweep` for clean

AaronDCole:
This looks good and seems to work, Im just checking if I can see if there any any edge cases now

AaronDCole (approved):
This may cause some issues with my PR re collapsing single task workflows by default as Ive changed a fair amount of the markup in the cylc/gscan/GScan.vue and cylc/tree/TreeItem.vue, but ill sort them out seperately

src/utils/aotf.js
wxtim: It was deliberate - open to discussing changing it later - I'm not sure that the broom is obviously a broom - a trash can might be better. I considered a hoover and a spray bottle too!

wxtim:
```suggestion
   trigger: mdiCursorPointer 
```

src/components/cylc/gscan/GScan.vue
```diff
@@ -373,12 +372,6 @@ export default {
     }
   },
   methods: {
-    getWorkflowClass (node) {
-      return {
-        'c-workflow-stopped': node && node.status && node.status === WorkflowState.STOPPED.name
-      }
-    },
-
```
oliver-sanders, line 381: Not sure it was the best idea to move this out of GScan but no biggie.

oliver-sanders (approved):
Works nicely.

src/components/cylc/gscan/GScan.vue
MetRonnie: I agree but it was a quick solution, and the SCSS targeting the `c-workflow-stopped` class is nested inside the `c-gscan-workflow` class selector so it works without a problem (for now)

PR #931, opened 2022-02-16
Fix incorrect nesting of hierarchical dirs in GScan

MetRonnie:
These changes close #930, fixing a bug where the GScan sidebar was only showing the top level of workflow directory trees.

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [ ] Appropriate tests are included (unit and/or functional).
- [ ] Appropriate change log entry included.
- [x] No documentation update required.

PR #936, opened 2022-02-17
test hierarchical gscan

oliver-sanders:
Add a simple test for #931

### DIFF
--- a/tests/unit/components/cylc/gscan/gscan.vue.spec.js
+++ b/tests/unit/components/cylc/gscan/gscan.vue.spec.js
@@ -110,8 +110,24 @@ describe('GScan component', () => {
           [TaskState.FAILED.name]: 1
         }
       }
+
+      // nested mode
       const nodeHierarchy = createWorkflowNode(workflow, true)
-      expect(nodeHierarchy.name).to.equal('a')
+      const ret = [nodeHierarchy.name]
+      const stack = [nodeHierarchy]
+      let snode
+      while (stack.length) {
+        snode = stack.pop()
+        if (snode && snode.children) {
+          for (const child of snode.children) {
+            ret.push(child.name)
+            stack.push(child)
+          }
+        }
+      }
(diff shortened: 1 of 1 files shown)


datamel (approved):
I have run the test locally, passed for me.

tests/unit/components/cylc/gscan/gscan.vue.spec.js
```diff
@@ -110,8 +110,24 @@ describe('GScan component', () => {
...
+      while (stack.length) {
+        snode = stack.pop()
+        if (snode && snode.children) {
+          for (const child of snode.children) {
+            ret.push(child.name)
+            stack.push(child)
+          }
+        }
+      }
+      expect(ret).to.deep.equal(['a', 'b', 'c'])
```
MetRonnie, line 128: Simpler to do?
```suggestion
      expect(nodeHierarchy).to.deep.nested.include({
        name: 'a',
        'children[0].name': 'b',
        'children[0].children[0].name': 'c'
      })
```

MetRonnie (commented):
I tested this on 8dd7e52b9d9bc1d7ac355952a4f1baae54f511d7 (commit before the bug was fixed) and it passed 🤔

I have suggested a simplification that passes on master but not on 8dd7e52b9d9bc1d7ac355952a4f1baae54f511d7

tests/unit/components/cylc/gscan/gscan.vue.spec.js
MetRonnie: It's basically testing for the subset
```js
{
    name: 'a',
    children: [{
        name: 'b',
        children: [{
            name: 'c'
        }]
    }]
}
```
in the object.

Whereas, before #931, it looked like
```js
{
    name: 'a',
    children: [{
        name: 'b',
        children: []
    }, {
        name: 'c',
    }]
}
```
I don't think there should be more than 1 child per level in this function.

I suppose we would also need to check that the `c` node does not have a `children` property

PR #943, opened 2022-02-18
Add useful unit tests opts to README

MetRonnie:
Also make chai print whole objects in gscan test failures instead of the incredibly useful `{ ...(5) }`


hjoliver (approved):

wxtim (approved):
Now I want Chai.

Issue #737, opened 2021-08-24
Disable Cylc Hub button for `jupyter-cylc`


datamel:
After the trio: https://github.com/cylc/cylc-uiserver/pull/230, https://github.com/cylc/cylc-ui/pull/711 and https://github.com/cylc/cylc-flow/pull/4331 have been merged, upon running `jupyter-cylc` and clicking the cylc hub button, you will get a 404 error. This button should be disabled instead.

kinow:
Good catch @datamel. Do we want to disable that button always, or only when running `jupyter-cylc`? If the latter, I think we will need some indication from the server as to whether to show or not the button (or show it greyed out?).

datamel:
The latter, I think, makes most sense.

PR #969, opened 2022-03-22
Remove cylc hub button in single user mode

datamel:
The cylc hub button on the dashboard used to be available in single user mode (`cylc gui`). If clicked this resulted in a 404 page.
This change adds a condition that the gui mode should be multi-user if the button is to display.

I am not sure the preference on the links in the dashboard if the cylc hub button is not visible.

I have stuck with keeping the help links to the right, and the functional links to the left; as currently is the case. I think this looks a little lopsided in single user mode. I’m happy to move the Cylc UI Quickstart to the left list and move the cylc hub to the right for aesthetics. 
 
In single user mode the dashboard now looks as follows: 

 
image
In multi-user mode: 
 image

These changes close https://github.com/cylc/cylc-ui/issues/737 
Sibling: https://github.com/cylc/cylc-uiserver/pull/331

- [x] No change log entry required (why? invisible to users).

### DIFF
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -106,19 +106,26 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               </v-list-item-subtitle>
             </v-list-item-content>
           </v-list-item>
-          <v-list-item id = "cylc-hub-button" v-if=multiUserMode :href=hubUrl>
+          <v-tooltip :disabled=multiUserMode bottom>
+            <template v-slot:activator="{ on }">
+          <div v-on="on" >
+          <v-list-item id = "cylc-hub-button" :disabled=!multiUserMode :href=hubUrl>
             <v-list-item-avatar size="60" style="font-size: 2em;">
               <v-icon large>{{ svgPaths.hub }}</v-icon>
             </v-list-item-avatar>
             <v-list-item-content>
-              <v-list-item-title class="title font-weight-light">
-                Cylc Hub
-              </v-list-item-title>
+                <v-list-item-title class="title font-weight-light">
+                  Cylc Hub
+                </v-list-item-title>
               <v-list-item-subtitle>
                 Visit the Hub to manage your running UI Servers
               </v-list-item-subtitle>
             </v-list-item-content>
           </v-list-item>
+          </div>
+            </template>
+            <span>Only available in multi user mode.</span>
+          </v-tooltip>
         </v-list>
       </v-flex>
       <v-flex xs12 md6 lg6>
@@ -129,7 +136,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             </v-list-item-avatar>
             <v-list-item-content>
               <v-list-item-title class="title font-weight-light">
-                Cylc UI Quickstart
+                Cylc UI Quickstart2
               </v-list-item-title>
               <v-list-item-subtitle>
                 Learn how to use the Cylc UI
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -106,7 +106,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               </v-list-item-subtitle>
             </v-list-item-content>
           </v-list-item>
-          <v-list-item :href=hubUrl>
+          <v-list-item :id = "cylc-hub-button" v-if=multiUserMode :href=hubUrl>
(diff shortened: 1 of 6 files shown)


hjoliver (commented):
Looks good.

However, how about instead of removing it entirely, display a greyed-out hub icon and *You are not running Cylc UI via Cylc Hub* or similar.  That would hint at additional multi-user capability that's available, and also explain why the Hub button isn't there, in case the lack of it (in comparison to screenshots and documentation, say) is confusing.

oliver-sanders (approved):

oliver-sanders:
> However, how about instead of removing it entirely, display a greyed-out hub icon

Sounds like a good idea, @datamel there is a prop builtin to the list component we are using which can do this

https://vuetifyjs.com/en/api/v-list-item/#props-disabled

I think it needs this diff:

```diff
- v-if=multiUserMode
+ :disabled=multiUserMode
```

datamel:
I have made it so a tool tip comes up if you hover over the disabled button for cylc hub.

I am having difficulty testing this, cypress doesn't appear to have a hover function, though is a currently open issue. Instead it is recommend that we use `.trigger` instead but this does not function for disabled items.

In addition, vue.js renders this cylc-hub-button as an anchor tag 
image
 ; which does not pick up on the `be.disabled` check of cypress (again recognised in issue on cypress). The recommended work around is to check `cy.get(...).should('have.attr', 'disabled')` which I can not get to work.
It seems that having a tooltip on a disabled v-list-item is tricky to test. If anyone has ideas then I'm happy to give an suggestions a go.

src/views/Dashboard.vue
```diff
@@ -129,7 +136,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             </v-list-item-avatar>
             <v-list-item-content>
               <v-list-item-title class="title font-weight-light">
-                Cylc UI Quickstart
+                Cylc UI Quickstart2
```
oliver-sanders, line 139: ?

oliver-sanders:
> I am having difficulty testing this

Perhaps check/scan for the `v-list-item--disabled` class.

src/views/Dashboard.vue
datamel: Sorry! Changing the page as my reloads were caching even with ctrl F5.

src/views/Dashboard.vue
```diff
@@ -106,19 +106,26 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+                <v-list-item-title class="title font-weight-light">
+                  Cylc Hub
+                </v-list-item-title>
               <v-list-item-subtitle>
                 Visit the Hub to manage your running UI Servers
               </v-list-item-subtitle>
             </v-list-item-content>
           </v-list-item>
+          </div>
+            </template>
```
MetRonnie, line 126: The indentation is off (functionally doesn't matter but proper indentation makes it easier to see the HTML hierarchy)

src/views/Dashboard.vue
```diff
@@ -257,6 +264,13 @@ export default {
             count: count[state.name] || 0
           }
         })
+    },
+    ...mapState('user', ['user']),
+    multiUserMode () {
+      if (this.user.mode === 'single user') {
+        return false
+      }
+      return true
```
MetRonnie, line 273:
```suggestion
      return this.user.mode !== 'single user'
```

tests/e2e/specs/dashboard.js
```diff
@@ -44,5 +44,11 @@ describe('Dashboard', () => {
       })
       .should('equal', [...WorkflowStateOrder.entries()][0][0])
   })
+  it('Should have disabled cylc hub button', () => {
```
MetRonnie, line 47:
```suggestion
  it('Should have disabled cylc hub button for single user mode', () => {
```

src/views/Dashboard.vue
MetRonnie: Indentation better now, but best to avoid putting tags on the same line like this:
```
<template v-slot:activator="{ on }"> <div v-on="on" >
```

datamel: @MetRonnie I'm not great at these things so have put it through an auto-prettify.

If you are happy then I think we can merge this.

MetRonnie: I have no idea why it would want to leave those on the same line. Anyway, it doesn't matter too much

MetRonnie (approved):

Issue #960, opened 2022-03-04
Occasional mutation menu failure - TypeError: object null is not iterable


MetRonnie:
#### Steps to reproduce the bug

This only happens sometimes. It's possible it only happens when building in development mode.

Load up the UI and click on a workflow icon in GScan to open the mutation menu.

Instead you might get "- state unknown"

![image](https://user-images.githubusercontent.com/61982285/156827472-ddfc91dd-afff-4b0e-8fbe-341d7afaf405.png)

and this error in the console:
```
Uncaught TypeError: object null is not iterable (cannot read property Symbol(Symbol.iterator))
    at _createForOfIteratorHelper (createForOfIteratorHelper.js:3:46)
    at filterAssociations (aotf.js:438:68)
    at HTMLSpanElement.listener (plugin.js:35:50)
```

The only solution seems to be hard reloading (possibly several times) or rebuilding. 

~This seems to be something to do with Babel.~ It is actually a race condition due to an apollo graphql query whose promise might not have been resolved by the time the workflow icon/button is binded

**Pull requests welcome!**
This is an Open Source project - please consider contributing a bug fix
yourself (please read `CONTRIBUTING.md` before starting any work though).

MetRonnie:
> It is actually a race condition due to an apollo graphql query whose promise might not have been resolved by the time the workflow icon/button is binded

This bug can be made more reproducible by introducing an artificial delay into the loading of the mutations from the graphql query:
```diff
diff --git a/src/services/workflow.service.js b/src/services/workflow.service.js
index dba4487..cf1a542 100644
--- a/src/services/workflow.service.js
+++ b/src/services/workflow.service.js
@@ -104,9 +104,11 @@ class WorkflowService {
       query: getIntrospectionQuery(),
       fetchPolicy: 'no-cache'
     }).then((response) => {
-      this.mutations = response.data.__schema.mutationType.fields
-      this.types = response.data.__schema.types
-      processMutations(this.mutations, this.types)
+      setTimeout(() => {
+        this.mutations = response.data.__schema.mutationType.fields
+        this.types = response.data.__schema.types
+        processMutations(this.mutations, this.types)
+      }, 5000)
     })
   }
```

PR #978, opened 2022-03-30
Use eslint-plugin-no-only-tests

MetRonnie:
There were 2 tests where `.only()` was seemingly accidentally committed, leading to other tests in those files being excluded. This PR removes those `.only()`s and adds https://github.com/levibuzolic/eslint-plugin-no-only-tests to prevent it happening again.

**Requirements check-list**
- [x] Does not need tests 


MetRonnie:
The e2e test failure is due to #960, which I am working on

PR #1024, opened 2022-05-13
Remove unnecessary dev dependencies

MetRonnie:
This is a small change with no associated Issue.

- autoprefixer
- postcss
- regenerator-runtime
- zen-observable

Autoprefixer & postcss seem to be included with vue-cli-service, so no need for a direct dependency. https://cli.vuejs.org/guide/css.html#postcss

I have checked:
- `yarn install`
- `yarn build`
- Looked at `dist/css` to see that vendor prefixes are still present

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Does not need tests 
- [x] No change log entry required
- [x] No documentation update required.

PR #1025, opened 2022-05-13
Deps

oliver-sanders:
Quick attempt to pull out unused deps. I used `depcheck`, it got some false-positives but flagged a few which appear legitimate.

There may well be more in the dep deps, can move onto them when we have the time.


oliver-sanders:
I also rebuilt the `yarn.lock` which may or may not be a good idea, ideas?

MetRonnie:
Rebuilt as in ran `yarn install`? That should be fine

hjoliver:
If a rebuild works from scratch, without these, presumably that's a definitive test that we don't need them.

MetRonnie (approved):

oliver-sanders:
> Rebuilt as in ran yarn install? That should be fine

Rebuilt as in `rm yarn.lock; yarn install` so non-additive, basically a new environment.

The existing `yarn.lock` seems to have a lot of cruft in it, after the rebuild (and the dep changes above) it's about 1300 lines shorter.

MetRonnie:
I have tested this in combination with #1024, everything seems to be working

Issue #1035, opened 2022-05-27
scan side-panel width


hjoliver:
Not sure if this design issue has come up yet.

If workflow names are too long, and you install with `--no-run-name`, the name will push the job icons off to the right. Even more so if you crank the font size up in UI preferences.

Can we make the row of icons drop down to the next line, if it overruns the panel? Otherwise we'll need to allow users to adjust the width.

![MicrosoftTeams-image](https://user-images.githubusercontent.com/2362137/170618987-1c5e4a3f-aa6e-4c71-9fda-9ba97fd1ff1f.png)

oliver-sanders:
> Can we make the row of icons drop down to the next line, if it overruns the panel?

We tried this, it's quite jarring.

> Otherwise we'll need to allow users to adjust the width.

:+1:

PR #1043, opened 2022-06-07
allow the left hand drawer to re resized by the user if required

AaronDCole:
These changes address #1035

**Requirements check-list**
- [X] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [X] Appropriate tests are included (unit and/or functional).
- [X] Appropriate change log entry included.
- [X] No documentation update required.


hjoliver:
Tested!  The resizing works nicely, but it doesn't play well with lumino tabs on the right.  Resizing the browser window causes some kind of reset of the lumino tab state, so I guess we need that to happen when resizing the drawer too?

AaronDCole:
> Tested! The resizing works nicely, but it doesn't play well with lumino tabs on the right. Resizing the browser window causes some kind of reset of the lumino tab state, so I guess we need that to happen when resizing the drawer too?

Thanks Hilary, no worries I'll take a look at that. Im currently trying to figure out how to test some parts of the resize code anyway as its failing the quality gate.

hjoliver:
Interesting two-step redraw now, if I have several tabs side-by-side. But it gets the right result, and users won't be constantly resizing so I think that's fine.  :+1:

hjoliver:
Couple of small conflicts to fix.

hjoliver:
Is testing this reasonable/feasible/worth-the-effort?

AaronDCole:
> Is testing this reasonable/feasible/worth-the-effort?

Yes the two set resize is slightly annoying, but is needed because if the resize happens before lumino tabs resize handler runs it wont correctly resize.

Ive fixed the conflicts, thankfully minor!

I think it is, Ive been banging my head into it for a few days actually. If nothing else it'll be a good learning for me

AaronDCole:
rebuild

AaronDCole:
@hjoliver Hey Hilary, Ive fixed the pesky test issues so I think this is ready to review and merge in now

hjoliver:
Great, having a look now...

hjoliver:
Resizing works great, thanks @AaronDCole 

My only comment is: can selection be turned off while dragging, to stop the spurious selection and highlighting of text? (Do you see that too?)

Not a show stopper as it's easily unselected and users won't be constantly resizing the drawer. I've no idea if it's easy to fix though.

AaronDCole:
> Resizing works great, thanks @AaronDCole
> 
> My only comment is: can selection be turned off while dragging, to stop the spurious selection and highlighting of text? (Do you see that too?)
> 
> Not a show stopper as it's easily unselected and users won't be constantly resizing the drawer. I've no idea if it's easy to fix though.

Ive pushed a change to disable any other mouse interactions while dragging (hopefully the codecov gate passes, that thing is the bane of my existence 😂)

hjoliver:
It passed! :tada:

hjoliver (approved):
Mainly functional review, looks great to me. :+1:

tests/unit/components/cylc/toolbar.vue.spec.js
```diff
@@ -31,7 +30,30 @@ Vue.use(Vuex)
 describe('Toolbar component', () => {
   let vuetify
   let $route
-  const store = new Vuex.Store(storeOptions)
+  // for some obscene reason, using the actual "options.js" store, gets contaminated data from other tests during the course of the whole test running
```
datamel, line 33: Is this something that you think would need fixing/further investigation; I'm still a bit novice with the ui - I would guess yes? If so, I could open an issue.

AaronDCole:
No, not for now, I’m doing a quiet background investigation into why it’s happening and will open another ticket when I work out why.

datamel (commented):
I have checked this out and had a really good play around with the resizing. 
One small query from me, I managed to resize the left hand drawer, pulling it totally to the left which made it fully disappear. I could get it back with a refresh but the top left icon (which usually opens the gscan) no longer pulls the left bar in (as it has been resized away). This may be suitable for a different issue.
Otherwise, no issues found from me, excellent feature and it is working really smoothly with the lumino tabs.

AaronDCole:
> I have checked this out and had a really good play around with the resizing.
> One small query from me, I managed to resize the left hand drawer, pulling it totally to the left which made it fully disappear. I could get it back with a refresh but the top left icon (which usually opens the gscan) no longer pulls the left bar in (as it has been resized away). This may be suitable for a different issue.
> Otherwise, no issues found from me, excellent feature and it is working really smoothly with the lumino tabs.

Thank you for your review Melanie, thats a good catch

AaronDCole:
> > I have checked this out and had a really good play around with the resizing.
> > One small query from me, I managed to resize the left hand drawer, pulling it totally to the left which made it fully disappear. I could get it back with a refresh but the top left icon (which usually opens the gscan) no longer pulls the left bar in (as it has been resized away). This may be suitable for a different issue.
> > Otherwise, no issues found from me, excellent feature and it is working really smoothly with the lumino tabs.
> 
> Thank you for your review Melanie, thats a good catch

Ive pushed a change to enforce the default if you make the drawer too small.. I could make it smaller, see what you think.

datamel (commented):
@AaronDCole that last change isn't working for me. I've raised a PR https://github.com/AaronDCole/cylc-ui/pull/28, see what you think of moving the restriction to the resize.

AaronDCole:
> @AaronDCole that last change isn't working for me. I've raised a PR [AaronDCole#28](https://github.com/AaronDCole/cylc-ui/pull/28), see what you think of moving the restriction to the resize.

Thanks for this suggestion, the reason I did it on the setter is so that the check always fires when the user tries to open the panel (as a true boolean is passed to the setter when that happens). I understood the issue to be that if the user makes the sidebar too thin, then closes and opens the panel they might not be able to see it / grab it again? But I may have misunderstood.

Setting a min width that the panel can be resized is another option, but perhaps Im not the right person to ultimately decide?

In any case Ive updated the original MR, thank you for catching that, I have to completely restart the UI for your issue to show up but Ive added a fix for it.

datamel (commented):
Hi @AaronDCole
This latest change isn't working for me, not sure if I'm doing something wrong with my build or if I have not made my issue clear so I've made a video, hopefully you'll be able to see what I mean. 
https://user-images.githubusercontent.com/37735232/189343093-8b315595-dffa-49c6-865c-618d5488dde1.mov
Is this the case for you?
I don't think this is a major blocker as I'm sure users will probably not do this, and the refresh button pulls the bar back over although limiting the resizing would fix this issue.

tests/e2e/specs/drawer.cy.js
```diff
@@ -22,6 +22,9 @@ describe('Drawer component', () => {
       .get('.v-navigation-drawer')
       .should('be.visible')
   })
+  it('it should have a width of 360', () => {
+    cy.get('.v-navigation-drawer').invoke('innerWidth').should('be.lt', 360)
```
MetRonnie, line 26: Why does the test say "have a width of 360" but then checks that it is less than that?

tests/e2e/specs/drawer.cy.js
```diff
@@ -35,4 +38,13 @@ describe('Drawer component', () => {
...
   })
+  it('should drag to trigger resize', () => {
+    cy.get('.v-navigation-drawer').invoke('innerWidth').should('be.lt', 360)
+    cy.get('.v-navigation-drawer__border')
+      .trigger('mousedown', { which: 1 })
+      .trigger('mousemove', { clientX: 100 })
+      .trigger('mouseup', { force: true })
+    cy.get('.v-navigation-drawer__border')
+      .invoke('innerWidth').should('be.lt', 460)
+  })
```
MetRonnie, line 49: This test does not seem to be quite working unfortunately. It does not appear to resize 

Before mousemove:

image

After mousemove:

image

MetRonnie (commented):
Works pretty well from my testing. However I can reproduce Mel's issue of resizing the drawer into oblivion.

> Setting a min width that the panel can be resized is another option, but perhaps Im not the right person to ultimately decide?

I think that is sensible

hjoliver:
> Works pretty well from my testing. However I can reproduce Mel's issue of resizing the drawer into oblivion.

Summary of discussion with @AaronDCole 
- resize to oblivion is a desirable feature, to maximize screen real-estate
- setting a minimum size isn't great (ugly, and wastes a little space)
- **if resized to zero, the burger icon should appear** and clicking on that will restore the default-width drawer
   -  a good compromise: *doesn't restrict users actions, does provide a way to recover when they accidentally drive off a cliff*

datamel:
> * **if resized to zero, the burger icon should appear, and clicking on that should restore the default-width drawer**
>   
>   * a good compromise: _doesn't restrict users actions, does provide a way to recover when they accidentally drive off a cliff_

Great idea, better than restricting the width :)

tests/e2e/specs/drawer.cy.js
AaronDCole: Theres something really weird going on here, im looking into it.

tests/e2e/specs/drawer.cy.js
AaronDCole: Will fix this with issue below

src/components/cylc/Drawer.vue
```diff
@@ -99,19 +101,84 @@ export default {
...
+      el.style.width = f + 'px'
+    },
+    setEvents () {
+      const minSize = this.navigation.borderSize
+      const el = this.getDrawerElement()
+      const drawerBorder = el.querySelector('.v-navigation-drawer__border')
+      drawerBorder.addEventListener(
+        'mousedown',
+        (e) => {
+          if (e.offsetX < minSize) {
```
MetRonnie, line 153: I have figured out this line is what is causing the cypress test to not work properly. The cypress-triggered event is an `Event` rather than `MouseEvent` for a real mouse-down, and so is missing the `offsetX` property.

What is the reason for this `if` statement?

AaronDCole: This is a really good catch, thank you. I did actually look into the events but didn't make the connection with offsetX. I cant honestly remember why the offsetX was added, but with testing it doesn't seem to be being used so I'll remove it.

tests/e2e/specs/drawer.cy.js
AaronDCole: Fixed with latest test updates ( think )

tests/e2e/specs/drawer.cy.js
```diff
@@ -35,4 +38,14 @@ describe('Drawer component', () => {
       .get('#toggle-drawer')
       .should('be.visible')
   })
+  it('should drag to trigger resize', () => {
```
MetRonnie, line 41: Reloading the page is a good idea to prevent the previous `it` from polluting this one
```suggestion
  it('should drag to trigger resize', () => {
    cy.visit('/#/')
```

tests/e2e/specs/drawer.cy.js
```diff
@@ -35,4 +38,14 @@ describe('Drawer component', () => {
...
+    cy.get('.v-navigation-drawer').invoke('innerWidth').should('be.eq', 260)
+    cy.get('.v-navigation-drawer__border')
+      .trigger('mousedown', { which: 1 })
+      .trigger('mousemove', { clientX: 0, clientY: 500 })
+      .trigger('mouseup', { force: true })
+    cy.get('.v-navigation-drawer__border')
+      .invoke('innerWidth').should('be.lt', 460)
```
MetRonnie, line 48: I think this makes more sense
```suggestion
    cy.get('.v-navigation-drawer')
      .invoke('innerWidth').should('be.eq', 0)
```

MetRonnie (commented):
Couple more suggestions for the e2e test but otherwise looks good to go. We should probably also display the drawer toggle button at all times but that can come in a follow-up PR

AaronDCole:
@MetRonnie You can merge this whenever your ready, thanks for reviewing

datamel:
> Thanks! I'm happy with this, @datamel do you want to merge if happy?

Yes, I'll just do a review since the last changes.

datamel (approved):
Checked out and code read. Great change, thanks @AaronDCole.

PR #1044, opened 2022-06-07
Remove dependency on eslint-plugin-node

MetRonnie:
This is a small change with no associated Issue. Follow-up to #1016

From the changelog of eslint-config-standard:

> **`eslint-config-node` has been replaced with the up to date fork `eslint-config-n`.** If you have used comments like `// eslint-disable-line node/no-deprecated-api` you now have to reference the `n/` rules instead.

**Requirements check-list**
- [x] Does not need tests 


MetRonnie:
Actually, seeing as eslint-plugin-n is a dependency of eslint-config-standard anyway, we don't need that dependency either

PR #1050, opened 2022-06-16
Use task icons in gscan, not job icons.

hjoliver:

As discussed in Element chat, the gscan panel shows task states represented by job icons. Technically this is a bug, even though it was deliberate, because task and jobs states are not quite the same thing. We apparently discussed this early in the project and decided to use job icons in gscan because:
- for the important states, there is a one-to-one mapping between task and job state
- coloured icons are better at-a-glance indicators
- existing users have been conditioned to seeing coloured icons, by the old Cylc 7 gscan GUI

However, the one-to-one mapping is somewhat beside the point in the case of retrying tasks, and I have evidence (a "bug" report) that users may find this confusing.

ss1

User: *"gscan showed no failed tasks, but when I opened the workflow there was a failed task[sic] in it"*.

On this branch:

ss2

(NB the task progress bar is not relevant here, would need to get rid of that of course).

Understanding this will still require understanding the difference between task and job icons, but unlike in the first case the result will not be confusing to those who do understand the difference.

So IMO we should change this. If others are super attached to a colourful gscan, we should consider making that a user preference (next to which we can document the failed job retrying task "subtlety" that results).

**Requirements check-list**

- [ ] No documentation update required.

### DIFF
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -166,7 +166,10 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                           dark
                           icon
                         >
-                          <job :status="state" />
+                          <task :status="state"
+                            :startTime="Date.now()"
+                            :estimatedDuration="30"
+                          />
                         </v-btn>
                       </template>
                       <!-- tooltip text -->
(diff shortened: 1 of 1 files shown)


hjoliver:
Relatedly, we should consider adding a waiting task icon the gscan summaries. 

Waiting tasks indicate there is more to come in the workflow, which is useful information because it is not always the case. There will be no waiting tasks if:
- the schedule is running the last task(s), after which it will shut down or stall
- the is stalled (which doesn't necessarily involve failed tasks)

It would be useful for gscan to distinguish been the above and downtime waiting between real time cycle points, for example - i.e. between "nothing happening, for the moment" and "nothing happening, end of".

(a succeeded icon remains pretty useless however, since the lack of that would only distinguish the first tasks of the run).

hjoliver:
(August meeting discussion: agreed that we need to resolve the current inconsistent use of icons, but the way forward still needs more thought.)

PR #1053, opened 2022-06-20
Add Cylc 7 job colour theme.

hjoliver:
I'm a convert to the new default colour theme myself, but have had several requests already for the old Cylc 7 "green == running" theme.

Screenshot from 2022-06-20 13-39-13

**Requirements check-list**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,10 @@ menu would show the wrong workflow.
 [#1027](https://github.com/cylc/cylc-ui/pull/1027) - Remove development
 "Mutations View".
 
+[#1053](https://github.com/cylc/cylc-ui/pull/1053) - Add optional Cylc 7 colour
+theme for job icons.
+
+
 ### Fixes
 
 [#979](https://github.com/cylc/cylc-ui/pull/979) - Fix bug where the mutations
--- a/src/styles/cylc/_job.scss
+++ b/src/styles/cylc/_job.scss
@@ -69,6 +69,41 @@ $cjob: ".c-job svg.job rect";
     }
 }
 
+@mixin job_theme_cylc7() {
+    $grey: rgb(173,165,165);
+    $green: rgb(0,193,64);
+    $green2: rgb(161,207,37);
+    $red: rgb(255,0,0);
+    $red2: rgb(255,136,204);
+
+    #{$cjob} {
+        &.submitted {
+            fill: $green2;
+            stroke: $green2;
+        }
+
+        &.running {
+            fill: $green;
+            stroke: $green;
+        }
+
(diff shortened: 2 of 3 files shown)


oliver-sanders:
> we'll probably need to make it totally user configurable in due course.

When we discussed this previously I think the outcome was, no Cylc 7 preset, but easy to re-create with user-defined themes, which haven't arrived yet.

> If a user knowingly opts in to a particular theme, it should not confuse them.

Not confusing for them, initially, but when they try to work with someone else...

> And, I might quote/paraphrase your primary explanation

By that logic we can't change *anything*?!

hjoliver:
> By that logic we can't change anything?!

I'm just pointing out a glaring inconsitency :grin: - you used exactly that logic to justify our ~incorrect use of coloured job icons in Cylc 8 gscan, but you are now arguing against the very same logic when extended very slightly to the actual colours of the icons (and optional colours at that).

hjoliver:
> Not confusing for them, initially, but when they try to work with someone else...

I guess, but that's a small problem and we really can't avoid it without being unreasonably restrictive (colour themes for anything have to be configurable!)

MetRonnie (approved):

oliver-sanders (approved):
I had successfully fended this one off for over two years only to give in on the last day, dammit.

IMO: We should not offer this is as a preset. I don't think reluctance to change is a valid justification for keeping it, as with most UI change people will quickly forget the old once they've used the new for a couple weeks. The new colours are easy to understand and adapting to them is a part of migrating to the new task/job separation (and new states for that matter). This will keep the legacy scheme kicking around for ever with lots of scope for confusion, especially with the green running/succeeded difference. There are only five colours, TBH [accessibility aside] there really isn't a good reason to allow them to be configurable at all! Other systems e.g. GitHub actions don't have job colour themes, they don't need them, it's culturally understand what red and green mean.

hjoliver:
@oliver-sanders - 

> I had successfully fended this one off for over two years only to give in on the last day, dammit.

Well, thanks for giving in :grin: ... but since you feel so strongly about it, and you get massive kudos for the UI design spec, in a staggering reversal I've had a rethink and will close this without merging it.

For the record, my thinking was as follows:
- Some users have requested the old theme already, and I had thought that others on the team were more concerned than me about making non-essential changes that might annoy users!
- We have to provide several themes for accessibility reasons
- And I had it in my head that we planned to make it user-configurable anyway, in which case holding out on this for the stated reasons seemed a bit perverse

BUT if we do not make the job state theme generally configurable, then actually I'm quite happy to take a principled stand on this and just tell users to get used to the new theme because **it is better than the old one** :boom:

PR #1058, opened 2022-06-28
aotf: add help field for BroadcastSetting

oliver-sanders:


MetRonnie (approved):
Ok as a temporary measure, but will need some work

PR #1087, opened 2022-08-09
Quick commit to test the table deltas

### LINKED ISSUE #1066: Offline mode: table view incorrect most recent job icon
Spotted in offline testing, the collated job icon does not match the most recent job status.

Genuine issue with the table view (1.3.0 milestone) or smaller issue with the offline test mode (1.4.0 milestone)?

**Steps to reproduce:**

```console
$ yarn run serve
```

Open the table view, the job icon next to the task icon should show the most recent job (where multiple jobs are present a multi-job stack icon should also show the status of the previous job).

At the moment the most recent job is shown as submitted for all tasks:

Screenshot from 2022-07-20 10-21-19

**Pull requests welcome!**

kinow:
These changes close #1066 and #1099

**Requirements check-list**
- [x] Appropriate tests are amended
- [x] No change log entry included (hard to notice bug).

Hi! I saw some activity in Cylc UI today, went through the tickets now before dinner, and found one that looked simple. Took me longer to actually remember about the code base than to find where the bug (possibly) was :smile: 

I think the deltas for tables is always using the same element, if we get the element with highest `submitNum` first (which happens in the mock/offline table deltas [see JSON in service/mock folder]). Didn't have time to do a more through investigation and write tests, but figured you guys would be able to take over and confirm if it's indeed a bug and maybe use this as reference for a fix (feel free to discard this PR :+1: )

:vulcan_salute:

### DIFF
--- a/src/components/cylc/table/Table.vue
+++ b/src/components/cylc/table/Table.vue
@@ -22,7 +22,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
   >
     <!-- Toolbar -->
     <v-row
-        class="d-flex flex-wrap table-option-bar no-gutters flex-grow-0"
+      class="d-flex flex-wrap table-option-bar no-gutters flex-grow-0"
     >
       <!-- Filters -->
       <v-col
@@ -79,8 +79,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                   v-if="slotProps.index === maximumTasks"
                   class="grey--text caption"
                 >
-            (+{{ tasksFilter.states.length - maximumTasks }})
-          </span>
+                  (+{{ tasksFilter.states.length - maximumTasks }})
+                </span>
               </template>
             </v-select>
           </v-col>
@@ -126,9 +126,9 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               </thead>
             </template>
             <template
-                slot="item"
-                slot-scope="{ item }"
-              >
+              slot="item"
+              slot-scope="{ item }"
+            >
               <tr>
                 <td>
                   <div class="d-flex align-content-center flex-nowrap">
@@ -145,8 +145,15 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                     </div>
                     <div class="mr-1">
                       <Job
+                        v-if="item.jobs.length === 0"
                         v-cylc-object="item.node"
                         :status="item.node.state"
+                        :previous-state="item.jobs.length > 1 ? item.jobs[1].state : ''"
+                      />
+                      <!-- most recent job summary -->
+                      <Job
+                        v-else
+                        :status="item.jobs[0].state"
                         :previous-state="item.jobs.length > 1 ? item.jobs[1].state : ''"
                       />
                     </div>
@@ -180,7 +187,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                 <td>
                   <div class="d-flex align-content-center flex-nowrap">
                     <div class="mr-1">
-                      <Job
+                      <job
                         v-cylc-object="job"
                         :key="`${job.id}-summary-${index}`"
                         :status="job.state"
--- a/src/components/cylc/table/Table.vue
+++ b/src/components/cylc/table/Table.vue
@@ -126,8 +126,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               </thead>
             </template>
             <template
-              slot="item"
-              slot-scope="{ item }"
+              v-slot:item="{ item }"
             >
               <tr>
                 <td>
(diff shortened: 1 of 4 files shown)


kinow:
image

hjoliver:
Thanks @kinow - good to see you here :grin:

src/components/cylc/table/Table.vue
```diff
@@ -145,10 +145,17 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                     </div>
                     <div class="mr-1">
                       <Job
+                        v-if="item.jobs.length === 0"
                         v-cylc-object="item.node"
                         :status="item.node.state"
                         :previous-state="item.jobs.length > 1 ? item.jobs[1].state : ''"
```
kinow, line 151: Oh, just realized previous-state here should be `''` (empty) actually. No point in checking its length again.

I left these two jobs because I **think** when we don't have jobs we display the state of the task? I think :grimacing:

wxtim:
I think that these changes completely address #1066, which I think is a problem with the test data. 😄 @kinow  - did you want to un-draft.

kinow:
> I think that these changes completely address #1066, which I think is a problem with the test data. smile @kinow - did you want to un-draft.

Hi @wxtim !

I left it as a draft as I didn't have time to fix the last part about the previous state (see my comment above in a code-review), and also because I didn't have time for the tests :grimacing: 

image

Happy if it's merged as-is, if someone takes over, or if it's left open for a while (I will add the tests & finish it eventually, shouldn't take long).

src/components/cylc/table/deltas.js
```diff
@@ -136,8 +136,10 @@ function applyDeltasAdded (added, table, lookup) {
           const latestJob = latestJobSubmitNum < job.submitNum
             ? lookup[job.id]
             : existingEntry.latestJob
+          // FIXME: Couldn't we avoid setting table[existing.id].latestJob multiple times here?
+          //        If a task has 100 jobs, in the worst case we will call Vue.set 100 times?
```
MetRonnie, line 140: From the [docs](https://v2.vuejs.org/v2/api/#Vue-set), `Vue.set()` is only needed to add a property to a reactive object. But if the properties `jobs` and `latestJob` already exist, there should be no need to use `Vue.set()`, just assign as normal?

src/components/cylc/table/Table.vue
```diff
@@ -180,7 +187,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                 <td>
                   <div class="d-flex align-content-center flex-nowrap">
                     <div class="mr-1">
-                      <Job
+                      <job
```
MetRonnie, line 190: Why make this lowercase, out of curiosity?

MetRonnie (commented):
Thanks Bruno!

src/components/cylc/table/deltas.js
```diff
@@ -133,14 +133,10 @@ function applyDeltasAdded (added, table, lookup) {
...
-            : existingEntry.latestJob
-          // FIXME: Couldn't we avoid setting table[existing.id].latestJob multiple times here?
-          //        If a task has 100 jobs, in the worst case we will call Vue.set 100 times?
-          Vue.set(existingEntry, 'latestJob', latestJob)
-          existingEntry.jobs.push(job)
-          Vue.set(existingEntry, 'jobs', existingEntry.jobs)
+          if (job.submitNum > latestJobSubmitNum) {
+            existingEntry.latestJob = lookup[job.id]
+          }
+          existingEntry.jobs.unshift(job)
```
MetRonnie, line 139: With your commit, the aggregate icon was showing the older job on top. So I changed `.push()` to `.unshift()` to add the job to the front of the list, which is where it gets picked up on
https://github.com/cylc/cylc-ui/blob/a1a9a2f10a2749a13b33ca9ee4300d893d4f4fec/src/components/cylc/table/Table.vue#L146-L149

MetRonnie:
I think this closes #1099 too?

src/components/cylc/table/Table.vue
kinow: >I think the previous-state is valid; it's the status that should be the latest job instead of the task? I'll push a commit with a simplification

Could be. I was referring to the `v-if="item.jobs.length === 0"` above. If it is added, then it wouldn't make sense to have the `:previous-state="item.jobs.length > 1`, since `item.jobs.length is zero at that point.

src/components/cylc/table/Table.vue
kinow: Sorry my OCD. I believe I did that because it was lowercase in... the tree component? But no functional need to change it here :grimacing:

src/components/cylc/table/deltas.js
kinow: Good catch! Thanks!

kinow:
Thanks for the review @MetRonnie !

src/components/cylc/table/Table.vue
MetRonnie: Actually I would like to change the lowercase instances to uppercase, as it makes them stand out from normal HTML tags!

kinow: +1! Good idea, and vive la consistance! 😄

AaronDCole (approved):
No complaints, works as expected now. Good job Bruno

hjoliver (approved):
Looks like this is good to go now. Thanks @kinow

PR #1088, opened 2022-08-10
Update the Quick Start guide a bit.

hjoliver:
- document task and job states better
- remove the ugly `v-card` drop-shadow (we don't have other 3D design elements).

**Requirements check-list**
- [x] Does not need tests (simple text tweak).

### DIFF
--- a/src/views/Guide.vue
+++ b/src/views/Guide.vue
@@ -17,7 +17,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 <template>
   <v-container>
-    <h1 class="ma-0">Guide</h1>
+    <h1 class="ma-0">Cylc UI Quick Start</h1>
     <!--
       TODO: make sections linkable
 
@@ -29,20 +29,29 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
     -->
 
     <h2 id="task-and-job-states">
-      Task And Job States
+      Understanding Tasks &amp; Jobs
     </h2>
-    <!-- TODO document difference between tasks and jobs -->
+
+    <p>
+      <b>Tasks</b> represent single units of activity in a Cylc workflow.<br/>
+      <b>Jobs</b> are real processes submitted by Cylc to perform a task's activity.<br/>
+      One task can have multiple jobs due to automatic retries or manual triggering.
+    </p>
+
     <div class="card-grid">
 
       <v-flex
         md6
         xs12
       >
-        <v-card>
+        <v-card outlined>
           <v-card-title primary-title>
-            <p class="display-1 text--primary">Task State Vs Job State</p>
+            <p class="display-1 text--primary">Task &amp; Job States</p>
           </v-card-title>
+
           <v-card-text>
+            Tasks represent future and past jobs as well as current jobs.<br/>
+            A <b>waiting</b> task that already has job(s) will <b>retry</b>.
             <table id="task-job-state-table">
               <tr>
                 <td>Task</td>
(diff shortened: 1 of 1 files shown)


hjoliver:
shot

src/views/Guide.vue
```diff
@@ -29,20 +29,29 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+      Understanding Tasks &amp; Jobs
     </h2>
-    <!-- TODO document difference between tasks and jobs -->
+
+    <p>
+      <b>Tasks</b> represent single units of activity in a Cylc workflow.<br/>
+      <b>Jobs</b> are real processes submitted by Cylc to perform a task's activity.<br/>
+      One task can have multiple jobs due to automatic retries or manual triggering.
+    </p>
+
```
oliver-sanders, line 40: Sorry, this TODO was in the wrong place, could we move this into the "Task &amp; Job States" card.

src/views/Guide.vue
```diff
@@ -29,20 +29,29 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-        <v-card>
+        <v-card outlined>
           <v-card-title primary-title>
-            <p class="display-1 text--primary">Task State Vs Job State</p>
+            <p class="display-1 text--primary">Task &amp; Job States</p>
           </v-card-title>
+
           <v-card-text>
+            Tasks represent future and past jobs as well as current jobs.<br/>
+            A <b>waiting</b> task that already has job(s) will <b>retry</b>.
```
oliver-sanders, line 54: The spacing after this text is a bit short. I think it either needs to be wrapped in a `<p>` or followed by a `<br />`.

src/views/Guide.vue
hjoliver: I wasn't following the TODO so much as putting the general description of tasks and jobs at the top of this (currently the only, but whatever) section, because it applies to both the following cards (i.e., including the modifiers one).  Alternatively, we could move modifiers into the main task/job card?

src/views/Guide.vue
hjoliver: Done, it looks OK now (screenshot not updated).

src/views/Guide.vue
oliver-sanders: I think this is clarification for the task-job separation card?

The idea was for the cards to be self-contained topics. Part of the reason for this is that we could re-use these elsewhere in the GUI e.g. startup-tips.

hjoliver: I'm not really against the cards being entirely self-contained topics, but if so we shouldn't have a heading that applies to multiple cards which is currently the case on master as well as here:

> Task And Job States
> [card: Task Vs Job States]
> [card: Special Task States]

If we have cards categorized under headings like this, then a short topic description under the category heading would be fine.

If not, then I guess we should remove the heading as well.

Clearly you prefer the second option. But then it seems a bit odd to have task state modifiers in a separate card from the one that describes what task states are in the first place.

MetRonnie (approved):

src/views/Guide.vue
hjoliver: OK I've moved the "section heading" content into the "Tasks and Jobs" card, and reframed the content of the "modifiers" card so that it doesn't seem quite so much like it should be inside the other one.  I'll post a screen shot...

hjoliver:
Now tweaked to work as independent cards (i.e., not grouped under headings).

Screenshot from 2022-08-23 11-19-31

oliver-sanders (approved):
:+1:

PR #1097, opened 2022-08-15
made mutation edit button x-large to make it easier to click.

### LINKED ISSUE #1096: Make hit area of mutation (aka command) edit button bigger

The "edit" button in the command menu has an unreasonably small hit area. It is too easy to submit the command when trying to click the edit button.

image

**Pull requests welcome!**

wxtim:
May close #1096 

**Requirements check-list**
- [x] No change log entry required (v small change)

### DIFF
--- a/src/styles/cylc/_menu.scss
+++ b/src/styles/cylc/_menu.scss
@@ -38,7 +38,4 @@
       line-height: 1.35;
     }
   }
-  .pad-button {
-    padding: 30px;
-  }
 }
--- a/src/styles/cylc/_menu.scss
+++ b/src/styles/cylc/_menu.scss
@@ -38,4 +38,7 @@
       line-height: 1.35;
     }
   }
+  .pad-button {
+    padding: 30px;
+  }
 }
--- a/src/components/cylc/cylcObject/Menu.vue
+++ b/src/components/cylc/cylcObject/Menu.vue
@@ -74,13 +74,12 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             </v-list-item-content>
             <v-list-item-action>
               <v-btn
+                icon
                 :disabled=!authorised
                 x-large
                 class="float-right"
                 @click.stop="openDialog(mutation)"
                 data-cy="mutation-edit"
-                outlined
-                width=80%
               >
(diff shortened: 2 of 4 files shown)


MetRonnie (changes requested):
Icon is now outsized

image

I have opened https://github.com/wxtim/cylc-ui/pull/4 against this with some improvements

src/styles/cylc/_menu.scss
```diff
@@ -38,4 +38,7 @@
       line-height: 1.35;
     }
   }
+  .pad-button {
+    padding: 30px;
+  }
```
MetRonnie, line 43: There is no element with the class `pad-button` so this does nothing

wxtim: I did not want to commit this file - this was something I played with and discarded.

wxtim:
@MetRonnie  I've removed the `icon` property from the button so it goes back to being square, and I've added an outline so it's completely clear what is and is not the button. I think either of these measures would be sufficient on their own, but in combination...

MetRonnie:
Not sure I like the rectangular button or outline. IMHO draws too much attention for a secondary control.

image

My proposal:

image

wxtim:
> Not sure I like the rectangular button or outline. IMHO draws too much attention for a secondary control.
> 
> image
> 
> My proposal (reverts [f079595](https://github.com/cylc/cylc-ui/commit/f079595e5742a7f6af8540b10fcb7ec0f7c1a95c)):
> 
> image

How about the rectangular button without the edge? 

I don't think anyone will move their mouse over there if they don't want the edit button, and pressing the edit button wrongly is the fail-safe way...

hjoliver:
I think the subtly-shaded round button looks best.  Is @MetRonnie's proposal still bigger/more-clickable than the status quo?

wxtim:
> I think the subtly-shaded round button looks classiest. Is @MetRonnie's proposal bigger/more-clickable than the status quo? [Looking back at the issue, it is bigger - is that not enough in your view @wxtim ?]

Whatever, I took this on as a learning experience, I'm not militant - If the consensus is the round button, I'll do the round button.

Issue #82, opened 2019-05-27
static graph visualization


hjoliver:
We probably want to replace `cylc graph` with web technology too, not just the control GUI graph view.

hjoliver:
This graph demo (found by @MartinRyan ) allows manual node repositioning (as many do, I think): https://cytoscape.org/cytoscape.js-euler/.  Could be useful when automatic layout fails us.

MartinRyan:
Cytoscape looks great, it has a dagre extension, and a navigator 'bird's eye view and pan control' extension (mini-map). It consumes JSON though their are convertors for graph objects.

hjoliver:
Also from @MartinRyan [py2cytoscape[(https://py2cytoscape.readthedocs.io/en/latest/) might be useful for static vis.  (Cytoscape itself might be useful for the live control GUI as well)

oliver-sanders:
This is an interesting challenge because:

* Users typically run `cylc graph` to visualise a suite whilst developing the *working copy*.
* The UI-Server is concerned with the *installed copy* (in the `cylc-run` directory).

The UI-Server is only concerned with the *installed copy* and is not necessarily able to see the *working copy*.

It is OK to require users to register a suite for `cylc graph` functionality (as the suite wouldn't appear in the web UI otherwise) but we wouldn't want users to have to repeatedly re-install a suite in order to visualise it.

hjoliver:
Well, yes and no.

If you're working on the working copy then you are able to install it repeatedly pretty easily, and if the payoff is kick-ass web base viz that seems a small price to pay,  And besides you currently have to do that anyway if your `suite.rc` takes Jinja2 inputs from `rose-suite.conf` (it won't validate or graph without being installed first).

Also what's the alternative?  We don't want to keep a GTK-based static graph GUI.  An electron app might allow us to kill both birds with one stone, I guess.

matthewrmshin:
Assuming that we have done enough logical migration from `rose suite-run` to Cylc in the coming months, then we should have enough logic in Cylc to deal with the `rose-suite.conf` stuffs. The web UI can have an local file system browser so users should be able to open a suite from its source location. Otherwise, we can also build support for the UI to search for named suites from various configurable locations on the file system. (E.g. If a suite cannot be found under `~/cylc-run/SUITE/` then look at `~/roses/SUITE/`.)

oliver-sanders:
> If you're working on the working copy then you are able to install it repeatedly pretty easily

[...]

Issue #178, opened 2019-08-12
Visual testing?


kinow:
@hjoliver pointed we had a pending item from a previous meeting, about e2e tests. We have e2e set up already in the project, but no useful test for now. That's mainly due to the fact that what we had before was more of a prototype.

Soon we will have the final screens for the application UI, and then we should be ready to write good e2e functional tests.

However, during the same meeting, @oliver-sanders mentioned an issue with the progress display in the `Task` component. The issue we had was related to a CSS expression that wasn't working on Chrome. In theory, having a e2e test would cover it, but by default would not detect if that was broken.

Cypress (similarly to Selenium and Nightwatch) provide only functional automated web tests. So only tests that confirm that the web application has the features from its requirements, and that they work as expected (things like the colors are right, the page loads fast enough, the gauge arrow is displayed, are normally considered "non-functional requirements").

Cypress documentation mentions "Visual Testing", see this part of their docs

>Luckily, Cypress gives a stable platform for writing plugins that can perform visual testing.
>...
>Typically such plugins take an image snapshot of the entire application under test or a specific element, and then compare the image to a previously approved baseline image.
https://docs.cypress.io/guides/tooling/visual-testing.html#Functional-vs-visual-testing

@oliver-sanders are these the kind of tests you would like for Cylc UI?

I don't think we are quite ready to introduce these tests at the moment. IMO, we better wait a bit longer, until we are certain the components and views are ready, and then choose whether to use this testing technique or not, and if we decide to use, then choose what to test.

But in anyway, I'd prefer to cover just having e2e tests in Cylc now as part of #170, and then discuss visual testing later if necessary :+1:

oliver-sanders:
>> Typically such plugins take an image snapshot of the entire application under test or a specific element, and then compare the image to a previously approved baseline image.

> @oliver-sanders are these the kind of tests you would like for Cylc UI?

Absolutely.

A visual diff tool should subtract one image from another leaving you with an image of the things which have moved / changed e.g:

http://diffee.me/?http://una.im?http://una.im/diffeedemo

kinow:
>The big issue is storing/displaying those reports (can't be done on Travis) and GH integration. There are systems out there but they cost a fortune.

https://www.cypress.io/pricing

Cypress has a service with a free tier. 3 users and 500 recordings.

matthewrmshin:
The open source plan https://www.cypress.io/oss-plan/ allows 5 users and unlimited recordings.

[...]

PR #1108, opened 2022-09-27
graph view and central data store

### LINKED ISSUE #74: Graph View

### LINKED ISSUE #595: Display non-standard non-output messages
... in job leaves. See description of #596

Requires changes to back-end and schema, to supply the data.

### LINKED ISSUE #694: Revisit sorting, should it be done in render or when adding/updating data?

Placeholder for this comment from GScan's PR: https://github.com/cylc/cylc-ui/pull/543#discussion_r658767856

image

**Additional context**

**Pull requests welcome!**

### LINKED ISSUE #803: Create views off the data store (instead of only using initial-data-burst)

At the moment the views like Tree and Table view are created using Vuex store properties. These properties are only ever initialized from GraphQL initial-data-burst messages.

Note that the data in the Vuex store that we use is derived from GraphQL data. e.g. the Tree View doesn't use GraphQL data directly. The GraphQL endpoint doesn't provide hierarchical data. Instead, there is some code that creates a hierarchical data structure with pointers to the GraphQL original data.

**Additional context**

The challenge here, then, is to think in a way to create the derived data structure in an intelligent manner. 

We could identify we already have a compatible subscription (i.e. same query used in 2 views), and then just call the code that creates the derived data structure.

Or we could find a way to tie-together the original GraphQL data & the derived data, so that components/views only use the derived data directly (as they are doing now).

The last option is to always refresh a subscription.

>NOTE: we identified this issue in #631 where we added a Table View, but already had a Tree View; the two queries are pretty much the same, so merging them results in the same query. See the Table View PR's & issue for more.

**Pull requests welcome!**

### LINKED ISSUE #838: GScan sort running workflows before expandable (containing run dirs) workflows
**Describe the bug**

At the moment, no-run-name workflows that are running are displayed before stopped non-run-name workflows, but after stopped workflows containing numbered/named run dirs.

If sorting is to be done properly, running workflows should be at the very top of the GScan sidebar.

Also, why do workflows containing run dirs come before no-run-name workflows in the sort order anyway?

**Pull requests welcome!**

### LINKED ISSUE #903: Received a Table delta before the workflow initial data burst
When you have the Table View open and a workflow shuts down automatically due to finishing, an error appears:
> Received a Table delta before the workflow initial data burst

**Pull requests welcome!**

### LINKED ISSUE #1078: task: clockface alignment broken in webkit/safari
The clockface alignment has become broken in webkit/safari when zooming in/out:

Screenshot 2022-07-27 at 09 14 37

At 100% the alignment is correct.

See also #178 

**Pull requests welcome!**

### LINKED ISSUE #1110: subscriptions: merged query is not used
Discovered whilst trying to understand why I can't get the Tree and Graph views running at the same time.

When you open a workflow in the UI it creates a subscription. When you open another "view" on the workflow the subscription gets stopped, the new subscription is merged into it and the new subscription is re-issued.

However, it doesn't look like the new query is actually being picked up, you just get whatever subscription was issued first.

I jammed in a breakpoint into the UIS here:

https://github.com/cylc/cylc-uiserver/blob/549afbd99389a5e29571b3f5f71cc6e4505a4270/cylc/uiserver/handlers.py#L388-L389

And inspected the websocket messages like so:

```python
(Pdb) import json
(Pdb) msg = json.loads(message)
(Pdb) print(msg['payload']['query'])
```

To inspect the query, which confirmed that the updated query was not making it through to the UIS.

This means that the subscription merging isn't actually hooked into anything at the moment. The Tree & Table views are able to co-exist only because their queries overlap.

**Pull requests welcome!**

### LINKED ISSUE #1116: workflow table: add workflow state indicators
The neglected workflow table is missing workflow state icons which means we can't easily see the state or control the workflow.

**Pull requests welcome!**

### LINKED ISSUE #1129: outputs: task outputs can be mapped onto all jobs
The tree view associates task outputs with the jobs which created them.

In this example I have a task which fails twice before succeeding, each time it generates *one* output.

```ini
[runtime]
    [[foo]]
        script = """
            if [[ $CYLC_TASK_TRY_NUMBER -eq 1 ]]; then
                cylc message -- loscil
                false
            elif [[ $CYLC_TASK_TRY_NUMBER -eq 2 ]]; then
                cylc message -- biosphere
                false
            else
                cylc message -- gas
            fi
        """
```

If I watch this task in the GUI I see one output associated with the first job (correct), two with the second (incorrect) and three with the third (incorrect).

Curiously when I refresh the page I see all three outputs associated with all three jobs (really incorrect).

outputs-bug

**Pull requests welcome!**

oliver-sanders:
Features:
* Closes #74 
  Basic graph view now implemented, will push family expand / collapse and other features into future issues.
* Closes #803 
  Once finished all views will be powered off of the central data store with no local stores or callbacks.
* Closes #694
  Sorting is now performed in the central data store. Where views require a different sort order (e.g. GScan) they use computed properties.
* Closes #1116 
  Added workflow icon and status column whilst converting to run on the central data store.
* Closes #595 
  The way custom task outputs are listed had to be re-written as a result of the migration to the central data store. The easiest way to close #1129 was to use messages, closes #595 in the process.
* ~1800 fewer lines of code!

Bugfixes:
* Closes #903
  Once finished will have removed the local callbacks from which this error can occur.
* Closes #1110
  Fixed bug by re-creating the query object before submitting it.
* Closes #838
  GScan sorting has been re-implemented.
* Closes #1078
  This had the same root-cause as the issue which made the old task component fall apart when used in the graph. Solved in the re-write.
* Closes #1129
  Task outputs are now filtered by job messages to ensure they are correctly associated with the job which created them.

Partially addresses:
* Partially address #1126
  All workflow are now present in a single tree. From the perspective of most views, it doesn't really matter if they are given a workflow (e.g. `~u/w/run1//`) or workflow-part (e.g. `~u/w`) node.
  The tree view now supports multiple workflow nodes. If there is only one workflow (as at present) it will strip that workflow node (so that cycle points are at the top level).
* Partially addresses #82 
  The graph view added here will also run off of offline data when it is added at the UIS.
* Partially addresses #471
  Needed a toolbar for the graph view so created a reusable component.

~Proof of concept~ Implementation of the graph view running directly off of a central data store.

> ~**⚠️ Warning:** Proof of concept code with TODOs, don't look too closely at the hacks too closely, some things you can't unsee.~

etchasketch

### Overview

* Graph layout performed by a web assembly (WASM) port of [GraphViz](https://github.com/hpcc-systems/hpcc-js-wasm).
* Graph rendering performed in a perpetual dynamic SVG document.
* Nodes and edges listed directly from a central data store with no local callbacks.
* Perpetual components and partial re-rendering, no need to destroy the whole graph every update.
* Layout matches the design document, we have total control over node rendering.
* We also have control over edge rendering which might come in useful...

### Features

* Pan.
* Zoom.
* Task mutation menus.
* Auto refresh (only performed when graph changes).
* Graph edges routed to and from task icons.

### Vue ❤️ SVG

Vue can work with SVG document just as well as HTML ones.

This means we can reuse our existing "Task" and "Job" components with an SVG approach. We can even use the mutation menus in exactly the same way as we would normally.

```vue
<template>
  <g>
    <symbol :id="nodeID" viewBox="0 0 100 100">
      <task
        :svg="true"
        :status="task.node.state"
      />
    </symbol>
    <use
      :href="`#${nodeID}`"
      x="0" y="0"
      width="100" height="100"
      v-cylc-object="task.node"
    />
  </g>
</template>
```

Reactive SVG 💥

### How It Works

The graph is contained in a single reactive SVG document which lives for the life of the graph view.

A timer calls a refresh function (every 2 seconds at the time of writing).

The refresh function lists nodes and edges from the data store and constructs a hash representing the current graph. The hash is compared to the pervious one, if they differ the layout needs to be updated.

The graph nodes (i.e. tasks) are added into the SVG document. There's no layout information so they sit on top of each other in a stack.

Next we define the graph in GraphViz "dot" format. In Cylc 7 we gave nodes text labels e.g. `answer\n42`, here I've given nodes HTML-like labels which allows us to control their structure e.g:

```
          "~user/workflow//42/answer" [
            label=<
              <TABLE HEIGHT="110.99747467041016">
                <TR>
                  <TD PORT="in" WIDTH="100"></TD>
                </TR>
                <TR>
                  <TD PORT="task" WIDTH="100" HEIGHT="110.99747467041016">icon</TD>
                  <TD WIDTH="218.9765625">~user/workflow//42/answer</TD>
                </TR>
                <TR>
                  <TD PORT="out" WIDTH="100"></TD>
                </TR>
              </TABLE>
            >
          ]
```

You'll notice there are width and height values in the table. These have been extracted from from the rendered nodes in the SVG document. This way the dimensions of the nodes in the SVG document exactly matches the dimensions in the GraphViz document.

Also note those "PORT" attributes, these control where GraphViz routes edges from and to. I've placed these ports above and below the the task icon which constrains the edges to a fairly small space making for more regular graphs.

We then get GraphViz to layout the graph, but rather than rendering it, we get GraphViz to give us the layout in JSON format. If it were rendered it would look like this:

Screenshot 2022-09-27 at 16 57 00

The graph holds the structure we want but with the details missing.

We then extract the node coordinates from the GraphViz output and convert them into SVG transforms which can be applied to the nodes in the SVG document.

Finally we extract the edge information from the GraphViz output, convert them into SVG format and add them to the SVG. We now have a graph that looks like this:

Screenshot 2022-09-27 at 16 57 26

It's exactly the same graph as GraphViz produced but we've taken control of the rendering. This way we can have a perpetual SVG document which we update rather than continuously creating / destroying it. This is more efficient but it also means that we can elegantly preserve things like user-selection and zoom/pan coordinates.

### Data Store

Currently (on master) the data store contains a flat store of nodes (the lookup).

The tree view wants these nodes in a tree structure so it registers a callback which allows it to intercept every added/updated/pruned delta and maintain its own local data store.

The GScan view also wants these nodes in a tree structure so registers a similar callback.

Many views are likely to want a tree structure as this tree is quite fundamental to Cylc:

* user
  * workflow
    * cycle
      * task
        * job

Or to put it another way:

```
~user/workflow//cycle/task/job
```

I've generalised the approach to create a single callback which handles the whole universal ID (UID) and manages a new bit of the central data store to reflect this tree structure. It's kinda like a mix of the central data store and the tree view data store.

It's completely de-coupled from the existing store as I plan to either update the callbacks for the other views to feed off of the new bit of the store, or, if possible, run them off of the new bit of the store directly.

> **Update:** The tree view is now running directly off of the central data store with no local store or callback.

Most of the approach is inherited from the current implementation, it uses the same node merging etc, but the structure has changed. E.G. if we added this node to an empty store:

```
~user/path/to/workflow//cycle/task/job
```

We would get a data store which looks like this:

```
* user
  * path
    * to
      * workflow
        * cycle
          * task
            * job  
```

* A flat index exists at the top level (aka the lookup).
* Namespaces (i.e. TaskDefs) and graph edges are stored in special indexes on the workflow node so that we can fit them into the same structure.

The graph view is running directly off of this data store, no local callbacks needed, the one generic callback can power all views. This reduces the delta handling overheads as we currently have two callbacks per view meaning each delta is processed twice per open view (GScan counts as a view).

### Running It

There are a couple of issues at present which you'll have to work around if you want to play with this.

1) **Now Fixed** ~I haven't managed to get webpack to play with WASM nicely
   The first time you build on this branch run the following command (remember to `yarn install` first):~
   ```bash
   cp node_modules//@hpcc-js/wasm/dist/graphvizlib.wasm dist/js/
   ```
2) **Now Fixed** ~I've run into an issue with subscription merging, it will either run the tree subscription OR the graph one.
    I think something is going wrong in the GraphQL subscription merging (or I wrote it wrong, equally likely!). Haven't taken a look yet, for now the workaround is to first close the tree view and then open the graph view.~
3) **Now Fixed** ~The graph doesn't auto centre (yet), you'll need to zoom out and pan around.~
4) **Now Fixed** ~Ignore `Error: <path> attribute d: Unexpected end of attribute.` warnings in the console.
    They appear to be harmless, need to sort them out.~

**Requirements check-list**
- [x] Tests (basic testing done, testing a graph layout is going to be, ..., tricky)


oliver-sanders:
Transpose and reset implemented.

transpose-and-reset

oliver-sanders:
Task modifiers and progress indicators have been disabled so far due to a couple of issues with them:

1. They overflow the 100x100 viewBox.
2. The task modifiers fly away from the task icons when incorporated into a larger document. (some transformation problem I expect).

I've re-worked the task icon to address this. It's a WIP, I'm aware that the modifiers are too small and the progress indicator isn't calibrated. More fun for later.

oliver-sanders:
I have identified the issue which is preventing the Tree and Graph views from co-existing, written up here - https://github.com/cylc/cylc-ui/issues/1110

oliver-sanders:
Update:

* Fixed the issue with subscription merging which meant that you couldn't' have the tree and graph views open at the same time.
* The tree view is now running directly off of the central data store with no local data or callbacks.
* The graph is now auto-centred.
* The graph view can now display task modifiers.

Screenshot 2022-10-06 at 17 08 46

Up next:

* Get the table view running off of the central data store.
* Remove the old callback code that's no longer needed.
* Design fiddling of the new task icon component & remove the old CSS that's no longer needed.

hjoliver:
Fantastic :tada:

oliver-sanders:
And then there were three:

Screenshot 2022-10-12 at 16 15 44

oliver-sanders:
Have just discovered a roadblock after converting GScan to use the central data store. The UIS is issuing an erroneous updated delta *after* the pruned delta issued when a workflow is removed. Consequently if you remove a workflow it will briefly disappear from GSscan before being subsequently re-added.

oliver-sanders:
Semi-accidental side-effect of the changes made here (and also the current lack of data store housekeeping 🙈), if you have hierarchical workflows, load one of them, then change the URL to go up a level, the tree view will now display a multi-workflow tree:

Screenshot 2022-10-27 at 14 45 18

This will break once housekeeping is hooked up again, however, it demonstrates that this should be easy to implement later on. Have written up the next steps for this on in #1126.

oliver-sanders:
Update: This ballooned quite a lot as a few issues caused roadblocks, but nearly there.

* Most interfaces in the UI should now be working correctly with GScan, Tree, Graph & Table views all running off of the central data store with search/filter working in all (except Graph where it's not implemented).
* Some runtime errors still likely.
* Housekeeping (i.e. memory leak protection) now implemented for the central data store.

Up Next:
* Remove unused callbacks and related code.
* Remove unused stylesheet and other detritus.
* Fix the progress indicator in the task icon.
* Make the task state modifiers more visible.
* Clear up any lingering errors.
* Tidy, tidy, tidy!
* Test, test, test!
* Polish off the graph view itself rather than messing about with the data store.
* Document!

oliver-sanders:
Task progress indicators now working in the graph view, graph nodes now show up to 6 jobs, then show a `+<n>` to represent overflowing jobs:

Screenshot 2022-10-31 at 13 41 05

oliver-sanders:
Rebased and deconflicted.

oliver-sanders:
A bit more spit and polish required so I've not put this up for review yet, but this should now be fully functional!

Screenshot 2022-11-04 at 15 54 10

I've not managed to solve the WASM packaging issue (see the top of the README), if anyone has any ideas please do say as the best I can think to do is to shim a `cp ...` into the build script 🤦.

oliver-sanders:
Putting this up as ready for [functional] review.

I've still got some code tidying and testing, however, everything is supposed to work at this stage so it's ready for functional review.

A good approach for testing is to have two versions of the GUI open simultaneously, one with the UI before these changes and one after. Then click around and explore the UI looking for any differences.

FYI: You can do that something like this:

```console
$ git checkout master
$ yarn run build
$ cp dist dist-master
$ git checkout oliver-sanders/etchasketch.dev
$ yarn run build
$ cylc gui --ui-build-dir="$PWD/dist-master"
$ cylc gui --ui-build-dir="$PWD/dist"
```

oliver-sanders:
(updated the wiki to match - https://github.com/cylc/cylc-ui/wiki/Architecture)

wxtim:
I count 12 issues addressed.

> 1800 fewer lines of code

🎉 

> We also have control over edge rendering which might come in useful...

Optional outputs? The few remaining Suicide triggers?



Sorry for this horrific code dump, the data store work was quite atomic and hard to break down. The PR ballooned when I hit broken interfaces and had to do things like re-implement the Task icon because of strange SVG behaviour I couldn't explain.

Here's how I would break down the review to make it more manageable:

## My Review in Progress

(I may or may not do all items.)

* [ ] Compare the UI before and after change to check for unexpected changes.
* [ ] The new Task component.
   * Run the component tests in cypress (`cypress/component/cylc-icons.cy.js`).
   * Then inspect the results in `tests/e2e/screenshots/`.
* [ ] The new GraphNode component.
   * Run the component tests in cypress (`cypress/component/cylc-graph-node.cy.js`).
   * Then inspect the results in `tests/e2e/screenshots/`.
* [ ] The new ViewToolbar component.
   * Source: `src/components/cylc/ViewToolbar.vue`.
   * Used by: `src/view/Graph.vue`.
* [ ] The new Graph view.
   * Source: `src/views/Graph.vue`.
* [ ] The new central data store.
   * Source: `src/store/workflows.module.js`
   * Tests: `tests/unit/store/workflows.spec.js`
* [ ] Check the issued marked as closed have been addressed:
   * Note, there's an explanation for each in the PR header.
   * [ ] #74
   * [ ] #803
   * [ ] #694
   * [ ] #1116 
   * [ ] #595 
   * [ ] #1129 
   * [ ] #903
   * [ ] #1110
   * [ ] #838
   * [ ] #1078
   * [ ] #1129
* [ ] Functional test of the existing views to ensure I haven't severely buggered anything:
   * [ ] Tree view.
   * [ ] Table view. 
   * [ ] GScan.
   * [ ] Dashboard.
   * [ ] Workflow table (linked from the dashboard).

package.json
```diff
@@ -38,13 +39,15 @@
...
+    "svg-pan-zoom": "^3.6.1",
     "vue": "^2.6.14",
     "vue-i18n": "^8.27.0",
     "vue-meta": "^2.4.0",
     "vue-router": "^3.5.3",
     "vue-the-mask": "^0.11.1",
     "vuetify": "^2.6.10",
-    "vuex": "^3.6.2"
+    "vuex": "^3.6.2",
+    "wasm-loader": "^1.3.0"
```
MetRonnie, line 50: Don't webpack loaders normally only need to be dev dependencies?

README.md
```diff
@@ -4,6 +4,16 @@
...
+## TODO
+
+I haven't managed to configure webpack to pass through the WASM files yet
+so for the moment you have to manually copy the wasm file over into dist
+the first time you build.
+
+```bash
+cp node_modules//@hpcc-js/wasm/dist/graphvizlib.wasm dist/js/
+```
+
```
MetRonnie, line 16: Suggest moving this to the "Developers" section

wxtim:
# My thoughts on playing with GUI Graph View:

### Possible bugs

1. I've got some overlapping arrows: image
2. Scroll bar appears by default
3. Weird failure of edges to appear on first opening/ multiple copies of each edge therafter!
4. Suggested transform co-ords for transposed version: (15, 30)



### Possible enhancements 

feel free to boot to later

- Could the controls for increasing and decreasing spacing be sliders or integer input?

src/components/cylc/GraphNode.vue
```diff
@@ -0,0 +1,157 @@
...
+      x="180" y="80"
+      font-size="50"
+    >
+      {{ task.name }}
+    </text>
+
+    <!-- the cycle point -->
+    <text
+      x="180" y="110"
+      font-size="25"
```
MetRonnie, line 57: I'm finding the cycle point text size to be very small
image

src/components/cylc/GraphNode.vue
```diff
@@ -0,0 +1,157 @@
...
+  },
+  computed: {
+    nodeID () {
+      return `graph-node-${this.task.id}`
+    },
+    startTime () {
+      if (this.jobs.length) {
+        return this.jobs[0].node.startedTime
+      }
+      return undefined
```
MetRonnie, line 133:
```suggestion
      return this.jobs?.[0].node.startedTime
```

src/components/cylc/GraphNode.vue
```diff
@@ -0,0 +1,157 @@
...
+      const ret = []
+      let ind = 0
+      for (const job of this.jobs) {
+        if (ind >= this.maxJobs) {
+          break
+        }
+        ret.push(job)
+        ind++
+      }
+      return ret
```
MetRonnie, line 146:
```suggestion
      return this.jobs.slice(0, this.maxJobs)
```

src/components/cylc/GraphNode.vue
```diff
@@ -0,0 +1,157 @@
...
+  components: {
+    SVGTask,
+    Job
+  },
+  props: {
+    task: {
+      required: true
+    },
+    jobs: {
+      require: true
```
MetRonnie, line 117:
```suggestion
      type: Array,
      required: true
```

src/components/cylc/GraphNode.vue
```diff
@@ -0,0 +1,157 @@
...
+    },
+    maxJobs: {
+      // maximum number of jobs to display before using an overflow indicator
+      default: 6,
+      require: false
```
MetRonnie, line 122:
```suggestion
      required: false
```

src/components/cylc/GraphNode.vue
```diff
@@ -0,0 +1,157 @@
...
+
+export default {
+  name: 'GraphNode',
+  components: {
+    SVGTask,
+    Job
+  },
+  props: {
+    task: {
+      required: true
```
MetRonnie, line 114:
```suggestion
      type: Object,
      required: true
```

MetRonnie (commented):
35/97 files viewed

package.json
oliver-sanders: Makes sense, will move.

README.md
oliver-sanders: Planning on solving this before merge, somehow...

src/components/cylc/GraphNode.vue
oliver-sanders: Agreed, will boost along with the job squares.

MetRonnie (commented):
58/97 files viewed

MetRonnie (commented):
97/97 files viewed 😮

src/views/Graph.vue
```diff
@@ -0,0 +1,625 @@
...
+        this.refreshTimer = null
+      }
+    },
+    increaseSpacing () {
+      // increase graph layout node spacing by 10%
+      this.spacing = this.spacing * 1.1
+    },
+    decreaseSpacing () {
+      // decrease graph layout node spacing by 10%
+      this.spacing = this.spacing * (10 / 11)
```
wxtim, line 361: This is decreasing by 10% of the final value - is there a reason for this?

src/views/Graph.vue
```diff
@@ -0,0 +1,625 @@
...
+        clearInterval(this.refreshTimer)
+        this.refreshTimer = null
+      }
+    },
+    increaseSpacing () {
+      // increase graph layout node spacing by 10%
+      this.spacing = this.spacing * 1.1
+    },
+    decreaseSpacing () {
+      // decrease graph layout node spacing by 10%
```
wxtim, line 360: This is reducing spacing by 10% of the final value, so that if you increase the final value by 10% you get back to where you started. I wonder if it might be worth pointing this. If it isn't web-boilerplate obvious to everyone who's done this sort of thing before.

src/store/workflows.module.js
```diff
@@ -63,41 +49,519 @@ const state = {
...
+function createTree (state) {
+  // console.log('@@ Create')
+  if (state.cylcTree) {
+    return
+  }
+  const tree = {
+    $index: {},
+    id: '$root'
+  }
+  Vue.set(tree, 'children', [])
```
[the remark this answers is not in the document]
oliver-sanders: I think the reason we need to do this here is because Vue needs to hook into Array objects to make them reactive. So the object is fine, it's the thing we're adding to the object that we need to make reactive. I've got a feeling that this is different at Vue3.

src/views/Graph.vue
oliver-sanders: It's pretty irrelevant, I just like the numerical niceness of getting back to where we started.

src/services/mock/graphql.js
```diff
@@ -24,6 +24,7 @@ const data = require('./json/index')
  */
 const DATA_MAPPING = {
   IntrospectionQuery: data.IntrospectionQuery,
+  App: data.App,
   OnWorkflowTreeDeltasData: {
     deltas: {
       added: data.OnWorkflowTreeDeltasData
```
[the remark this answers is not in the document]
oliver-sanders: Done.

hjoliver:
I've been attempting a functional review today, in Firefox. The task icons disappear for me, soon after start up. The edges remain visible though.  I'll try a redo from scratch before the meeting, before posting details.

oliver-sanders:
> I've been attempting a functional review today, in Firefox. The task icons disappear for me, soon after start up. The edges remain visible though. I'll try a redo from scratch before the meeting, before posting details.

Looks like one of the review suggestions needed tweaking, should be fixed now.

hjoliver:
Works for me now, brilliant :tada:

oliver-sanders:
And that should be the last bug fixed, next batch please!

hjoliver:
Preliminary comments:

- layout jumping is much less of a problem with the `n=1` window, than it was for the typically much bigger Cylc 7 graphs (it's now much easier to find tasks of interest again, at a glance)
- similarly disconnected graphs, due to non-overlapping `n=1` sub-windows, aren't much of a problem and are easy to understand. Maybe we can improve with something like the old scissor nodes, but no hurry on that IMO
- tested with several kinds of workflow, looks good
- in fact, I frickin' love it :beers:

hjoliver:
On this branch, but not on master, I get these errors at start-up (with gscan and dashboard):
```
Warning: fragment with name WorkflowData already exists.
graphql-tag enforces all fragment names across your application to be unique; read more about
this in the docs: http://dev.apollodata.com/core/fragments.html#unique-names [chunk-vendors.a58588bb.js:22:152423](http://localhost:8889/cylc/js/chunk-vendors.a58588bb.js)
Warning: fragment with name Deltas already exists.
graphql-tag enforces all fragment names across your application to be unique; read more about
this in the docs: http://dev.apollodata.com/core/fragments.html#unique-names [chunk-vendors.a58588bb.js:22:152423](http://localhost:8889/cylc/js/chunk-vendors.a58588bb.js)
Warning: fragment with name AddedDelta already exists.
graphql-tag enforces all fragment names across your application to be unique; read more about
this in the docs: http://dev.apollodata.com/core/fragments.html#unique-names [chunk-vendors.a58588bb.js:22:152423](http://localhost:8889/cylc/js/chunk-vendors.a58588bb.js)
Warning: fragment with name UpdatedDelta already exists.
graphql-tag enforces all fragment names across your application to be unique; read more about
this in the docs: http://dev.apollodata.com/core/fragments.html#unique-names [chunk-vendors.a58588bb.js:22:152423](http://localhost:8889/cylc/js/chunk-vendors.a58588bb.js)
Warning: fragment with name AddedDelta already exists.
graphql-tag enforces all fragment names across your application to be unique; read more about
this in the docs: http://dev.apollodata.com/core/fragments.html#unique-names [chunk-vendors.a58588bb.js:22:152423](http://localhost:8889/cylc/js/chunk-vendors.a58588bb.js)
Warning: fragment with name UpdatedDelta already exists.
graphql-tag enforces all fragment names across your application to be unique; read more about
this in the docs: http://dev.apollodata.com/core/fragments.html#unique-names [chunk-vendors.a58588bb.js:22:152423](http://localhost:8889/cylc/js/chunk-vendors.a58588bb.js)
Warning: fragment with name PrunedDelta already exists.
graphql-tag enforces all fragment names across your application to be unique; read more about
this in the docs: http://dev.apollodata.com/core/fragments.html#unique-names [chunk-vendors.a58588bb.js:22:152423](http://localhost:8889/cylc/js/chunk-vendors.a58588bb.js)
```
And similar additional errors, for the tree view fragments, on clicking on a workflow.

hjoliver:
Also saw this in Firefox dev tools console when playing with a small-ish workflow:

```
Will-change memory consumption is too high. Budget limit is the document surface area multiplied by 3 (1108080 px). Occurrences of will-change over the budget will be ignored.
```
https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

Weirdly, `git grep` doesn't show any use of `will-change` in the repo. ??

hjoliver:
The only graph view display issue I've found so far is:

If I set n-window extent down to zero, the graph view gets (almost) down to single separate nodes (as it should), but then stops updating. The tree view carries on normally (with `n=0`). Setting `n` back to `1` again restores normal behaviour.

oliver-sanders:
> Warning: fragment with name WorkflowData already exists.

I'm aware of this, I renamed all of the fragments to match in order to ensure they got merged. Might be able to change, this warning doesn't appear to impact the output subscription.

> Will-change memory consumption is too high.

Dammit, I've not seen any browser warnings yet, can you reproduce it?

> If I set n-window extent down to zero, the graph view gets (almost) down to single separate nodes (as it should), but then stops updating. 

That actually makes sense as the graph wont re-layout if there are no nodes or edges. I guess I should remove the edge test otherwise orphan-only graphs won't get rendered.

oliver-sanders:
As a follow on I've added a new development-only view in #1139 as functional documentation to help with view writing.

oliver-sanders:
* Graphs with no edges should now update.
* Added offline data (so you can test the graph with `yarn run serve`).
* Added a basic functional test.
* Fixed the WASM packaging problem.

oliver-sanders:
(force pushed to fix the lint error)

oliver-sanders:
Firefox is no longer installed the new Ubuntu 22 image causing e2e test failures.

Same failure on master: https://github.com/cylc/cylc-ui/actions/runs/3498091931/jobs/5858007858

wxtim:
I cannot reproduce the bugs I had before in Graph view! 👍🏼

oliver-sanders:
@wxtim can you confirm the list of bugs in the description are closed by this PR?

datamel (commented):
I have had a really good play with this branch and it is working really smoothly. I have been working through checking bug fixes. 

Just wondering if anyone is seeing this traceback in uiserver logs on this branch? 

```
[E 2022-11-21 11:55:06.752 CylcUIServer] 'families'
    Traceback (most recent call last):
      File "~/cylc-uiserver/cylc/uiserver/data_store_mgr.py", line 291, in _apply_all_delta
        if delta_time >= self.data[w_id]['delta_times'][field.name]:
    KeyError: 'families'
[E 2022-11-21 11:55:06.753 CylcUIServer] 'tasks'
    Traceback (most recent call last):
      File "~/cylc-uiserver/cylc/uiserver/data_store_mgr.py", line 291, in _apply_all_delta
        if delta_time >= self.data[w_id]['delta_times'][field.name]:
    KeyError: 'tasks'
[E 2022-11-21 11:55:06.753 CylcUIServer] 'families'
    Traceback (most recent call last):
      File "~/cylc-uiserver/cylc/uiserver/data_store_mgr.py", line 291, in _apply_all_delta
        if delta_time >= self.data[w_id]['delta_times'][field.name]:
    KeyError: 'families'
[E 2022-11-21 11:55:06.753 CylcUIServer] 'tasks'
    Traceback (most recent call last):
      File "~/cylc-uiserver/cylc/uiserver/data_store_mgr.py", line 291, in _apply_all_delta
        if delta_time >= self.data[w_id]['delta_times'][field.name]:
    KeyError: 'tasks'

```

I got this when restarting a workflow, which hadn't been run with the ui before. Very possible it is unrelated to these changes though.

oliver-sanders:
(rebased to pick up e2e test fix)

wxtim:
I have now checked all the bugfixes except the one which requires safari (I'll see if I can get it on home laptop, but not going to worry if it takes long).

I have found a new issue (https://github.com/cylc/cylc-ui/issues/1145 ) whilst checking #838.

IMO this issue is small(ish), and should _not_ block this PR from merging.

oliver-sanders:
The last commit should fix the Firefox test.

The one before helps with debugging Cypress tests by uploading any screenshots of the test failures as artefacts.

tests/e2e/specs/graph.cy.js
```diff
@@ -0,0 +1,71 @@
...
+    .then(($el) => { checkGraphLayoutPerformed($el) })
+}
+
+function checkGraphLayoutPerformed ($el, depth = 0) {
+  // Check if the graphID has been set (indicating successful layout) or wait
+  // one second.
+  // This is a recursive function which will be called up to 10 times. We can't
+  // do this in a for loop as otherwise the waits aren't chained correctly.
+  if (depth > 10) {
+    expect('graph loaded').to.equal(true)
```
MetRonnie, line 31: Unless I am missing something, this is `'graph loaded' === true` which will always fail. Presumably it's not being hit?

oliver-sanders: Yep, this test is guaranteed to fail. It gets called if the graph doesn't load in time.

cypress/component/cylc-graph-node.cy.js
```diff
@@ -0,0 +1,214 @@
...
+  for (const jobState of jobStates) {
+    const jobTokens = tokens.clone({ job: `0${itt}` })
+    job = {
+      id: jobTokens.id,
+      name: jobTokens.job,
+      node: {
+        state: jobState
+      }
+    }
+    if (jobState === 'running') { // TODO constant
```
wxtim, line 40: Comment is at best cryptic, at worst suggests somthing else you want to do.

src/components/cylc/SVGTask.vue
```diff
@@ -0,0 +1,486 @@
...
+          y="25"
+          width="16"
+          height="50"
+          rx="10"
+          ry="10"
+        />
+      </g>
+      <!-- queued
+
+        Burger bar-like icon representing isQueued.
```
wxtim, line 208: Same symbol as "fog, sky not visible"

wxtim (approved):
My review is largely functional, but I have had a quick (per file, it took _ages_) poke through the changes.

src/components/cylc/Task.vue
```diff
@@ -19,468 +19,53 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-      type: Boolean,
-      require: true
-    },
     startTime: {
-      type: String,
-      required: false,
-      default: ''
+      // TODO: aim to remove this in due course
+      // (we should be able to obtain this directly from the task)
+      required: false
```
datamel, line 47: Does this need an issue created?

src/components/cylc/gscan/GScan.vue
```diff
@@ -228,9 +230,7 @@ export default {
         GSCAN_DELTAS_SUBSCRIPTION,
         {},
         'root',
-        [
-          new GScanCallback()
-        ]
+        []
```
datamel, line 233: Nice that these callbacks have gone. I was wondering about these when doing the log view.

src/components/cylc/tree/Tree.vue
```diff
@@ -203,10 +231,37 @@ export default {
...
-      }
+      },
+      cyclePointsOrderDesc: true
     }
   },
+  mounted () {
+    // set cyclePointsOrderDesc
+    // NOTE: this isn't reactive, however, changing the value requires
+    // navigating away from this view so it doesn't have to be
+    // TODO: make this a view-specific configuration
```
datamel, line 242: Does this need an issue?

datamel (approved):
I've had another good functional review yesterday and this morning. Nothing I have found will block this merge and so approving. Couple of small notes, whether they require fixing is up for debate and can certainly be done in follow-ups...

- Transpose, I noticed, in my case where my simple linear graph (`task1 => task2...=>last-task`) had stopped the transpose no longer worked and simply shifted the arrow a little. I couldn't replicate this while the workflow was playing so I expect it is no big deal.
- When pulling the graph view to a vertical pane on the right of the window, the graph didn't auto-centre and so it was tricky to see the tasks. Again, no big deal, since clicking the centre button myself fixed this issue. 
- The timer countdown animation continues to go down when the task is held.
 
Code-wise, thanks for all the comments. It has helped my understanding of the changes and made things much easier. My vue is still in the learning stages, I've not spotted anything that needs changing.
Data-store change-wise, the changes look good to me, I have looked at the structure and it makes sense to me.
Subscription changes-wise, I'm still a little lacking in understanding as to why we have an onAdded and onUpdated as it looks to me that the same info is coming in on both, although this is not really related to your pr and is just something I don't fully understand and will pick your brains about another time.

All in all, very impressive pr 🥇 👏 Any comments I have are very minor.

oliver-sanders:
@datamel could you elaborate on this one:

> The timer countdown animation continues to go down when the task is held.

Is that a running task that was held whilst it was running? Or was the task status not running but the clockface showing erroneously?

tests/e2e/specs/graph.cy.js
MetRonnie: Ah, I understand now. According to the chai docs you should be able to do
```suggestion
    expect.fail('graph did not load')
```

src/components/cylc/tree/Tree.vue
oliver-sanders:
```suggestion
    // TODO: make this a view-specific configuration
    // https://github.com/cylc/cylc-ui/issues/1146
```

src/components/cylc/Task.vue
oliver-sanders:
```suggestion
```

datamel:
> Is that a running task that was held whilst it was running?  
This one, I guess maybe problem with the held icon being there rather than the clockface?

oliver-sanders:
Could do with a screenshot or something as I'm not sure I understand.

"Holding" a task doesn't actually pause its execution, it just suppresses new job submissions so I would expect the clockface to continue to count-down for a running task even if it's held.

oliver-sanders:
The last commit should enable component tests in CI. Will add comparisons to known-good-output with #178 to ensure task/job/graph-node icons aren't accidentally changed.

oliver-sanders:
Ok, that was a total failure, I'll get the component tests sorted out in the next PR.

PR #1112, opened 2022-10-14
Fix table view delta-T column.

hjoliver:
The mean delta-T (mean elapsed run time) column in the table view was not getting populated.

Also tidy the table header a bit.

Screenshot from 2022-10-14 17-50-27

**Requirements check-list**

### DIFF
--- a/src/components/cylc/table/Table.vue
+++ b/src/components/cylc/table/Table.vue
@@ -110,6 +110,12 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             item-key="id"
             show-expand
           >
+
+            <!-- eslint-disable-next-line -->
+            <template v-slot:header.node_task_meanElapsedTime="{ header }">
+              <span v-html="header.text"></span>
+            </template>
+
             <template
               v-slot:item="{ item }"
             >
@@ -153,7 +159,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                 <td>{{ item.latestJob.submittedTime }}</td>
                 <td>{{ item.latestJob.startedTime }}</td>
                 <td>{{ item.latestJob.finishedTime }}</td>
-                <td>{{ item.meanElapsedTime }}</td>
+                <td>{{ (item.node.task.meanElapsedTime).toFixed() }}</td>
               </tr>
             </template>
             <template v-slot:expanded-item="{ item }">
@@ -233,7 +239,7 @@ export default {
       headers: [
         {
           text: 'Task',
-          value: 'node.name',
+          value: 'node_name',
           sort: DEFAULT_COMPARATOR
         },
         {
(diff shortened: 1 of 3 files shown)


src/components/cylc/table/Table.vue
```diff
@@ -233,7 +239,7 @@ export default {
       headers: [
         {
           text: 'Task',
-          value: 'node.name',
+          value: 'node_name',
```
oliver-sanders, line 242: Where is `node_name` defined?

src/components/cylc/table/Table.vue
```diff
@@ -110,6 +110,12 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             item-key="id"
             show-expand
           >
+
+            <!-- eslint-disable-next-line -->
+            <template v-slot:header.node_task_meanElapsedTime="{ header }">
+              <span v-html="header.text"></span>
+            </template>
+
```
oliver-sanders, line 118: The view seems to work the same with or without this template.

```suggestion
```

src/components/cylc/table/Table.vue
```diff
@@ -243,42 +249,42 @@ export default {
         },
         {
           text: 'Cycle Point',
-          value: 'node.cyclePoint',
+          value: 'node_cyclePoint',
```
oliver-sanders, line 252: I don't think these changes work, it breaks sorting the table columns.

src/components/cylc/table/Table.vue
```diff
@@ -243,42 +249,42 @@ export default {
...
-          text: 'T-finish',
-          value: 'latestJob.finishedTime',
+          text: 'T-Finish',
+          value: 'latestJob_finishedTime',
           sort: datetimeComparator
         },
         {
-          text: 'dT-mean',
-          value: 'meanElapsedTime',
+          text: '&Delta;T-Mean',
```
oliver-sanders, line 286: `&Delta;` doesn't get expanded into a delta symbol. If you wanted the unicode delta symbol here, I think you can write it in plain text, no escape needed, the web runs on UTF-8.

oliver-sanders (changes requested):
The "T-Mean" column works but this breaks column sorting (but not the tests, yay).

Raised a PR: https://github.com/hjoliver/cylc-ui/pull/60

oliver-sanders:
I've applied the fix from this PR onto master and added some tests in #1164.

MetRonnie:
I'm closing as superseded by #1164

hjoliver:
Thanks for taking this over in my absence @oliver-sanders

src/components/cylc/table/Table.vue
hjoliver: BTW this did render a Delta symbol - see the screenshot on this issue. The reason was the `v-html` template you commented on above https://github.com/cylc/cylc-ui/pull/1112/files#r1045695288 ... However, you're right about just using UTF-8!

PR #1123, opened 2022-10-25
Centralise task filtering into own component; list task states in logical order

### LINKED ISSUE #1118: view toolbar: task state filter should list task states in logical order
The task state filter used in some of the views (e.g. tree view) lists task states in alphabetical order.

This isn't easy to read, suggest using logical state order intended for display purposes which is defined here:

https://github.com/cylc/cylc-ui/blob/35a39e3c55912cecddd07be94067a9e24d092568/src/model/TaskState.model.js#L47-L56

**Pull requests welcome!**

MetRonnie:
- The task filtering HTML and code was duplicated in the Tree and Table components. I have extracted it into its own component and simplified the code
- Closes #1118

**Requirements check-list**
- [x] No change log entry required (mostly invisible to users).

### DIFF
--- a/src/components/cylc/common/filter.js
+++ b/src/components/cylc/common/filter.js
@@ -1,0 +1,38 @@
+/**
+ * Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* Logic for filtering tasks. */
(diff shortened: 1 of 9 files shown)


oliver-sanders:
Due to the refactoring this will be a tricky one to reconcile with the graph view / data store work as filtering works in a different way when views are powered from the central data store.

hjoliver:
> Due to the refactoring this will be a tricky one to reconcile with the graph view / data store work ...

What's your suggestion - shelve this one?

MetRonnie:
I have looked at #1108 and I don't see too much surface area for conflict, only a handful of lines

oliver-sanders:
I haven't got filtering working in all views on that branch yet. The main difference is that in order to drive the views off of the central data store we cannot change the data or set attributes on it directly. Instead we must store the filter state in the view locally.

I'm going to try to get back to that branch this week to see if I can clear out the remaining issues and find out.

MetRonnie:
Note that currently this new filter component does not house the logic of the filtering, that is left up to the parent components that use/embed the filter component

MetRonnie:
Converting to draft, will be easier to deconflict this one after #1108 has gone in

src/components/cylc/common/filter.js
```diff
@@ -0,0 +1,38 @@
+/**
```
oliver-sanders, line 1: Perhaps move into src/utils?

oliver-sanders (approved):
LGTM

wxtim (approved):
- [x] Read code.
- [x] Tried functionality.

PR #1124, opened 2022-10-25
more items per page options in table view

### LINKED ISSUE #1121: table: increase rows per page
The default number of rows-per-page is 10, which is quite low. From the dropdown we have the option to change this to 5, 10, 15 or all.

I would suggest values like 50, 100, 200 (& all) would be more appropriate for this view.

When attempting this do some quick testing to ensure the table view performance is acceptable for the pre-defined values.

**Pull requests welcome!**

wxtim:
These changes close #1121 

### Comments on changes

As well as modifying the numbers of tasks per page I turned on the table [dense mode](https://vuetifyjs.com/en/components/data-tables/#slots) - I don't think it loses owt, and gains more space:

dense

I've provided 20, 50, 100 and 200 - I couldn't find any way of defaulting to 20 but having 10 available - It's fine even on a small screen:

Screen Shot 2022-10-25 at 15 40 11

Additionally I tried it with a workflow with 600 tasks per cycle and it seemed to be as performant as one might reasonably expect (200 took under a second, 2400 took about 5).

Finally, since it's a simple change I added a first and last page icon to the table.

**Requirements check-list**

- [x] Does not need tests - manually tested

### DIFF
--- a/src/components/cylc/table/Table.vue
+++ b/src/components/cylc/table/Table.vue
@@ -114,6 +114,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               itemsPerPageOptions: [10, 20, 50, 100, 200, -1],
               showFirstLastPage: true
             }"
+            :options="{ itemsPerPage: 50 }"
           >
             <template
               v-slot:item="{ item }"
--- a/src/components/cylc/table/Table.vue
+++ b/src/components/cylc/table/Table.vue
@@ -109,6 +109,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             :sort-desc.sync="sortDesc"
             item-key="id"
             show-expand
+            dense
+            :footer-props="{
+              itemsPerPageOptions: [10, 20, 50, 100, 200, -1],
+              showFirstLastPage: true
+            }"
           >
             <template
               v-slot:item="{ item }"
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,11 @@ ones in. -->
 
 ## __cylc-ui-1.4.0 (<span actions:bind='release-date'>Pending</span>)__
 
+### Enhancements
+
+[#1124](https://github.com/cylc/cylc-ui/pull/1075) - Table view: More options
+for number of tasks per page.
+
 ### Fixes
 
 [#1075](https://github.com/cylc/cylc-ui/pull/1075) - Reverse default sort order


src/components/cylc/table/Table.vue
```diff
@@ -109,6 +109,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             :sort-desc.sync="sortDesc"
             item-key="id"
             show-expand
+            dense
```
MetRonnie, line 112: It might be good to allow the user to choose dense mode or not in Settings or perhaps via a checkbox in the table footer

MetRonnie (approved):

hjoliver:
I would still prefer to have 10 as an option, albeit not the default. For [this reason](https://github.com/cylc/cylc-ui/issues/1121#issuecomment-1291046247).

src/components/cylc/table/Table.vue
hjoliver: I prefer dense mode, having just tried it.  No harm in having a choice though, I guess.

hjoliver (commented):
Looks good, but I think we do still need 10 lines as an option.  It's not just small screens, but also this sort of tab layout (which will be common because of the horizontal extent of the table view):

(Again, in a paginated view, it is best to have - at least as an option - a whole page per page, otherwise you have to use both scrolling AND pagination).

shot

src/components/cylc/table/Table.vue
wxtim: I considered it - I'll see if it's straightforward.

wxtim:
> I would still prefer to have 10 as an option, albeit not the default. For [this reason](https://github.com/cylc/cylc-ui/issues/1121#issuecomment-1291046247).

Happy to have another look, but I couldn't work out how to include 10 but not make it the default.

src/components/cylc/table/Table.vue
MetRonnie: Decision: forget about it for now

src/components/cylc/table/Table.vue
```diff
@@ -109,6 +109,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+            :footer-props="{
+              itemsPerPageOptions: [10, 20, 50, 100, 200, -1],
+              showFirstLastPage: true
+            }"
```
MetRonnie, line 116: Looks like you can set the default like so
```suggestion
            }"
            :options="{ itemsPerPage: 50 }"
```

MetRonnie (approved):

oliver-sanders (approved):

oliver-sanders:
> Looks good, but I think we do still need 10 lines as an option

10 line option has been restored.

PR #1151, opened 2022-11-30
router: strip the ?token when the app is first loaded

### LINKED ISSUE #1147: jupyter-server: strip token from URL
When using `cylc gui` in single-user mode, Jupyter Server provides a URL containing a `?token`. This is used to create a secure cookie which is then used for authentication.

Once this cookie has been created the token can be safely removed from the URL as the cookie is now in charge of authentication. I thought I had put in logic to do this at some point, however, from experimentation the token appears to remain in the URL.

**Pull requests welcome!**

oliver-sanders:
* Closes #1147.
* The token is used to set a secure cookie.
* We don't need it after that so we may as well remove it from the URL.
* This is cleaner and reduces the temptation for users to copy it with the URL.

**Requirements check-list**

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -35,6 +35,16 @@ Vue.prototype.$eventBus = mitt()
 
 Vue.config.productionTip = false
 
+if (location.search) {
+  // Vue Router puts the "hash" (#) after the "search" (?) for some reason.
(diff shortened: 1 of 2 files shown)


datamel (approved):
Working smoothly for me on test. Thanks @oliver-sanders.

PR #1153, opened 2022-12-01
Revert "router: strip the ?token when the app is first loaded"

oliver-sanders:
Reverts cylc/cylc-ui#1151

Sorry all, I think I've jiggered up here.

I've just tried launching `cylc gui`, but it just gave me the "Connection to server lost. You are offline." notice. The second time I opened the link it worked fine.

I think there's a race condition going on here where the token may, or may not still be present in the URL when it's required, dammit.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -35,16 +35,6 @@ Vue.prototype.$eventBus = mitt()
 
 Vue.config.productionTip = false
 
-if (location.search) {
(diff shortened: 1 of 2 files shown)


datamel (approved):
Yes, must be a race condition. I have not noticed it using the change but happy to revert since you are getting the error.

PR #1164, opened 2022-12-12
table: fix dt-mean

oliver-sanders:
Applies the fix from #1112 onto master and adds tests.

**Requirements check-list**


src/components/cylc/table/Table.vue
```diff
@@ -372,6 +372,13 @@ export default {
       // I don't really like this, but we need to somehow force the 'change detection' to run again once the clear has taken place
       this.tasksFilter.name = null
       this.$refs.filterNameInput.$el.querySelector('input').dispatchEvent(new Event('keyup'))
+    },
+    dtMean (taskNode) {
+      const ret = taskNode.node?.task?.meanElapsedTime
+      if (ret) {
+        return ret.toFixed()
+      }
+      return undefined
```
MetRonnie, line 381:
```suggestion
      return taskNode.node?.task?.meanElapsedTime?.toFixed()
```

MetRonnie (commented):
What about the delta symbol instead of "d"?

oliver-sanders:
> What about the delta symbol instead of "d"?

Could do but I'm not sure that's a positive change. Should probably rethink the timing headers some time.

MetRonnie:
Oh, test failing after my suggestion was committed because it was expecting a ΔT of 0 to be converted to ''. I have had a look at a real workflow and it seems that runahead tasks that have not run in a previous cycle get a mean elapsed time of 0 when it should probably be nullish (running tasks that have not run in a previous cycle get a nullish value). In the mock data there are several tasks that have a mean elapsed time of 0 which should probably be >0 if the task has a T-finish, else nullish.

Or, you could just undo my suggestion

oliver-sanders:
Going to revert the suggestion in order to keep this as robust as possible.

src/components/cylc/table/Table.vue
```diff
@@ -372,6 +372,13 @@ export default {
       // I don't really like this, but we need to somehow force the 'change detection' to run again once the clear has taken place
       this.tasksFilter.name = null
       this.$refs.filterNameInput.$el.querySelector('input').dispatchEvent(new Event('keyup'))
+    },
+    dtMean (taskNode) {
+      const ret = taskNode.node?.task?.meanElapsedTime
+      if (ret) {
```
MetRonnie, line 378:
```suggestion
      // Only return a number if > 0, because 0 may be used to indicate the task has not run before
      if (ret) {
```

MetRonnie (approved):
1 suggestion for clarifying comment

wxtim:
@MetRonnie  - this now conflicts with master after merger of #1123

oliver-sanders:
Rebased

MetRonnie (approved):
I have tested it out and it's working for me. Just the changelog that is conflicting

wxtim:
- [x] Proof of pudding.

wxtim (approved):
- [x] Read code.

PR #1172, opened 2022-12-14
eslint

oliver-sanders:
A one off run of "[prettier](https://prettier.io)" over the codebase as mentioned on Element.

Change the style rules to something a little "prettier" and more familiar.

* Drop "standard" rules.
      The "standard" style is somewhat questionable and a bit of a pain to work with.
* Add replacement rules.
      Enable a few eslint rules (all with auto-fixers) configured to match the prevailing style / expectations from other languages.
* Change operator order.
      Place operators after line breaks rather than before to match prevailing opinion.
* Reduce errors to warnings when in development mode.

Run linting with production rather than development rules.

Obvs this is going to conflict like hell with other work so I've waited for major branches to get out of the way.


oliver-sanders:
Could consider doing a one-off run of `prettier` here too?

src/components/cylc/common/sort.js
```diff
@@ -68,21 +63,29 @@ export const DEFAULT_COMPARATOR = (left, right) => {
...
   // If given a function, use it. Otherwise, simply use identity function.
   const iterateeFunction = iteratee || ((value) => value)
-  // If given a function, use it. Otherwise, simply use locale sort with numeric enabled
-  const comparatorFunction = options.comparator || ((leftObject, leftValue, rightObject, rightValue) => DEFAULT_COMPARATOR(leftValue, rightValue))
+  // If given a function, use it. Otherwise, simply use locale sort with
+  // numeric enabled
+  const comparatorFunction
+    = options.comparator
+    || ((leftObject, leftValue, rightObject, rightValue) =>
+      DEFAULT_COMPARATOR(leftValue, rightValue))
```
MetRonnie, line 88: Can't say I'm a fan of line break before assignment operators 🤮 But I've had a look and can't see any way to set `operator-linebreak` to `after` for assignment operators only.

(P.S. apparently I am in the minority but I generally prefer line breaks after logical operators too, reason being my brain reads a line as a "unit", so if there is no operator at the end my brain assumes that's the end of the statement, only to discover there is another condition when my eyes jump to the next line. Whereas if there is an operator at the end I know in what context I need to process the next line (including whether it could be skipped to due short-circuiting))

MetRonnie (commented):
I have some further changes:

> * Fix an import that got broken
> * The line breaks in JSDoc comments that had `- ` added to them show up as bullet points in VSCode, so I got rid of the dash
> * Made it so function calls should not have trailing commas

https://github.com/oliver-sanders/cylc-ui/pull/97

oliver-sanders:
Too late to save this branch, will revisit this again when we have less work going on in the UI.

PR #1175, opened 2022-12-20
Attempt 2 collapsing gscan

AaronDCole:
Attempt number 3 at the auto collapse workflows, this may not work flawlessly yet but would be good to get some review on based on the feedback from Oliver on the previous pull request (https://github.com/cylc/cylc-ui/pull/1037).

This is a ground up rewrite of the original feature request, based on the latest changes on developer (including the most recent changes from Oliver), with (hopefully) a lot of code quality improvements and general readability updates (since I found the structure of the g-scan element incredible hard to follow / change in a recursive fashion (as suggested in Olivers comments on the previous MR, listed above).

Tests will be broken, but an initial review would be helpful (even if just a sighting pass), then Ill make final adjustments and then fix the tests.

**Requirements check-list**

### DIFF
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -139,8 +139,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                   </v-col>
                   <!-- We check the latestStateTasks below as offline workflows won't have a latestStateTasks property -->
                   <v-col
-                    v-if="!branchingLineage && (lastDescendent.type === 'workflow' && lastDescendent.node.latestStateTasks)"
-                    class="text-right c-gscan-workflow-states flex-grow-0"
+                    v-if="!branchingLineage && (lastDescendent.type === 'workflow' && lastDescendent.node.latestStateTasks) && !expansionStatus"
+                    class="text-right c-gscan-workflow-states"
                   >
                     <!-- task summary tooltips -->
                     <span
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -111,7 +111,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           <template v-slot:node="{node, latestDescendantTasks, lastDescendent, descendentLabel, branchingLineage, expansionStatus}">
             <workflow-icon
               class="mr-2"
-              v-if="lastDescendent.type === 'workflow' && !branchingLineage"
+              v-if="!branchingLineage && lastDescendent.type === 'workflow'"
               :status="lastDescendent.node.status"
               v-cylc-object="lastDescendent"
             />
@@ -139,7 +139,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                   </v-col>
                   <!-- We check the latestStateTasks below as offline workflows won't have a latestStateTasks property -->
                   <v-col
-                    v-if="(!branchingLineage && (lastDescendent.type === 'workflow' && lastDescendent.node.latestStateTasks) && !expansionStatus) || (!expansionStatus && branchingLineage)"
+                    v-if="!expansionStatus || node.type === 'workflow'"
                     class="text-right c-gscan-workflow-states"
                   >
                     <!-- task summary tooltips -->
--- a/src/components/cylc/gscan/GScan.vue
+++ b/src/components/cylc/gscan/GScan.vue
@@ -108,10 +108,10 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           class="c-gscan-workflow ma-0 pa-0"
           ref="tree"
         >
-          <template v-slot:node="{node, latestDescendantTasks, lastDescendent, descendentLabel, branchingLineage, expansionStatus}">
+          <template v-slot:node="{node, descendantTaskTotals, latestDescendantTasks, lastDescendent, descendentLabel, branchingLineage, expansionStatus}">
             <workflow-icon
               class="mr-2"
-              v-if="!branchingLineage && lastDescendent.type === 'workflow'"
+              v-if="!branchingLineage && !expansionStatus && lastDescendent.type === 'workflow'"
               :status="lastDescendent.node.status"
               v-cylc-object="lastDescendent"
             />
@@ -142,11 +142,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                     v-if="!expansionStatus || node.type === 'workflow'"
                     class="text-right c-gscan-workflow-states"
                   >
-                    <!-- task summary tooltips -->
+<!--                     task summary tooltips-->
                     <span
                       v-for="[state, tasks] in getLatestStateTasks(Object.entries(latestDescendantTasks))"
                       :key="`${node.id}-summary-${state}`"
-                      :class="getTaskStateClasses(latestDescendantTasks, state)"
+                      :class="getTaskStateClasses(descendantTaskTotals, state)"
                     >
                     <v-tooltip color="black" top>
                       <template v-slot:activator="{ on }">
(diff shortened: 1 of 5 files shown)


src/components/cylc/gscan/GScan.vue
```diff
@@ -102,79 +96,45 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-              class="mr-2"
-              v-if="scope.node.type === 'workflow'"
-              :status="scope.node.node.status"
-              v-cylc-object="scope.node"
-            />
-            <v-list-item
-              :to="workflowLink(scope.node)"
-            >
+
+            <workflow-icon class="mr-2" v-if="scope.node.type === 'workflow' || !checkForBranchingLineage(scope.node)" :status="getLastDescendent(scope.node).node.status || 'waiting'" v-cylc-object="getLastDescendent(scope.node)" />
```
oliver-sanders, line 104: `checkForBranchingLineage` and `getLastDescendent` seem to be used multiple times which means there will be duplicate computation. It would be worth using computed properties to cache these values to avoid unnecessary computation.

src/components/cylc/gscan/GScan.vue
```diff
@@ -102,79 +96,45 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
               <v-list-item-title>
                 <v-layout align-center align-content-center d-flex flex-nowrap>
-                  <v-flex
-                    v-if="scope.node.type === 'workflow-part'"
-                    class="c-gscan-workflow-name"
-                  >
-                    <span>{{ scope.node.name || scope.node.id }}</span>
+
+                  <v-flex v-if="scope.node.type === 'workflow-part'" class="c-gscan-workflow-name">
+                    <span>{{ checkForBranchingLineage(scope.node) ? scope.node.name || scope.node.id : createDescendentLabel(scope.node) }}</span>
```
oliver-sanders, line 111: I think this test might need to use `this.isExpanded` so that the name collapsing is only performed when the node is collapsed.

```
<span>{{ (this.isExpanded && checkForBranchingLineage(scope.node)) ? scope.node.name || scope.node.id : createDescendentLabel(scope.node) }}</span>
```

E.G. the collapsed state should look like this:

```
b/b1/b11
```

The expanded state should look as it did before:

```
b/
  b1/
    b11/
```

Rather than:

```
b/b1/b1
  b1/b11
    b11/
```

oliver-sanders (commented):
Thanks @AaronDCole.

I'm not terribly keen on the HTML line break removals which are also making it hard to work out exactly what this PR has changed. It would be easier to consider code style changes in a separate issue/PR (see also #1172), we may also consider using prettier (or similar) with Vue support for HTML/SCSS formatting.

The auto-collapsing works well for the examples :+1:, although for some strange reason, when I tried it with a workflow named `generic/run1` it didn't work and the workflow icon broke?

gscan

I also spotted an issue with the workflow names when these nodes are expanded (more in later comment):

gscan-3

The "Tree Component" `src/components/cylc/tree/Tree` is used by both the "GScan Component" and the "Tree View" (`src/views/Tree`) which has the unfortunate side effect that this auto-collapsing also impacts the tree view which is undesirable. For now we can use a simple toggle switch to turn this off (more in later comment).

gscan-2

src/components/cylc/gscan/GScan.vue
```diff
@@ -102,79 +96,45 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-          :filterable="false"
-          :expand-collapse-toggle="false"
-          :workflows="workflows"
-          :stopOn="['workflow']"
-          :autoExpandTypes="['workflow-part', 'workflow']"
-          class="c-gscan-workflow ma-0 pa-0"
-          ref="tree"
-        >
+      <div v-if="!isLoading" class="c-gscan-workflows flex-grow-1">
+        <tree :filterable="false" :expand-collapse-toggle="false" :workflows="workflows" :stopOn="['workflow']" :autoExpandTypes="['workflow-part', 'workflow']" class="c-gscan-workflow ma-0 pa-0" ref="tree">
```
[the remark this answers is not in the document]
AaronDCole: Sorry you're totally right, I didn't actually intend to include that and I will fix/revert that in the next commit.

src/components/cylc/gscan/GScan.vue
AaronDCole: Good point!

src/components/cylc/gscan/GScan.vue
AaronDCole: 'The auto-collapsing works well for the examples 👍, although for some strange reason, when I tried it with a workflow named generic/run1 it didn't work and the workflow icon broke?'

I actually worked out why this was happening after I left the office last night, I will fix it today

oliver-sanders:
This feature can be tested in a component test but at the moment the code structure doesn't allow this (because we would have to mock all the GraphQL/VueX data stuff). I've split the GraphQL/VueX stuff out of the GScan component into a GScan view to allow the component to be tested standalone.

Feel free to cherry-pick this commit, it contains a blank component test with the `a/a1` example workflow pre-loaded.

https://github.com/cylc/cylc-ui/compare/master...oliver-sanders:cylc-ui:gscan-component?expand=1

There are a couple of simple component tests to look at but we haven't started using them widely yet. You should be able to `cy.get` the expand/collapse button and `cy.click` it, you can tell if a row is collapsed using `cy.should('be.visible')` or by inspecting the vm and looking at the `isExpanded` attribute.

AaronDCole:
> This feature can be tested in a component test but at the moment the code structure doesn't allow this (because we would have to mock all the GraphQL/VueX data stuff). I've split the GraphQL/VueX stuff out of the GScan component into a GScan view to allow the component to be tested standalone.
> 
> Feel free to cherry-pick this commit, it contains a blank component test with the `a/a1` example workflow pre-loaded.
> 
> https://github.com/cylc/cylc-ui/compare/master...oliver-sanders:cylc-ui:gscan-component?expand=1
> 
> There are a couple of simple component tests to look at but we haven't started using them widely yet. You should be able to `cy.get` the expand/collapse button and `cy.click` it, you can tell if a row is collapsed using `cy.should('be.visible')` or by inspecting the vm and looking at the `isExpanded` attribute.

Thanks Oliver, I am refactoring a lot of the commit (to, for example, be able to pass through computed properties from the TreeItem component to use where the slot overrides are defined in the gScan component), so I will look into this when that is finished.

AaronDCole:
There is an issue where when you switch to the tree view, it seems to change the format of the node object which breaks a lot of stuff within the gscan component. I will investigate tomorrow.

hjoliver:
Taking a look today ...

hjoliver:
@AaronDCole - the placement of job icons looks good now as I collapse and expand the tree, but I've noticed a couple of problems.

I tested it side-by side with master, by moving the `dist` directory from building this branch to `dist-scan`, and on master to `dist-master`, then starting two UIs:

```console
$ cylc gui --CylcUIServer.ui_build_dir="/home/oliverh/cylc/cylc-ui/dist-gscan"
$ cylc gui --new --CylcUIServer.ui_build_dir="/home/oliverh/cylc/cylc-ui/dist-master"
```

Then I started 4 copies of this workflow:
```ini
[scheduling]
   cycling mode = integer
   runahead limit = P2
   [[graph]]
      P1 = "a => b1 & b2 & b3 "
[runtime]
   [[a, b1, b2, b3]]
       script = sleep $((10 + RANDOM % 5))
```
... at different levels of run-dir nesting, like this:

```console
$ cylc vip  # bgg/run1
$ cylc vip  # bgg/run2
$ cylc vip --workflow-name=top/mid/one  # top/mid/one/run1
$ cylc vip --workflow-name=top/mid/two  # top/mid/two/run1
```

The two problems are:
- the workflow icon shouldn't appear at the 2nd level from the bottom (see red arrow, below)
- the job icon state totals aren't responding properly at the lowest level (see red loop, below). In my test cases here, for instances, the tooltip always says 5 running, even if there are none running (watch out for this bug if you pause a workflow, but the problem is more than that: https://github.com/cylc/cylc-flow/issues/5417

gscan-1

To get the state totals right you could put some really long sleep values in some tasks, so that the totals stay stable for a long time.  And use `cylc scan -t rich --colour-blind` or `cylc dump -t [workflow-ID]` to see the right numbers per workflow.

hjoliver:
Tests pass 🎉  I'm going to build and test later this afternoon...

hjoliver:
This is looking good. I'll try to find time to leave some review comments this afternoon, and to prompt others on the team to review it. (Bearing in mind the Vue3 changes in master now).

hjoliver:
superseded by #1406

Issue #888, opened 2022-01-20
Filtering: case sensitivity and family names


MetRonnie:
- [ ] Filtering in the table/tree views is currently case sensitive. Should this be the case?
- [ ] You can't filter by family name. Should we be able to?

**Pull requests welcome!**

hjoliver:
> Filtering in the table/tree views is currently case sensitive. Should this be the case?

Options:
  1. case insensitive  (OK if caps are rarely used)
  2. case insensitive with checkbox to make it case sensitive
  3. "intelligent":  if all lowercase: case insensitive; otherwise (any uppercase chars): case sensitive

Options 3 is nice IMO (I use this for vim search). Most of the time it's super convenient, but occasionally you might net more results than you want.  (Assumes most search targets are all lowercase, which is reasonable).

> You can't filter by family name. Should we be able to?

Yes.

oliver-sanders:
I think "case insensitive, with checkbox to make it case sensitive" is fine.

I think search should work on all levels of the tree, cycle points and families included.

PR #1182, opened 2023-01-04
Allow filtering by ID instead of only task name

### LINKED ISSUE #1220: Task filtering in tree and table view raising TypeError
**Describe the bug**

Task filtering is broken for me in tree and table view on master branch. I get `TypeError: node.name is undefined`. 

image

Can anyone confirm before I have a go at fixing?

**Pull requests welcome!**

MetRonnie:
These changes partially address #888

- Filtering in the tree/table views now supports relative IDs so you can filter by cycle point, family name, task name or a combination thereof
- Changed the "clear" button on filter inputs from an X in a circle to just an X, to differentiate from the failed task icon

Closes #1220
- Fix bug where filtering by task name was broken

**Requirements check-list**
- [x] Appropriate tests are included


hjoliver (commented):
I see you've now removed `__slots__` declarations from the Interval subclasses. Python docs seems to say this means instances will get an attributes dict as well:

> The action of a __slots__ declaration is not limited to the class where it is defined. __slots__ declared in parents are available in child classes. However, child subclasses will get a [__dict__](https://docs.python.org/3/library/stdtypes.html#object.__dict__) and __weakref__ unless they also define __slots__ (which should only contain names of any additional slots).

I guess that means subclasses with no additional attributes should declare an empty slots tuple?

datamel (approved):
Checked out, code read and manually tested filtering by id. Fixes the issue for me. Thanks @MetRonnie

hjoliver (approved):
Looks good, tests as working 👍

PR #1184, opened 2023-01-05
show mean times as iso durations

### LINKED ISSUE #1180: table: mean time human readable format
Currently we display the mean time in seconds, this isn't very helpful as many tasks may take hours to run.

The time should be displayed in a human readable format I.E. [HH:]MM:SS, note this needs to play with table sorting.

**Pull requests welcome!**

wxtim:
These changes close #1180

In table view convert mean time from a number of seconds to an iso duration.

Question - do we want to make it mm:ss if duration < 3600?

**Requirements check-list**

### DIFF
--- a/src/utils/tasks.js
+++ b/src/utils/tasks.js
@@ -105,3 +105,28 @@ export {
   latestJob,
   jobMessageOutputs
 }
+
+export function dtMean (taskNode) {
+  // Convert to an easily read duration format:
+  const dur = taskNode.node?.task?.meanElapsedTime
+  if (dur) {
+    const seconds = dur % 60
+    const minutes = ((dur - seconds) / 60) % 60
+    const hours = ((dur - minutes * 60 - seconds) / 3600) % 24
+    const days = (dur - hours * 3600 - minutes * 60 - seconds) / 86400
+
+    let dayss = ''
+    if (days > 0) {
+      dayss = days.toString() + 'd '
+    }
+
+    const ret = dayss +
+      hours.toString().padStart(2, '0') +
+      ':' + minutes.toString().padStart(2, '0') +
+      ':' + Math.round(seconds).toString().padStart(2, '0')
+    return ret
+  }
+  // the meanElapsedTime can be undefined (e.g. task has not run before)
+  // return "undefined" rather than a number for these cases
+  return undefined
+}
--- a/src/utils/tasks.js
+++ b/src/utils/tasks.js
@@ -120,11 +120,10 @@ export function dtMean (taskNode) {
       dayss = days.toString() + 'd '
     }
 
-    const ret = dayss +
+    return dayss +
       hours.toString().padStart(2, '0') +
       ':' + minutes.toString().padStart(2, '0') +
       ':' + Math.round(seconds).toString().padStart(2, '0')
-    return ret
   }
   // the meanElapsedTime can be 0/undefined (i.e. task has not run before)
   // return "undefined" rather than a number for these cases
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -34,6 +34,7 @@
     "graphql": "~15.8.0",
     "graphql-tag": "~2.12.6",
     "lodash": "^4.17.21",
+    "luxon": "^3.2.1",
```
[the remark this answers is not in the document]
wxtim: Happy to change to this, but might we not want isodatetime functions elsewhere in future?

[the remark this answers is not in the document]
oliver-sanders: It is likely that we might want other date-time functionality later but it would be better to push that decision down the road if possible.

tests/unit/utils/tasks.spec.js
```diff
@@ -90,4 +90,50 @@ describe('tasks', () => {
...
+              }
+            }
+          },
+          expected: '01:10:42'
+        }
+      ]
+      tests.forEach(test => {
+        console.log(dtMean(test.taskNode))
+        console.log(typeof dtMean(test.taskNode))
+
```
oliver-sanders, line 134:
```suggestion
```

oliver-sanders (approved):
LGTM, haven't built yet.

src/utils/tasks.js
```diff
@@ -105,3 +105,21 @@ export {
...
+
+export function dtMean (taskNode) {
+  // Use luxon.Duration to parse the mean as milliseconds,
+  // then convert to a nice iso formatted duration:
+  const dur = taskNode.node?.task?.meanElapsedTime
+  if (dur) {
+    const seconds = dur % 60
+    const minutes = (dur - seconds) % 60
+    const hours = dur - minutes - seconds
+    const ret = hours.toString().padStart(2, '0') +
```
hjoliver, line 117: I haven't tested, but does this work OK if hours > 99?  And relatedly, tasks (e.g. climate models) could potentially run for more 24 hours, so should we use days too if necessary?

wxtim: There's a whole conversation in the resolved comment about not using an external dependency where I said

> it's a local assumption that Cylc tasks won't be > 24 hours long.

However, this is a seriously minority case. I will investigate what happens if hours > 99.

I do think that days is probably unnecessary though.

oliver-sanders: How about:

```bash
[DDd] HH:MM:SS

01:02:03     # PT1H2M3S
2d 01:02:03  # P2DT1H2M3S
```

wxtim: Seems like a legit upgrade. I'll bite.

tests/unit/utils/tasks.spec.js
```diff
@@ -90,4 +90,57 @@ describe('tasks', () => {
       })
     })
   })
+  describe('dtMean', () => {
+    it('should format seconds to nice isodatetime format', () => {
+      const tests = [
+        {
+          taskNode: { node: null },
+          expected: undefined
+        },
```
MetRonnie, line 99: Should test the 0 case too
```suggestion
        },
        {
          taskNode: {
            task: {
              meanElapsedTime: 0
            }
          },
          expected: undefined
        },
```

MetRonnie (approved):
I have tested that the sorting still works correctly (it sorts based on the number of seconds rather than the string displayed in the table)

MetRonnie:
Happy to merge once tests pass

PR #1186, opened 2023-01-09
Remove deprecated Vue & Vuetify syntax

MetRonnie:
Remove stuff that is already deprecated as of Vue 2 / Vuetify 2

Vue:
- Deprecated slot syntax

Vuetify:
- Old grid/layout components
  - `v-layout` -> `v-row` or `div`
  - `v-flex` -> `v-col` or `div`

Vue-Router:
- Remove broken navigation guard that was not doing anything since first being added in #543

**Requirements check-list**
- [x] Does not need tests 

### DIFF
--- a/src/mixins/subscriptionView.js
+++ b/src/mixins/subscriptionView.js
@@ -36,10 +36,5 @@ export default {
     next(vm => {
       vm.$workflowService.startSubscriptions()
     })
-  },
-  beforeRouteUpdate (to, from, next) {
-    next(vm => {
-      vm.$workflowService.startSubscriptions()
-    })
   }
 }
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -21,8 +21,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
     grid-list
     class="c-dashboard mt-4"
   >
-    <v-layout wrap>
-      <v-flex xs12 md6 lg6>
+    <v-row wrap>
+      <v-col md="6" lg="6">
         <p class="display-1">Workflows</p>
         <v-data-table
           :headers="workflowsHeader"
@@ -32,7 +32,6 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           hide-default-header
           id="dashboard-workflows"
         >
-          <v-progress-linear slot="progress" color="grey" indeterminate></v-progress-linear>
           <template v-slot:item.count="{ item }">
             <v-skeleton-loader
               :loading="isLoading"
@@ -47,8 +46,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             <span class="title font-weight-light">{{ item.text }}</span>
           </template>
         </v-data-table>
-      </v-flex>
-      <v-flex xs12 md6 lg6>
+      </v-col>
+      <v-col md="6" lg="6">
         <p class="display-1">Events</p>
         <v-data-table
           :headers="eventsHeader"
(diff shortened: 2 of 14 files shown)


src/views/WorkflowsTable.vue
```diff
@@ -21,47 +21,21 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-            <span
-              class="subheading font-weight-light text-success text--darken-3"
-              v-text="header.text"
-            />
-          </template>
-          <v-progress-linear slot="progress" color="green" indeterminate></v-progress-linear>
-          <template
-            slot="item"
-            slot-scope="{ item }"
-          >
```
MetRonnie, line 64: These things are either unused or adequately handled by default Vuetify component behaviour

src/mixins/subscriptionView.js
```diff
@@ -36,10 +36,5 @@ export default {
     next(vm => {
       vm.$workflowService.startSubscriptions()
     })
-  },
-  beforeRouteUpdate (to, from, next) {
-    next(vm => {
-      vm.$workflowService.startSubscriptions()
-    })
```
oliver-sanders, line 43: Need to test this one very carefully.

src/views/Dashboard.vue
```diff
@@ -32,7 +32,6 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           hide-default-header
           id="dashboard-workflows"
         >
-          <v-progress-linear slot="progress" color="grey" indeterminate></v-progress-linear>
```
oliver-sanders, line 35: Undocumented change?

src/mixins/subscriptionView.js
MetRonnie: From my description in https://github.com/cylc/cylc-ui/pull/1186/commits/cc68f90eaa5e30c7e6361cf6381cb49ed06f8359

> Passing a callback to `next()` is only supported in `beforeRouterEnter()`, not `beforeRouterUpdate()`, so that was not doing anything this whole time.
> See https://v3.router.vuejs.org/guide/advanced/navigation-guards.html#in-component-guards

oliver-sanders (commented):
LGTM, need to build and have a click through.

src/views/Dashboard.vue
MetRonnie: Removed because a) I've never seen it while using the UI and b) I think the only difference to the default progress bar is the colour, not really worth defining a custom one imho

datamel (approved):
I have checked this out and run the ui, looking at the various changes and in particular at the subscription change. LGTM, one small comment about the TODO added.

src/mixins/subscriptionView.js
MetRonnie: > specifically covered by a unit test

The problem with that test was that it was passing in a function to the `next` arg that called the arg provided to that function, but vue-router does not necessarily do this - presumably it just passes in a function that is called without args.

src/views/WorkflowsTable.vue
```diff
@@ -21,48 +21,22 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-          <v-alert
-            :icon="svgPath.table"
-            prominent
-            color="grey lighten-3"
-          >
-            <h3 class="headline">{{ $t('Workflows.tableHeader') }}</h3>
-          </v-alert>
+    <v-row class="align-self-start">
+      <v-col>
+        <!-- TODO: this is not really an alert, it's a heading -->
```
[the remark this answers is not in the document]
oliver-sanders: I think either link an issue in the comment or let it slide.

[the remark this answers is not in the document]
MetRonnie: I feel like the comment as-is is at least helpful because otherwise you might be confused as to why there's a `v-alert` component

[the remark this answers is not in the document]
oliver-sanders: Then it's a `NOTE`?

oliver-sanders (approved):
Tested, LGTM.

src/mixins/subscriptionView.js
MetRonnie: From investigating #1248, this will have to be brought back in the correct form as
```suggestion
  },
  beforeRouteUpdate () {
    this.$workflowService.startSubscriptions()
  }
```
(however this is still broken until #1248 is fixed)

src/views/WorkflowsTable.vue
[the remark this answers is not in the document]
MetRonnie: Noted

PR #1187, opened 2023-01-10
Improve GScan filter menu

### LINKED ISSUE #1119: gscan: improve filter menu
This relates to the filter pop-up/modal thinggy which allows us to filter workflows by workflow and task state.

Current issues:
* Most information is cut off by scrolling.
* It's not immediately obvious that we can filter by both workflow and task states.
* Hard to select / unselect all.

Suggestions:
* Add select/unselect all toggle button.
* Either use a multi-column layout or make the filter sections collapseable.

Note we have an intention to add additional filters later.

**Pull requests welcome!**

MetRonnie:
These changes close #1119
- Replace single list of checkboxes with two select inputs
- Filters are "opt-in" rather than "opt-out"
- Added number badge to filter button to indicate when filters are active

Seeking feedback on usability before adding/fixing any tests

**Before:**

image

**After:**

image

**Requirements check-list**


oliver-sanders:
> Seeking feedback on usability before adding/fixing any tests

Looks like a nicer approach to me.

oliver-sanders:
One advantage the old approach has was the ability to select all. This might be useful shortcut e.g. for filtering out "stopped" workflows without having to click through each of the states individually.

oliver-sanders:
I like the number badge showing the number of filter items, slick!

Especially given that we use task state icons in the tree/table view task filters I think it would be a good idea to do that here too (it serves as functional documentation if nothing else). We can also do this with the workflow icons.

Worth considering dropping the "Error" workflow state from the filter list. I think we should remove that in due course though so no biggie.

oliver-sanders (approved):

hjoliver (approved):
I like it. @MetRonnie - do you want to enhance as suggested by @oliver-sanders before merge, or as a follow-up?

MetRonnie:
I'd rather leave any further tweaks till after the Vue 3 upgrade

PR #1216, opened 2023-01-17
deps: remove unnecessary babel-loader dev dep

oliver-sanders:
Alternative to #1192

Doesn't look like we are using this dev dependency directly.


MetRonnie (approved):
Have confirmed it still builds without direct dependency. So it must be `@vue/cli-plugin-babel` that does it for us

PR #1185, opened 2023-01-05
replace alert with snackbar

wxtim:
These changes close #1127 
This is a small change with no associated Issue.

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Appropriate tests are included (unit and/or functional).
- [x] No change log entry included - this is a like-for like functionality replacement of one vue component for another. Users do not need to made specifically aware of it.
- [x] No documentation update required.

PR #1218, opened 2023-01-19
Replace alert bar with snackbar

### LINKED ISSUE #1127: alert: error icon looks like close button
The error icon in alerts looks somewhat like a close icon:

Screenshot 2022-10-27 at 16 08 26

Suggest using something like ⚠️, ⛔️ or 😞.

**Pull requests welcome!**

MetRonnie:
These changes close #1127

Built upon #1185
- Replace error message bar with a snackbar to avoid pushing content down

Changes I've added:
- Prevent duplicate snackbars when a mutation fails
- Remove provision for icon in alert - no real point and was not being used much

**Requirements check-list**
- [x] No change log entry required (this is a like-for like functionality replacement of one vue component for another. Users do not need to made specifically aware of it.)


src/components/core/Alert.vue
```diff
@@ -43,43 +42,35 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
      */
     closeAlert () {
-      this.snackbar = false
       this.setAlert(null)
     }
+  },
+
+  icons: {
+    mdiClose
   }
```
MetRonnie, line 74: There was no need for `svgPaths` to be in `data` because it doesn't need to be reactive. I have moved it to an option and renamed, accessible via `$options.icons`

src/components/core/Alert.vue
```diff
@@ -43,43 +42,35 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import { mdiClose } from '@mdi/js'
 import { mapActions, mapState } from 'vuex'
 
+// TODO: remove later when https://github.com/vuetifyjs/vuetify/issues/11021 is fixed
+const colors = new Map([
+  ['error', 'red'],
+  ['success', 'green'],
+  ['warning', 'amber']
+])
```
MetRonnie, line 50: This didn't need to be in `data` either, nor in options because it's not accessed in the template (only a computed property), so I've moved it out of the default object altogether

src/components/core/Alert.vue
oliver-sanders: So anything that Vue doesn't recognise as a regular config section (e.g. `data`, `computed`, `methods`, etc) is automatically made available as `$options`?

src/utils/aotf.js
```diff
@@ -848,15 +856,14 @@ async function _mutateError (mutationName, message, response) {
...
-    'error')
+  await store.dispatch(
+    'setAlert',
+    new Alert(`Command failed: ${mutationName} - ${message}`, 'error')
   )
 
   // format a response
   return {
-    status: TaskState.SUBMIT_FAILED,
+    status: mutationStatus.FAILED,
```
oliver-sanders, line 866: FYI: submit-failed makes more sense for most commands (e.g. broadcast) because an error response means that the scheduler didn't receive the command at all.

It's different for the UIS mutations (e.g. play) because they actually run the command synchronously so can return the runtime error.

We could probably do with being able to differentiate these one day but that will have to wait for - https://github.com/cylc/cylc-flow/issues/3329

oliver-sanders (approved):

src/components/core/Alert.vue
```diff
@@ -16,65 +16,61 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-
   computed: {
     ...mapState(['alert'])
   },
 
   methods: {
     ...mapActions(['setAlert']),
     getColor (type) {
-      return this.colors.get(type) || ''
+      return colors.get(type) || ''
```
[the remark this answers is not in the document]
MetRonnie: I think default is dark grey

wxtim (approved):

Issue #868, opened 2021-12-14
mutations menus should change based on workflow state


hjoliver:

oliver-sanders:
Blocked by https://github.com/cylc/cylc-flow/issues/4726

MetRonnie:
Is this about doing stuff like disabling/hiding pause if the workflow is already paused?

oliver-sanders:
Closed by #1224

PR #1224, opened 2023-02-06
Disable Buttons in Menu, based on workflow state 

datamel:
Sibling PRs: https://github.com/cylc/cylc-flow/pull/5347, https://github.com/cylc/cylc-uiserver/pull/418
Closes https://github.com/cylc/cylc-flow/issues/4726, https://github.com/cylc/cylc-ui/issues/868
Mutation menu now disables mutations based on workflow status.

image


datamel:
I have disabled the buttons for now. I am happy to change this to a filter instead, although I thought a disable was more appropriate, since it may be tricky for users to workout what state their workflow needs to be in to perform a desired mutation if the mutations are hidden from them.

oliver-sanders:
Maybe we could consider removing the options from the short list but leaving them in as "disabled" in the long list (i.e. the one you get when you press "see more") to avoid confusion..

src/utils/aotf.js
```diff
@@ -432,9 +432,30 @@ export function processMutations (mutations, types) {
...
 }
+/**
+ * Get the workflow states that the mutation is valid for.
+ *
+ * @export
+ * @param {string=} text - Full mutation description.
+ * @return {Array<String>}
+ */
+export function getStates (text) {
+  const defaultStates = ['running', 'paused', 'stopping', 'stopped']
```
oliver-sanders, line 447: These states are enumerated here: https://github.com/cylc/cylc-ui/blob/master/src/model/WorkflowState.model.js

src/utils/aotf.js
```diff
@@ -432,9 +432,30 @@ export function processMutations (mutations, types) {
...
+  if (!text) {
+    return defaultStates
+  }
+  const re = /Valid\sfor:\s(.*)\sworkflows./
+  // default to all workflow states
+  const validStates = text.match(re)
+  if (validStates) {
+    return validStates[1]
```
oliver-sanders, line 455: maybe consider splitting on `,`.

src/utils/aotf.js
```diff
@@ -432,9 +432,30 @@ export function processMutations (mutations, types) {
...
   }
 }
+/**
+ * Get the workflow states that the mutation is valid for.
+ *
+ * @export
+ * @param {string=} text - Full mutation description.
+ * @return {Array<String>}
+ */
+export function getStates (text) {
```
oliver-sanders, line 446: Would be good to unit test this something like:

```js
it('gets valid states', () => {
  expect(get_states('foo\nValid states: running')).to.deep.equal(['running'])
})
```

It would also be good to add a function test for this. To do so edit [this file](https://github.com/cylc/cylc-ui/blob/master/src/services/mock/json/IntrospectionQuery.json) to add a few `Valid for: ...` lines into the descriptions. Then add a test to [this file](https://github.com/cylc/cylc-ui/blob/master/tests/e2e/specs/aotf.cy.js) to make sure that mutations which aren't valid for a running workflow are filtered out (I think the "one" workflow is running).

oliver-sanders (commented):
LGTM.

datamel:
Some tests added and the cycle/task menu bar now updates based on workflow status too, as I had broken those menus.

I had a bash at changing the list of mutations e.g. `see more` having the disabled ones in it rather than the short list, but at the moment the list is created before the disabled status (dependent on the status and authorisation) is calculated. I got myself in a bit of a mess refactoring it so have stashed that for follow-up until I think about the best approach.

oliver-sanders:
Small suggestion, filter out state-disabled mutations from the shortlisted items - https://github.com/datamel/cylc-ui/pull/114

oliver-sanders (approved):
:+1:

datamel:
Thanks for the pr, it seems obvious now I see it, I had tried moving isDisabled to displayMutations but hadn't thought of setting the authorised to true when I didn't have access to it 🤦‍♀️. I also tried setting an attribute on the mutation object to try and not have to go through isDisabled twice but got myself in a tangle doing that too.  Nice addition.

wxtim (commented):
- Should the description of play in the UI be "Start, resume or restart", given that resume is now a separate button?

Otherwise, pretty cool. 💯

wxtim (approved):

PR #1245, opened 2023-02-20
auto-publish: pin node to v16

oliver-sanders:


oliver-sanders:
Added some whitespace into the package.json file to allow release to proceed.

MetRonnie (approved):

Issue #1146, opened 2022-11-24
views: collect view options and make them configurable


oliver-sanders:
> See also: https://github.com/cylc/cylc-ui/issues/662#issuecomment-1303772166 and https://github.com/cylc/cylc-ui/pull/1108#discussion_r1031241403

Each view has some options associated with it, e.g. the graph view has a "transpose" option and the tree/table views have a "cycle point sort order" option (currently configured globally but should probably be separate).

#471 (view toolbar) will provide a standard way for these options to be presented to the user. We should also collect the defaults for these options so that they can be provided to the view in a standard way.

This issue is about:
* Collecting the default options for each component and allowing them to be provided to the view via a standard prop.
* Making these defaults configurable from the settings page, ideally using the same view toolbar component.

**Pull requests welcome!**

PR #1246, opened 2023-02-20
options mixin: add a pattern for setting options on views

oliver-sanders:
* Addresses #1146
* Collects the view options so that they are contained in a single object.
* Defines a pattern for initiating views using specified options.

This opens up future enhancements:
1. We can now record the options a view was opened with allowing us to re-open it in the future. E.G. if the user switch workspaces, then switches back again we have the required info to reconstruct the options the views were open with.
2. We now have a standard interface for initiating components with a set of options which allows us to configure the defaults in the user-profile page, see the remainder of #1146

**Requirements check-list**


tests/unit/mixins/options.spec.js
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+import { expect } from 'chai'
+import { createLocalVue, shallowMount } from '@vue/test-utils'
+
+import optionsMixin from '@/mixins/options'
+
+const localVue = createLocalVue()
+
+describe.only('options mixin', () => {
```
oliver-sanders, line 25:
```suggestion
describe('options mixin', () => {
```

datamel (approved):
Much nicer code. Thanks for adding the comments in the mixin too, makes it nice and clear how it is used.
I have checked this out and manually tested the aspects of the ui which are affected by this change and functionality has been preserved. Thanks @oliver-sanders. 

Edit: just looking into test problems.

datamel (commented):
I've had a bit more look at this.
I think this has broken the standalone graph view for me. In the lumino-tab view, it requires a click of the transpose button to have the edges draw correctly for me. Sorry I hadn't spotted that before. 
I think perhaps the initialOptions need setting.

oliver-sanders:
I've seen that too but I think it might be unrelated / random. Is it repeatable for you?

datamel:
> I've seen that too but I think it might be unrelated / random. Is it repeatable for you?

Yep, repeatable unfortunately. And the e2e graph test is also failing for me locally 😞

oliver-sanders:
Ok, I've figured out what was going on here. I had changed the default for `autoRefresh` from `true` to `false`.

When you open with graph with `autoRefresh = false` it renders immediately and never updates, hence why we don't see edges initially in `#/workspace` and don't see anything in `#/graph`.

Changing the default back to `true` will resolve this, however, it does flag a bit of an issue with `autoRefresh` which could do with fixing.

datamel:
Ah brill, I was back on this branch a few mins ago to see what was going on and on master to see if I could get the problem identified but it looks like you have got it done 👍 .

oliver-sanders:
Ok, I think I've largely resolved this by:

* Changing the default for `autoRefresh` back to `true`.
* Delaying graph render until `isLoading` is `false`.

If you change the default for `autoRefresh` back to `false` you should see:

* The graph should load correctly in standalone mode (`#/graph`).
* The graph should render without edges in workspace mode (`'#/workspace`) due to the same issue causing https://github.com/cylc/cylc-ui/issues/1248.

src/mixins/options.js
```diff
@@ -0,0 +1,57 @@
...
+
+import Vue from 'vue'
+
+/**
+ * A mixin which provides a pattern for setting and storing options.
+ *
+ * Options stored here should be suitable for use as default values and may
+ * be saved as part of a tab layout. So keep derived settings, view state or
+ * workflow specific values in the components data rather than putting them
+ * into these options.
```
oliver-sanders, line 26:
```suggestion
 Options stored here will be preserved when the tab layout is restored.
```

src/services/workflow.service.js
```diff
@@ -354,7 +356,6 @@ class WorkflowService {
         )
         this.subscriptions[subscription.query.name] = subscription
         // All done!
-        subscription.handleViewState(ViewState.COMPLETE, null)
```
datamel, line 357: Moving this has broken some unit tests.

oliver-sanders: Good!

Will fix them to the new behaviour.

src/views/Graph.vue
```diff
@@ -216,18 +218,29 @@ export default {
       title: this.getPageTitle('App.workflow', { name: this.workflowName })
     }
   },
+  props: {
+    initialOptions: {
+      type: Object,
+      required: false,
+      default: () => {}
+    }
+  },
```
MetRonnie, line 227: Provided by `optionsMixin`?
```suggestion
```

src/views/Graph.vue
```diff
@@ -216,18 +218,29 @@ export default {
...
+  defaults: {
+    // the spacing between nodes
+    spacing: 1.5,
+    // if true layout is left-right is false it is top-bottom
```
MetRonnie, line 231: This took me a minute to figure out
```suggestion
    // if true, layout is left-right; if false, it is top-bottom
```

src/views/Graph.vue
```diff
@@ -387,11 +400,11 @@ export default {
     },
     increaseSpacing () {
       // increase graph layout node spacing by 10%
-      this.spacing = this.spacing * 1.1
+      this.options.spacing = this.options.spacing * 1.1
```
MetRonnie, line 403:
```suggestion
      this.options.spacing *= 1.1
```

src/views/Graph.vue
```diff
@@ -387,11 +400,11 @@ export default {
...
     },
     decreaseSpacing () {
       // decrease graph layout node spacing by 10%
-      this.spacing = this.spacing * (10 / 11)
+      this.options.spacing = this.options.spacing * (10 / 11)
```
MetRonnie, line 407:
```suggestion
      this.options.spacing *= (10 / 11)
```

src/views/Log.vue
```diff
@@ -135,6 +139,10 @@ export default {
       default: () => {}
     }
   },
```
MetRonnie, line 141: Again, `initialOptions` provided by `optionsMixin`

oliver-sanders:
Closing this now, it will be superseded by other work

PR #1162, opened 2022-12-09
Log view mutation

oliver-sanders:
There are a couple of things the log view will need which we hadn't implemented interfaces for yet:

1. The log view will need to change its subscription on the fly.
   * E.G. if the user opens the log view on one job, then switches to viewing another, then we need the subscription to be closed and re-opened.
   * The view "query" is already a computed property, we just needed to get the WorkflowService to respect that.
2. The log view will need to be open-able from the mutation menu.
   * E.G. the user should be able to click on a task, then select "log" to open the log view for that task.
   * This means views require some for of "initialData" to carry arguments related to how they were opened.

**Requirements check-list**
- [ ] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [ ] Contains logically grouped changes (else tidy your branch by rebase).
- [ ] Does not contain off-topic changes (use other PRs for other changes).

- [ ] Appropriate tests are included (unit and/or functional).
- [ ] Already covered by existing tests.
- [ ] Does not need tests (why?).

- [ ] Appropriate change log entry included.
- [ ] No change log entry required (why? e.g. invisible to users).

- [ ] I have opened a documentation PR at cylc/cylc-doc/pull/XXXX.
- [ ] No documentation update required.

PR #1249, opened 2023-02-21
Fix bug where subscriptions would not update on query change

### LINKED ISSUE #1248: No tasks load when navigating between workflows in the standalone tree/table views
**Describe the bug / steps to reproduce**

This bug affects the standalone tree/table/graph views accessed by e.g. `#/table/myworkflow` rather than the workspace view accessed by `#/workspace/myworkflow`.

When you navigate from `#/table/woof` to `#/table/meow` by changing the URL, no tasks load for workflow `meow` and the table becomes empty. This is because the subscriptions for workflows are kept track of by the view's UID, and the view is not changing during this navigation even though the workflow changes. So the workflow service ends up thinking it is already subscribed and does not actually subscribe to `meow`.

Note this does not affect the workspace view as the table/tree/graph views live inside lumino tabs there, and the tabs are destroyed on navigation so the views are re-rendered and thus subscribe to the correct workflow.

1.5.0

**Pull requests welcome!**

MetRonnie:
These changes close #1248

Builds upon https://github.com/cylc/cylc-ui/pull/1162/commits/4841f6a47c76001c944f602067fc869d9b8256db

When the view stays the same but the query changes (e.g. you navigate from `#/table/woof` to `#/table/meow`), we need to unsubscribe and re-subscribe.

Also, for a subscription to successfully start, you need to call, *in order*:
1. `workflowService.subscribe(componentOrView)`
2. `workflowService.startSubscriptions()`

Rather than split these between a mixin for component lifecycle hooks and a mixin for vue-router navigation guards respectively, I have combined them into the mixin for component lifecycle hooks. 

**Requirements check-list**
- [x] Appropriate tests are included (unit).


src/mixins/subscriptionComponent.js
```diff
@@ -35,8 +35,17 @@ export default {
...
   beforeDestroy () {
     this.$workflowService.unsubscribe(this)
+  },
+  watch: {
+    query () {
+      // if the query changes, unsubscribe & re-subscribe
+      this.$workflowService.unsubscribe(this)
+      this.$workflowService.subscribe(this)
+      this.$workflowService.startSubscriptions()
+    }
```
oliver-sanders, line 49: Ah heck, something I hadn't thought about is what happens when there are multiple "views" in the workspace.

Currently we always initiate a workspace with a single view (because we don't remember the previous tab layout and haven't got default layouts sorted yet). But soon we will be initiating views with multiple components.

I think that calling `startSubscriptions` in either `beforeMount` or `watch.query` would result in subscribing and unsubscribing multiple times as each view registers its subscription in turn. This is probably why this call was bound to navigation events rather than component lifecycle events (as we did briefly remember tab layout by accident in an earlier version).

So we want to delay running `$workflowService.startSubscriptions()` until `$workflowService.subscribe(this)` has run for all views, not just the first. I'm not sure how to do this without introducing an arbitrary timeout.

Perhaps the parent view should be responsible for registering subscriptions before the child component is mounted?

wxtim (approved):
- [x] Read the code - It seems sane, but I've been learning JS concepts on the fly.
- [x] Checked it out and tried it locally - I was able to reproduce the bug on master and confirm that this branch fixes it. :)

src/mixins/subscriptionComponent.js
MetRonnie: Is this ok to go in, with a follow-up issue for dealing with multiple subscriptions on page load?

MetRonnie: Aye but in the Vue 3 upgrade this also fixes the worse problem of tasks not loading when navigating between different workflows in the Workspace view

hjoliver: Where did this discussion end up?

oliver-sanders: https://github.com/cylc/cylc-ui/issues/1278

PR #1269, opened 2023-04-06
Upgrade to Vue 3 ⚡ 

### LINKED ISSUE #411: Vue 3
_Updated Jan 2023_

Placeholder to keep track of information for upgrading to Vue 3. https://v3-migration.vuejs.org/breaking-changes/

#### Already done with Vue 2
- [x] Remove event bus (replaced with `mitt` #504)
- [x] Vuex v4 - https://vuex.vuejs.org/guide/migrating-to-4-0-from-3-x.html#installation-process
- [x] vue-router v4 - https://router.vuejs.org/guide/migration/index.html#removed-star-or-catch-all-routes
- [x] vue-i18n v9 - https://vue-i18n.intlify.dev/guide/migration/breaking.html
- [x] Vue Dev Utils (browser extension) - https://github.com/vuejs/vue-devtools/issues/1199 & the `next` branch

#### Vue 3
- [x] Fix Lumino component
- [x] Re-implement resizable drawer (can't really use Vuetify navigation drawer anymore as border element has gone)
- [x] Fix missing data when switching workflows (#1249)
- [x] Replace vue-meta with @unhead/vue (provision for setting page title etc.)
- [x] Fix font size selection
- [x] Update cypress e2e tests
- [x] Update cypress component tests
- [x] Update unit tests
- [x] Ensure coverage is measured (code instrumented)
- [ ] Ensure Babel transcompilation is happening (#1270)
- [x] Vue compatibility build -> normal build
- [x] #1215 

#### Vuetify 3

- [x] Vuetify v3.1 - https://next.vuetifyjs.com/en/getting-started/upgrade-guide/
- [ ] Data Table out of ["labs"](https://next.vuetifyjs.com/en/labs/introduction/) and production-ready
- [x] Fix mutations menu
- [x] Fix mutation edit form
- [x] Fix dropdown selects (e.g. in task filtering)
- [x] Remove all obsolete Vuetify utility components (`v-list-item-content` etc.)
- [x] Fix all styling
- [x] Tree-shaking of Vuetify library (Should be handled by webpack-plugin-vuetify or whatever, but not working for some reason. Update: looks like this can only be solved by moving to vite?)
- [ ] Skeleton Loader #1272

### Additional context

We can plan a bit ahead, and try to make the code as easy to be ported as possible. For Vue 4, it will probably force more users to migrate to TypeScript. **So adopting TypeScript earlier than that might be a good idea**. A project doesn't need to be written only in ES6, or in TS.

Pinia is now the preferred Vue store module (replacing Vuex). However this can be tackled later.

MetRonnie:
Closes #411 

**Check List**

- [x] Tests are included
- [x] No docs PR needed


public/js/graphvizlib.wasm
oliver-sanders, line 1: :+1:

src/components/cylc/gscan/GScan.vue
```diff
@@ -16,84 +16,76 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 -->
 <template>
   <div
-    class="c-gscan h-100"
+    class="c-gscan"
   >
-    <v-skeleton-loader
+    <!-- TODO: replace v-progress-linear with v-skeleton-loader when
+    the latter is added to Vuetify 3 -->
```
oliver-sanders, line 22: Link an issue for this.

src/components/cylc/gscan/GScan.vue
```diff
@@ -279,14 +267,17 @@ export default {
...
       immediate: false,
       handler: 'filterWorkflows'
     },
     workflows: {
+      deep: true,
       immediate: true,
       handler: 'filterWorkflows'
     },
     filteredWorkflows: {
+      deep: true,
```
oliver-sanders, line 280: Dammit, does it break without this? I think it's a nasty performance hit.

oliver-sanders (commented):
:tada: brilliant, well done, this looks like some gnarly work!

I look forward to trying it out after Easter. I think I might need to read through the Vue 3 docs too, any good migration summaries out there?

src/components/cylc/gscan/GScan.vue
MetRonnie: https://v3-migration.vuejs.org/breaking-changes/watch.html

I don't think the deep watch will have any effect for `filteredWorkflows` because it is only ever replaced, not mutated. I'll try removing `deep`

MetRonnie: Deep means 100% deep as far as I know. However, for `workflows` I think it is pretty much necessary  as part of the functionality is filtering by task status

oliver-sanders (commented):
I've built from this branch and have had a play around, works nicely :tada:!

Just three minor issues...

#### 1) Colour Theme Selection

In the settings page, there is a strange issue where selecting a colour theme changes the display of the samples:

Screenshot from 2023-04-11 16-53-15
Screenshot from 2023-04-11 16-53-29

#### 2) Font Color

Some of the icons have changed from off-black to solid black which is a bit striking.

E.G, compare the workflow icon for generic/run1 to the home icon in this screenshot:

Screenshot from 2023-04-11 16-59-28

#### 3) Menu Animation

The new menu open animation, whilst flash, is a little slow. With regular use I expect this will become a bit irritating, can we speed it up a bit?

src/components/cylc/gscan/GScan.vue
MetRonnie: I think the best option is to use intermediate computed properties that pull out the relevant fields and watch those instead. Will take some refactoring

oliver-sanders:
Cam reduce the spacing here a bit too:

Screenshot from 2023-04-11 17-19-04

src/components/cylc/gscan/GScan.vue
oliver-sanders: Hmm, on second thought the `stateTotals` thinggy includes counts:

```
        "stateTotals": {
          "waiting": 0,
          "expired": 0,
          "preparing": 0,
          "submit-failed": 0,
          "submitted": 0,
          "running": 0,
          "failed": 0,
          "succeeded": 0
        }
```

So I guess task state changes would trigger this anyway because the count would change. But there are other things besides task state changes, job messages, graph edges, etc, which could trigger re-filtering with `deep: true`.

oliver-sanders:
Tried adding/removing 200 workflows at a time which stretched the old implementation badly. I wouldn't expect a speedup from Vue 3 but running a quick test to ensure performance hasn't deteriorated. It took ~20s to add them and ~15s to remove them which is in-line with the Vue 2 performance, possibly slightly better.

oliver-sanders:
Couple of build system comments, not sure what we can do about it but worth noting:

* Incremental builds now take ~25s for me in `build:watch` mode, they used to take <5s. It looks like Vite is re-doing all the chunking which might be the culprit?
* Auto-refresh doesn't seem to be working for me in `serve` mode. Just me?

hjoliver:
Wow, good job.  I haven't tried it yet, hopefully tomorrow...

wxtim:
The add view menu looks a bit odd when collapsed - like it's being covered by the ghost of the menu?

image

kinow:
> Tried adding/removing 200 workflows at a time which stretched the old implementation badly. I wouldn't expect a speedup from Vue 3 but running a quick test to ensure performance hasn't deteriorated. It took ~20s to add them and ~15s to remove them which is in-line with the Vue 2 performance, possibly slightly better.

Vue 3 should bring some performance improvements automatically to the UI, but I suspect some of the promised improvements from Vue3 could require first updating the code (e.g. haven't looked at the code here yet, but Vue3 brought the [teleport](https://vuejs.org/guide/built-ins/teleport.html) that could be used to manage the Lumino tabs components in the DOM - it should be faster & safer than the hack I wrote for that :smile: ).

kinow (commented):
First review, just looking at the IDE analysis and playing with a workflow. Found a couple minor issues in the IDE, but nothing introduced in this PR :tada: 

Quick question about the GScan task summary. If I hold the tasks on my workflow so that it cannot proceed, I see the tasks correctly held (and works great with the graph component too! That component is working flawlessly, super responsive :clap: :clap: ).

On GScan I see the workflow status as running. Without clicking on the workflow, just looking at the task icons, it shows 0 failed, 0 submit-failed, 0 submitted, and 0 running. Does that automatically imply that the workflow tasks are all held? Or was there some indicator on the workflow that all statuses had `0` tasks because some were held?

image

kinow (approved):
Had a look at the code over dinner, using IDE and Vue 3 docs. Everything looks great! Looked at what I remembered from migrating Vue2 to Vue3, like the `index.html` changes, importing Vue components including the `.vue` extension, etc.

Managed to review 49/179 files (but most of what I didn't were test files), picking components, util, views, service, so at least one or two files from each type of JS code. Found nothing that looked immediately wrong (actually learned a lot about current status of vuetify, and some JS tricks).

There is a lot of other great improvements in this PR, especially around build/vite & tests/cypress. Unfortunately most of that won't be seen by users, I think, but IMHO that deserves as much praising as the rest (especially Cypress that can be quite moody and hard to troubleshoot sometimes) :grimacing: excellent job @MetRonnie ! :trophy:

kinow:
Oh, BTW, also had a look with Vue dev tools, and the browser performance, but using a very simple workflow (only had that `five` workflow besides this simple one I wrote :laughing: , `complex` would be a good one to test).

Found nothing wrong, no frames dropped, no slowness reported. Would be good someone compare before & after this PR using a better workflow, especially since I suspect if there's nothing wrong with this PR, the reports for Vue3 must show some performance improvement (would be interesting to know where, less JS scripting time maybe because of better reactivity in vue3? better/faster frames because rendering is faster? etc)

image

hjoliver:
Thanks for the review @kinow !

kinow:
> Thanks for the review @kinow !

Not a problem. It was fun, and I missed using Cylc & Cylc UI (that graph view is really neat :+1: )

tests/unit/components/cylc/gscan/gscan.vue.spec.js
```diff
@@ -41,77 +36,12 @@ import {
...
-      computed: {
-        isLoading () {
-          return true
-        }
-      }
-    })
-    const skeletonLoader = wrapper.find('.v-skeleton-loader')
-    const isBusy = skeletonLoader.element.getAttribute('aria-busy')
-    expect(isBusy).to.equal('true')
-  })
```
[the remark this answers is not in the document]
MetRonnie: Yes, will add tests back in #1272

src/components/graphqlFormGenerator/components/NonNull.vue
```diff
@@ -16,31 +16,30 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
     </template>
-  </component>
+  </FormInput>
 </template>
 
 <script>
 import { formElement } from '@/components/graphqlFormGenerator/mixins'
 
+/** Disallow empty array/string or nullish */
+export const nonNullRule = (x) => Boolean(x?.length ?? x != null) || 'Required'
```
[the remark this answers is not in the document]
MetRonnie: I think there was a Babel plugin that handled it before, not sure when that was added

src/store/index.js
```diff
@@ -15,16 +15,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import Vuex from 'vuex'
-import Vue from 'vue'
+import { createStore } from 'vuex'
```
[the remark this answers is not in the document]
MetRonnie: Some day I think so, I've opened an issue at #1274

.github/workflows/auto_publish_release.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         run: yarn install
 
       - name: Build
-        run: yarn build --optimize
+        run: yarn build
```
wxtim, line 38: Why have we dropped the `--optimize` flag?

.github/workflows/main.yml
wxtim, line 1: Would this file be better called `tests.yml` or `CItests.yml` - "Main" isn't very descriptive IMO.

.github/workflows/auto_publish_release.yml
MetRonnie: That was a vue-cli flag; there's no such flag for vite

.github/workflows/main.yml
MetRonnie: I agree but hesitant to change it now as we rely on the filename for the cylc-admin status page etc

README.md
```diff
@@ -53,35 +56,32 @@ There are three groups of tests:
...
   * Simple unit tests for individual functions and classes.
-  * **Framework:** Mocha
-  * **Assertions:** Chai
+  * **Framework:** [Vitest](https://vitest.dev/)
+  * **Assertions:** [Chai](https://www.chaijs.com/)/[Jest](https://jestjs.io/docs/expect)
   * **Path:** `tests/unit`
-  * **Command:** `yarn run test:unit`
-  * (To watch add `--watch`)
+  * **Command:** `yarn run test:unit` (watches by default, only re-runs changed file)
+  * (To prevent watching, use `yarn vitest run`)
```
wxtim, line 63:
```suggestion
  * **Command:** `yarn run test:unit` (watches by default, only re-runs changed file:
    To prevent watching, use `yarn vitest run`)
```

README.md
```diff
@@ -53,35 +56,32 @@ There are three groups of tests:
...
 * Component tests
   * In-browser tests which mount a single Vue component standalone.
-  * **Framework:** Cypress
+  * **Framework:** [Cypress](https://docs.cypress.io/guides/overview/why-cypress)
   * **Assertions:** Chai
   * **Path:** `cypress/component`
-  * **Command:** `yarn run test:e2e` (navigate to component page)
-  * (For "headless" mode add `-- --headless --config video=false`)
+  * **Command:** `yarn run test:component`
+  * (For "headless" mode use `yarn cypress run --component --config video=false`)
```
wxtim, line 70: I think probably these last lines need to be continuation of the `**Command**` bullet point. Not militant about this though.

.github/workflows/main.yml
```diff
@@ -23,26 +26,18 @@ jobs:
         run: yarn install
 
       - name: Lint
-        env:
-          NODE_ENV: production
         run: |
-          yarn run lint --no-fix
+          yarn run lint
```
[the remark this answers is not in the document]
MetRonnie: On the one hand it's shorter but on the other it's only 4 chars extra and avoids any possible conflict with built-in yarn commands. Don't know what we prefer as a team

cypress/component/cylc-graph-node.cy.js
```diff
@@ -153,8 +145,8 @@ describe('graph node component', () => {
         state.name,
         jobStates
       )
-      console.log(jobs)
-      cy.mount(GraphNodeSVG, { propsData: { task, jobs } })
+      // console.log(jobs)
```
wxtim, line 148: Leftover debugging?

cypress/component/cylc-icons.cy.js
```diff
@@ -99,10 +102,10 @@ describe('Task component', () => {
         .should('have.css', 'animation-duration', `${MEAN_ELAPSED_TIME}s`)
         // the offset should be set to the "percent" of the expected job duration
         .should('have.css', 'animation-delay')
-        .and('match', /([\d\.]+)s/) // NOTE the delay should be negative
+        .and('match', /([\d.]+)s/) // NOTE the delay should be negative
```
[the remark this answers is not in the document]
oliver-sanders: This line is extracting the `number` variable which is passed into the `.then`.

.github/workflows/main.yml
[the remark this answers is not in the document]
oliver-sanders: I think the `run` is safer e.g. if you installed a top-level dependency called `lint`?

cypress.config.js
```diff
@@ -13,19 +15,45 @@ module.exports = defineConfig({
...
     }
-  }
+  },
+
+  env: {
+    // eslint-disable-next-line no-unneeded-ternary
+    coverage: process.env.COVERAGE ? true : false
+  },
+
+  morgan: false, // Disable XHR logging as it's very noisy
```
oliver-sanders, line 58: :laughing:

src/components/cylc/Drawer.vue
```diff
@@ -18,164 +18,144 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 <template>
   <v-navigation-drawer
     v-model="drawer"
+    id="sidebar"
```
oliver-sanders, line 21: Use the Cylc namespace prefix "c-" for top-level component classes/ids.

```suggestion
    id="c-sidebar"
```

src/components/cylc/Drawer.vue
```diff
@@ -18,164 +18,144 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-      return this.$refs.drawerRef.$el
-    },
-    setBorderWidth () {
-      const i = this.getDrawerElement().querySelector(
-        '.v-navigation-drawer__border'
-      )
-      i.style.width = this.navigation.borderSize + 'px'
-      i.style.cursor = 'ew-resize'
+      // Cannot use $refs.drawerRef.$el due to https://github.com/vuetifyjs/vuetify/issues/16766
+      return document.getElementById('sidebar')
```
oliver-sanders, line 118:
```suggestion
      return document.getElementById('c-sidebar')
```

src/styles/cylc/_drawer.scss
```diff
@@ -15,19 +15,37 @@
...
+@use '../settings.scss';
 @import '../util.scss';
 
-/* this is not in our styles/material-dashboard, so we need to force-override */
-.v-navigation-drawer {
+body.resizing-drawer {
+  cursor: ew-resize !important;
+}
+
+#sidebar {
```
oliver-sanders, line 25:
```suggestion
#c-sidebar {
```

src/components/cylc/gscan/GScan.vue
```diff
@@ -106,98 +99,99 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+                    v-for="[state, tasks] in getLatestStateTasks(Object.entries(scope.node.node.latestStateTasks))"
+                    :key="`${scope.node.id}-summary-${state}`"
+                    :class="getTaskStateClasses(scope.node.node, state)"
                   >
-                    <!-- task summary tooltips -->
-                    <span
-                      v-for="[state, tasks] in getLatestStateTasks(Object.entries(scope.node.node.latestStateTasks))"
-                      :key="`${scope.node.id}-summary-${state}`"
-                      :class="getTaskStateClasses(scope.node.node, state)"
+                    <!-- a v-tooltip does not work directly set on Cylc job component, so we use a dummy button to wrap it -->
```
oliver-sanders, line 147: Still true? Looks like the button is gone?

index.html
```diff
@@ -31,3 +31,5 @@
     <!-- built files will be auto injected -->
   </body>
 </html>
+
+<script type="module" src="/src/main.js"></script>
```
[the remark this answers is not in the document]
MetRonnie: 🤷‍♂️ Seems to work? I think it gets replaced anyway

package.json
```diff
@@ -3,33 +3,31 @@
...
+    "serve": "node ./scripts/concurrently.js serve:jupyterhub serve:vue",
+    "serve:vue": "vite --mode offline",
     "serve:jupyterhub": "nodemon src/services/mock/json-server.js",
-    "test:e2e": "bash -c 'concurrently --names \"SERVER,TESTS\" --prefix-colors \"yellow,blue\" --success \"first\" --kill-others \"yarn run serve:jupyterhub\" \"vue-cli-service test:e2e --mode offline ${0}\"'",
-    "test:unit": "vue-cli-service test:unit",
-    "test": "yarn run test:unit || exit 1; yarn run coverage:e2e --headless"
+    "test:component": "cypress open --component",
+    "test:e2e": "yarn run serve e2e:open",
+    "test:unit": "vitest",
+    "test": "vitest run && yarn run serve cy:run"
```
wxtim, line 21: Could we make the concurrently script do this too?

(In my head this command runs the e2e headless because it's a generic check that everything passes)

oliver-sanders: This is `&&` (do this, then do that) not `&` (do this in the background and do that)

cypress/component/cylc-graph-node.cy.js
```diff
@@ -190,21 +182,21 @@ describe('graph node component', () => {
...
       )
       // check the progress animation
-      .get('.c8-task:last .status > .progress')
+      cy.get('.c8-task:last .status > .progress')
         // the animation duration should be equal to the expected job duration
         .should('have.css', 'animation-duration', `${MEAN_ELAPSED_TIME}s`)
         // the offset should be set to the "percent" of the expected job duration
         .should('have.css', 'animation-delay')
-        .and('match', /([\d\.]+)s/) // NOTE the delay should be negative
+        .and('match', /([\d.]+)s/) // NOTE the delay should be negative
```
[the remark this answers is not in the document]
MetRonnie: This is being picked up as an unnecessary escape character by eslint. (`.` inside square brackets is matched literally)

src/components/cylc/Drawer.vue
```diff
@@ -18,164 +18,144 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-        </v-list-item>
 
         <v-list-item
           to="/"
         >
-          <v-list-item-action>
-            <v-icon>{{ svgPaths.home }}</v-icon>
-          </v-list-item-action>
+          <template v-slot:prepend>
+            <v-icon style="opacity: 1;">{{ svgPaths.home }}</v-icon>
```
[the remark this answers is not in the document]
MetRonnie: Vuetify gives icons in the prepend slot a dimmed opacity for some reason

index.html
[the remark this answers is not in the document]
kinow: Good question! I was curious and went to check what I used in Jena, there the script is inside `</html>`, and outside `</body>`: https://github.com/apache/jena/blob/c8bfccef341b85e2a5e1d347c33fc528e4aa7226/jena-fuseki2/jena-fuseki-ui/index.html#L33

Then checked the examples in the vue org, there it seems it's also inside `</html>`, but inside the `</body>` tag :smile: https://github.com/vuejs/vue/blob/a9ca2d85193e435e668ba25ace481bfb176b0c6e/examples/classic/select2/index.html#L26 & https://github.com/vuejs/devtools/blob/122207de2ed79d839c572438b095ce3f5ea1fed9/packages/shell-host/public/index.html#L37 (also some weird HTML without `</html>` https://github.com/vitejs/vite/blob/e7924d22f70fdfdab32b4aab224cd66317f3c768/playground/html/index.html)

>Seems to work?

:man_shrugging:² indeed, but never paid attention where I was inserting the scripts tag, will do so next time now :laughing:

src/views/Log.vue
wxtim, line 1: I don't seem to be able to view logs on this branch.

MetRonnie: (needed to check out cylc-flow & uiserver before https://github.com/cylc/cylc-flow/pull/5453 & https://github.com/cylc/cylc-uiserver/pull/439)

wxtim (approved):
I've been through this. I have played with any feature where code has changed.

I'm tired just reviewing it - well done @MetRonnie

oliver-sanders:
Tested on Safari, works well.

I spotted one issue where the clock-face progress indicator doesn't show in the graph view:

Screenshot 2023-04-26 at 15 20 57

I think this was working before but I remember there was some difference between Safari/others regards SVG rotation-centre coordinates. Working on other browsers?

MetRonnie:
> I spotted one issue where the clock-face progress indicator doesn't show in the graph view

Working for me on Chome & Firefox

oliver-sanders:
> Working for me on Chrome & Firefox

Ok, I'll have another go / compare to Vue 2 when I've got my Mac handy.

src/components/cylc/tree/TreeItem.vue
```diff
@@ -349,16 +366,13 @@ export default {
...
-      return this.depth * NODE_DEPTH_OFFSET
-      // Note: this should actually depend on the expand/collapse icon size, but
-      // fortuitously, the vuetify default icon size is 24px which is 1.5em for
-      // a font-size of 16px
+      return this.depth * this.indent
     },
     nodeStyle () {
       return {
-        'padding-left': `${this.node.type === 'job-details' ? 0 : this.nodeIndentation}em`
+        'padding-left': `${this.node.type === 'job-details' ? 0 : this.nodeIndentation}px`
```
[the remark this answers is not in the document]
MetRonnie: These need to be non-relative otherwise they are dependent on font size which can be changed by the user, leading to incorrect spacing. In Vuetify, spacing is not relative: https://vuetifyjs.com/en/styles/spacing/#how-it-works

src/components/cylc/workflow/Toolbar.vue
```diff
@@ -38,88 +38,87 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
     </v-btn>
     <!-- title -->
     <v-toolbar-title
-      class="text-md-h6 text-subtitle-1"
+      class="c-toolbar-title text-md-h6 text-subtitle-1 font-weight-medium text-primary"
+      :class="showNavBtn ? 'ml-0' : null"
```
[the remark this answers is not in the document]
MetRonnie: Additive - see https://vuejs.org/guide/essentials/class-and-style.html#binding-html-classes

src/components/cylc/workflow/Toolbar.vue
```diff
@@ -38,88 +38,87 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-          {{ svgPaths.menu }}
-        </v-icon>
+          :icon="svgPaths.menu"
+          size="small"
+        />
 
-        <v-icon
+        <v-btn
           id="workflow-play-button"
-          color="#5E5E5E"
```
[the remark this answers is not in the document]
MetRonnie: Why? I think the black colour helps distinguish from disabled state

src/model/Subscription.model.js
```diff
@@ -66,18 +67,17 @@ class Subscription {
    * @param {*} context
    */
   handleViewState (viewState, context) {
-    const _this = this
-    if (viewState !== ViewState.ERROR) {
-      Object.values(this.subscribers).forEach(subscriber => {
+    if (toRaw(viewState) !== ViewState.ERROR) {
```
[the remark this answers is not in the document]
MetRonnie: I'm pretty sure I tried that and it doesn't work unfortunately

src/components/cylc/tree/TreeItem.vue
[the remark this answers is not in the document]
oliver-sanders: >  otherwise they are dependent on font size which can be changed by the user

But this **should** be dependent on the font size set by the user right?

src/components/cylc/workflow/Toolbar.vue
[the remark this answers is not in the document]
oliver-sanders: The solid black is too harsh which is why it was toned down.

src/components/cylc/tree/TreeItem.vue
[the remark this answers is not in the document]
oliver-sanders: Vuetify uses a mixture of absolute and relative units pending this slow-moving issue - https://github.com/vuetifyjs/vuetify/issues/3401

tldr; absolute units are fragile and considered bad for reasons including responsive design and accessibility so we've tried to keep them out of our code. Either `em` or `rem` should suffice for most purposes.

src/model/Subscription.model.js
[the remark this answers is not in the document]
oliver-sanders: Cheers, just trying to understand what this equality does.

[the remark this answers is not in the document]
MetRonnie: A Proxy does not equal the object it is proxying, this is a bit of a gotcha and certainly worth being aware of

src/components/cylc/tree/TreeItem.vue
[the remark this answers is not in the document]
MetRonnie: Ah ok, the only thing that changes with font-size is the task/job icon size, so that means the arrow/triangle thingy should scale with font-size too. But the indent doesn't depend on font-size.

> absolute units are fragile and considered bad for reasons including responsive design and accessibility

Sorry but I do not think there is a clear consensus on this. See https://stackoverflow.com/a/609561/3217306. While it makes sense to use `em`/`rem` for font sizes, or other relative units such as `vw`/`vh` for certain things that depend on responsive design, for the indentation of the tree view I see no particular reason this should scale with font size.

MetRonnie:
I've addressed 2/3 points.

> The menu pop-up animation is better (no more flickering) but possibly too much on the slow side?

I've got a branch ready to go as a follow-up PR that adds a setting in the user profile for turning off menu animations

hjoliver (approved):
Great job. (Mainly functional review as I'm still no JS/Vue expert, but the code looks fine to me.)

hjoliver:
@oliver-sanders - do you want to do a final pre-merge check and approval?

src/views/Graph.vue
```diff
@@ -622,14 +623,15 @@ export default {
         await this.$nextTick()
       }
     },
+    /** re-layout the graph after any new nodes have been rendered */
     async layout (nodes, edges, nodeDimensions) {
-      // re-layout the graph after any new nodes have been rendered
+      const graphviz = await Graphviz.load()
```
oliver-sanders, line 628: What are the implications of calling `Graphviz.load` for every layout?

Could we do this on mount?

src/views/Graph.vue
```diff
@@ -622,14 +623,15 @@ export default {
...
 
       // generate the GraphViz dot code
       const dotCode = this.getDotCode(nodeDimensions, nodes, edges)
 
       // run the layout algorithm
-      const jsonString = await graphviz.layout(dotCode, 'json')
+      const jsonString = graphviz.layout(dotCode, 'json')
```
oliver-sanders, line 634: Does this make `layout` blocking?

oliver-sanders (commented):
Made it through the source code :sweat:

src/mixins/toolbar.js
```diff
@@ -24,38 +24,20 @@ import { mapMutations } from 'vuex'
...
+    this.setDrawer(!this.showNavBtn)
   },
 
-  beforeDestroy () {
-    window.removeEventListener('resize', this.onResponsiveInverted)
-    document.querySelector('.v-navigation-drawer')?.removeEventListener('resize', this.onResponsiveInverted)
+  computed: {
+    showNavBtn () {
+      return this.$vuetify.display.mobile || !this.$store.state.app.drawer
+    }
```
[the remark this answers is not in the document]
oliver-sanders: Tested, good.

src/components/cylc/workflow/Toolbar.vue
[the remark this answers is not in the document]
oliver-sanders: A bunch of icons are now defaulting to solid black which is jarring with the overall design.

*Before:*

Screenshot from 2023-05-11 15-19-15

*After:*

Screenshot from 2023-05-11 15-19-04

src/components/cylc/Drawer.vue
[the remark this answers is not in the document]
oliver-sanders: I'm not convinced that's wrong?

src/components/cylc/gscan/GScan.vue
oliver-sanders: This needs investigation/solving. Please put up an issue if not doing it here.

src/components/cylc/tree/TreeItem.vue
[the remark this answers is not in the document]
oliver-sanders: * The indentation is based on the expand/collapse icon size.
* The v-icon default size is `1em` [apparently](https://vuetifyjs.com/en/api/v-icon/#sass)
* Therefore the indent size should be in em!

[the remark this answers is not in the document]
MetRonnie: `v-btn` doesn't scale with root font size

src/components/cylc/Drawer.vue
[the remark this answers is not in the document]
MetRonnie: Seems very arbitrary though? I would expect a slot called prepend to only deal with positioning

src/components/cylc/workflow/Toolbar.vue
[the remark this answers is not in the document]
MetRonnie: I've looked into it and Vuetify 2 had a default color for all `v-icon`s whereas Vuetify 3 leaves it up to DOM inheritance (i.e. font color). Ironically, the added `#5E5E5E` made it _darker_ than the default color! Now that I'm used to the dark icons it doesn't seem jarring at all, if we want to have a comb through I think it would best to in a follow-up PR?

src/views/Graph.vue
oliver-sanders: > The await in 1.x meant this was blocking anyway

???

`await` waits for a promise to be completed which should be async i.e. non-blocking right?

I think because the layout is happening in WASM the execution happens outside of the JS thread, but I've no idea how that works with blocking calls.

oliver-sanders: It looks like HPCC is using the "web worker" pattern which should allow it to run in another thread:

https://github.com/hpcc-systems/hpcc-js-wasm/blob/d757dba9f6511bc81ad264b573de8dc6400492ea/index-worker.js

MetRonnie: > `await` waits for a promise to be completed which should be async i.e. non-blocking right?

Oh you mean blocking in the overall stack as opposed to this context

src/styles/cylc/_toolbar.scss
```diff
@@ -15,29 +15,19 @@
...
-  min-height: inherit !important;
-
-  .v-toolbar__content {
-    min-height: inherit !important;
-    column-gap: 0.5rem;
-  }
-}
 
 .c-toolbar {
+  $icon-size: 24px;
```
oliver-sanders, line 23:
```suggestion
  $icon-size: 1rem;
```

src/components/cylc/workflow/Toolbar.vue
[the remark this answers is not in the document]
oliver-sanders: For the default, it looks like multiple Vuetify components override the `body { color }` e.g. the draw and toolbars so this is just the pain of trying to bend Vuetify to match our design sketches, fair enough.

src/components/cylc/gscan/GScan.vue
MetRonnie: #1296

src/components/cylc/workflow/Toolbar.vue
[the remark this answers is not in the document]
MetRonnie: The Vuetify update is a style change anyway (all the icons have changed colour), so I don't see this as sneaking something through. Rather than leaving behind an inconsistent icon colour, I think the better option is to open an issue for deciding what colour the icons should be?

src/components/cylc/tree/TreeItem.vue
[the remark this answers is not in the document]
oliver-sanders: Ok, will address later.

src/views/Graph.vue
oliver-sanders: Blocking means prevents the program from proceeding e.g in Python land:

```python
async def foo():
    blocking()
    await non_blocking()
```

I don't understand the JS implementation details but I think it's vaguely equivalent to Python's cooperative coroutine  approach:

* `await` suspends the coroutine and yields control back to the event-loop to pick another coroutine to run/resume.
* Returning from an `async` function lists the calling coroutine as ready to resume when the event-loop gets around to it.
* In JS `Promise` objects are used under the hood as proxies for the async execution.

In this code my worry is that there's no `await` on the `layout` which suggests that this operation is blocking. If so that would block the entire application preventing any UI updates whilst the layout is running. If that's true it's really bad. But I don't understand how webworkers interface with JS or the details of the JS implementation at all so I could be totally wrong.

oliver-sanders: Bump to https://github.com/cylc/cylc-ui/issues/1298

src/components/cylc/workflow/Toolbar.vue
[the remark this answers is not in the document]
oliver-sanders: I disagree, but I'll cave in.

Issue #1279, opened 2023-04-21
1.6.x sync


oliver-sanders:
Merge/rebase changes on the 1.6.x branch into master post the merge of the Vue 3 upgrade PR.

PR #1303, opened 2023-05-22
1.6.x -> master (log view improvements)

MetRonnie:
Partially addresses #1279

Includes:
- #1249
- #1275

Plus:
- Fix uncaught TypeError in log view
- Minor style & test improvements for log view

**Check List**

- [x] Tests are updated
- [x] Changelog entry brought through
- [x] No docs needed


src/mixins/subscriptionComponent.js
```diff
@@ -49,11 +49,11 @@ export default {
     this._updateQuery(null, this.query)
   },
   methods: {
-    _updateQuery (newVal, oldVal) {
-      if (oldVal) {
-        this.$workflowService.unsubscribe(this)
+    _updateQuery (newQuery, oldQuery) {
+      if (oldQuery) {
+        this.$workflowService.unsubscribe(oldQuery, this._uid)
```
MetRonnie: TypeError was caused by trying to unsubscribe the new query instead of the old

tests/e2e/specs/log.cy.js
```diff
@@ -16,92 +16,70 @@
...
       .should('be.visible')
-      .should('have.attr', 'disabled', 'disabled')
-      .should(($input) => {
-        expect($input.val()).to.equal('~user/one')
-      })
-
-      // the job log files list should have been populated by the query
-      .get('.c-log')
-      .should(($el) => {
-        expect($el[0].__vue__.logFiles).to.have.length(3)
```
MetRonnie: This doesn't work in Vue 3, and is better suited to unit testing that e2e Cypress testing

tests/unit/views/log.vue.spec.js
```diff
@@ -0,0 +1,101 @@
...
+    $workflowService = sinon.createStubInstance(WorkflowService)
+  })
+
+  it('issues the subscription', async () => {
+    const wrapper = mountFunction()
+    expect(wrapper.vm.jobLog).toEqual(0)
+    expect(wrapper.vm.query.variables).toMatchObject({
+      id: workflowID,
+      file: initialFile,
+    })
```
MetRonnie, line 68: Vitest supports both Chai assertions (e.g. `expect(...).to.deep.equal`) and these which are "Jest compatible" assertions: https://vitest.dev/api/expect.html

The [`toMatchObject()`](https://vitest.dev/api/expect.html#tomatchobject) assertion is very nice if you don't care about certain fields of an object

src/components/cylc/log/Log.vue
```diff
@@ -28,21 +28,49 @@ export default {
...
+      }
+    }
+  },
+  methods: {
+    updateLogs () {
+      return this.logs.map((logLine) => {
+        return this.stripTimestamp(logLine)
+      })
+    },
+    stripTimestamp (logLine) {
```
wxtim, line 67: I feel that this ought to be unit-tested - have I missed it being tested somewhere?

src/views/Log.vue
```diff
@@ -14,96 +14,214 @@ GNU General Public License for more details.
...
+    files
   }
+}
+`
 
-  onAdded (added, store, errors) {
-    this.lines.push(...added.lines)
+// The preferred file to start with as a list of patterns
+// The first pattern with a matching file name will be choosen
+const LOG_FILE_DEFAULTS = [
```
wxtim, line 193: Putting aside the issue of users feeding more or less anything to stderr and stdout - I would have thought that if a `job.err` has been created this would be the user's priority?

MetRonnie:
> Though I had to install node 16 - it wasn't happy at node 14

Better than the other way round!

src/views/Log.vue
wxtim: Totally.

PR #1254, opened 2023-02-28
Add analysis view table containing workflow timing statistics

JAllen42:
Partially addresses #1170
Linked with cylc/cylc-uiserver#434
Based on the proof of concept #1171

This adds a table view containing statistics of the timings of tasks in a workflow. Currently this includes mean, standard deviation, minimum, maximum, median and quartiles, which can be calculated using the queue times for tasks, run times or the total time. These can be filtered based on the name of the task and which platform the tasks was run on.

image

The proof of concept query which loads job information from the workflow database has been changed to load task information instead, and the subscription has been deleted and replaced with a button to rerun the query.

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [x] Tests are included (or explain why tests are not needed).
- [x] `CHANGES.md` entry included if this is a change that can affect users
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

PR #1308, opened 2023-05-31
1.6.x -> master (analysis view)

MetRonnie:
Partially addresses https://github.com/cylc/cylc-ui/issues/1279

Includes:
- #1254

**Check List**

- [x] Tests are updated
- [x] Changelog entry brought through
- [x] No docs needed


MetRonnie:
The real changes here are in https://github.com/cylc/cylc-ui/pull/1308/commits/7cb7ab4582909a7befd6e9fa0f56155e082a11aa

tests/e2e/specs/analysis.cy.js
```diff
@@ -16,16 +16,19 @@
  */
 
 describe('Analysis view', () => {
-  it('Should display the mocked workflow', () => {
+  beforeEach(() => {
     cy.visit('/#/analysis/one')
+  })
```
oliver-sanders: FYI, the intention was to support other workflows in the offline data, which we may still do to help test views for which the simple example isn't enough. This view is a good example of one which could benefit from a more advanced example.

It's also not a bad idea to test a views interaction with the workspace page as we will add more interactions between views in due course. E.G. we already have the ability to open log files in a new tab, we should also acquire the ability to select a task in one view and use that to find the same task in another view (e.g. click+drag type functionality), keyboard controls, that sort of stuff.

No need to go collecting `cy.visit`s in all tests is the point, we may end up reversing these diffs.

oliver-sanders (approved):
LGTM, tested.

wxtim (approved):
I've thoroughly tested it manually and had a quick flick through the code.

PR #1311, opened 2023-06-02
Update renovate config

MetRonnie:
We are not currently getting updates for dependencies when the new version falls within the semantic specifier. E.g. for `^1.2.3` we would only get a dependency PR if they released `2.0.0`, but not `1.3.0` or `1.2.4`.
- https://docs.renovatebot.com/configuration-options/#rangestrategy
- https://docs.renovatebot.com/presets-config/#configjs-app

If the dependency PRs get too noisy we can use a custom schedule to manage them.

**Check List**

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,14 +1,16 @@
 {
   "extends": [
-    "config:base"
+    "config:js-app"
   ],
   "labels": ["dependencies", "javascript"],
   "packageRules": [
     {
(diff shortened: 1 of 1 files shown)


oliver-sanders (approved):
> If the dependency PRs get too noisy we can use a custom schedule to manage them.

I suspect it probably will for some of the busier projects, but lets give it a go and adapt on the fly.

PR #1139, opened 2022-11-17
simple tree: bare-bones tree view for documentation purposes

oliver-sanders:
> ~Built on #1108~

SimpleTree: a simple view a functional documentation
    
* A new bare-bones live tree view not available in production mode.
* Indended to serve as documentation and a guide for future view
  development.

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [ ] Appropriate tests are included (ok for laters, this is docs?)
- [x] No change log entry required (why? e.g. invisible to users).
- [x] No documentation update required.

PR #1318, opened 2023-06-08
Use dynamic async components for workspace view tabs + 1.6.x -> master (SimpleTree view)

### LINKED ISSUE #1221: workspace: avoid importing views until needed
We currently import all of the views into the workspace component.

I think this means the resources required for *all* of the views are loaded when you open the workspace component irrespective of whether you need them or not. **Needs confirmation**.

It was an ok approach to start with, but as the list of views grows this will cause loading pain. It is possible to dynamically import components in Vue something like:

```js
components: {
  TreeView: () => import('...'),
  ...
}
```

This will, however, require moving the view metadata (title and icon) outside of the view component so that it can be imported separately e.g something like:

```js
export treeViewMetadata = {
}

export default {
  ...
}
```

**Pull requests welcome!**

### LINKED ISSUE #1273: Investigate using `Teleport` for Lumino widgets
From https://github.com/cylc/cylc-ui/pull/1269#issuecomment-1511691902

### Problem

The current implementation of the views inside Lumino tabs involves mounting the component in a hidden div then moving it to the corresponding Lumino tab's div. Which is a bit hacky.

### Proposed Solution

See if https://vuejs.org/guide/built-ins/teleport.html can make it easier and perhaps more efficient

MetRonnie:
In order to incorporate the dynamically imported SimpleTree view it seemed necessary to make all the views dynamically imported in Vue 3. Closes #1221.

I also took this opportunity to use [`<Teleport>`](https://vuejs.org/api/built-in-components.html#teleport) instead of the hacky solution for displaying these views in the Lumino tabs. (Hat tip to @kinow for the suggestion, it is rather nice!). Closes #1273.

Sync:
- #1139

**Check List**

- [x] Tests are not needed.
- [x] No changelog entry needed.

### DIFF
--- a/src/components/cylc/workflow/Lumino.vue
+++ b/src/components/cylc/workflow/Lumino.vue
@@ -18,23 +18,23 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
   <div ref="main" class="main pa-2 fill-height">
     <!-- Lumino box panel gets inserted here -->
   </div>
-  <div v-show="false">
-    <!-- Hidden div acts as staging area for views before they are
-    moved into a lumino widget -->
-    <component
-      v-for="(item, id) of views"
-      :key="id"
-      :ref="(ref) => setViewRef(id, ref)"
-      :is="item.view"
-      :tab-title="getTabTitle(item.view)"
-      :workflow-name="workflowName"
-      :initialOptions="item.initialOptions"
-      class="h-100"
-    />
-  </div>
+  <template
+    v-for="(item, id) in views"
+    :key="id"
+  >
+    <Teleport :to="`#${id}`">
+      <component
+        :is="item.view"
+        :workflow-name="workflowName"
+        :initialOptions="item.initialOptions"
+        class="h-100"
+      />
+    </Teleport>
+  </template>
 </template>
 
 <script>
+import { startCase } from 'lodash'
 import LuminoWidget from '@/components/cylc/workflow/lumino-widget'
 import { BoxPanel, DockPanel, Widget } from '@lumino/widgets'
 
@@ -74,15 +74,6 @@ export default {
       type: Array,
       required: true
     },
-    /**
-     * Prop to customize the tab title. Defaults to name.
-     * If a component does not have the $component.$tabTitleProp
-     * set, then we still revert to the old default $component.name.
-     */
-    tabTitleProp: {
-      type: String,
-      default: 'name'
-    }
   },
 
   emits: [
@@ -90,11 +81,10 @@ export default {
     'lumino:deleted'
   ],
 
-  data () {
-    return {
-      /** Keep track of views' refs separate from $refs in order to access
-       * by ID */
-      viewRefs: {}
+  beforeCreate () {
+    // Populate components
+    for (const { name, component } of this.allViews) {
+      this.$options.components[name] = component
     }
   },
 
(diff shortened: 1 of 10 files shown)


src/views/Workspace.vue
```diff
@@ -37,19 +36,45 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
 import ViewState from '@/model/ViewState.model'
 import Lumino from '@/components/cylc/workflow/Lumino.vue'
 import Toolbar from '@/components/cylc/workflow/Toolbar.vue'
-import TableView from '@/views/Table.vue'
-import TreeView from '@/views/Tree.vue'
-import GraphView from '@/views/Graph.vue'
-import LogView from '@/views/Log.vue'
-import AnalysisView from '@/views/Analysis.vue'
+
+// Use dynamic async components for lazy loading:
```
kinow, line 57: :ok_hand:

src/views/SimpleTree.vue
```diff
@@ -0,0 +1,238 @@
...
+  taskProxies {
+    ...TaskProxyData
+  }
+  jobs {
+    ...JobData
+  }
+}
+
+# We must list all of the types we request data for here to enable automatic
+# hosekeeping.
```
kinow, line 156: housekeeping? :fire_engine:

src/components/cylc/workflow/Lumino.vue
```diff
@@ -18,23 +18,23 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+  >
+    <Teleport :to="`#${id}`">
+      <component
+        :is="item.view"
+        :workflow-name="workflowName"
+        :initialOptions="item.initialOptions"
+        class="h-100"
+      />
+    </Teleport>
+  </template>
```
kinow, line 33: :clap: :clap: :clap:

kinow (approved):
Looks grand to me! :bow: :clap:

src/views/Workspace.vue
```diff
@@ -37,19 +36,45 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+const AnalysisView = defineAsyncComponent(() => import('@/views/Analysis.vue'))
+const SimpleTreeView = defineAsyncComponent(() => import('@/views/SimpleTree.vue'))
+
+const allViews = [
+  { name: 'Tree', component: TreeView, icon: mdiFileTree },
+  { name: 'Table', component: TableView, icon: mdiTable },
+  { name: 'Graph', component: GraphView, icon: mdiGraph },
+  { name: 'Log', component: LogView, icon: mdiFileDocumentMultipleOutline },
+  { name: 'Analysis', component: AnalysisView, icon: mdiChartLine },
+]
```
oliver-sanders, line 71: For the long term we should think about how to make views more "plugable". It would be helpful for this information to remain tethered with the view, perhaps in some kind of metadata / sidecar file that could be imported separately. Would be worth taking a look at how Jupyter Lab make their "views" plugable as they face a similar (but more advanced) problem here.

No need for changes, this is great for now, but something to store in the back of the mind for later work...

oliver-sanders (approved):
:+1:

src/views/Workspace.vue
oliver-sanders: Added a "pending" issue to preserve this, https://github.com/cylc/cylc-ui/issues/1319

PR #1336, opened 2023-06-15
Fix tree view task state filtering bug

### LINKED ISSUE #1335: Tree view task state filtering bug
### Description / Reproducible Example

Filtering by task state shows tasks of non-matching state under matching families

image

### Expected Behaviour

Only tasks with matching state should show. Tasks with non-matching state should not show under families of matching state.

(But _do_ show tasks of non-matching ID under matching family ID, e.g. filtering by "WVGM" should show all tasks under the WVGM family even if they do not have "WVGM" in their ID.)

MetRonnie:
Closes #1335

Fix bug where choosing a state filter in the tree view would wrongfully include tasks that don't match the state if they were members of a family matching the state.

**Check List**

- [x] Tests are included
- [x] No docs


oliver-sanders (approved):
Tested, does the trick.

src/store/workflows.module.js
```diff
@@ -35,7 +35,7 @@ export function merge (data, update) {
   }
 }
 
-const state = {
+const state = () => ({
```
wxtim, line 38: Excuse my JS ignorance, but why this change?

tests/unit/components/cylc/common/filter.spec.js
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
```
wxtim, line 17: If module level docs are a js think then I think a quick plain-English run down at the start would be nice.

tests/unit/components/cylc/common/filter.spec.js
```diff
@@ -0,0 +1,94 @@
...
+  { node: taskNode, states: ['succeeded', 'failed'], expected: true },
+  { node: taskNode, states: ['failed'], expected: false },
+])('matchState(<$node.node.state>, $states)', ({ node, states, expected }) => {
+  it(`returns ${expected}`, () => {
+    expect(matchState(node, states)).toBe(expected)
+  })
+})
+
+describe.each([
+  { node: taskNode, id: '', states: [], expected: true },
```
wxtim, line 84: Do you need this many tests for this function? I'd have thought that since it just chains `matchState` and `matchID` that you only need to test for 0, 1 & 2 `true` responses for components. Even then it's somewhat close to testing the language's `&&` operator.

tests/unit/components/cylc/common/filter.spec.js
oliver-sanders: Test descriptions should be included in the `describe` and `it` blocks.

wxtim (approved):
- [x] Read code stylistic comments and suggestions only.
- [x] Tried it out functionally.

src/store/workflows.module.js
oliver-sanders: I assume this is similar to the way component `data` is provided as a function.

This is providing the initial data for the store. The way it was done before is perfectly legit due to the way we are using it, however, if we tried to use multiple workflow stores on the app it would break. Using a function avoids this problem and is probably documented as best practice.

tests/unit/components/cylc/common/filter.spec.js
MetRonnie: > If module level docs are a JS thing then I think a quick plain-English run down at the start would be nice.

The run-down of this file is simply that it contains the tests for `src/components/cylc/common/filter.js`

tests/unit/components/cylc/common/filter.spec.js
MetRonnie: Just in case it gets changed in future

wxtim (approved):

PR #1340, opened 2023-06-20
Fix family tasks discrepancy in tree view

### LINKED ISSUE #1322: tree: tree view differs from table view
We have received multiple reports of discrepancies between the tree and table views.

In the most recent report a task was showing in tree view as submitted, but in the table view as succeeded (correct).

In all cases the table view was showing the correct data so this would appear to be a tree-view bug. It has not yet been confirmed whether the graph view matches the tree or table but I suspect it will match the table.

The tree view follows the "node.familyTree" down to the task whereas table/graph views are not interested in this information so use "node.children". So this is likely related to inheritance. We had a previous bug where task ancestors were coming out in an unstable order, this may be similar.

To debug suggest:
* Take a reported example and attempting to reproduce on the reported task (this might not be reproducible in any old workflow with any old task).
* Try restarting the workflow if struggling to reproduce in case this is a random order / race condition bug in cylc-flow.
* Once reproduced, inspect the vuex store drilling down to the familyTree of the cycle point the task is in.
* If the tree view matches the familyTree, then the tree view has done what it was told to.
  * The bug is either in the ui-data-store(`src/store/workflows.module.js`), or the cylc-flow data store.
  * Use GraphiQL to inspect `ancestors { name }` and `child tasks { id }` to determine which.
* If the tree view differs from the familyTree then it is a reactivity problem in `src/components/cylc/tree/Tree[Item].vue`.

MetRonnie:
Closes #1322.

Fixes a bug where tasks belonging to a family would sometimes disappear for a second only to reappear (typically when changing state), and would sometimes reappear permanently after having finished and moved out of the n=1 window.

There was an assumption the `childTasks` field of each updated-family delta always contained all current child tasks, but it only includes updated tasks (or maybe even arbitrary tasks?). Instead, use the pruned-task delta to remove the task from the family.

**Check List**

- [x] Tests are included

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,6 +29,10 @@ Tree view: show mean run time in job details.
 
 ### Fixes
 
+[#1340](https://github.com/cylc/cylc-ui/pull/1340) -
+Fixed bug in tree view where tasks belonging to families would disappear
+and reappear eroneously.
+
 [#1312](https://github.com/cylc/cylc-ui/pull/1312) -
 Fixed incorrect latest job info in table view.
 
--- a/src/components/cylc/table/Table.vue
+++ b/src/components/cylc/table/Table.vue
@@ -66,7 +66,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                 <div class="mr-1">
                   <Job
                     v-cylc-object="item.value.task"
-                    :status="item.value.task.node.state"
+                    :status="item.value.latestJob?.node?.state"
                     :previous-state="item.value.previousJob?.node?.state"
                   />
                 </div>
--- a/src/graphql/queries.js
+++ b/src/graphql/queries.js
@@ -81,6 +81,9 @@ fragment TaskProxyData on TaskProxy {
   task {
     meanElapsedTime
   }
+  firstParent {
+    id
+  }
 }
 `
 
--- a/src/store/workflows.module.js
+++ b/src/store/workflows.module.js
@@ -270,28 +270,20 @@ function cleanParents (state, node) {
   }
 }
 
-/* Build / housekeep the familyTree based on the family information containined
+/**
+ * Build the familyTree based on the family information containined
  * in this node.
  */
 function applyInheritance (state, node) {
-  if (node.type === 'family') {
-    const childTasks = node.node.childTasks || []
-
+  if (node.type === 'family' && node.node.childTasks) {
     // add new tasks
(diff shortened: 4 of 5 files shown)


oliver-sanders:
This approach looks good to me.

MetRonnie:
I've got a little trouble with the unit tests. At this line:

https://github.com/cylc/cylc-ui/blob/5026d751d742c9edcfc60cbf9399381e4c799195/src/store/workflows.module.js#L484-L486

```
TypeError: Cannot read properties of undefined (reading 'id')
```

which seems to be due to this line when creating the node, there is no `firstParent` field:

https://github.com/cylc/cylc-ui/blob/5026d751d742c9edcfc60cbf9399381e4c799195/src/store/workflows.module.js#L410-L415

It looks like I need to make the creation of the tree in the test more explicit?

```diff
diff --git a/tests/unit/store/workflows.spec.js b/tests/unit/store/workflows.spec.js
index f39f9b4e..e696775f 100644
--- a/tests/unit/store/workflows.spec.js
+++ b/tests/unit/store/workflows.spec.js
@@ -170,9 +167,21 @@ describe('cylc tree', () => {
 
     // add some nodes to the tree
     function addNodes () {
-      store.commit('workflows/UPDATE', { id: '~a/b//c/d/e' })
-      store.commit('workflows/UPDATE', { id: '~a/b//c/d/f' })
-      store.commit('workflows/UPDATE', { id: '~a/b//g' })
+      store.commit('workflows/UPDATE', {
+        id: '~a/b//c/d',
+        firstParent: { id: '~a/b//c' },
+      })
+      store.commit('workflows/UPDATE', {
+        id: '~a/b//c/d/e',
+        firstParent: { id: '~a/b//c/d' },
+      })
+      store.commit('workflows/UPDATE', {
+        id: '~a/b//c/d/f',
+        firstParent: { id: '~a/b//c/d' },
+      })
+      store.commit('workflows/UPDATE', {
+        id: '~a/b//g',
+      })
       expect(getTree(store)).to.deep.equal({
         a: {
           b: {
```

oliver-sanders:
I think that's fine

With these changes the `firstParent { id }` becomes a mandatory field which must be requested by any view which wishes to construct the family-tree.

So long as the store is still able to function for subscriptions which do not bring in family data (e.g. the table view) then it's good.

wxtim (approved):

oliver-sanders (approved):
:1+:

PR #1347, opened 2023-06-22
Remove unused dependency on bufferutil

MetRonnie:


hjoliver (approved):
(I'll save Oliver the bother).

PR #1355, opened 2023-06-27
Analysis view: box & whiskers plot

MetRonnie:
Supersedes #1290

I've extracted out the relevant diff from #1290 and tidied it up / built upon it

image

**Check List**

- [x] No docs needed


MetRonnie:
@ChrisPaulBennett Back to you - anything else to add? I think a test for the box plot sorting in `tests/e2e/specs/analysis.cy.js` would be good; happy to give some pointers if needed.

oliver-sanders (commented):
Looks good, works well, seems robust.

A few small points:

It would be good to give units with these times (i.e. seconds), worth bearing in mind that we might want to customise the display of these times in the future (e.g. `45s`, `10m`, `1:20`, `PT4H31M`, etc):

Screenshot from 2023-08-09 17-19-39

I managed to get a `null` Q1 time where it didn't make sense:

Screenshot from 2023-08-09 17-20-44

It's not completely clear what the chart is displaying, e.g. what do the blue/green numbers mean, especially when there is little spread in the times so far:

Screenshot from 2023-08-09 17-21-57

ChrisPaulBennett:
The null entries are caused by a bug in the SQL in the UI Server.
A pull request has been created for this (https://github.com/cylc/cylc-uiserver/pull/483).
Currently investigating the other issues

oliver-sanders (approved):
:+1:

Issue #1364, opened 2023-06-29
Faster task menu navigation (set outputs)


SGallagherMet:
### Problem

When I'm working on a new branching workflow I like to test it by running the suite in simulation or dummy mode and manually setting outputs. Currently this is cumbersome as there are a large number of steps required.

To set a task to succeed.

click task -> click see more -> scroll down -> click set outputs

To set a task to failed
click task -> click see more -> scroll down -> click the edit button on set outputs -> type failed  -> scroll down -> click ok

### Proposed Solution

1. Automatically expand the menu when the mouse wheel is scrolled (instead of clicking on "see more")
2. Add buttons for frequently used options to the menu items. In the example above this would mean an option to set the output to succeeded or failed directly in the menu, without having to open the edit box.

This would mean the sequence of events to set the output would be:
click task -> scroll down -> click to set output to failed or succeeded

oliver-sanders:
> 1. Automatically expand the menu when the mouse wheel is scrolled (instead of clicking on "see more")

Interesting idea, I'm not sure we can capture a scroll event unless the content is scrollable, but there might be something we can do there.

> 2. Add buttons for frequently used options to the menu items

The good news is that this is already planned. Here's a screenshot from an early design sketch that shows what we are envisaging:

![Screenshot from 2023-07-05 16-48-47](https://github.com/cylc/cylc-ui/assets/16705946/6b2a0ca5-1e26-4a15-a760-912e5084283d)

(although the reset button only had a "succeeded" value in that sketch)

The bad news is that it may take us a while to implement this.

oliver-sanders:
In the short term we could consider adding `set-outputs` to the default list of commands.

Note: The `set-outputs` command is currently subject to some discussions which could see it evolve into something a little more powerful (and hopefully intuitive) on the 8.3.0 timescale.

MetRonnie:
I think we could also make the menu items take up less height

oliver-sanders:
> I think we could also make the menu items take up less height so there's less scrolling and mouse movement involved

Done in #1392

oliver-sanders:
Opened a separate issue for hardcoded shortcuts (e.g. set-outputs "failed") #1404.

oliver-sanders:
The remaining suggestion from this issue was:

> Automatically expand the menu when the mouse wheel is scrolled (instead of clicking on "see more")

[...]

PR #1380, opened 2023-07-05
aotf: Make set-outputs a primary mutation for namespaces.

oliver-sanders:
* Partially addresses #1364
* The `set-outputs` command is used reasonably often and current discussions could see its importance grow in the near future.

**Check List**

### DIFF
--- a/src/utils/aotf.js
+++ b/src/utils/aotf.js
@@ -183,7 +183,8 @@ export const primaryMutations = {
     'release',
     'trigger',
     'kill',
-    'log'
+    'log',
+    'set-outputs'
   ]
 }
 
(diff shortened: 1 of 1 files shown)


MetRonnie (commented):
Do we want to consider adding trigger and edit runtime to the primary list?

src/utils/aotf.js
```diff
@@ -183,7 +183,8 @@ export const primaryMutations = {
     'release',
     'trigger',
```
oliver-sanders, line 184: Trigger is already on the list.

Edit runtime is a bit more specialist so probably reasonable to put behind a button. The idea behind the button is to avoid clutter and keep the interface simple for less advanced users.

MetRonnie: > Trigger is already on the list.

Duh, I'm blind

MetRonnie (approved):

PR #1389, opened 2023-07-11
Fix unnecessary reactive data

### LINKED ISSUE #1247: Avoid creating unnecessary reactive variables

Every element in `data` is made reactive. That creates unnecessary [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) objects in Vue. While the performance improvement won't be much, I guess if you have a large workflow every nano/micro second of processing, and every scrap of memory matters, and since it's a cheap fix…

Here's an example: https://github.com/cylc/cylc-ui/blob/2907d64656181ba95e2693c98a7b4de5b8ebcbc3/src/views/Log.vue#L140-L146

I believe that list of files is static and not meant to be changed in the UI, so it can be replaced by `jobLogFiles: Object.freeze(['job.out',`, or moved outside as a constant somewhere :+1: 

**Additional context**

I saw a post on Discourse and checked out the code to play with the old workflow `five`, and while having a look at the code I noticed that list. Since I had just changed that in Jena UI's code, I started freezing some `data` properties but never finished to get a good pull request. Thus this PR :grimacing: :wave: 

Some other good tips to improve performance: https://old.reddit.com/r/vuejs/comments/10piox0/what_vue_specific_performance_tips_do_you_have/j6lm3f4/

(Great job on the graph & log views! They are fun!)

**Pull requests welcome!**

MetRonnie:
Closes #1247

Can't say I've noticed any change in performance however. Can bump this to later milestone.

**Check List**

- [x] No changelog or docs etc.

### DIFF
--- a/src/components/Markdown.vue
+++ b/src/components/Markdown.vue
@@ -26,6 +26,8 @@
 <script>
 import MarkdownIt from 'markdown-it'
 
+const md = new MarkdownIt()
+
 export default {
   name: 'Markdown',
 
@@ -36,13 +38,9 @@ export default {
     }
   },
 
-  data: () => ({
-    md: new MarkdownIt()
-  }),
-
   computed: {
     html () {
-      return this.md.render(this.markdown)
+      return md.render(this.markdown)
     }
   }
 }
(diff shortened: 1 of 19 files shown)


oliver-sanders (commented):
LGTM, haven't given a functional test yet.

MetRonnie:
@wxtim up for a functional test? (Basically do a quick try out of every aspect of the app you can think of)

wxtim:
> @wxtim up for a functional test? (Basically do a quick try out of every aspect of the app you can think of)

I've done this.

wxtim (approved):
Functionally tested

PR #1390, opened 2023-07-13
Strip token from URL when app is first loaded

### LINKED ISSUE #1147: jupyter-server: strip token from URL

MetRonnie:
Closes #1147, re-do of #1151

After the browser has sent the token as part of the URL in the first GET (e.g. `https://localhost:1234/cylc/?token=ab1c234de`) the server responds with a cookie that is stored in the browser so the token is no longer needed.

I have not encountered the race condition described in #1153 but reviewers should test carefully

**Check List**

- [x] Tests are included
- [x] No changelog entry as fairly minor change

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -33,31 +33,45 @@ import { createVuetify } from 'vuetify'
 import mitt from 'mitt'
 import { createHead, VueHeadMixin } from '@unhead/vue'
 
-const app = createApp(App)
+if (location.search) {
+  /* Remove token from the querystring - we only need it on first load.
+  After the browser has sent the token as part of the URL in the first GET, the
+  server responds with a cookie that is stored in the browser so the token is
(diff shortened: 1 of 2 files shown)


oliver-sanders (commented):
LGTM, need to test.

wxtim (approved):
- [x] Manually Tested as working.
- [x] Haven't come across race condition.
- [x] Read Diff. Amused by `token=l0r3m1p5um`.

PR #1391, opened 2023-07-14
Renovate: automerge Vite dependencies

MetRonnie:
I think these should be fine to automerge as the CI is extremely likely to fail if there's any issue

**Check List**

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -4,11 +4,23 @@
   ],
   "labels": ["dependencies", "javascript"],
   "platformAutomerge": true,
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -21,7 +33,6 @@
         "nodemon"
       ],
       "automerge": true,
-      "automergeStrategy": "squash",
```
oliver-sanders, line 24: ?

MetRonnie: I've set it at the top-level which hopefully works and means not repeating it every time

PR #1406, opened 2023-07-27
Merge gscan auto collapsing changes with the latest vue3 branch

AaronDCole:
Hey guys, please have a look at this MR which is a lift and shift of the changes from the main gscan branch, onto a fresh checkout of master (as it was getting a bit tricky to merge in place).

Supersedes https://github.com/cylc/cylc-ui/pull/1175

Let me know if anything else is required.

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [x] Tests are included (or explain why tests are not needed).
- [ ] `CHANGES.md` entry included if this is a change that can affect users
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

Issue #952, opened 2022-03-01
Add a hierarchical workflow to the offline/testing mode


MetRonnie:
Currently we have only one workflow (`one`) in the demo/offline mode (`yarn run serve`). This limits the ability to manually test out GScan in this mode. We should add a hierarchical workflow (e.g. `a/b/c/run1`).

**Pull requests welcome!**

PR #1416, opened 2023-08-02
Auto-flatten single-child directories in GScan sidebar

### LINKED ISSUE #865: gscan: auto collapse nodes where appropriate
GScan displays a hierarchical view of workflow registrations e.g:

```
foo/
  run1
  run2
bar/
  pub
baz/
  run1
```

The numbered runs (i.e. the run1, run2, etc) are a new Cylc 8 feature.

For many cases there will only be one run (run1), so displaying this full hierarchy is kinda annoying e.g:

```
foo/
  run1
bar/
  run1
baz/
  run1
pub/
  run1
```

To make GScan more compact and to help ease users into the numbered runs we should collapse these nodes somehow (kinda like what GitHub does when a directory contains only one directory e.g. cylc/flow).

```
foo/run1
bar/run1
baz/run1
pub/run1
```

**Pull requests welcome!**

### LINKED ISSUE #1417: `TreeItem` component: computed properties that are only relevant for GScan
### Problem

From #1416:

> We have to be careful with computed properties for the `TreeItem` component as some workflows have thousands of tasks and having several computed properties per item that are only relevant to GScan is not good performance-wise.

### Proposed Solution

Not sure what the best solution here is. One possibility is a separate `GScanTreeItem` component.

MetRonnie:
Closes #865
Closes #1417
Partially addresses #952

Supersedes #1175 & #1406

This is @AaronDCole's #1175 squash rebased onto master plus some tidying up and optimisation and fixes for a couple of small issues.

**Update:**
- I've split the GScan part of the TreeItem component into its own component to address the performance concerns of #1417
- I went on to actually flatten single-child workflow-part nodes at the time of sorting them

image

#### Check List

- [x] Tests are included 
- [x] No docs


wxtim:
Initial testing looks good 

image


Is it OK (probably tangential) that the task state LEDs dissapear under the edge of the GScan if the sidebar isn't wide enough? The do reappear if you resize the bar. I have not yet checked this on master, but will do so once I've finished this review and am happy to put the branch I'm reviewing away.

CHANGES.md
```diff
@@ -19,8 +27,8 @@ ones in. -->
 [#1269](https://github.com/cylc/cylc-ui/pull/1269) -
 Upgraded Vue and Vuetify frameworks to v3.
 
-[1240](https://github.com/cylc/cylc-ui/pull/1240) - Allow edit-runtime
-for ``[root]`` family by clicking on Cycle icon.
+[#1240](https://github.com/cylc/cylc-ui/pull/1240) - Allow edit-runtime
```
wxtim, line 30: Nice Spot - this was my error.

MetRonnie:
> If I change
> 
> ```ini
> # global.cylc
> [install]
>   max depth = 8
> ```
> 
> Workflows installed at a depth > 4 don't show at all!

They do for me

src/components/cylc/tree/TreeItem.vue
```diff
@@ -382,26 +382,20 @@ export default {
       traverseChildren(this.node)
       return tasks
     },
-    // this has be used in conjunction with branchingLineage since it will always return the first child otherwise
-    lastDescendent () {
+    /** Last descendant node that is the only child of its parent. */
+    lastSingleDescendant () {
```
oliver-sanders, line 388: Rather than having this as a computed property of the TreeItem and passing that property into the template, could we have it as a method of the GScan component and call it on the (reactive) node?

That would keep the GScan logic in the GScan component, avoiding the overheads on the tree view.

MetRonnie: Not completely straightforward as other computed properties depend on this one. And I think the caching of computed properties is desirable. So the only real solution I can think of is to split this component in two - a GScan tree item and tree view tree item (#1417)

oliver-sanders: If one computed property depends on another then they can be combined into a single function?

Isn't this:

```js
myFunction(reactiveProperty)  // gets recomputed when reactiveProperty changes
```

Essentially equivalent to this:

```js
myDerivedProperty // gets recomputed when reactiveProperty changes
// ...
computed: { myDerivedProperty () { return myFunction(reactiveProperty) } }
```

There are a couple of other solutions too:

1. Solve this at the node level (keeps the logic out of the tree component).
2. Use a mixin (avoids splitting the tree component).

MetRonnie: Computed properties only get re-evaluated when their reactive dependencies change, whereas methods get re-evaluated on every re-render, see https://vuejs.org/guide/essentials/computed.html#computed-caching-vs-methods

Looking at `TreeItem.vue` there's actually very little in common between the tree view related stuff and the GScan related stuff, hence why I think splitting is the best option. But we can certainly use a mixin (or composable) for the common stuff.

oliver-sanders: Good to know on the reactivity front.

IMO, there should be a simple abstract `Tree` component which just does tree thing, like indentation, recursion, expand-collapse, etc. GScan and the Tree view should both use this Tree component. We should be able to swap for another off-the-shelf tree component if it is desirable to do so e.g. [the Element virtualised Tree](https://element-plus.org/en-US/component/tree-v2.html). Unfortunately the logic has leaked and the Tree component is far from abstract.

Note we can solve this at the node level as suggested above, the GScan component is already defining its own node tree. We should be able to change the data going into the tree to reflect the state-totals trickling stuff.

MetRonnie: (Copied some of this to #1417, suggest we continue discussion there)

oliver-sanders: I'm a bit concerned about the possible performance impact on the tree view, especially given that the current performance is insufficient for some examples.

oliver-sanders (commented):
Code looks good (was a fun diff to unravel).

Workflow collapsing looks good, haven't tested deeply yet, will give that a go shortly.

I did spot one issue where the workflow icons stopped being clickable after state change.

Also, another one, but I'm not sure if it's related to this PR. The counter doesn't make a whole lot of sense to me:

Screenshot from 2023-10-23 13-23-11

This is made more pronounced by the collated totals for a workflow installation hierarchy as the counters don't appear to be summed?

A possible enhancement to the state total summing which would be really handy if it's easy to squeeze into this PR. Keep the collapsed summary open when the node is expanded. I.E. always show the aggregate state totals.

src/components/cylc/tree/GScanTreeItem.vue
```diff
@@ -0,0 +1,228 @@
...
+        :class="nodeClass"
+        class="flex-shrink-0"
+        style="margin: 0 2px;"
+      />
+      <v-list-item
+        :to="workflowLink"
+        :class="nodeClass"
+        class="flex-grow-1 flex-shrink-1 px-2 ml-1"
+      >
+        <div class="d-flex align-center align-content-center flex-nowrap">
```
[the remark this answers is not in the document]
MetRonnie: `v-row`/`v-col` is mainly meant for [grid layouts](https://vuetifyjs.com/en/components/grids) from what I can gather. Here I find using `div`s with flexbox is a simpler way to go

src/components/cylc/tree/TreeItem.vue
```diff
@@ -38,200 +41,153 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-        <div :class="nodeDataClass" @click="nodeClicked">
-          <!-- Task summary -->
-          <Task
-            v-cylc-object="node"
-            :key="node.id"
-            :task="node.node"
-            :startTime="(latestJob(node) || {}).startedTime"
+      <slot v-bind="{ isExpanded }">
+        <!-- the node value -->
+        <!-- TODO: revisit these node.type values that can be replaced by constants later (and in other components too). -->
```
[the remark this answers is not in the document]
MetRonnie: I think it means 
```diff
-v-if="node.type === 'cycle'"
+v-if="node.type === CYCLE"
```

src/components/cylc/tree/TreeItem.vue
```diff
@@ -38,200 +41,153 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+          <template v-if="node.type === 'cycle'">
+            <!-- NOTE: cycle point nodes don't have any data associated with them
+              at present so we must use the root family node for the task icon.
+              We don't use this for the v-cylc-object as that would set the node
+              type to family. -->
+            <Task
+              v-cylc-object="node"
+              v-if="node.familyTree?.length"
+              :key="node.id"
+              :task="node.familyTree[0].node"
```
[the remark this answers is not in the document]
MetRonnie: I thought about it but decided to hold off. Performance-wise, I don't think this is too much of a problem because there is not likely to be too many workflows in GScan compared to cycle points/families/tasks in the Tree view which could be hundreds. (Having said that, we did see that one case of run249!)

src/components/cylc/tree/TreeItem.vue
```diff
@@ -313,17 +266,26 @@ export default {
...
+        return this.manuallyExpanded ?? this.autoExpandTypes.includes(this.node.type)
+      },
+      set (value) {
+        this.manuallyExpanded = value
       }
+    },
+
+    renderChildren () {
+      // Toggles to true when this.isExpanded first becomes true and doesn't get recomputed afterwards
+      return this.renderChildren || this.isExpanded
```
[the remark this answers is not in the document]
MetRonnie: I tested this out with a breakpoint. 
```diff
     renderChildren () {
       // Toggles to true when this.isExpanded first becomes true and doesn't get recomputed afterwards
+      if (this.node.type === 'task' && this.node.id.includes('eventually')) debugger
       return this.renderChildren || this.isExpanded
     },
```
1. Workspace view load: `renderChildren: undefined -> false`
1. Click chevron button to expand task "eventually_succeeded": `renderChildren: false -> true`
1. Click chevron button to collapse it: `renderChildren: true -> true`
2. Click chevron button again ad infinitum: breakpoint is not hit

src/components/cylc/tree/TreeItem.vue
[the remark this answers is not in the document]
oliver-sanders: So, it probably should go there, but there's no pressure to do so ATM?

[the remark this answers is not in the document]
MetRonnie: Want me to open an issue for it?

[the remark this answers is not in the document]
oliver-sanders: Can do, I was mostly just trying to understand the motive for the changes.

MetRonnie:
> Is it OK (probably tangential) that the task state LEDs dissapear under the edge of the GScan if the sidebar isn't wide enough? The do reappear if you resize the bar.

Fixed btw

MetRonnie:
> [MetRonnie#6](https://github.com/MetRonnie/cylc-ui/pull/6)

I think this is the case you are talking about - a single-child workflow-part that is itself a child of a many-child workflow-part?

image

Fixed in https://github.com/cylc/cylc-ui/pull/1416/commits/90c92dd8f02c480e0360dca8909a80739199f3d4

oliver-sanders (approved):
:rocket:

wxtim (approved):
- [x] Tried to break locally.
- [x] Read the code.

PR #1428, opened 2023-08-09
Feature mutation form sticky buttons

### LINKED ISSUE #835: Close mutation editor on submission

[edit] updated, see comments bellow

- [x] Remove the old task icon status thinggy (was never intended to be released). (#1108)
- [x] Make the dialogue box close on successful submission of a mutation. (#1160)
- [x] Keep the dialogue box open for an unsuccessful submission and raise an Alert (if not already done). (#1160)
- [ ] Add a close "x" button to the top right of the mutation editor form, as well as the mutation menu

[/edit]

The following dialog is fully functional but some users may find having to close with the cancel button confusing.

In my opinion these dialogs should have a cross which copies the functionality of the cancel button at either the top right or top left corner.

image

**Pull requests welcome!**

markgrahamdawson:
Closes #835

Added additional sub-components to v-card component including v-card-title and v-card-text.
Added styling to v-card-text to make it 50% of view port height and have scroll in y direction.

Added Mark Dawson to contributing.md

**Check List**

- [x] Tests are included
- [x] no docs needed

### DIFF
--- a/src/components/cylc/Mutation.vue
+++ b/src/components/cylc/Mutation.vue
@@ -16,14 +16,12 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 -->
 
 <template>
-    <v-card class="d-inline-block pa-4">
+    <v-card>
       <!-- the mutation title -->
-      <v-card-title>
-        <h3 :style="{ 'text-transform': 'capitalize' }">
+      <v-card-title class="py-3">
           {{ mutation._title }}
-        </h3>
       </v-card-title>
-      <v-card-text class="card-text">
+      <v-card-text class="card-text py-0 px-4">
         <!-- the mutation description -->
         <v-expansion-panels
           variant="accordian"
@@ -69,8 +67,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           v-model="isValid"
         />
       </v-card-text>
-      <br />
-      <v-card-actions>
+      <v-card-actions class="pa-3">
         <v-spacer></v-spacer>
         <v-btn
           color="grey"
--- a/src/components/cylc/Mutation.vue
+++ b/src/components/cylc/Mutation.vue
@@ -222,8 +222,8 @@ export default {
 }
 </script>
 
-<style>
+<style scoped>
 .card-text{
-  overflow-y:scroll
+  overflow-y:auto
 }
 </style>
--- a/src/components/cylc/Mutation.vue
+++ b/src/components/cylc/Mutation.vue
@@ -224,6 +224,6 @@ export default {
 
 <style>
 .card-text{
-  height:50vh; overflow-y:scroll
+  overflow-y:scroll
 }
 </style>
--- a/src/components/cylc/Mutation.vue
+++ b/src/components/cylc/Mutation.vue
@@ -18,55 +18,57 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 <template>
     <v-card class="d-inline-block pa-4">
       <!-- the mutation title -->
-      <h3 :style="{ 'text-transform': 'capitalize' }">
-        {{ mutation._title }}
-      </h3>
-
-      <!-- the mutation description -->
-      <v-expansion-panels
-        variant="accordian"
-        v-bind="extendedDescription ? { hover: true } : { readonly: true }"
-      >
-        <v-expansion-panel
-          class="mutation-desc"
-          elevation="0"
+      <v-card-title>
+        <h3 :style="{ 'text-transform': 'capitalize' }">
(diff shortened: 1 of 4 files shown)


src/components/cylc/Mutation.vue
```diff
@@ -18,55 +18,57 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 <template>
     <v-card class="d-inline-block pa-4">
```
MetRonnie, line 19:
```suggestion
    <v-card>
```

There's a gap between the scrollbar and the edge of the card. We can remove this by removing this padding class (`pa-4` meaning `padding` `all sides` `4 * 4px = 16px`) from the card.

image

Also I noticed that if we remove the `inline-block` style we don't need to explicitly set the height of the `v-card-text` component

src/components/cylc/Mutation.vue
```diff
@@ -18,55 +18,57 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-        variant="accordian"
-        v-bind="extendedDescription ? { hover: true } : { readonly: true }"
-      >
-        <v-expansion-panel
-          class="mutation-desc"
-          elevation="0"
+      <v-card-title>
+        <h3 :style="{ 'text-transform': 'capitalize' }">
+          {{ mutation._title }}
+        </h3>
```
MetRonnie, line 24: I've noticed the `h3` element and the text-transform are not really needed 

```suggestion
      <v-card-title class="py-3">
        {{ mutation._title }}
```

src/components/cylc/Mutation.vue
```diff
@@ -18,55 +18,57 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+            mutation,
+            types,
+            initialData
+          }"
+          ref="form"
+          v-model="isValid"
+        />
+      </v-card-text>
       <br />
       <v-card-actions>
```
MetRonnie, line 73: While you're here I've noticed this break element is not needed. And we can add back in some padding I've removed from the parent card above.

```suggestion
      <v-card-actions class="pa-3">
```

src/components/cylc/Mutation.vue
```diff
@@ -222,3 +224,9 @@ export default {
   }
 }
 </script>
+
+<style>
```
MetRonnie, line 228: Tip: use [scoped style](https://vuejs.org/api/sfc-css-features.html#scoped-css) to avoid this CSS leaking out where we don't expect it
```suggestion
<style scoped>
```

MetRonnie (commented):
Good work 👍 

I've left some suggestions for changes, most of which are removing some cruft I've spotted - don't worry, it was lying around before you got here!

tests/e2e/specs/mutation.cy.js
```diff
@@ -199,4 +199,19 @@ describe('Mutations component', () => {
...
+      const rect = $el[0].getBoundingClientRect() // https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect
+      const isInViewport = (
+        rect.top >= -2 &&
+        rect.left >= 0 &&
+        rect.bottom <= Cypress.$(cy.state('window')).height() &&
+        rect.right <= Cypress.$(cy.state('window')).width()
+      )
+      expect(isInViewport).to.be.true
+    })
+  })
```
MetRonnie, line 216: We need to test on a mutation that has a lot of stuff in the form so that overflows. Unfortunately with the mocked mutations all of them are very short so they don't overflow, and so this test passes on master.

Also, there's no need to manually check the element is within the viewport; the assertion `.should('be.visible')` does this for you.

Here's how I would go about this
```suggestion
    cy.get('.c-interactive:first')
      .click()
      .get('#less-more-button')
      .click()
      .get('.c-mutation-menu-list')
      // choose mutation with a long form so it overflows
      .contains('Edit Runtime')
      .click()
      .get('.v-card-actions')
      .should('be.visible')
```

tests/e2e/specs/mutation.cy.js
```diff
@@ -199,4 +199,18 @@ describe('Mutations component', () => {
       .get('@errTooltip')
       .should('not.have.css', 'display', 'none')
   })
+  it('should display action buttons in menu card actions component', () => {
```
MetRonnie, line 202: Just a slight clarification
```suggestion
  it('has actions buttons pinned to bottom of form', () => {
```

MetRonnie (commented):
Nearly there 👍 Just a few remaining comments to be addressed

MetRonnie (approved):
Nice

oliver-sanders (approved):
:+1:

PR #1430, opened 2023-08-10
Log view: update file list when opening file select

MetRonnie:
A small usability enhancement: update the list of log files when you open the input, instead of having to click the refresh button beforehand.

demo12

**Check List**

- [x] Existing tests should be fine
- [x] No docs needed


oliver-sanders:
Looks good but I'm no so sure about this, it's a bit over-the-top. Listing log files involves an SSH to the remote server so is best avoided when not required, the list of log files seldom changes and most tasks don't make use of custom log files in the first place.

MetRonnie:
The main advantage to this is for viewing the newest scheduler log after restarting a workflow. We could debounce the request to only happen at most every e.g. 30s?

PR #1432, opened 2023-08-14
Update README

MetRonnie:
Correction for how to run Vite dev server with Cylc UI Server live data


markgrahamdawson (approved):
Thanks that's a lot clearer

PR #1437, opened 2023-08-17
Added noAuth page and routing

### LINKED ISSUE #1394: hub: display something if the user does not have read permissions
The Cylc UI Server authorisation system blocks restricts API calls to authorised users, static file requests, however, are not subject to authorisation.

This means that if a user does not have `READ` permissions, the Cylc UI app will load, but no workflows will show up. This makes it look like the Cylc UI is broken.

> Note: It's the same story with Jupyter Lab, which will also load the app, but all API calls will fail.

We should try to make this situation a little more obvious to the authenticated user by displaying a static page saying something like "you are not authorised to access this server" in the Cylc UI.

markgrahamdawson:
Closes #1394

**Check List**


tests/e2e/specs/url.cy.js
```diff
@@ -35,4 +35,21 @@ describe('URL handling', () => {
         expect(url.endsWith('?a=1&b=2')).to.be.true
       })
   })
+
+  it('reroutes to noAuth page if user isnt authenticated', () => {
```
MetRonnie, line 39: Difference between authorisation and authentication is basically
- Authentication is ensuring the identity of the user
- Authorisation is determining the permissions of the user

In this case the user is authenticated but not authorised to view anything
```suggestion
  it('reroutes to noAuth page if user isnt authorised', () => {
```

oliver-sanders (approved):

MetRonnie:
For some reason this `analysis.cy.js` test seems to be reliably failing on this branch but only for Firefox

Analysis view -- sorts the chart entries -- before each hook (failed)

src/router/paths.js
```diff
@@ -145,5 +145,15 @@ export default [
...
     props: true
+  },
+  {
+    path: '/noAuth',
+    view: 'NoAuth',
+    name: 'noAuth',
+    meta: {
+      layout: 'noAuth',
+      toolbar: false,
+      showSidebar: true
```
MetRonnie, line 156: (Don't need these for this custom layout)
```suggestion
```

oliver-sanders:
> For some reason this analysis.cy.js test seems to be reliably failing on this branch but only for Firefox

It passed on retry, this test is flaky, I've put up an issue, can ignore for now.

Issue #1393, opened 2023-07-17
multi-user: better display of user information and easier server switching


oliver-sanders:
> **Terminology**
>
> * User - The authorised user, i.e. the person looking at the web browser.
> * Owner - The user account running the server, i.e. the person who's workflows the user is looking at.
> * Deployment - The Cylc Hub instance we are logged in to. Note, there may be more than one, e.g. research and production deployments might be separate.

Multi user functionality is a little rough around the edges at the moment:

* It isn't especially obvious who the server owner is. Their username is in the URL but that's not especially obvious.
* The user's username is displayed in the top left but it's not especially obvious that it's **their** username not the owner.
* There is no way to switch users without editing the URL.
* There is not way to switch deployments without editing the URL.

Suggestions:

1. Display the user in the top right in a circle avatar thinggy.
   * This has become the standard way to display the authenticated user which is now the way it is displayed in Jupyter Lab.
   * Put the user's initials in a circle.
   * Attach a link to the user-profile to this avatar for easy access to the Cylc "settings".
2. Display the owner in the top left.
4. Make the owner field editable.
   * When the value is changed, the page should redirect or open a new browser tab as a way of changing server.
   * We can't necessarily list user accounts, but we can remember previous values to cut down on typing.
   * We may open up the ability to configure this, or hook into the Jupyter Hub API for user listing where functional and authorised, so keep in mind that we may want to inject initial values for this field.
5. Make the deployment editable (some day)
    * One day we may allow sites to configure the URLs of their different `cylc hub` deployments.
    * If and when, we could also make this field editable for easy deployment switching.

Illustrative Example:

![auth](https://github.com/cylc/cylc-ui/assets/16705946/ca644172-71ed-4e31-b60a-283492607463)

PR #1444, opened 2023-08-23
Feature multiuser display

markgrahamdawson:
Closes https://github.com/cylc/cylc-ui/issues/1393

Added combobox inputs for survey owner and deployment  
- these are disabled on single user user mode
- when user selects new path - 'Go' button appears
- clicking go button takes user to new web address 
- 'Go button' is hidden in single user mode
- - difficult to test routing as cant test multi-user mode

Added avatar to top right on workflow view
- avatar shows users full username and admin status

**Check List**

### DIFF
--- a/src/components/cylc/Header.vue
+++ b/src/components/cylc/Header.vue
@@ -37,8 +37,9 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         </g>
       </svg>
     </div>
-    <div class="c-environment-info w-75 d-flex-col justify-center">
+    <div id="cylc-select-options" class="c-environment-info w-75 d-flex-col justify-center">
       <v-combobox
+        id="cylc-owner-combobox"
         :disabled = "store.state.user.user.mode !=='multi user'"
         bg-color="white"
         label="server owner"
@@ -50,6 +51,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         @keyup.enter="addOwner(owner)"
       ></v-combobox>
       <v-combobox
+        id="cylc-deployment-combobox"
         :disabled = "store.state.user.user.mode !=='multi user'"
         bg-color="white"
         label="deployment"
--- a/src/components/cylc/Header.vue
+++ b/src/components/cylc/Header.vue
@@ -39,6 +39,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
     </div>
     <div id="cylc-select-options" class="c-environment-info w-75 d-flex-col justify-center">
       <v-combobox
+        class="my-3"
         id="cylc-owner-combobox"
         :disabled = "store.state.user.user.mode !=='multi user'"
         bg-color="white"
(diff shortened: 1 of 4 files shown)


hjoliver:
Nice!

MetRonnie (changes requested):
Got some review suggestions at #1454

tests/e2e/specs/header.cy.js
```diff
@@ -15,19 +15,63 @@
...
+        name: 'userTest',
+        groups: [
+          'cylc',
+          'developer'
+        ],
+        created: '2021-03-23T23:26:23.606Z',
+        admin: true,
+        server: '/user/cylc/',
+        permissions: [
+        ],
```
MetRonnie, line 42: Tests are failing because of the no-auth landing page you added
```suggestion
        permissions: [
          'read',
        ],
```

MetRonnie (approved):

src/components/cylc/Header.vue
```diff
@@ -37,27 +37,79 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+        <v-combobox
+          class="w-100"
+          id="cylc-deployment-combobox"
+          :disabled="store.state.user.user.mode !== 'multi user'"
+          label="deployment"
+          :default="deployment"
+          :items="Array.from(deployments)"
+          v-model="deployment"
+          @keyup.enter="deployments.add(deployment)"
+        />
```
MetRonnie, line 69: Actually would "host" make more sense than "deployment"?

oliver-sanders: Deployment may refer to a collection of hosts e.g. exvcylc01-17.

src/components/cylc/Header.vue
```diff
@@ -37,27 +37,79 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+// route logic
+const url = computed(() => `//${deployment.value}/user/${owner.value}/cylc/#`)
+
+const isNewRoute = computed(() => {
+  return deployment.value !== deploymentOnLoad || owner.value !== ownerOnLoad
+})
+
 </script>
+
+<style>
```
oliver-sanders, line 110: Should this be `scoped`?

src/components/cylc/Header.vue
```diff
@@ -37,27 +37,79 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+        }
+      }">
+        <v-combobox
+          class="w-100"
+          id="cylc-owner-combobox"
+          :disabled="store.state.user.user.mode !== 'multi user'"
+          label="server owner"
+          :default="owner"
+          :items="Array.from(owners)"
+          v-model="owner"
```
oliver-sanders, line 57: I was hoping the `v-combobox` would remember values the user had input previously (i.e. from a previous session) which it doesn't.

We should look at moving the `owner` and `deployment` values into `LocalStorage` so that they are remembered for future sessions, but we can do that later.

src/components/cylc/Header.vue
MetRonnie: Yes, or better this style can be moved to `src/styles/index.scss`

oliver-sanders (approved):

PR #1446, opened 2023-08-24
aotf: ensure log mutation is listed when appropriate

### LINKED ISSUE #1435: command menu: include the "log" command when user does not have write perms
The command menu (also called mutation menu) lists the actions you can perform on a node (e.g. stop, play, trigger).

If you do not have write permissions, then it filters out all commands i.e:

Screenshot from 2023-08-15 13-25-02

These days the command menu includes a link to the log view which only requires read permissions so should not be filtered out.

oliver-sanders:
* Closes #1435
* Ensure the log mutation is listed, even when a user only has read permissions
* List the log mutation for workflows too.

The log mutation was getting filtered out when a user only had read permissions. The reason turned out to be this:

https://github.com/cylc/cylc-ui/commit/95f14f5fad6646fea683f3a11412125e069562f4#diff-ed0c1a902b9b47f63c1058d8ab66442fe9410209e762c70ddf15623b4b60dd1eR170

I expect this was a simple stopgap solution to hiding mutations for users who don't have write perms before we had the current solution implemented.

**Check List**

### DIFF
--- a/src/components/cylc/cylcObject/Menu.vue
+++ b/src/components/cylc/cylcObject/Menu.vue
@@ -199,7 +199,7 @@ export default {
     ...mapState('user', ['user']),
 
     displayMutations () {
-      if (!this.mutations.length || this.user.permissions.length < 2) {
+      if (!this.mutations.length) {
         return []
       }
       const shortList = this.primaryMutations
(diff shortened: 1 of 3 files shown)


MetRonnie (approved):

MetRonnie:
I'm assuming no changelog entry necessary?

oliver-sanders:
It's a very minor change and the log view was always available, just not listed as an action.

wxtim (approved):
LGTM

PR #1454, opened 2023-08-29
Review suggestions for multi user functionality

MetRonnie:
PR against #1444

### DIFF
--- a/src/components/cylc/Header.vue
+++ b/src/components/cylc/Header.vue
@@ -37,42 +37,47 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         </g>
       </svg>
     </div>
-    <div id="cylc-select-options" class="c-environment-info w-75 d-flex-col justify-center">
-      <v-combobox
-        class="my-3"
-        id="cylc-owner-combobox"
-        :disabled = "store.state.user.user.mode !=='multi user'"
-        bg-color="white"
-        label="server owner"
-        placeholder="server owner"
-        :default="owner"
-        :items="owners"
-        variant="outlined"
-        v-model="owner"
-        @keyup.enter="addOwner(owner)"
-      ></v-combobox>
-      <v-combobox
-        class="my-3"
-        id="cylc-deployment-combobox"
-        :disabled = "store.state.user.user.mode !=='multi user'"
-        bg-color="white"
-        label="deployment"
-        placeholder="deployment"
-        :default="deployment"
-        :items="deployments"
-        variant="outlined"
-        v-model="deployment"
-        @keyup.enter="addDeployment(deployment)"
-      ></v-combobox>
-      <v-btn
-        @click="window.location.href = url.value"
-        class="mx-12 my-2"
-        width="100"
-        color="green"
-        :hidden="isNewRoute == false || store.state.user.user.mode == 'single user'"
-      >
-        Go
-      </v-btn>
+    <div
+      id="cylc-select-options"
+      class="c-environment-info w-100 d-flex flex-column align-center px-4 row-gap-3"
+    >
+      <v-defaults-provider :defaults="{
+        VCombobox: {
+          bgColor: 'white',
+          rules: [(val) => Boolean(val) || 'Required'],
+        }
+      }">
+        <v-combobox
+          class="w-100"
+          id="cylc-owner-combobox"
+          :disabled="store.state.user.user.mode !== 'multi user'"
+          label="server owner"
+          :default="owner"
+          :items="Array.from(owners)"
+          v-model="owner"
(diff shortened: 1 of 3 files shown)


src/components/cylc/Header.vue
```diff
@@ -37,42 +37,47 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-        Go
-      </v-btn>
+    <div
+      id="cylc-select-options"
+      class="c-environment-info w-100 d-flex flex-column align-center px-4 row-gap-3"
+    >
+      <v-defaults-provider :defaults="{
+        VCombobox: {
+          bgColor: 'white',
+          rules: [(val) => Boolean(val) || 'Required'],
```
MetRonnie, line 47: The `rules` prop allows you to specify validation for vuetify inputs

src/components/cylc/Header.vue
```diff
@@ -37,42 +37,47 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-        class="mx-12 my-2"
-        width="100"
-        color="green"
-        :hidden="isNewRoute == false || store.state.user.user.mode == 'single user'"
-      >
-        Go
-      </v-btn>
+    <div
+      id="cylc-select-options"
+      class="c-environment-info w-100 d-flex flex-column align-center px-4 row-gap-3"
```
MetRonnie, line 42: I've tweaked the styles to make more use of the horizontal space available

src/components/cylc/Header.vue
```diff
@@ -86,24 +91,18 @@ const store = useStore()
 // owner logic
 const ownerOnLoad = store.state.user.user.owner
 const owner = ref(ownerOnLoad)
-const owners = ref([ownerOnLoad])
-const addOwner = (owner) => {
-  if (owners.value.indexOf(owner) === -1) { owners.value.push(owner) }
-}
+const owners = ref(new Set([ownerOnLoad]))
```
MetRonnie, line 95: [Sets](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) are a useful container for when you don't want duplicate items, both in JS and Python

src/components/cylc/Header.vue
```diff
@@ -86,24 +91,18 @@ const store = useStore()
...
-  if (deployments.value.indexOf(deployment) === -1) { deployments.value.push(deployment) }
-}
+const deployments = ref(new Set([deploymentOnLoad]))
 
 // route logic
-const url = computed(() => '//' + deployment.value + '/user/' + owner.value + 'cylc/#')
+const url = computed(() => `//${deployment.value}/user/${owner.value}/cylc/#`)
 
 const isNewRoute = computed(() => {
-  return !(deployment.value === window.location.host & owner.value === window.location.href.split('/')[4])
```
MetRonnie, line 106: You accidentally used the _bitwise_ AND operator here so it wasn't working

src/components/cylc/workflow/Toolbar.vue
```diff
@@ -117,28 +117,35 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-          </v-avatar>
-        </template>
-        <div class="text-h5">{{this.user.username}}</div>
-        <div class="d-flex justify-space-between" ><span class="mr-4">Admin</span> <span>{{this.user.admin}}</span></div>
-      </v-tooltip>
-    </router-link>
+      <v-btn icon size="small">
+        <v-avatar color="primary" size="small">
+          {{ userInitials }}
+        </v-avatar>
```
MetRonnie, line 123: I've converted the tooltip into a menu, as it is a common pattern for clicking the profile pic to open a menu rather than redirect you away from the page you were on, which could be unexpected for the user

tests/e2e/specs/header.cy.js
MetRonnie, line 1: I've made these tests a little more exhaustive

src/components/cylc/Header.vue
```diff
@@ -37,42 +37,47 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+          label="deployment"
+          :default="deployment"
+          :items="Array.from(deployments)"
+          v-model="deployment"
+          @keyup.enter="deployments.add(deployment)"
+        />
+        <v-btn
+          v-if="store.state.user.user.mode !== 'single user' && isNewRoute"
+          data-cy="multiuser-go-btn"
+          :href="url"
```
MetRonnie, line 73: Using `v-if` rather than `hidden` means the browser doesn't have to render the DOM node at all, which is generally better for performance.

Also can just set `href` instead of using a click handler.

markgrahamdawson (approved):

PR #1456, opened 2023-08-31
Fix message chips for task outputs

MetRonnie:
The chip text should be the output; the tooltip should be the message

**Check List**

- [x] No docs needed

### DIFF
--- a/src/services/mock/json/App.json
+++ b/src/services/mock/json/App.json
@@ -263,7 +263,7 @@
           "finishedTime": "2020-11-08T22:57:41Z",
           "state": "succeeded",
           "submitNum": 4,
-          "messages": ["started", "foo", "bar", "baz", "pub", "qux", "x", "y"],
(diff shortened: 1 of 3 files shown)


hjoliver:
> The chip text should be the output; the tooltip should be the message (this was backwards before)

Oops, not sure how we got that wrong!

src/services/mock/json/App.json
```diff
@@ -263,7 +263,7 @@
           "finishedTime": "2020-11-08T22:57:41Z",
           "state": "succeeded",
           "submitNum": 4,
-          "messages": ["started", "foo", "bar", "baz", "pub", "qux", "x", "y"],
+          "messages": ["started", "foo message", "bar message", "x", "baz message", "pub message", "qux message", "y"],
```
MetRonnie, line 266: (I moved `x` up so we get a mix of outputs and non-outputs visible in the offline mode; previously `x` and `y` were hidden as beyond the max number shown)

oliver-sanders (approved):

Issue #181, opened 2019-08-12
Create Gantt chart view


kinow:
Not a high priority, but I am creating this issue now to move the work from #81 (to be closed) here.

kinow:
Sketch attached in #75 by @oliver-sanders 

![](https://user-images.githubusercontent.com/16705946/45328727-112fe800-b555-11e8-9fe3-8c798e5181e7.png)

kinow:
And @oliver-sanders ' comment from #81:

>Also introducing the gant chart:
>
>* In live mode view projcted workflow timings.
>* In anlysis mode view averaged timings for resource planning.

![gnt](https://user-images.githubusercontent.com/16705946/45328012-d1b3cc80-b551-11e8-88ed-3a1c6ad9f75f.png)

>would probably need to run left to right and collapse nodes for real suites.

![gnt 2](https://user-images.githubusercontent.com/16705946/45328727-112fe800-b555-11e8-9fe3-8c798e5181e7.png)

oliver-sanders:
* Show dependencies to help suite developers identify bottlenecks.
* Show future cycles (using average job time perhaps with error bars) for basic Cylc forecasting.
* Task collapsing (same as graph, tree, dot views)

oliver-sanders:
Screenshot Tim took of the [Prefect](https://www.prefect.io/) GUI's equivalent of the Gantt view. Not the prettiest, but the inclusion of dependencies is interesting.

![screenshot](https://github.com/cylc/cylc-ui/assets/16705946/5a7b4fc3-1258-42eb-ab3b-88d68538b98f)

PR #1466, opened 2023-09-04
Added Gantt chart

ChrisPaulBennett:
Partially addresses https://github.com/cylc/cylc-ui/issues/181

This is the first iteration of the Gantt chart view.
We wanted to add a few various other features (zooming, scaling of the data and pagination) but I will be going on parental leave earlier than expected. So would like to get some version out there for us to get some much needed feedback.
This also contains some small additions to the box plot.
The view itself is a full working state, just looks a bit squished when viewing a big suite.

**Check List**


hjoliver:
>  So would like to get some version out there for us to get some much needed feedback. 

Thanks @ChrisPaulBennett , I'm looking forward to trying this!  I pull the branch today and got fatal build errors in production mode. Just linter style issues I think. The build completed in dev mode, but I just got a blank screen in the new view. Can you do a build from scratch to check this? (If you haven't gone on leave yet!).

ChrisPaulBennett:
Hi @hjoliver. It was linting issues. Not sure what happened there, but I have amended them now

ChrisPaulBennett:
That was a bit more complicated than I had envisaged. But it is now ready for review.

This branch needs changes to UI server for the query to work properly. This is currently in a pull request
https://github.com/cylc/cylc-uiserver/pull/511

package.json
MetRonnie, line 1: Please address these changes and run `yarn install` to update the lockfile

MetRonnie (changes requested):

src/views/Gantt.vue
```diff
@@ -0,0 +1,220 @@
...
+          { workflows: this.workflowIDs }
+        )
+        this.callback.onAdded(ret.data)
+      },
+      200 // only re-run this once every 0.2 seconds
+    ),
+    onPageChange (page) {
+      console.log(page)
+      this.currentPage = page
+    }
```
oliver-sanders, line 217: Unused?

oliver-sanders (commented):
I'm getting this error when opening the gantt view:

```
TypeError: this.seriesRangeStart[o] is undefined
```

src/views/Analysis.vue
MetRonnie, line 1: All the changes in this file seem to be regressions

src/components/cylc/gantt/GanttChart.vue
MetRonnie, line 1: Is the idea not to include this in the Analysis view as a mode alongside the table and box/whiskers modes?

oliver-sanders: I think the gantt view is more akin to a live view like tree or graph than an analysis view so it makes reasonable sense for it to be a separate view.

package.json
```diff
@@ -50,6 +51,7 @@
     "vue-router": "4.2.5",
     "vue-the-mask": "0.11.1",
     "vue3-apexcharts": "1.4.1",
+    "vue3-pagination": "^1.0.16",
```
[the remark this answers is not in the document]
ChrisPaulBennett: Previously a pagination was going to be included. But some complications have caused my to remove it and add it in as a pull request later. Missed a few deps

src/views/Gantt.vue
ChrisPaulBennett: Previously a pagination was going to be included. But some complications have caused my to remove it and add it in as a pull request later. Missed a few deps

src/components/cylc/gantt/GanttChart.vue
ChrisPaulBennett: Originally this was going to be in the analysis view. But as development continued it became apparent that it needed to be in a separate tab.

package.json
```diff
@@ -85,7 +87,7 @@
     "standard": "17.1.0",
     "vite": "5.0.2",
     "vite-plugin-eslint": "1.8.1",
-    "vite-plugin-istanbul": "5.0.0",
+    "vite-plugin-istanbul": "4.1.0",
```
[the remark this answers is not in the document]
ChrisPaulBennett: Not sure what happened there. Resolved

src/views/Analysis.vue
ChrisPaulBennett: Yes, it was decided that the Gantt chart was to be taken out of the analysis view placed in it's own tab

package.json
```diff
@@ -40,6 +40,7 @@
     "lodash": "4.17.21",
     "markdown-it": "13.0.2",
     "mitt": "3.0.1",
+    "moment": "^2.29.4",
```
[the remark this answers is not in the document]
ChrisPaulBennett: Done

package.json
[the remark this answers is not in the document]
MetRonnie: Note that Vuetify provides a pagination component (which is used for the box & whiskers view)

MetRonnie:
Looking at the commit history here, this could all be squashed into one and rebased onto master, while discarding the unwanted changes to `yarn.lock` that have happened. I'm happy to go through this with you this week

MetRonnie:
Seems to work from a quick test, but do get a couple of uncaught errors

image

MetRonnie:
Ooh, also could do with a reload button I think

src/components/cylc/gantt/GanttChart.vue
```diff
@@ -0,0 +1,212 @@
...
+        }
+      }
+      return [{ data }]
+    },
+
+    chartOptions () {
+      return {
+        chart: {
+          animations: {
+            enabled: this.$store.state.app.reducedAnimation ? false : this.animate,
```
MetRonnie, line 143: This has changed to use a composable rather than the store

```diff
diff --git a/src/components/cylc/gantt/GanttChart.vue b/src/components/cylc/gantt/GanttChart.vue
index 9b5ce490..6341db89 100644
--- a/src/components/cylc/gantt/GanttChart.vue
+++ b/src/components/cylc/gantt/GanttChart.vue
@@ -33,4 +33,5 @@ import {
   mdiSortVariant,
 } from '@mdi/js'
+import { useReducedAnimation } from '@/composables/localStorage'
 
 export default {
@@ -141,5 +142,5 @@ export default {
         chart: {
           animations: {
-            enabled: this.$store.state.app.reducedAnimation ? false : this.animate,
+            enabled: useReducedAnimation().value ? false : this.animate,
             easing: 'easeinout',
             speed: 300,
```

src/components/cylc/gantt/GanttChart.vue
```diff
@@ -0,0 +1,212 @@
...
+            data.push({
+              x: this.jobs[i].name,
+              y: [
+                new Date(this.jobs[i].submittedTime).getTime(),
+                new Date(this.jobs[i].startedTime).getTime(),
+              ],
+              fillColor: jobColours[this.jobs[i].name],
+            })
+          }
+        }
```
MetRonnie, line 134: More DRY-ly:
```suggestion
        const timingOptions = {
          total: { start: 'submittedTime', end: 'finishedTime' },
          run: { start: 'startedTime', end: 'finishedTime' },
          queue: { start: 'submittedTime', end: 'startedTime' },
        }
        const { start, end } = timingOptions[this.timingOption]
        data.push(...this.jobs.map((job) => {
          return {
            x: job.name,
            y: [
              new Date(job[start]).getTime(),
              new Date(job[end]).getTime(),
            ],
            fillColor: jobColours[job.name],
          }
        }))
```

src/components/cylc/gantt/GanttChart.vue
```diff
@@ -0,0 +1,212 @@
...
+    { value: 100, title: '100' },
+    { value: 200, title: '200' },
+    { value: -1, title: 'All' }
+  ],
+}
+</script>
+
+<style>
+.apexcharts-text {
+  font-size: 1.1rem;
```
MetRonnie, line 210: For consistency with the box & whiskers view?
```suggestion
  font-size: 0.9rem;
```

src/components/cylc/gantt/GanttChart.vue
```diff
@@ -0,0 +1,212 @@
...
+    sortDesc: mdiSortVariant,
+  },
+  itemsPerPageOptions: [
+    { value: 10, title: '10' },
+    { value: 20, title: '20' },
+    { value: 50, title: '50' },
+    { value: 100, title: '100' },
+    { value: 200, title: '200' },
+    { value: -1, title: 'All' }
+  ],
```
MetRonnie, line 204: Unused
```suggestion
```

src/components/cylc/gantt/GanttChart.vue
```diff
@@ -0,0 +1,212 @@
...
+      required: true,
+    },
+    timingOption: {
+      type: String,
+      required: true,
+    },
+    itemsPerPage: {
+      type: Number,
+      default: 20,
+    },
```
MetRonnie, line 55: Unused
```suggestion
```

oliver-sanders (commented):
Nice, looking forward to this one.

**Refreshing:**

It could definitely do with a refresh button for now. For future work we can subscribe to the live data and have it update automatically which would be pretty cool.

**Too many tasks:**

We do need to protect this new view against large and or long running workflows to avoid issues. I tried running the OS35 global to see how the gantt view coped, it took a while:

Screenshot from 2024-01-04 14-46-07

But it did eventually load, although the resulting chart wasn't the most usable:

Screenshot from 2024-01-04 14-50-47

But once I used task name filtering to whittle down the number of rows, it worked rather well:

Screenshot from 2024-01-04 14-58-00

I would suggest restricting the number of tasks displayed to a configured value (maybe default ~30) and displaying a message explaining that there are too many tasks to display, please use the task filter to reduce the number. Or alternatively, doing something similar to #1510 and requiring users to manually specify the tasks they're interested in.

**Too many cycles:**

We have the same issue with long-running workflows where there may have been a large number of cycles. E.G. This workflow has racked up a few cycles:

Screenshot from 2024-01-04 14-58-48

Some workflows run for several months and cycle as often as 5 mins which will probably cause issues. I'm not sure how best to counter against this issue, ideas:

* Max cycle limit (i.e. show the most recent 30 cycles and display a message if there are more than 30).
* Get the UI Server to intelligently pick the time window to display and only load earlier / later cycles on request.

**Simulation mode:**

Cylc supports a simulation mode (e.g. `cylc play --mode=simulation`).
 in which jobs are simulated rather than run. Simulated jobs don't add their timings to the DB which means the gantt view doesn't work with simulation mode. This is ok I think, but we might want to detect this so we can display a message explaining what's going on rather than a blank tab.

Alternatively, we could just get Cylc to stick the timings into the database? Will ask the team.

src/views/Gantt.vue
```diff
@@ -0,0 +1,216 @@
...
+          cols="12"
+          md="4"
+          class="pr-md-2 mb-2 mb-md-0"
+        >
+          <v-text-field
+            id="c-gantt-filter-job-name"
+            clearable
+            placeholder="Filter by job name"
+            v-model.trim="jobsFilter.name"
+            ref="filterNameInput"
```
oliver-sanders, line 33: This is filtering by *task* name (e.g. foo, bar, baz) rather than job ID (i.e. the Cylc submission number e.g. 01, 02, 03).

ChrisPaulBennett: Done.
I'm aware of performance / size issues you mention. 
The next thing to implement is pagination. This was going to be done in this pull request, but it turned out to be more work than I envisaged, so it was cut.

ChrisPaulBennett:
I'm not sure how the code coverage is calculated. I have lots of tests in 'gantt-chart.vue.spec.js' that tests this component. What am I missing?

oliver-sanders:
I don't think you've `git add`'ed that test file on this branch.

Screenshot from 2024-01-08 09-52-02

ChrisPaulBennett:
I confused myself a bit there. You are correct that file does indeed not exist. 
What I meant was ganttchart.cy.js
This tests the gantt component. Does more does codecov need?

oliver-sanders:
Here's what Codecov says:

Screenshot from 2024-01-09 13-35-27

So `src/components/cylc/gantt/GanttChart.vue` is well tested (by the component test) but `src/views/Gantt.vue` is untested, which is causing the ~45% patch coverage. An e2e test (`tests/e2e/specs`) would help to ensure the query works along with view features like filtering.

cypress/component/ganttchart.cy.js
```diff
@@ -71,7 +71,46 @@ describe('GanttChart', () => {
...
+    cy.mount(GanttChart, merge(mountOpts, {
+      props: {
+        jobs: [job1, job2],
+        timingOption: 'total'
+      },
+    }))
+    cy.get('.apexcharts-rangebar-area')
+      .first()
+      .invoke('attr', 'barWidth')
+      .should('eq', '161.1500015258789')
```
MetRonnie, line 85: Rather than test the bar size (which is browser-specific) in a component test, you can test the component computed property "`series`" directly in a unit test

ChrisPaulBennett:
> > We do need to protect this new view against large and or long running workflows to avoid issues.
> 
> How about paginating, showing the most recent tasks on the first page?

Pagination was originally going to be in this pull request. But it ended up being a fairly involved process so we thought it best to do it separately. (Actually some of the issue's in this pull request were due to the pagination efforts being removed).
As soon as this PR is in I will begin pagination.

oliver-sanders:
I'm a bit worried about putting this out there before we have some form of pagination / node-limit as there are quite a lot of workflows out there of a sufficient size to cause issues.

Happy for this to be solved in a follow-up, but probably won't want to merge this to master until some form of damage limitation is in so that we can continue to deploy the master branch as there are lots of bugfixes (mostly efficiency fixes for the other views) going in at the moment that we wouldn't want to hold up behind other work.

MetRonnie:
We could create a feature branch for this gantt view in the meantime and merge into that, if that would help break up the work into smaller PRs?

oliver-sanders:
Alternatively, we could stick in a crude solution now and improve on it later, e.g. only show the first N tasks and pop up a notice if there are more tasks than can be displayed at present.

cypress/component/ganttchart.cy.js
ChrisPaulBennett: Agreed that browser specific tests are "icky".
Can you give me a hand creating the test? I'm struggling a bit with them

MetRonnie:
A bunch of duplicate commits showed up here because it looks like you merged your remote branch into your local one. I've force-pushed (7c15333 to 669867a) to remove that unnecessary merge commit - if you compare there is no diff for this push.

oliver-sanders:
Pagination looks nice.

I think it might be interfering with the height of the graph, might need some CSS jiggery pokery to get the graph to fill the available space:

Screenshot from 2024-01-30 14-21-43

The filters will need re-working into the paginated code:

Screenshot from 2024-01-30 14-22-51

Would be good to give over a little more space for the task name too as they can get quite long.

MetRonnie:
> Would be good to give over a little more space for the task name too as they can get quite long.

The truncated task name gets revealed in tooltip if you hover over it

oliver-sanders:
> The truncated task name gets revealed in tooltip if you hover over it

That's good. But it might not be that useful if you have a bunch of tasks with the same prefix where the suffix is obscured.

E.G. If I searched for `ensemble_res_create` I could end up with pages of results which would all display the same.

So if we could increase this width a little it would be very helpful.

ChrisPaulBennett:
This adds the task filter option. This concludes all requirements in our initial iteration of the analysis tools.
We'd like to make these tools available and get feedback before adding any more features.

src/views/Gantt.vue
```diff
@@ -0,0 +1,267 @@
...
+/** The query which retrieves historical Job timing statistics */
+const QUERY = gql`
+query ganttQuery ($workflows: [ID]) {
+  jobs(live: false, workflows: $workflows) {
+    ${jobFields.join('\n')}
+  }
+}
+`
+/** The callback which gets called when data comes in from the query */
+class GanttCallback {
```
MetRonnie, line 147: @oliver-sanders should this be?
```suggestion
class GanttCallback extends DeltasCallback {
```

oliver-sanders: I don't think it makes a functional difference, but it's good practice to inherit from the base class.

oliver-sanders:
Looks good, spotted some small issues:

#### Widget doesn't expand to fill vertical space

Screenshot from 2024-02-15 16-25-44


#### Pagination doesn't reset when the tasks per-page option is changed

This can make it look like there aren't any tasks to display:

vokoscreen-2024-02-15_16-21-26


#### Pagination options

The 10, 25 and 50 tasks per-page options make sense, but the 2 and 5 tasks per-page don't have an obvious purpose.


#### Suggestions for possible future work

* Filtering tasks by globs / families is likely to be a popular request.
* Consider displaying both queue and run times in a single chart?
* The date time format used in the tooltips is a little verbose and the start / finish times aren't horizontally aligned which makes it hard to work out where they differ.
* Hook this view into live data so it updates on the fly (can help with this).

ChrisPaulBennett:
> Looks good, spotted some small issues:
> 
> #### Widget doesn't expand to fill vertical space

Believe it or not, that tiny chart is when it is set to "100%". Setting it ~280% produces a chart that fills the screen. But if you use a monitor and a laptop, you can render the page on your monitor and move the page to your laptop and it's far too big (Windows scaling I think causes this). I've set it to "auto" which doesn't have this issue. But always renders slightly too big


> #### Pagination doesn't reset when the tasks per-page option is changed
> This can make it look like there aren't any tasks to display:

Good spot. Added a watcher to fix this 
 
> #### Pagination options
> The 10, 25 and 50 tasks per-page options make sense, but the 2 and 5 tasks per-page don't have an obvious purpose.

Done

> #### Suggestions for possible future work
> * Filtering tasks by globs / families is likely to be a popular request.

Yup, that's probably the next thing to implement
> * Consider displaying both queue and run times in a single chart?

This never occurred to me. Good idea, should be fairly easy to implement

> * The date time format used in the tooltips is a little verbose and the start / finish times aren't horizontally aligned which makes it hard to work out where they differ.

I've created a function that creates the tooltip text. So it's fairly easy to make it say whatever you want

> * Hook this view into live data so it updates on the fly (can help with this).

Yes, I think I'll need help with that.

src/components/cylc/gantt/GanttChart.vue
```diff
@@ -0,0 +1,214 @@
...
+            offsetX: -10,
+          },
+        },
+      }
+    },
+  },
+}
+</script>
+
+<style>
```
oliver-sanders, line 210: @metronnie, should this be "scoped"?

src/views/Gantt.vue
```diff
@@ -0,0 +1,267 @@
...
+        </v-col>
+      </v-row>
+      <div
+        id="gantt-toolbar"
+        class="d-flex align-center flex-wrap my-2 col-gap-2 row-gap-4">
+
+        <!-- Toolbar -->
+        <v-defaults-provider>
+
+          <!-- Box plot sort input teleports here -->
```
oliver-sanders, line 105: Not quite sure what this means?

src/views/Gantt.vue
```diff
@@ -0,0 +1,267 @@
...
+  onUpdated (updated, store, errors) {
+    this.add(updated)
+  }
+
+  // other hooks we don't need but must declare (for now)
+  before () {}
+  after () {}
+  onPruned () {}
+  commit () {}
+  tearDown () {}
```
oliver-sanders, line 188: MB. If we inherit from the parent class as Ronnie suggested, then we don't have to define these blank methods.

src/views/Gantt.vue
oliver-sanders, line 1: Excellent split of view / component code BTW.

tests/e2e/specs/gantt.cy.js
```diff
@@ -0,0 +1,29 @@
...
+
+describe('Gantt view', () => {
+  beforeEach(() => {
+    cy.visit('/#/gantt/one')
+  })
+  describe('Gantt view', () => {
+    it('Should display the mocked workflow', () => {
+      cy
+        .get('.vue-apexcharts')
+        .should('be.visible')
```
oliver-sanders, line 26: Would be good to test something here, but admittedly it's difficult to do this without testing the internal implementation of Apex Charts which is not the purpose of these tests.

@MetRonnie any ideas?

oliver-sanders (commented):
Cheers.

There should be a solution to the height problem that doesn't overspill the container. Flexbox layouts are a popular way to handle problems like this these days. As it stands, it's only a minor irritation, not a major usage issue so happy to punt this to future work.

https://www.w3schools.com/csS/css3_flexbox.asp

I spotted one more minor issue when loading in a larger workflow. It took a few seconds to fetch the data during which time it looked like there was nothing to display:

vokoscreen-2024-02-16_13-37-52

It would be good to wrap the graph component inside a "skeleton loader" which is a widget that hides the content behind a loading animation to make it clear that something is happening.

https://dev.vuetifyjs.com/en/components/skeleton-loaders/

tests/e2e/specs/gantt.cy.js
ChrisPaulBennett: We have already discussed this and came to the same conclusion. We have unit tests that make sure that apex charts is given the correct input.
I did originally did have tests looking at bar length in the charts. But this was fiddly and different results from different browsers.
I think we just have to assume that apex charts displays things correctly.

src/views/Gantt.vue
ChrisPaulBennett: Flattery will get you everywhere!

src/views/Gantt.vue
ChrisPaulBennett: Remnants of old code that has now been removed.

ChrisPaulBennett: The skeleton loader has been added. I don't have a workflow big enough to cause long enough loads.
Can you test it quickly for me

oliver-sanders: The skeleton loader needs to be toggled according to the load state. Here's an example from the log view:

https://github.com/cylc/cylc-ui/blob/e846f8646582c1f00906af358bf646ceef56670a/src/views/Log.vue#L116-L120

The skeleton loader will only display here if "results.connected" is falsey.

I'll send you a link to my workflow DB.

src/components/cylc/gantt/GanttChart.vue
MetRonnie: Yes, good spot

oliver-sanders: @ChrisPaulBennett, we should add the `scoped` prop to `<style>` elements defined in components to restrict the effects of the style to the component it's defined in:

```suggestion
<style scoped>
```

Details here: https://vue-loader.vuejs.org/guide/scoped-css.html

src/views/Gantt.vue
ChrisPaulBennett: Ronnie was kind enough to implement skeleton loaders properly for me

src/views/Gantt.vue
```diff
@@ -0,0 +1,243 @@
...
+      return [this.workflowId]
+    },
+    filteredJobs () {
+      if (this.jobsFilter.name.length === 0) {
+        return this.jobs
+      }
+      return [matchTasks(this.jobs[0], this.jobsFilter)]
+    },
+    platformOptions () {
+      return platformOptions(this.jobs)
```
MetRonnie, line 215: This function does not seem to be suited to the data format of `this.jobs`. The platform filtering is currently broken

src/components/cylc/gantt/GanttChart.vue
```diff
@@ -0,0 +1,219 @@
...
+          },
+        },
+      }
+    },
+  },
+}
+</script>
+
+<style scoped>
+.apexcharts-text {
```
MetRonnie, line 216: Oh, it turns out you have to do
```suggestion
:deep(.apexcharts-text) {
```
to get this to apply within the `VueApexCharts` child component

https://vuejs.org/api/sfc-css-features.html#deep-selectors

src/views/Gantt.vue
```diff
@@ -0,0 +1,243 @@
...
+      v-else
+      fluid
+      class="pa-2"
+    >
+      <!-- Filters -->
+      <v-row no-gutters>
+        <v-col
+          cols="12"
+          md="4"
+          class="pr-md-2 mb-2 mb-md-0"
```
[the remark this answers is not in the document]
ChrisPaulBennett: Not going to lie. That has been annoying me for a while. Thank you!
Done

src/views/Gantt.vue
ChrisPaulBennett: Ah yes I had forgotten to mention that. I had to reshape the data for the gantt chart (Which has broken the platform filtering)
Since we have the task selection box, do we need platform filtering? Or can we remove it?

src/views/Gantt.vue
```diff
@@ -0,0 +1,243 @@
...
+            id="c-gantt-tasks-per-page"
+            :items="$options.taskChoices"
+            prefix="Tasks Per Page"
+            v-model="tasksPerPage"/>
+        </v-col>
+      </v-row>
+      <div
+        id="gantt-toolbar"
+        class="d-flex align-center flex-wrap my-2 col-gap-2 row-gap-4">
+      </div>
```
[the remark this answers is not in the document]
ChrisPaulBennett: I think that was going to be used by the refresh button. But that has not been implemented yet due to not having live jobs yet. That will be part of the next iteration of features.
I'll remove it for now

src/components/cylc/gantt/GanttChart.vue
ChrisPaulBennett: done

src/views/Gantt.vue
```diff
@@ -0,0 +1,243 @@
...
+        >
+          <v-select
+            id="c-gantt-filter-job-platforms"
+            :items="platformOptions"
+            prefix="Platform:"
+            v-model="jobsFilter.platformOption"
+          />
+        </v-col>
+      </v-row>
+      <v-row no-gutters>
```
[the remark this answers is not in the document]
ChrisPaulBennett: Done

oliver-sanders:
I think we're getting close here? Will re-review when I get the chance. Please complete the checklist in the PR description to prepare for sign off.

src/views/Gantt.vue
oliver-sanders: Up to you.

ChrisPaulBennett: I've fixed platform options. (and refactored filtering)

src/views/Gantt.vue
```diff
@@ -0,0 +1,226 @@
...
+   */
+
+  add (data) {
+    const taskNameList = Array.from(
+      new Set(data.jobs.map((job) => job.name))
+    )
+    this.uniqueTasks = Array.from(
+      new Set(data.jobs.map((job) => job.name))
+    )
+    const sortedData = Object.fromEntries(taskNameList.map(key => [key, []]))
```
MetRonnie, line 139:
```suggestion
    this.uniqueTasks = Array.from(
      new Set(data.jobs.map((job) => job.name))
    )
    const sortedData = Object.fromEntries(this.uniqueTasks.map(key => [key, []]))
```

src/components/cylc/gantt/filter.js
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Function to determine if a task should be displayed given a certain filter
+ * Checks the name includes a search string and if the platform is equal to
+ * that chosen
+ *
+ * @export
+ * @param {object} tasks - Object containing tasks to evaluate
+ * @param {object} tasksFilter - The filter to apply to the tasks
+ * @return {boolean} An Object with the tasks that made it through the filter
```
MetRonnie, line 26:
```suggestion
 * @param {Record<string,Object[]>} tasks - Object containing tasks to evaluate
 * @param {{ name: string, platformOption: string|-1 }} tasksFilter - The filter to apply to the tasks
 * @return {Record<string,Object[]>} An Object with the tasks that made it through the filter
```

tests/unit/components/cylc/gantt/filter.spec.js
```diff
@@ -0,0 +1,78 @@
...
+      task_name: 'task_2',
+      task_platform: 'test_platform_2',
+      some_other_key: 'some_other_test_key'
+    }],
+  }
+
+  it('should filter out the tasks that do not match', () => {
+    const filter = {
+      name: ['task_1', 'task_2'],
+      platformOption: -1,
```
MetRonnie, line 57: Ought to have another test that tests filtering by platform as well

ChrisPaulBennett: Done

oliver-sanders (approved):
LGTM, tested, task selection is good, platform filters work. Thanks!

Would be great if you could put together a screenshot or GIF for the [changes section in cylc-doc](https://cylc.github.io/cylc-doc/nightly_8.3/html/reference/changes.html#cylc-8-3-0), we'll announce every entry in this section one by one.

@MetRonnie, please squash merge.

MetRonnie (changes requested):
Spotted an issue with the filtering: https://github.com/ChrisPaulBennett/cylc-ui/pull/4

src/components/cylc/gantt/GanttChart.vue
```diff
@@ -0,0 +1,218 @@
...
+  },
+  components: {
+    VueApexCharts,
+  },
+
+  props: {
+    jobs: {
+      type: Array,
+      required: true,
+    },
```
MetRonnie, line 58: Just spotted this type is incorrect
```suggestion
    /** @type {Record<string, Object[]>} */
    jobs: {
      type: Object,
      required: true,
    },
```

MetRonnie (approved):
I'm happy with this now 👍  I did notice a couple more issues which I don't know if you want to address now or later.

Edit: address later - https://github.com/cylc/cylc-ui/issues/1698

Issue #1334, opened 2023-06-15
GraphQL 403 message in Chrome


wxtim:
Seen in Windows/Chrome/Version 114.0.5735.134 (Official Build) (64-bit)

### Description

Accessing Cylc Hub from Chrome on windows I'm getting a 403 error from the graphql. I've uploaded the query value.
This does not happen on Firefox 114.0.1 (64-bit) running on the same system.

[query.txt](https://github.com/cylc/cylc-ui/files/11755928/query.txt)

The most noticeable effect is the failure of mutation menus to ever replace the placeholder.

oliver-sanders:
We saw this error with the Jupyter Server v2 branch when the UI is loaded in a fresh browser.

The error in the UIS log showed that the request was missing the xsrf header meaning that it was sent too early.

One option would be to delay the request until the xsrf stuff has happened. Otherwise I would have thought something like this would do it:

```diff
diff --git a/src/services/workflow.service.js b/src/services/workflow.service.js
index 6991d680..dbb27515 100644
--- a/src/services/workflow.service.js
+++ b/src/services/workflow.service.js
@@ -156,10 +156,17 @@ class WorkflowService {
   async loadTypes () {
     // TODO: this assumes all workflows use the same schema which is and
     //       isn't necessarily true, not quite sure, come back to this later.
-    const response = await this.apolloClient.query({
-      query: getIntrospectionQuery(),
-      fetchPolicy: 'no-cache'
-    })
+    try {
+      const response = await this.apolloClient.query({
+        query: getIntrospectionQuery(),
+        fetchPolicy: 'no-cache'
+      })
+    } catch (err) {
+      console.error(err)
+      console.log('retrying')
+      await new Promise(r => setTimeout(r, 2000));
+      return this.loadTypes()
+    }
     const mutations = response.data.__schema.mutationType.fields
     const queries = response.data.__schema.queryType.fields
     const { types } = response.data.__schema
```

PR #1490, opened 2023-09-25
Bug fix graphql 403

markgrahamdawson:
Closes https://github.com/cylc/cylc-ui/issues/1334

I don't have a good understanding of the issue on this one so I have just implemented the solution that was suggested on the [issue ](https://github.com/cylc/cylc-ui/issues/1334#issuecomment-1678971956)by @oliver-sanders .
This issue is hard to reproduce so its difficult to understand if this solution fixes the problem, but the tests still run locally at least.

**Check List**

### DIFF
--- a/src/services/workflow.service.js
+++ b/src/services/workflow.service.js
@@ -156,10 +156,18 @@ class WorkflowService {
   async loadTypes () {
     // TODO: this assumes all workflows use the same schema which is and
     //       isn't necessarily true, not quite sure, come back to this later.
-    const response = await this.apolloClient.query({
-      query: getIntrospectionQuery(),
-      fetchPolicy: 'no-cache'
-    })
+    let response = null
+    try {
+      response = await this.apolloClient.query({
+        query: getIntrospectionQuery(),
+        fetchPolicy: 'no-cache'
+      })
+    } catch (err) {
+      console.error(err)
+      console.log('retrying')
+      await new Promise(resolve => setTimeout(resolve, 2000))
+      return this.loadTypes()
+    }
     const mutations = response.data.__schema.mutationType.fields
     const queries = response.data.__schema.queryType.fields
     const { types } = response.data.__schema
--- a/src/services/workflow.service.js
+++ b/src/services/workflow.service.js
@@ -156,7 +156,7 @@ class WorkflowService {
   async loadTypes () {
     // TODO: this assumes all workflows use the same schema which is and
     //       isn't necessarily true, not quite sure, come back to this later.
-    let response = null
+    let response
     try {
       response = await this.apolloClient.query({
         query: getIntrospectionQuery(),
--- a/src/services/workflow.service.js
+++ b/src/services/workflow.service.js
@@ -164,7 +164,7 @@ class WorkflowService {
       })
     } catch (err) {
       console.error(err)
-      console.log('retrying')
+      console.log('retrying introspection query')
       await new Promise(resolve => setTimeout(resolve, 2000))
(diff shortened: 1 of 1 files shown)


oliver-sanders:
To test, we could try patching the UI Server so that it always raises a 403 error, then inspect the network tab in the developer console to ensure that it is retrying the query as intended.

oliver-sanders:
Tested with the following cylc-uiserver patch which simulates four failed requests before working:

```diff
diff --git a/cylc/uiserver/handlers.py b/cylc/uiserver/handlers.py
index 0ec9341..0a6f334 100644
--- a/cylc/uiserver/handlers.py
+++ b/cylc/uiserver/handlers.py
@@ -269,6 +269,9 @@ class UserProfileHandler(CylcAppHandler):
         self.write(json.dumps(user_info))
 
 
+COUNT = 0
+
+
 class UIServerGraphQLHandler(CylcAppHandler, TornadoGraphQLHandler):
     """Endpoint for performing GraphQL queries.
 
@@ -321,7 +324,11 @@ class UIServerGraphQLHandler(CylcAppHandler, TornadoGraphQLHandler):
 
     @web.authenticated  # type: ignore[arg-type]
     async def execute(self, *args, **kwargs) -> 'ExecutionResult':
+        global COUNT
         # Use own backend, and TornadoGraphQLHandler already does validation.
+        COUNT += 1
+        if COUNT < 5:
+            raise web.HTTPError(403, reason='authorization insufficient')
         return await self.schema.execute(
             *args,
             backend=self.backend,
```

This PR does the job, in the network tab we can see the four failed requests (note we get the error code 400 rather than 403 due to internal implementation details):

Screenshot from 2023-10-02 11-38-26


The fifth request succeeds. I tested this by opening the mutation menu (which shows the loading status), and waiting for the successful request. The mutation list then populated automatically.

oliver-sanders:
There are some funky commits on this branch which shouldn't be here.

The easiest way to get rid of them IMO, is a rebase.

```bash
# make a safe copy of the branch
git checkout bug-fix-graphql-403
git branch bug-fix-graphql-403.safe

# rebase onto the latest upstream/master
$ git fetch upstream/master
$ git rebase -i upstream/master
```

In the editor window that opens, delete any commits which aren't related to this branch.

In this case there should only be one commit "added try catch when calling IntrospectionQuery".

Then save and close the rebase file.

oliver-sanders (approved):

MetRonnie (approved):
Seems to work well. However, instead of retrying indefinitely, might it be better to retry a certain number of times then present an error if it still hasn't worked by that point?

oliver-sanders:
Maintaining the retry state may be tricky, and websockets will constantly try to reconnect too so I don't think the infinite retry thing is a big issue.

MetRonnie:
I was thinking more about alerting the user to the problem otherwise there will just be a loading bar instead of the menu, rather than worrying about the cost of retrying indefinitely. But I guess other requests would also fail if it can't fetch the introspection query at all.

MetRonnie:
> But I guess other requests would also fail if it can't fetch the introspection query at all.

Actually this has turned out to be false, there are problems with jupyterhub version 4.1.x to do with the XRSF cookie that mean only the introspection query fails and so it keeps retrying indefinitely with no success and no error message.

As for what is the cause of the problems with this range of jupyterhub versions, I've dug around and not really found anything illuminating, just that they have been fiddling with the XRSF cookie handling in several quick releases: https://jupyterhub.readthedocs.io/en/stable/reference/changelog.html

Issue #1487, opened 2023-09-22
log: overzealous timestamps option


oliver-sanders:
The "timestamps" option can be a little overzealous:

![Screenshot from 2023-09-22 10-52-47](https://github.com/cylc/cylc-ui/assets/16705946/2078e6fd-aa5b-455d-8eaf-c68b4e8326cf)
![Screenshot from 2023-09-22 10-53-01](https://github.com/cylc/cylc-ui/assets/16705946/8183d996-8944-4a24-a447-9fe93536fd09)

The intention of this option is to strip timestamps from logs on command to make them more readable. I.E. it turns this:

```
TIME1 some message
TIME2 other message
```

Into this:

```
some message
other message
```

It is not supposed to strip things that look like timestamps from elsewhere in log output.

PR #1496, opened 2023-09-29
Bug overzealous timestamps

markgrahamdawson:
Closes https://github.com/cylc/cylc-ui/issues/1487

I have simply added a caret to the front of the Regex to ensure it only matches the beginning of the log entry text. This solves the problem in the issue - whether it solves every situation with matching regex properly - I'm not sure.

**Check List**


MetRonnie (approved):
(There are some extra commits on this branch. When creating a branch make sure to first fetch from upstream and create the branch off `upstream/master`)

hjoliver (approved):

hjoliver:
I think those extra commits will disappear on merge...

hjoliver:
> I think those extra commits will disappear on merge...

Hmm, not really. Fail. 🤦

PR #1501, opened 2023-10-03
Graph node: nudge the task icons down

MetRonnie:
Follow-up to #1489


oliver-sanders (approved):
Dang, can't get anything past you!

Issue #1504, opened 2023-10-04
Source user avatar initials from Jupyter Server


markgrahamdawson:
### Problem

The workflow toolbar avatar is currently getting the users first initial as the first value in the username string.
 `this.user.username[0].toUpperCase()`
This is ok, but sometimes this name is an unexpected value, resulting in a confusing letter being presented in the avatar.

### Proposed Solution
An alternative option would be to get the users initials[ from Jupyter Server](https://jupyter-server.readthedocs.io/en/latest/operators/security.html#jupyter_server.auth.User).

The `initials` field may or may not be provided so we need to be robust to its absence (using the first letter of the username is a good option if initials aren't provided).

PR #1505, opened 2023-10-05
Fetch user initials from UIServer

markgrahamdawson:
Closes https://github.com/cylc/cylc-ui/issues/1504

The userprofile endpoint has been extended in https://github.com/cylc/cylc-uiserver/pull/510 so it now provides user initials.

If for whatever reason the initials field is empty the avatar will be a user symbol
image

And if the initials field is populated the avatar will look like one of the two 
image
image

**Check List**

### DIFF
--- a/src/components/cylc/workflow/Toolbar.vue
+++ b/src/components/cylc/workflow/Toolbar.vue
@@ -93,7 +93,9 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         color="primary"
         data-cy="add-view-btn"
       >
-        <v-icon class="icon">{{ $options.icons.add }}</v-icon>
+        <v-icon class="icon">
+          {{ $options.icons.add }}
+        </v-icon>
         <span class="label">
           {{ $t('Toolbar.addView') }}
         </span>
@@ -101,7 +103,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         <v-menu
           activator="parent"
           location="bottom"
-          >
+        >
           <v-list>
             <v-list-item
               v-for="view in views"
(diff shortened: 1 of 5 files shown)


markgrahamdawson:
Is a changes.md entry required for this one?

tests/unit/model/user.model.spec.js
```diff
@@ -27,14 +27,18 @@ describe('User model', () => {
...
       expect(user.username).to.equal(name)
       expect(user.groups.length === 2).to.equal(admin)
       expect(user.admin).to.equal(true)
       expect(user.created).to.equal(created)
       expect(user.server).to.equal(server)
       expect(user.owner).to.equal(owner)
       expect(user.permissions).to.equal(permissions)
+      expect(user.mode).to.equal(mode)
+      expect(user.initials).to.equal(initials)
     })
```
markgrahamdawson, line 42: There are more things coming back from the userprofile endpoint now although they are not actually being used, for example...
avatar_url
color
should these be added to the Vue user model and service and tests as well?

oliver-sanders:
No need for a changelog entry for this ticket (no one is going to check the cylc-ui changelog in the first place to major things only)

tests/unit/model/user.model.spec.js
oliver-sanders: Maybe, we can do that later.

Opened an issue for it: https://github.com/cylc/cylc-ui/issues/1519

src/components/cylc/workflow/Toolbar.vue
```diff
@@ -119,7 +119,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
       </v-btn>
       <v-btn icon size="small">
         <v-avatar color="primary" size="small">
-          {{ userInitials }}
+          <v-icon v-if="!user.initials" :icon="$options.icons.mdiAccountEdit"></v-icon>
```
oliver-sanders, line 122: I'm not sure how initials could end up undefined, so I don't think it's strictly necessary to handle this case, but there's no harm in it.

oliver-sanders (approved):
Looks good.

Note to reviewers, PAM doesn't return a helpful initials field, but other authenticators can.

MetRonnie (approved):

PR #1518, opened 2023-10-17
GH Actions: upload coverage to Codecov in separate job

MetRonnie:
Workaround https://github.com/codecov/codecov-action/issues/837

- Reduces how hard we hit Codecov API
- Allows easy re-attempt to upload
- Fail CI check if coverage fails to upload

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,9 +51,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest]
-        browser: [firefox, chrome]
         type: [e2e, component]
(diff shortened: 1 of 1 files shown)


wxtim:
Minor and tangential thought - changing the order of the matrix:

```diff
      matrix:
-         os: [ubuntu-latest]
        browser: [firefox, chrome]
        type: [e2e, component]
+         os: [ubuntu-latest]
```

Means that in the test screen you can see which tests are which without burrowing down:

image

wxtim (approved):
- [x] Looks sensible.
- [x] [Works for me](https://github.com/wxtim/cylc-ui/actions/runs/6559939033)

PR #1529, opened 2023-10-25
Properly log errors that are presented as alerts

MetRonnie:
This makes it easier to view the stack trace for caught errors

**Check List**

- [x] Tests are not needed
- [x] No `CHANGES.md` entry included (invisible to users)


src/model/Subscription.model.js
```diff
@@ -17,7 +17,7 @@
 
 import { toRaw } from 'vue'
 import ViewState from '@/model/ViewState.model'
-import Alert from '@/model/Alert.model'
+import { Alert } from '@/model/Alert.model'
```
MetRonnie, line 20: Non-functional change. For some reason the type recognition in VSCode wasn't working for a default export. So that's why I've changed it

tests/e2e/specs/userprofile.cy.js
```diff
@@ -106,6 +106,8 @@ describe('User Profile', () => {
         })
         // Close menu:
         .get('noscript').click({ force: true })
+        .get('@mutationMenu')
+        .should('not.be.visible')
```
MetRonnie, line 110: Based on https://stackoverflow.com/a/76784540/3217306

> ResizeObserver is often used to provide animation, so once you find the line that causes the error, add a visibility check to allow the page to settle before testing

oliver-sanders (approved):

Issue #1253, opened 2023-02-27
Use task state icons for the workflow filter menu in the GScan sidebar


MetRonnie:
From https://github.com/cylc/cylc-ui/pull/1187#issuecomment-1438724461

>Especially given that we use task state icons in the tree/table view task filters I think it would be a good idea to do that here too (it serves as functional documentation if nothing else). We can also do this with the workflow icons.
>
>Worth considering dropping the "Error" workflow state from the filter list. I think we should remove that in due course though so no biggie.

PR #1532, opened 2023-10-26
refactored filters to use icons

markgrahamdawson:
Closes https://github.com/cylc/cylc-ui/issues/1253

## The issue

The task filter v-select is used in both the tree and table views as well as for the gscan component. Previously this has been implemented separately in the two places.

There is also a workflow filter on on the gscan component which should be consistent in style and functionality with the task filter.

For the tree/table views the v-select used task icons.
image

For the gscan component the v-select used closable chips containing the names of the tasks/workflows
image

## The solution

I moved the functionality into a single v-select component (TaskFilterSelect.vue). 
This can be used for tasks and workflows - configured by props.

image

image

There is a bug with the way the vuetify v-select works in that it opens and closes the drop down whenever a value is selected - this was present before the refactor in this pr. I have opened a new issue to look into this further.
https://github.com/cylc/cylc-ui/issues/1535

**Check List**


src/components/cylc/TaskFilterSelect.vue
```diff
@@ -0,0 +1,58 @@
...
+defineEmits(['update:modelValue'])
+
+const localValue = ref()
+localValue.value = computed(() => props.modelValue)
+
+const maxVisibleStates = 4
+
+function removeItem (key) {
+  localValue.value.value.states = localValue.value.value.states.filter(item => item !== key)
+}
```
markgrahamdawson, line 57: Still to do: Should figure out what this value.value is all about :confused:

tests/e2e/specs/tree.cy.js
```diff
@@ -218,7 +218,7 @@ describe('Tree view', () => {
           .contains(name)
```
markgrahamdawson, line 218: Im not sure if this really needs additional tests. All the old tests for GScan - tree and table are passing

src/components/cylc/TaskFilter.vue
```diff
@@ -37,48 +37,20 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-          </div>
-          <span
-            v-if="index === $options.maxVisibleStates"
-            class="text-grey text-caption"
-          >
-            (+{{ localValue.states.length - $options.maxVisibleStates }})
-          </span>
-        </template>
-      </v-select>
+    <TaskFilterSelect :modelValue="localValue" :type="'task state'" :items="$options.allStates"></TaskFilterSelect>
```
MetRonnie, line 40: The conventional Vue style is to wrap if there is more than 1 attribute per tag: https://eslint.vuejs.org/rules/max-attributes-per-line.html (at some point we should probably turn on more `eslint-plugin-vue` rules).

Please could you go through and wrap these as I think the readability benefits
```suggestion
    <TaskFilterSelect
      :modelValue="localValue"
      type="task state"
      :items="$options.allStates"
    />
```

src/components/cylc/gscan/GScan.vue
```diff
@@ -59,22 +59,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-                <v-select
-                  v-model="filters[title]"
-                  :items="$options.allStates[title]"
-                  :label="`Filter by ${title}`"
-                  density="default"
-                  chips
-                  closable-chips
-                  clearable
-                  multiple
-                  class="my-2"
```
MetRonnie, line 75: The loss of this margin is causing display issues

image

src/components/cylc/TaskFilterSelect.vue
```diff
@@ -0,0 +1,57 @@
...
+
+  </v-chip>
+  <span
+      v-if="index === maxVisibleStates"
+      class="text-grey text-caption"
+    >
+      (+{{ localValue.states.length - maxVisibleStates }})
+  </span>
+</template>
+</v-select>
```
MetRonnie, line 34: There is some missing indentation here

oliver-sanders: Yet linting doesn't complain?

MetRonnie: #1536 - it's included in `eslint-plugin-vue` but only if you turn on "strongly-recommended"

src/components/cylc/TaskFilterSelect.vue
```diff
@@ -0,0 +1,57 @@
...
+
+<script setup>
+import { computed } from 'vue'
+import Task from '@/components/cylc/Task.vue'
+import Workflowicon from '@/components/cylc/gscan/WorkflowIcon.vue'
+
+const props = defineProps({
+  modelValue: Object, // { id, states }
+  items: Array,
+  type: Text
```
MetRonnie, line 45: Getting
> [Vue warn]: Invalid prop: type check failed for prop "type". Expected Text, got String with value "task state"

```suggestion
  type: String
```

src/components/cylc/TaskFilter.vue
```diff
@@ -54,7 +58,10 @@ export default {
   },
 
   props: {
-    modelValue: Object // { id, states }
+    modelValue: {
+      type: Object,
+      default: () => ({})
```
oliver-sanders, line 63: If this prop is required, you can set `required: true` rather than using a blank default.

MetRonnie (commented):
I've got some suggestions I've spun out into https://github.com/markgrahamdawson/cylc-ui/pull/1

src/components/cylc/TaskFilterSelect.vue
```diff
@@ -0,0 +1,95 @@
+<template>
+  <v-select
+    :label="`Filter by ${type}`"
+    :data-cy="`filter ${type}`"
+    :items="items"
+    clearable
+    multiple
+    :placeholder="`Filter by ${type}`"
```
MetRonnie, line 8: In some cases we want to have the label and in others we want to have the placeholder. So if we don't specify these props they will be [fallthrough attributes](https://vuejs.org/guide/components/attrs.html) and you can specify them on the `<TaskFilerSelect>` tag instead

src/components/cylc/TaskFilter.vue
```diff
@@ -37,52 +37,31 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-          <span
-            v-if="index === $options.maxVisibleStates"
-            class="text-grey text-caption"
-          >
-            (+{{ localValue.states.length - $options.maxVisibleStates }})
-          </span>
-        </template>
-      </v-select>
+        type="task state"
+        :items="$options.allStates"
```
MetRonnie, line 43: The text field before this uses placeholder not label, so for consistency 
```suggestion
        :items="$options.allStates"
        placeholder="Filter by task state"
```

src/components/cylc/gscan/GScan.vue
```diff
@@ -26,97 +26,97 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+          <v-list>
+            <v-list-item
+              v-for="(items, title) in filters"
+              :key="title"
+            >
+              <TaskFilterSelect
+                v-model="filters[title]"
+                :type="title"
+                :items="$options.allStates[title]"
+                class="my-2"
```
MetRonnie, line 72: To match how it looked previously
```suggestion
                :label="`Filter by ${title}`"
                density="default"
                class="my-2"
```

MetRonnie (approved):

oliver-sanders (approved):
:+1:

PR #1549, opened 2023-11-09
Fix GScan bug when filtering hierarchical workflows by state

MetRonnie:
Follow-up to #1416 which introduced a bug where filtering by workflow state would show all workflows in hierarchies that had a match:

image

**Check List**

- [x] Tests are included 


wxtim (approved):
IMO This is pretty cool:

image

src/components/cylc/tree/Tree.vue
```diff
@@ -168,80 +119,45 @@ export default {
...
+    // Reactively run filtering
+    if (this.nodeFilterFunc) {
+      this.$watch(
+        () => [this.filterState, this.rootChildren],
+        ([active, nodes], [wasActive, oldNodes]) => {
+          if (active) {
+            // Need to wipe cache due to mem leak bug in Vue < 3.4
+            // https://github.com/vuejs/core/issues/9419
+            // TODO: can remove this line after upgrade to Vue 3.4:
+            this.filteredOutNodesCache = new WeakMap()
```
MetRonnie, line 122: vuejs/core#9419

I tested out tree view filtering without this line and found a memory leak with our current Vue version 3.3

I then tested with Vue 3.4.0-alpha.1 and did not get a memory leak

oliver-sanders: Could you put up an issue for this and stick it into the comment to ensure we get round to it.

src/components/cylc/tree/TreeItem.vue
```diff
@@ -257,15 +253,8 @@ export default {
...
 
-  emits: [
-    ...passthroughEvents
-  ],
-
   data () {
     return {
-      active: false,
-      selected: false,
-      filtered: true,
```
oliver-sanders, line 268: FYI: The intention for this component was to support [multiple] selection, hence these (presumably unused) attributes.

We can presumably use the same weakmap for these when we get around to it later?

MetRonnie: Yes I suppose so

src/components/cylc/tree/Tree.vue
MetRonnie: I'm not sure if it's important enough to track in its own issue - creating a new map can't have that much of a performance penalty, can it?

oliver-sanders: No, an issue for the TODO!

oliver-sanders:
Bumping efficiency issues to https://github.com/cylc/cylc-ui/issues/1555

src/components/cylc/tree/Tree.vue
MetRonnie: That's what I'm saying - if we forget to act on this TODO it won't matter too much. But if someone comes across the TODO after the Vue version has been bumped they can address it. I may even get rid of this soon if I can get the computed properties approach working.

oliver-sanders (approved):

wxtim (approved):

Issue #1617, opened 2024-01-08
tree: more efficient expand/collapse icons/buttons


oliver-sanders:
A large proportion of the component mount time for the Tree view comes down to the expand/collapse buttons/icons.

Surprisingly, these take longer to mount than the tree and job icons, see https://github.com/cylc/cylc-ui/issues/1614#issuecomment-1881068722.

Suggest investigating a native implementation of this trival component i.e. one which does not use intermediate Vue components.

Issue #1614, opened 2024-01-04
efficiency: investigate bottleneck


oliver-sanders:
This workflow has proven to be remarkably difficult for the UIS & UI to handle:

```ini
#!Jinja2

{% set members = 10 %}
{% set hours = 100 %}

[scheduler]
    allow implicit tasks = True

[task parameters]
    member = 0..{{members}}
    fcsthr = 0..{{hours}}
  [[templates]]
    member = member%(member)03d
    fcsthr = _fcsthr%(fcsthr)03d

[scheduling]
  initial cycle point = 2000
  runahead limit = P3
  [[xtriggers]]
    start = wall_clock(offset=PT7H15M)
  [[graph]]
    T00,T06,T12,T18 = """
        @start & prune[-PT6H]:finish => prune & purge
        @start => sniffer:ready_<member,fcsthr> => <member,fcsthr>_process? => finish
        <member,fcsthr>_process:fail? => fault
      """

[runtime]
    [[sniffer]]
        [[[outputs]]]
{% for member in range(0, members + 1) %}
    {% for hour in range(0, hours + 1) %}
            ready_member{{ member | pad(3, 0) }}_fcsthr{{ hour | pad(3, 0) }} = {{ member }}{{ hour }}
    {% endfor %}
{% endfor %}
```

For more information see: https://cylc.discourse.group/t/slow-load-of-cylc-workflows-disconnects/823/19

Investigation so far has confirmed:
* The scheduler is **not** a source of delay.
* The UIS chokes on the update for several seconds.
  * During this time, updates to other workflows are suspended
* The UI chokes on the deltas of several seconds.
* The browser takes a couple of seconds to update.

This issue focuses on the UI side of things.

Suggested remediation (UI only, please update with new suggestions):

*

oliver-sanders:
The bulk of the time is being taken in the data store processing the deltas, this should be the first target for improvement. Profiling required to highlight problem areas, given that the table view is only slightly faster to load than the tree view, family tree computation is unlikely to be the cause.

oliver-sanders:
Early profiling results

<details>
<summary>Methodology</summary>

Firefox profiler in custom mode with a a 1ms sample rate.

The workflow from the OP with `hours` turned down to 20

</details>

Results:

* 5.68 seconds of scripting time.
  * 1.5s seconds of which is accounted for by the UPDATE_DELTAS call chain (i.e. workflow data store stuff)
[...]

PR #1619, opened 2024-01-08
tree: more efficient expand/collapse buttons

oliver-sanders:
* Closes https://github.com/cylc/cylc-ui/issues/1617
* There is one expand/collapse button per tree item.
* There are many tree items in the tree view.
* We were using two Vuetify components to implement the expand/collapse button, these components are expensive to mount.
* This PR switches to using native components to implement the expand/collapse button.
* This saves ~2/5 of mount+render time.

To test:
* Start the workflow from https://github.com/cylc/cylc-ui/issues/1614 with `hours` set to `20` in paused mode.
* Open the tree view on this workflow and wait for the UI to stabilise.
* Time how long it takes to open a second tree view.

Results:
* Before: ~5s
* After: ~2.5s

Caveats:
* This approach does not support standard Vuetify things like the mouseover "ripple". Personally I see this as more of a pro than a con but it's a change either way.
* This approach does not re-implement the rotate animation on state change. This would be fairly easy to do with CSS, personally I'm happier without, but it's another change.

**Check List**

### DIFF
--- a/src/components/cylc/tree/TreeItem.vue
+++ b/src/components/cylc/tree/TreeItem.vue
@@ -18,7 +18,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 <template>
   <div
     v-show="!filteredOutNodesCache.get(node)"
-    class="treeitem"
+    class="c-treeitem"
   >
     <div
       class="node d-flex align-center"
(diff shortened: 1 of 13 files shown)


wxtim (approved):
- [x] Read
- [x] Profiled

src/styles/cylc/_tree.scss
```diff
@@ -55,14 +55,20 @@ $icon-size: 1.2em;
...
   display:inline-block;
   vertical-align:middle;
   width: 100%;
   .node {
     line-height: $line-height;
     display: block;
 
+    .node-expand-collapse-button {
+      width: 1.3em;
+      height: 1.3em;
```
MetRonnie, line 68: Alignment is slightly off

image

This fixes it:
```suggestion
      width: 1.5em;
      height: 1.5em;
      margin: 0 2px;
```

src/components/cylc/tree/TreeItem.vue
```diff
@@ -18,29 +18,26 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
       :class="nodeClass"
       :style="nodeStyle"
     >
       <!-- the node's left icon; used for expand/collapse -->
-      <v-btn
+      <svg
         v-if="renderExpandCollapseBtn"
         aria-label="Expand/collapse"
-        aria-hidden="false"
         class="node-expand-collapse-button flex-shrink-0"
```
MetRonnie, line 32: This automatically adds the rotation transition back in, which I feel is a useful bit of visual feedback
```suggestion
        class="node-expand-collapse-button flex-shrink-0 v-btn"
```

markgrahamdawson (approved):

src/components/cylc/tree/TreeItem.vue
oliver-sanders: Not sure it's a good idea to apply v-btn styling to something which isn't even a `<button />`.

I'm not overly keen on the animation anyway, but would be happy to add this with CSS if desired.

MetRonnie (commented):
I have fixed the indentation using relative units at https://github.com/oliver-sanders/cylc-ui/pull/111

(If you combine #1642 with it, everything aligns properly)

MetRonnie (approved):

PR #1623, opened 2024-01-09
store: more efficient family tree building

oliver-sanders:
* Reduce the overheads of `applyInheritance` for families with lots of children.
* Use a map for O(1) lookup times when checking whether child nodes are present in the store.
* This should change scaling from `O(nm)` to `O(m)` where `n` is the number of tasks in a family when created and `m` is the number of children after an update (n~=m for most cases).
* Partially addresses https://github.com/cylc/cylc-ui/issues/1614
* For the reduced example (`hours = 20`) this reduces the time taken by `applyInheritance` by ~85% from ~0.6s to <0.1s (though I'm sure these times will be subject to much jitter).
* Due to the non-linear scaling, I think this will make a larger difference to the full example.

Note, if you have no families in your workflow, then all tasks inherit from `root`.

Before:

Screenshot from 2024-01-09 16-04-47

After:

Screenshot from 2024-01-09 16-04-59

To profile for yourself:
* Install and start the reduced example from #1614 in paused mode.
* Open the GUI and start the profiler.
* Open the workflow (assumes the Tree view is the default view).
* Stop the profiler when the tree view is populated.
* In the profiler results, dig for `applyInheritance`.
* Checkout this branch and repeat.

**Check List**

### DIFF
--- a/src/store/workflows.module.js
+++ b/src/store/workflows.module.js
@@ -134,17 +134,6 @@ function getIndex (state, id) {
   return state.cylcTree.$index[id]
 }
 
-/* Return true if a node has at least one child.
- *
- * Defaults to looking in node.children, set childAttr to use a different
- * tree (e.g. familyTree).
- */
-function hasChild (node, id, attr = 'id', childAttr = 'children') {
-  return node[childAttr].some(
-    item => item[attr] === id
-  )
-}
-
 /* Add a child node under a parent Node */
 function addChild (parentNode, childNode) {
   // determine which list to add this node to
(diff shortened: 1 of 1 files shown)


src/store/workflows.module.js
```diff
@@ -276,10 +265,11 @@ function cleanParents (state, node) {
  */
 function applyInheritance (state, node) {
   if (node.type === 'family' && node.node.childTasks) {
-    // add new tasks
+    // build a mapping of {childID: childNode} for faster lookup
+    const childIDs = node.children.reduce((map, obj) => { map[obj.id] = obj; return map }, {})
```
oliver-sanders, line 269: Similary to Python dictionaries, object lookups are `O(1)` whereas iterating over the array is `O(N)`.

wxtim:
> In the profiler results, dig for `applyInheritance`.

Surely you'll need to do `yarn run build:watch --no-minify` to enable this?

wxtim (commented):
- [x] Read
- [x] Checked profiling with Firefox tools:

### Profiling

Using `cylc vip bugs/bottleneck -s 'hours=X':

image

oliver-sanders:
Nice graph, attempting a highly scientific extrapolation of that for ``{% set hours = 100 %}`` gives an answer of ~10s!

oliver-sanders:
before - after ~= 10s

markgrahamdawson (approved):

wxtim (approved):

wxtim:
Sorry @oliver-sanders  - I was so busy making pretty graphs that I forgot to mark it approved.

src/store/workflows.module.js
kinow: Bravo! :clap:

PR #1631, opened 2024-01-17
data store: remove redundant safety check

oliver-sanders:
* This was checking whether a node already exists in the tree before creating a new one.
* However, the `$index` already serves this function (nodes will be retrieved from the index if they already exist) so this check was redundant.
* This reduces the cost of the `createTreeNode` routine when the number of sibling elements is large.

Coverage suggests the `return` statement was never hit, logically, it shouldn't be possible (except, possibly in the event of traceback).

For the example in this issue https://github.com/cylc/cylc-ui/issues/1614 with `{% set hours = 20 %}` this reduces the cost of the `createTreeNode` routine by ~2/3.

The `some` operation should  be `O(n)`, but it gets called once for each of `n` items being added to the tree so it ends up `O(n^2)` which causes heavy load for high `n`.

**Check List**

### DIFF
--- a/src/store/workflows.module.js
+++ b/src/store/workflows.module.js
@@ -310,12 +310,7 @@ function update (state, updatedData) {
 
   // create a new tree item
   let treeParent
-  const ret = createTreeNode(state, id, tokens, updatedData)
-  if (!ret) {
-    // node already exists, nothing more to do here
-    return
-  }
-  [treeParent, treeItem] = ret
+  [treeParent, treeItem] = createTreeNode(state, id, tokens, updatedData)
 
   // add the new item to the tree
   addChild(treeParent, treeItem)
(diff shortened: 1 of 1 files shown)


wxtim (approved):

MetRonnie (commented):
Looks like https://github.com/cylc/cylc-ui/blob/42f72c81759b5cac38fadc08754850a6f418fdfd/src/store/workflows.module.js#L314-L317 can also be deleted

```diff
diff --git a/src/store/workflows.module.js b/src/store/workflows.module.js
index 154f4e95..76673cdd 100644
--- a/src/store/workflows.module.js
+++ b/src/store/workflows.module.js
@@ -311,10 +311,5 @@ function update (state, updatedData) {
   // create a new tree item
   let treeParent
-  const ret = createTreeNode(state, id, tokens, updatedData)
-  if (!ret) {
-    // node already exists, nothing more to do here
-    return
-  }
-  [treeParent, treeItem] = ret
+  [treeParent, treeItem] = createTreeNode(state, id, tokens, updatedData)
 
   // add the new item to the tree
```

MetRonnie (approved):

PR #1655, opened 2024-01-31
Update vueuse & ensure composables are only used inside `setup`

MetRonnie:
Supersedes / unblocks #1620

- Ensure the default view loaded from local storage is valid else fall back to tree view

- #1620 

- Ensure composables are only used synchronously[^1] inside `setup`

    > The vue docs say:
    > 
    > ...
    > 
    > > [Usage restrictions](https://vuejs.org/guide/reusability/composables.html#usage-restrictions)
    > > Composables should only be called **synchronously in <script setup> or the setup() hook**.

    https://github.com/vueuse/vueuse/issues/2608#issuecomment-1464196417

[^1]: Synchronous means not inside callbacks of e.g. computed properties, lifecycle hooks etc.

**Check List**

- [x] Tests are included
- [x] No changelog entry needed as not user-facing


src/components/cylc/analysis/BoxPlot.vue
```diff
@@ -152,71 +222,6 @@ export default {
         { title: `Max ${this.timingOption} time`, value: `max${upperFirst(this.timingOption)}Time` },
       ]
     },
-
-    chartOptions () {
-      return {
-        chart: {
-          animations: {
-            enabled: useReducedAnimation().value ? false : this.animate,
```
oliver-sanders, line 160: Do we really have to shift this whole block just to handle this one usage?

markgrahamdawson:
https://vueschool.io/lessons/calling-composables-in-the-wrong-place

src/components/cylc/analysis/BoxPlot.vue
MetRonnie: Yes

src/components/cylc/analysis/BoxPlot.vue
```diff
@@ -97,6 +98,75 @@ export default {
     },
   },
 
+  setup (props) {
+    const reducedAnimation = useReducedAnimation()
+
+    const chartOptions = computed(() => ({
```
oliver-sanders, line 104: @cylc/data-workflows 

Heads up, this is just a different way of defining a `computed` property which is more kosher with the use of "composables" (aka things that start with the prefix `use`). No further changes required.

src/components/cylc/analysis/BoxPlot.vue
oliver-sanders: Dang.

oliver-sanders (approved):

markgrahamdawson (approved):

Issue #662, opened 2021-05-04
tab switching (preserve tab layout, configure defaults & share links)


oliver-sanders:
> [!NOTE]
> **Definition**: "Workspace"
>
> The part of the UI within the Lumino instance. Basically everything that isn't GScan in the "default" view.
>
> A workspace could be for:
> * An installed workflow (only workspace implemented at the moment).
> * A source workflow *(soon)*.
> * The user dashboard *(one day)*.

The aims the GScan tabbing system are:

* To simplify workspace selection/switching.
* To make the user's workflow between configuring > running > reviewing easy.
* To make it easy to work in a single browser tab.

Here are four steps to achieve these goals in order of priority.

1. Allow the current tab layout to be saved (urgent).
   * Add a component to the workflow toolbar which allows the current tab layout to saved as the default.
   * Apply the default layout when switching workflows.
   * NOTE: configuring multiple view layouts is blocked by #1278
1. Save the layout last used by a workspace (important). #1664
   * When the user navigates away from a workflow, save the tab layout.
   * When the user navigates back to the workflow, load the tab layout.
   * When the user navigates to a new run of the same workflow, copy the layout from the last run of the workflow.
1. Save view state (useful). _[In progress](https://github.com/cylc/cylc-ui/pulls?q=is%3Apr+initialoptions)_
   * Some aspects of the view state should be preserved with the layout.
   * E.G. Any task state filters should be preserved.
   * E.G. The "log" view should remember what log file it is looking at.
   * These things need to be saved as part of the layout information so that we can re-open the layout as it was before.
1. Named layouts (nice-to-have).
   * Allow users to save "named" layouts.
   * Allow users to re-configure the workspace to a previously saved layout.

oliver-sanders:
Long run we should provide:

* A way to "save" the current layout.
* A mechanism for dumping the view's data into the layout (e.g. if the view is filtered by task state, that should be included in the layout)
* A way to "load" a pre-saved layout.
* A configuration for setting the default layout.
* Possibly an additional configuration on each layout to allow setting it as the default layout based on the workflow hierarchy?

E.G. A layout menu icon which provides a "save current layout" button and lists all pre-saved layouts for quick switching.

MetRonnie:
IMO we should split the "Serialise lumino state into local storage" part off into a 1.4.0 issue?

oliver-sanders:
[...]

PR #1664, opened 2024-02-07
Save & restore workspace layout on navigation

### LINKED ISSUE #1278: subscriptions: prevent multiple subscribe/unsubscribe events
Currently we start subscriptions on a per-component basis, which creates the potential for problems where multiple subscribing components are involved.

E.G. if there are two components the sequence of events might look like this:

1. Component 1 initialises and registers its subscription.
2. The subscription is started.
3. Component 2 initialises and registers its subscription.
4. The subscription is stopped and re-started with the recomputed query.

We need to change this order of events to:

1. Component 1 initialises and registers its subscription.
2. Component 2 initialises and registers its subscription.
3. The subscription is started.

This bug doesn't bite at the moment because we only initialise the workspace with one component, however with #662 that will change.

See https://github.com/cylc/cylc-ui/pull/1249#discussion_r1119106019_ for more details

MetRonnie:
Closes #1278
Partially addresses #662

> 2. Save the layout last used by a workspace (important).
>     * When the user navigates away from a workflow, save the tab layout.
>     * When the user navigates back to the workflow, load the tab layout.
>     * When the user navigates to a new run of the same workflow, copy the layout from the last run of the workflow.

The layout is saved in the vuex store so only persists for the browser session. Ongoing investigation into saving it in local storage.

demo14

**Check List**

- [x] Tests are included
- [x] `CHANGES.md` entry included
- [ ] Changelog entry added in cylc-doc

### DIFF
--- a/src/components/cylc/workflow/Lumino.vue
+++ b/src/components/cylc/workflow/Lumino.vue
@@ -119,13 +119,13 @@ onMounted(() => {
   Widget.attach(boxPanel, mainDiv.value)
   // Watch for resize of the main element to trigger relayout:
   resizeObserver.observe(mainDiv.value)
-  $eventBus.on('add-view', addWidget)
+  $eventBus.on('add-view', addView)
   getLayout(props.workflowName)
 })
 
 onBeforeUnmount(() => {
   resizeObserver.disconnect()
-  $eventBus.off('add-view', addWidget)
+  $eventBus.off('add-view', addView)
   saveLayout()
   // Register with Lumino that the dock panel is no longer used,
   // otherwise uncaught errors can occur when restoring layout
@@ -138,7 +138,7 @@ onBeforeUnmount(() => {
  * @param {AddViewEvent} view
  * @param {boolean} onTop
  */
-const addWidget = (view, onTop = true) => {
+const addView = (view, onTop = true) => {
   const id = uniqueId('widget')
   const luminoWidget = new LuminoWidget(id, startCase(view.name), /* closable */ true)
   dockPanel.addWidget(luminoWidget, { mode: 'tab-after' })
@@ -155,7 +155,7 @@ const addWidget = (view, onTop = true) => {
 /**
  * Remove all the widgets present in the DockPanel.
  */
-const removeAllWidgets = () => {
+const closeAllViews = () => {
   for (const widget of Array.from(dockPanel.widgets())) {
     widget.close()
   }
@@ -168,7 +168,7 @@ const removeAllWidgets = () => {
  * @param {string} workflowName
  */
 const getLayout = (workflowName) => {
-  restoreLayout(workflowName) || addWidget({ name: defaultView.value })
+  restoreLayout(workflowName) || addView({ name: defaultView.value })
 }
 
 /**
@@ -212,7 +212,7 @@ const restoreLayout = (workflowName) => {
  */
 const changeLayout = (workflowName) => {
   saveLayout()
-  removeAllWidgets()
+  closeAllViews()
   // Wait if necessary for the workflowName prop to be updated to the new value:
   when(
     () => props.workflowName === workflowName,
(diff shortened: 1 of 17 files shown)


src/router/index.js
```diff
@@ -76,9 +76,7 @@ router.beforeEach(async (to, from) => {
   if (!store.state.user.user.permissions.includes('read') && to.name !== 'noAuth') {
     return { name: 'noAuth' }
   }
-})
 
-router.beforeResolve((to, from) => {
```
oliver-sanders, line 81: Interesting, I would like to understand this change.

src/store/app.module.js
```diff
@@ -27,6 +30,16 @@ const mutations = {
   setTitle (state, title) {
     state.title = title
   },
+  saveLayout (state, { workflowName, layout, views }) {
+    /* NOTE: use markRaw to prevent proxying of the Lumino layout in particular.
+    It is not necessary for this saved state to be reactive, and moreover
+    proxying the layout breaks some parts of the 3rd party Lumino backend. */
+    state.workspaceLayouts.set(workflowName, markRaw({ layout, views }))
+    if (state.workspaceLayouts.size > 10) {
```
oliver-sanders, line 38: Would be worth building a really complex layout and finding out how heavy it is.

I'm hoping that we can safely raise this number.

oliver-sanders (commented):
Slick.

src/router/index.js
MetRonnie: Discussed in person. When navigating from one workflow to another, the subscription would start using the old workflow name for a short while and then update to the new workflow name. Now the subscription starts using the new workflow name

MetRonnie:
Drafting while I see if I can fix #1278 before we merge this

MetRonnie:
@oliver-sanders and @markgrahamdawson I've opened a secondary PR against this one that modifies the log view's `intialOptions` prop to allow it to store the view state. Mark, this approach can be used on the other views but isn't implemented yet. https://github.com/MetRonnie/cylc-ui/pull/7

src/components/cylc/workflow/Lumino.vue
```diff
@@ -15,187 +15,243 @@ You should have received a copy of the GNU General Public License
...
+const views = ref(new Map())
+
+const defaultView = useDefaultView()
+
+// create a box panel, which holds the dock panel, and controls its layout
+const boxPanel = new BoxPanel({ direction: 'left-to-right', spacing: 0 })
+// create dock panel, which holds the widgets
+const dockPanel = new DockPanel()
+boxPanel.addWidget(dockPanel)
+BoxPanel.setStretch(dockPanel, 1)
```
oliver-sanders, line 111: Is that meant to be `BoxPanel.setStretch` with a capital `B`?

src/components/cylc/workflow/Lumino.vue
```diff
@@ -15,187 +15,243 @@ You should have received a copy of the GNU General Public License
...
-     */
-    syncWidgets (newVal, oldVal) {
-      for (const [id, item] of Object.entries(newVal)) {
-        if (!(id in oldVal)) {
-          this.addWidget(id, item.view)
-        }
+/**
+ * Remove all the widgets present in the DockPanel.
+ */
+const closeAllViews = () => {
```
oliver-sanders, line 158: IMO:

```js
function foo(bar) { ... }
```

Is cleaner than

```js
const foo = (bar) => { ... }
```

What's the motive for the `const` approach? Does it do something different?

src/components/cylc/workflow/Lumino.vue
MetRonnie: Yes, same as before

src/components/cylc/workflow/Lumino.vue
MetRonnie: No preference from me

oliver-sanders: So the only functional difference is the implicit `this` variable?

MetRonnie: Well, also declarations (`function foo () {}`) are hoisted but expressions (`const foo = () => {}`) are not.

PR #1672, opened 2024-02-09
Save view state not between workflows

markgrahamdawson:
Supersedes https://github.com/cylc/cylc-ui/pull/1669 and https://github.com/cylc/cylc-ui/pull/1246

I set up a composable for defining and managing all the view options.
The state persists when switching between workflows

Still to do - 
Set up a watcher that saves to local storage

**Check List**

### DIFF
--- a/src/composables/useViewState.js
+++ b/src/composables/useViewState.js
@@ -1,0 +1,17 @@
+import { ref } from 'vue'
+
+// options ref needs to be defined outside of export function to ensure
+// state persists between components
+const options = ref({
+  tree: {},
+  table: {},
+  graph: {},
+  log: {},
+  analysis: {}
+})
+
+export default function useViewState () {
+  return {
+    options,
+  }
+}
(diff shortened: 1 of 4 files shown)


src/views/Tree.vue
```diff
@@ -216,14 +218,57 @@ export default {
...
+  },
+
   data: () => ({
-    expandAll: null,
-    tasksFilter: {
-      id: null,
-      states: null,
-    },
+    options: {},
   }),
```
markgrahamdawson, line 234: I might be able to move all this into the composable?
https://vuejs.org/guide/reusability/composables

src/composables/useViewState.js
```diff
@@ -0,0 +1,29 @@
...
+// optionsArray ref needs to be defined outside of export function to ensure
+// state persists between workflows
+
+const optionsArray = ref([])
+
+export default function useViewState () {
+  const options = {
+    tree: {
+      tasksFilter: { type: Object, value: { id: null, states: null } },
+      expandAll: { type: Array, value: [] },
```
oliver-sanders, line 12: Ideally the definition of these options would lie in the components they come from rather than this compostable.

We can think of views as cylc-ui plugins and this compostable as part of the support infrastructure for said plugins. The code infrastructure shouldn't need to have any plugin logic hardcoded within it.

Issue #1200, opened 2023-01-10
slow initial connection


oliver-sanders:
Sometimes it takes a long time for the UI to connect with the backend.

It seems to get there in the end but it often takes in excess of 45 seconds to establish the subscription connection. The issue occurs randomly when the page is loaded and causes the no-connection error box hangs around.

Whilst we are waiting for the connection this message appears multiple times (Firefox):

```
 Firefox can’t establish a connection to the server at ws://localhost:8888/cylc/subscriptions.
```

Which suggests multiple failed attempts at connection, however, after jamming some logging into the UIS I can see no evidence of this. The UIS responds quickly once it receives the request.

It's not clear what's causing this, however, it is worth noting that we are using a deprecated framework for subscriptions, we should try moving to something more modern and see if the problem goes away by itself - #1199.

MetRonnie:
I don't think I have ever seen this on Chrome

oliver-sanders:
I'm assuming this is a general issue exacerbated by a laggy platform, you can see multiple connection attempts in the console.

hjoliver:
(What's VDI - virtual desktop?)

oliver-sanders:
I've done some investigation into this but have come out empty handed.

I've [also tried migrating](https://github.com/cylc/cylc-uiserver/pull/482) from `graphql-transport-ws` to `graphql-ws`, however, this did not fix the issue.

So far the error has only been reported in RedHat builds of Firefox where it presents like so:

```
 Firefox can’t establish a connection to the server at ws://localhost:8888/cylc/subscriptions.
```

This error will persist across multiple retries, it may last for over one minute before clearing.

This happens randomly on page load, sometimes it connects straight away.

kinow:
P.s. I used to see this when developing the UI. Never investigated the cause though. Sometimes ffox would keep showing the red icon at the top, showing that I was offline and there would be those js messages.

Other times it was after I opened the browser the first time. It happened with just that five workflow, or with multiple workflows (i.e probably not the amount of data).

However, when I used that fake uiserver (offline mode or random tests in python + pre made request/response) I would not get that.

So my guess would be something between browser, uis, and first xhr and/or wss request. Maybe setting a debugger in some parts of uis graphene or web request handling code that could be busy (handling an old request, waiting for some background task?) would replicate the issue...

oliver-sanders:
> Websocket ssl connection can be very slow to connect in Firefox
>
[...]

PR #1684, opened 2024-02-21
Increase initial WS timeout to try to mitigate slow connection problem

MetRonnie:
`subscriptions-transport-ws` uses [an exponential backoff function](https://github.com/apollographql/subscriptions-transport-ws/blob/36f3f6f780acc1a458b768db13fd39c65e5e6518/src/client.ts#L363) to cancel and retry the WebSocket connection when it first connects. The default initial timeout is 1s - https://github.com/apollographql/subscriptions-transport-ws/blob/master/src/defaults.ts#L1

I've increased it to 3s here to try and mitigate the slow initial connection problem #1200, though according to Oliver that problem may be caused by the connection closing unexpectedly (which triggers the WebSocket's `onclose`) rather than timing out (which doesn't). It's worth a shot anyway.

**Check List**

- [ ] `CHANGES.md` entry not included as we don't know if this fixes the problem

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,6 @@
     "eslint-plugin-vuetify": "2.1.1",
     "express": "4.18.2",
     "express-ws": "5.0.2",
-    "graphql-ws": "5.15.0",
     "istanbul-lib-coverage": "3.2.2",
     "jsdom": "24.0.0",
     "json-server": "0.17.4",
--- a/renovate.json
+++ b/renovate.json
@@ -32,7 +32,6 @@
         "concurrently",
         "express",
         "express-ws",
-        "graphql-ws",
         "json-server",
         "nodemon"
       ],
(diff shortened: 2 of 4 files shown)


oliver-sanders:
I have reproduced the issue in #1200 with a similar diff (see https://github.com/cylc/cylc-ui/issues/1200#issuecomment-1818842073), so I don't think this is a solution to that problem.

However, this may be a solution to an asymptomatic issue which you can sometimes see evidence of in the console where the first couple of connection attempts fail. This could be because the server is unable to send a response within the 1s window forcing timeout and reconnect.

oliver-sanders:
Tested this branch against master looking for evidence in the console of failed connection events. These show up like so:

Screenshot from 2024-02-26 16-35-42

I couldn't replicate #1200 on either branch after 10 tries. On master I save failures 2/10 times, on this branch 0/10 times.

oliver-sanders (approved):

kinow:
Ah, you kept the logs of when that happened, @oliver-sanders :point_up_2: 

I couldn't find any old screenshots from when this issue happened, but I remember there was something in the JS console. Your screenshot above might confirm a suspicious I had about this bug. GOing to comment in the main issue, linking to this one.

PR #1688, opened 2024-02-26
Log view: use `initialOptions` to save & restore view state on navigation

MetRonnie:
Follow-up to #1664
Moved from https://github.com/MetRonnie/cylc-ui/pull/7

The view state (filters, inputs, toggles etc.) can now be saved with the Lumino tab layout by using a prop called `initialOptions`. 

This is only implemented for the log view for now.

**Check List**

- [x] Tests are included
- [x] No changelog needed as entry for #1664 covers this
- [x] No docs PR needed


src/views/Log.vue
```diff
@@ -315,20 +371,23 @@ export default {
...
+      async ({ id }, old) => {
+        // update the query when the id or file change
+        this.updateQuery()
+        // refresh the file list when the id changes
+        if (id !== old?.id) {
+          await this.updateLogFileList()
+        }
+      },
+      { immediate: true }
+    )
```
MetRonnie, line 390: Rejigged the watchers for `file` and `id` to prevent race conditions. The usual order of events now is (for the example of opening the workflow log):
1. `id` gets set for the first time to `'~user/workflow'`. `file` is nullish.
    - `updateQuery()` gets called because the `id` changed, but because `file` is nullish it doesn't really do anything https://github.com/cylc/cylc-ui/blob/58f5d96f39f1e9968daa59cdfb6d2fd0c74ef84f/src/views/Log.vue#L429-L431
    - `updateLogFileList()` gets called because the `id` changed
 
1. `file` has been set to the default scheduler log by `updateLogFileList()`
    - `updateQuery()` gets called again because `file` changed, and causes the log file to display
    - `updateLogFileList()` doesn't get called because `id` is unchanged

src/views/Log.vue
```diff
@@ -398,36 +460,22 @@ export default {
...
-        logFiles = result.data.logFiles.files
-      } else {
-        logFiles = []
-      }
+      const logFiles = result.data.logFiles?.files ?? []
 
       // reset the file if it is not present in the new selection
       if (reset) {
-        if (this.file && !(this.file in logFiles)) {
+        if (this.file && !logFiles.includes(this.file)) {
```
MetRonnie, line 473: `x in arr` does not do what you think it does!

src/views/initialOptions.js
```diff
@@ -0,0 +1,62 @@
...
+  watch(
+    _ref,
+    (val, old) => emit(
+      updateInitialOptionsEvent,
+      { ...props.initialOptions, [name]: val }
+    ),
+    { immediate: true, deep: true }
+  )
+  return _ref
+}
```
MetRonnie, line 62: This is similar to using a writeable computed property e.g.
```js
computed({
  get: () => props.initialOptions[name],
  set: (val) => emit(
    updateInitialOptionsEvent,
    { ...props.initialOptions, [name]: val }
  )
})
```
but that seemed to lead to the ref value not updating until the next tick? Maybe the watcher approach works because it is backed up by a concrete ref? Not sure

src/views/Log.vue
```diff
@@ -264,34 +295,59 @@ export default {
...
+  ],
+
   props: {
-    initialOptions: {
-      type: Object,
-      required: false,
-      default: () => {}
-    }
+    initialOptions,
   },
```
markgrahamdawson, line 304: None of the other views have an initialOptions prop - is this a solution that can be rolled out for table, tree and graph views as well?

MetRonnie: Yes, the `initialOptions` prop can be added to them to allow storing their view states

src/views/Log.vue
```diff
@@ -264,34 +295,59 @@ export default {
     }
   },
 
+  emits: [
+    updateInitialOptionsEvent,
+  ],
+
```
markgrahamdawson, line 301: in the other views I think the initialOptions logic is inside src/components/cylc/`view-name`/`view-name`.vue
Should this be the case for log view as well for consistency?

MetRonnie: It is the views (in `src/views/`) that we mount in tabs in the workspace, so these are where we want the `initialOptions` prop to live.

You have correctly pointed out that in some cases, the component (in `src/components`) contains the state we want to save per view. We might need to refactor those to extract the state into the view, which would be passed into the component as a prop (probably using `v-model`).

oliver-sanders: On the split between `src/views` and `src/components`. IMO:

Views are cylc-ui plugins, the view code should contain all the boiler plate for the plugin, e.g. the views name & icon, the logic for subscription handling, mappings onto the data store, user facing options and their defaults.

Components are generic reusable items that could, conceivably, find application in multiple views. They should contain the logic for their own rendering but be as ignorant as possible of the cylc-ui application itself (but will have to make assumptions about the format of the data they are provided).

src/views/initialOptions.js
oliver-sanders, line 1: Should move this to `src/composables`.

oliver-sanders (approved):
LGTM, tested the debounce and preservation of initial options.

src/views/initialOptions.js
MetRonnie: I originally put it there but then it turned out out of the 3 exports, 1 is a composable, 1 is a prop and 1 is an event... And only components in this sub directory should be using this. 🤷‍♂️ Still can move it back there if you think it's a better place?

oliver-sanders: There are other utilities that only views use so this isn't a first. Think it makes more sense to put utils in the `composables` or `utils` dirs to keep things tidy.

tests/e2e/specs/log.cy.js
```diff
@@ -134,3 +134,52 @@ describe('Log View', () => {
...
+      .click()
+      .get('.c-log [data-cy=job-id-input] input').as('jobIDInput')
+      .type(jobID)
+      .get('[data-cy=log-viewer]')
+      .should('be.visible')
+      .get('.c-log [data-cy=file-input] input').as('fileInput')
+      .invoke('val')
+      .should('eq', jobFile)
+    // Navigate away
+    cy.visit('/#/workspace/two')
```
markgrahamdawson, line 173: Would we expect the state to persist when changing the route like this?

MetRonnie: The state is stored in the `vuex` store, which persists for the lifetime of the browser session until the user refreshes the page

MetRonnie:
If the commits are similar enough that it makes sense to squash, or there is only 1 commit, or the commits are not tidy. Otherwise in this case I think the granularity of the commits makes sense

PR #1695, opened 2024-03-06
Renovate: schedule lock file maintenance

MetRonnie:
> You can use `lockFileMaintenance` to refresh lock files to keep them up-to-date.
>
> When Renovate performs `lockFileMaintenance` it deletes the lock file and runs the relevant package manager. That package manager creates a new lock file, where all dependency versions are updated to the latest version. Renovate then commits that lock file to the update branch and creates the lock file update PR.

https://docs.renovatebot.com/configuration-options/#lockfilemaintenance

**Check List**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,7 +3,15 @@ name: CI
 on:
   push:
     branches: [master]
+    paths-ignore:
+      - 'renovate.json'
+      - '.github/dependabot.yml'
+      - '**.md'
   pull_request:
+    paths-ignore:
+      - 'renovate.json'
+      - '.github/dependabot.yml'
+      - '**.md'
   workflow_dispatch:
 
 concurrency:
--- a/renovate.json
+++ b/renovate.json
@@ -109,6 +109,11 @@
       "addLabels": ["analysis view"]
     }
   ],
+  "lockFileMaintenance": {
+    "enabled": true,
+    "automerge": true,
+    "schedule": ["on the last day of the month"]
+  },
   "vulnerabilityAlerts": {
     "addLabels": ["security"]
   }


oliver-sanders:
I'm not sure deleting and re-creating the lock file is a good idea, for starters it could revert fixes made in other PRs or jump versions we know to be problematic.

MetRonnie:
I was thinking about this for a while

> revert fixes made in other PRs

Unless the fix set a package resolution to a version outside of the `^` or `~` range, this won't happen. And I had a look through recent dependabot security fixes and couldn't find such an example.

> jump versions we know to be problematic

Not sure what you mean by this.

Note that regenerating the lockfile will only affect indirect/transitive dependencies as we pin our direct dependencies (no `^` or `~`).

oliver-sanders:
> couldn't find such an example.

We may fix a transitive dependency via a change to the yarn.lock file, e.g. to avoid a security issue.

These fixes apply only to the yarn.lock file so will be lost when it is deleted.

MetRonnie:
Old issue but we have a number of dependabot PRs that no-one is reviewing. This would be avoided by lockfile maintenance 

image

oliver-sanders:
This issue would not be *avoided* by lockfile maintenance, the changes would simply be grouped together along with others not on the list (note we could get Renovate to group all of these into one PR - but there still wouldn't be anyone to review it).

If you would like to come back to this, feel free. Ripping up the "whole" lockfile is a bit dubious (because it wipes out all of the careful management done to that point) but it could be completely reasonable to do it occasionally (e.g, after a release?).

MetRonnie:
Lockfile maintenance will automatically update all of these Dependabot-flagged dependencies.

Dependabot cannot update a dependency outside of its semver range, so lockfile maintenance will not undo any Dependabot changes.

To take a specific example:

image

```
> yarn why -R qs
└─ cylc-ui@workspace:.
   ├─ cypress@npm:15.18.0 (via npm:15.18.0)
   │  └─ @cypress/request@npm:4.0.1 (via npm:^4.0.0)
   │     └─ qs@npm:6.15.2 (via npm:^6.15.2)
   ├─ express@npm:5.2.1 (via npm:5.2.1)
   │  ├─ body-parser@npm:2.2.1 (via npm:^2.2.1)
   │  │  └─ qs@npm:6.15.2 (via npm:^6.14.0)
   │  └─ qs@npm:6.15.2 (via npm:^6.14.0)
   └─ json-server@npm:0.17.4 (via npm:0.17.4)
      ├─ body-parser@npm:1.20.3 (via npm:^1.19.0)
      │  └─ qs@npm:6.13.0 (via npm:6.13.0)
      └─ express@npm:4.21.2 (via npm:^4.17.1)
         ├─ body-parser@npm:1.20.3 (via npm:1.20.3)
         └─ qs@npm:6.13.0 (via npm:6.13.0)
```

Dependabot cannot update the version of `qs` used by `body-parser` and `express` because both pin to `6.13.0`.

In short, the "careful management" problem only exists _because_ we don't do regular lockfile maintenance.

Grouping all of these updates into a lockfile maintenance PR would ease the review burden, and if using automerge, remove it entirely.

oliver-sanders:
I think that's the same comment from a different perspective. We can either upgrade everything in one go, or in smaller batches.

Either way the same upgrade would have to be performed and reviewed in the same way, this would not remove the problem, just bunch it.

There may well be value in bunching - when I do review UI dep changes, I batch them - and I raise a `yarn up -R` PR at a *reasonably* appropriate point in the dev cycle - time permitting.

PR #1705, opened 2024-03-11
TreeView: Use `initialOptions` prop to save view state

markgrahamdawson:
These changes build on the work undertaken to save & restore workspace layout on navigation https://github.com/cylc/cylc-ui/pull/1664

Which is part of a larger issue https://github.com/cylc/cylc-ui/issues/662

There is an equivalent pull request for the logview https://github.com/cylc/cylc-ui/pull/1688

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [x] Tests are included (or explain why tests are not needed).
- [x] `CHANGES.md` entry included if this is a change that can affect users
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

PR #1711, opened 2024-03-15
TableView: Use initialOptions prop to save view state

markgrahamdawson:
These changes build on the work undertaken to save & restore workspace layout on navigation https://github.com/cylc/cylc-ui/pull/1664

Which is part of a larger issue https://github.com/cylc/cylc-ui/issues/662

There was an equivalent pull request for the log view https://github.com/cylc/cylc-ui/pull/1688
There is an equivalent pull request for the tree view which this builds on: https://github.com/cylc/cylc-ui/pull/1705

**Check List**


tests/e2e/specs/table.cy.js
```diff
@@ -124,3 +124,54 @@ describe('Table view', () => {
...
   })
 })
+
+function addView (view) {
+  cy.get('[data-cy=add-view-btn]').click()
+  cy.get(`#toolbar-add-${view}-view`).click()
+    // wait for menu to close
+    .should('not.be.exist')
+}
+
```
markgrahamdawson, line 134: I started a new `describe.` as - i think - the...
```
beforeEach(() => {
    cy.visit('/#/table/one')
  })
```
associated with the last one was stopping me from navigating away and back again

src/components/cylc/table/Table.vue
```diff
@@ -174,13 +178,18 @@ export default {
           order: cyclePointsOrderDesc.value ? 'desc' : 'asc'
         },
       ]),
-      tasksFilter: ref({})
+    }
+  },
+
```
markgrahamdawson, line 183: can't `v-model` a prop so put the `taskFilter` prop into `tasksFilterLocal`

src/components/cylc/table/Table.vue
```diff
@@ -174,13 +178,18 @@ export default {
...
+  data () {
+    return {
+      tasksFilterLocal: this.tasksFilter,
     }
```
markgrahamdawson, line 187: can't `v-model` a prop so put the `taskFilter` prop into `tasksFilterLocal`

src/views/Table.vue
```diff
@@ -18,6 +18,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 <template>
   <div class="h-100">
     <TableComponent
+      :tasksFilter="tasksFilter"
```
MetRonnie, line 21: Rather than do this, I think it would make sense to extract the filtering out of the component (`src/components/cylc/table/Table.vue`) into the view (`src/views/Table.vue`)

oliver-sanders: Sounds reasonable, we should follow the same approach with all views.

tests/unit/views/table.vue.spec.js
```diff
@@ -17,11 +17,18 @@
...
 import { createStore } from 'vuex'
+import { createVuetify } from 'vuetify'
 import sinon from 'sinon'
 import storeOptions from '@/store/options'
 import Table from '@/views/Table.vue'
 import WorkflowService from '@/services/workflow.service'
 import User from '@/model/User.model'
+import { nextTick } from 'vue'
+import CylcObjectPlugin from '@/components/cylc/cylcObject/plugin'
+import { simpleTableTasks } from '@/../tests/unit/components/cylc/table/table.data'
```
markgrahamdawson, line 28: These are being loaded from `@/../tests/unit/components/` which is maybe a bit weird as the test is in 'tests/unit/views' should simple tasks be stored somewhere more centrally?

oliver-sanders (approved):

tests/unit/views/table.vue.spec.js
oliver-sanders: So long as it's clear what code does and doesn't belong to the tests, I don't mind.

MetRonnie (commented):
Got some suggestions at https://github.com/markgrahamdawson/cylc-ui/pull/2

tests/unit/components/cylc/table/table.vue.spec.js
```diff
@@ -50,13 +49,12 @@ describe('Table component', () => {
   it('should sort cycle point column descending by default', async () => {
     const wrapper = mountFunction({
       props: {
-        tasks: simpleTableTasks
+        filteredTasks: simpleTableTasks
```
MetRonnie, line 52: Oops, I missed the test. Need to replace `filteredTasks` back to `tasks`

MetRonnie (approved):

MetRonnie (changes requested):
Uh oh, shouldn't have rebased onto master as this needs to go into 2.4.x

PR #1717, opened 2024-03-19
Graph view: use `initialOptions` to save & restore view state on navigation

markgrahamdawson:
These changes build on the work undertaken to save & restore workspace layout on navigation https://github.com/cylc/cylc-ui/pull/1664

Which is part of a larger issue https://github.com/cylc/cylc-ui/issues/662

There was an equivalent pull request for the log view https://github.com/cylc/cylc-ui/pull/1688
There is an equivalent pull request for the tree view which this builds on: https://github.com/cylc/cylc-ui/pull/1705
There is an equivalent pull request for the table view which this builds on: https://github.com/cylc/cylc-ui/pull/1711

**Check List**

### DIFF
--- a/src/views/Graph.vue
+++ b/src/views/Graph.vue
@@ -98,6 +98,10 @@ import { getPageTitle } from '@/utils/index'
 import { useJobTheme } from '@/composables/localStorage'
 import graphqlMixin from '@/mixins/graphql'
 import subscriptionComponentMixin from '@/mixins/subscriptionComponent'
+import {
+  initialOptions,
+  useInitialOptions
+} from '@/utils/initialOptions'
 import SubscriptionQuery from '@/model/SubscriptionQuery.model'
 // import CylcTreeCallback from '@/services/treeCallback'
 import GraphNode from '@/components/cylc/GraphNode.vue'
@@ -224,9 +228,34 @@ export default {
     }
   },
 
-  setup () {
+  props: { initialOptions },
+
+  setup (props, { emit }) {
+    /**
+     * The transpose toggle state.
+     * If true layout is left-right, else top-bottom
+     * @type {import('vue').Ref<boolean>}
+     */
+    const transpose = useInitialOptions('transpose', { props, emit }, false)
+
+    /**
+     * The auto-refresh toggle state.
+     * If true the graph layout will be updated on a timer
+     * @type {import('vue').Ref<boolean>}
+     */
+    const autoRefresh = useInitialOptions('autoRefresh', { props, emit }, true)
+
+    /**
+     * The node spacing state.
+     * @type {import('vue').Ref<number>}
+     */
+    const spacing = useInitialOptions('spacing', { props, emit }, 1.5)
+
     return {
       jobTheme: useJobTheme(),
+      transpose,
+      autoRefresh,
+      spacing
     }
   },
 
@@ -236,8 +265,6 @@ export default {
       orientation: 'TB',
       // the auto-refresh timer
       refreshTimer: null,
-      // the spacing between nodes
-      spacing: 1.5,
       // the nodes end edges we render to the graph
       graphNodes: [],
       graphEdges: [],
@@ -249,58 +276,8 @@ export default {
       graphID: null,
       // instance of system which provides pan/zoom/navigation support
       panZoomWidget: null,
-      // if true layout is left-right is false it is top-bottom
-      transpose: false,
-      // if true the graph layout will be updated on a timer
-      autoRefresh: true,
       // true if layout is in progress
       updating: false,
-      controlGroups: [
-        {
-          title: 'Graph',
-          controls: [
-            {
-              title: 'Refresh',
-              icon: mdiRefresh,
-              action: 'callback',
-              callback: this.refresh,
-              disableIf: ['autoRefresh']
-            },
-            {
-              title: 'Auto Refresh',
-              icon: mdiTimer,
-              action: 'toggle',
-              value: true,
-              key: 'autoRefresh'
-            },
-            {
-              title: 'Transpose',
-              icon: mdiFileRotateRight,
-              action: 'toggle',
-              value: false,
-              key: 'transpose'
-            },
-            {
-              title: 'Centre',
-              icon: mdiImageFilterCenterFocus,
-              action: 'callback',
-              callback: this.reset
-            },
-            {
-              title: 'Increase Spacing',
-              icon: mdiArrowExpand,
-              action: 'callback',
-              callback: this.increaseSpacing
-            },
-            {
-              title: 'Decrease Spacing',
-              icon: mdiArrowCollapse,
-              action: 'callback',
-              callback: this.decreaseSpacing
-            }
-          ]
-        }
-      ],
     }
   },
 
@@ -310,6 +287,7 @@ export default {
     this.graphviz = Graphviz.load()
     // allow render to happen before we go configuring svgPanZoom
     this.$nextTick(() => {
+      this.refresh()
       this.updateTimer()
     })
     this.mountSVGPanZoom()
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -15,7 +15,7 @@ ones in. -->
 
 ### Enhancements
 
-
+[#1717](https://github.com/cylc/cylc-ui/pull/1717) -
```
MetRonnie, line 18: Needs to go in a new entry under 2.5.0

src/views/Graph.vue
```diff
@@ -224,9 +229,34 @@ export default {
...
-  setup () {
+  props: { initialOptions },
+
+  setup (props, { emit }) {
+    /**
+     * The transpose toggle state.
+     * If true layout is left-right is false it is top-bottom
+     * @type {import('vue').Ref<Bool>}
+     */
+    const transpose = useInitialOptions('transpose', { props, emit }, ref(false))
```
MetRonnie, line 240: The default value should be a normal type, not a ref
```suggestion
    const transpose = useInitialOptions('transpose', { props, emit }, false)
```

src/views/Graph.vue
```diff
@@ -224,9 +229,34 @@ export default {
...
+
+    /**
+     * The auto-refresh toggle state.
+     * If true the graph layout will be updated on a timer
+     * @type {import('vue').Ref<Bool>}
+     */
+    const autoRefresh = useInitialOptions('autoRefresh', { props, emit }, ref(true))
```
MetRonnie, line 247:
```suggestion
    const autoRefresh = useInitialOptions('autoRefresh', { props, emit }, true)
```

src/views/Graph.vue
```diff
@@ -224,9 +229,34 @@ export default {
...
+
+    /**
+     * The node spacing state.
+     * @type {import('vue').Ref<Integer>}
```
MetRonnie, line 251:
```suggestion
     * @type {import('vue').Ref<number>}
```

src/views/Graph.vue
```diff
@@ -224,9 +229,34 @@ export default {
...
   },
 
-  setup () {
+  props: { initialOptions },
+
+  setup (props, { emit }) {
+    /**
+     * The transpose toggle state.
+     * If true layout is left-right is false it is top-bottom
+     * @type {import('vue').Ref<Bool>}
```
MetRonnie, line 238:
```suggestion
     * If true, layout is left-right, else top-bottom
     * @type {import('vue').Ref<boolean>}
```

src/views/Graph.vue
```diff
@@ -224,9 +229,34 @@ export default {
...
+     */
+    const transpose = useInitialOptions('transpose', { props, emit }, ref(false))
+
+    /**
+     * The auto-refresh toggle state.
+     * If true the graph layout will be updated on a timer
+     * @type {import('vue').Ref<Bool>}
```
MetRonnie, line 245:
```suggestion
     * @type {import('vue').Ref<boolean>}
```

tests/e2e/specs/graph.cy.js
```diff
@@ -37,6 +37,31 @@ function checkGraphLayoutPerformed ($el, depth = 0) {
...
 }
 
+function addView (view) {
+  cy.get('[data-cy=add-view-btn]').click()
+  cy.get(`#toolbar-add-${view}-view`).click()
+    // wait for menu to close
+    .should('not.be.exist')
+}
+
+function checkRemeberToolbarSettings (selector, stateBefore, stateAfter) {
```
MetRonnie, line 47:
```suggestion
function checkRememberToolbarSettings (selector, stateBefore, stateAfter) {
```

src/views/Graph.vue
```diff
@@ -249,13 +277,51 @@ export default {
...
       // true if layout is in progress
       updating: false,
-      controlGroups: [
+    }
+  },
+
+  mounted () {
+    if (!this.autoRefresh) {
+      // load the graph if the autorefresh is off
+      this.refreshTimer = setInterval(this.refresh, 1000)
```
oliver-sanders, line 288: This will wait 1 second, then attempt to refresh the graph.

This should cover most cases just fine, but if the data takes longer than 1 second to arrive, then the graph will never be drawn.  This might be ok, we tried and failed to render the graph, since it's on manual, it's down to you to press the button. We could potentially put in a placeholder telling the user to push the button.

Alternatively, we could attempt to call this refresh method every `x` seconds until the graph renders. I think the `graphID` attribute might work for this as it is `null` until there is something to render and it has been rendered.

MetRonnie: Would https://github.com/markgrahamdawson/cylc-ui/pull/3 be the right sort of approach?

oliver-sanders: I think that would work.

Ideally, we would do this in a way that isn't blocking to the `refresh` method as we should really be clearing the graph if there are no nodes to display rather than leaving the graph as it is and waiting for nodes to be created.

src/views/Graph.vue
```diff
@@ -249,13 +276,51 @@ export default {
...
```
oliver-sanders, line 287: For now how about this:

```suggestion
      this.refreshTimer = setInterval(this.refresh, 1000)
      this.refreshTimer = setInterval(this.refresh, 10000)  // make sure the graph is refreshed even if data is slow to arrive
```

Crude but effective.

Punt the rest to https://github.com/cylc/cylc-ui/issues/1756.

markgrahamdawson: Hmm, I thought this was fixed with this pr https://github.com/markgrahamdawson/cylc-ui/pull/3/files
That was merged into this branch? But the changes don't seem to be included?

oliver-sanders: That commit doesn't appear to be on this branch any more.

I think you [force pushed](https://github.com/cylc/cylc-ui/compare/a4dcf253506e89eea93045f235657dbc495ead57..9abf4dce0da7fc6cce6fac4aa9e43134d704eb11#diff-327c8ac58a04b20db03a8d9e07c142db2a3627d32226f9bb1c0855e03dba0bc2R284) over it because it was added to the remote branch but not the local one.

Here's one way to pull upstream changes (e.g. code suggestions, side-prs, etc) into your local branch

```console
$ git checkout <branch>
$ git fetch origin
$ git rebase origin/<branch>
```

Before force-pushing I often check the diff against origin like this:

```console
$ git fetch origin
$ git diff origin/<branch>
$ git push -f origin/<branch>
```

CHANGES.md
MetRonnie, line 1: Need to drop/revert these edits to `CHANGES.md`

markgrahamdawson:
It would be nice to get this in so I can finish this off...
https://github.com/cylc/cylc-ui/pull/1763

tests/e2e/specs/graph.cy.js
```diff
@@ -69,4 +94,18 @@ describe('Graph View', () => {
...
+    checkRememberToolbarSettings('[data-cy=control-autoRefresh]', 'have.class', 'not.have.class')
+  })
+
+  it('remembers transpose setting when switching between workflows', () => {
+    cy.visit('/#/workspace/one')
+    addView('Graph')
+    waitForGraphLayout()
+    checkRememberToolbarSettings('[data-cy=control-transpose]', 'not.have.class', 'have.class')
+  })
 })
```
markgrahamdawson, line 111: I could add an extra test to check that the graph loads when switching between workflows

```
  it('loads graph when switching between workflows', () => {
    cy.visit('/#/workspace/one')
    addView('Graph')
    waitForGraphLayout()
    cy.visit('/#/workspace/other/multi/run2')
    addView('Graph')
    waitForGraphLayout()
    cy.visit('/#/workspace/one')
    cy
    // there should be 7 graph nodes (all on-screen)
      .get('.c-graph:first')
      .find('.graph-node-container')
      .should('be.visible')
      .should('have.length', 7)
  })
  ```

File: src/views/Graph.vue [hunk 11]
```diff
@@ -558,10 +584,13 @@ export default {
       this.updating = true
 
       // extract the graph (non reactive lists of nodes & edges)
-      const nodes = this.getGraphNodes()
+      const nodes = await this.waitFor(() => {
+        const nodes = this.getGraphNodes()
+        return nodes.length ? nodes : false
```
MetRonnie, line 589:
```suggestion
        return this.getGraphNodes()
```

src/views/Graph.vue
```diff
@@ -558,10 +584,13 @@ export default {
...
+      })
       const edges = this.getGraphEdges()
 
-      if (!nodes.length) {
+      if (!nodes || !nodes.length) {
```
MetRonnie, line 593:
```suggestion
      if (!nodes.length) {
```

src/views/Graph.vue
markgrahamdawson: This causes the autorefresh issue to return
image

MetRonnie: Ah my bad

tests/e2e/specs/graph.cy.js
MetRonnie: Sounds good

MetRonnie (approved):
I'm having a bit of a poke about with the Graph view to see if the initial refresh can rely more on Vue's reactivity than manual polling, but this will do for the time being

markgrahamdawson:
@oliver-sanders If I make a new issue for [https://github.com/cylc/cylc-ui/pull/1717#pullrequestreview-2022714370](Ronnies note above). Can I merge this in? Does it really need a rebase first to tidy up?

oliver-sanders:
We can also squash-merge PRs (note, you don't need separate commits for changelog, tests, etc, they can go in together with the feature).

oliver-sanders:
Unfortunately the auto-refresh option appears to be flaky for me. Sometimes the graph comes back, sometimes it does not. I don't think the diff to the refresh method quite did the job.

markgrahamdawson (commented):
Ready for review

oliver-sanders (approved):
Nice.

oliver-sanders:
@MetRonnie, please check you're happy with new changes and whack the merge button.

PR #1755, opened 2024-04-16
toolbar: add n-window-selector

### LINKED ISSUE #864: configurable n-window
Ideally the user would be able to set the `n-window` for each view independently.

A good use case would be an operator working on a complex workflow. They might want the tree view in `n=1` and the graph view in `n=0`.

One day we would be able to construct an n-window around an arbitrary selection of start tasks which would enhance this further.

E.G. The operator could have the tree view open to `n=1` and the graph view to an `n=2` based around the tasks they are selecting in the tree view.

This would be useful in understanding the impacts of task failures, delays, bottlenecks, etc.

This is a complex thing to do and would require server-side assistance. I think the data store currently supports the concept of an arbitrary n-window selection.

Related:

* https://github.com/cylc/cylc-ui/issues/1093
* https://github.com/cylc/cylc-ui/issues/434
* https://github.com/cylc/cylc-ui/issues/534

**Pull requests welcome!**

oliver-sanders:
* Closes #864
* Add a selector for switching the n-window extent between 0 & 3 inclusive.
* The n-window value appears to be preserved when navigating back to the workflow.

n-window-changing

**Check List**

- [ ] Tests are included - seems to be hard to test meaningfully without true e2e tests (i.e. with live workflows).
- [x] docs issue - https://github.com/cylc/cylc-doc/issues/719

### DIFF
--- a/src/components/cylc/workflow/Toolbar.vue
+++ b/src/components/cylc/workflow/Toolbar.vue
@@ -86,6 +86,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
       <!-- n-window selector -->
       <v-chip
+        :disabled="isStopped"
         link
         size="small"
       >
@@ -356,9 +357,6 @@ export default {
         )
       }
     },
-    userInitials () {
-      return this.user.username[0].toUpperCase()
-    },
     nEdgeDistance () {
       // the graph window distance reported by the scheduler
       return this.currentWorkflow?.node?.nEdgeDistance
@@ -381,7 +379,7 @@ export default {
     },
     nEdgeDistance (newVal) {
       // the scheduler has reported that the window size has changed
-      if (newVal !== undefined) {
+      if (newVal != undefined) {
         this.nWindow = newVal
       }
     }
--- a/src/components/cylc/workflow/Toolbar.vue
+++ b/src/components/cylc/workflow/Toolbar.vue
@@ -51,6 +51,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           v-cylc-object="currentWorkflow"
           :icon="$options.icons.menu"
           size="small"
+          density="comfortable"
         />
 
         <v-btn
@@ -60,6 +61,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           v-if="!isRunning"
           @click="onClickPlay"
           size="small"
+          density="comfortable"
         />
 
         <v-btn
@@ -69,6 +71,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           v-if="isRunning"
           @click="onClickReleaseHold"
           size="small"
+          density="comfortable"
         />
 
         <v-btn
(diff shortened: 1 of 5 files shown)


oliver-sanders:
@MetRonnie, this setting is preserved when you navigate back, however, I have not done anything to achieve this. Should it be implemented differently or is this ok?

MetRonnie:
It's preserved because it's querying it, right?

oliver-sanders:
Derp, it's getting the n-window extent from the returned data :facepalm:

oliver-sanders:
As a true end-to-end feature (i.e. flow-uis-ui-uis-flow) it's a right pain to test. I had a look a testing, at the unit test level you end up testing vue watchers and at the e2e level, you end up testing the Vuetify Select API.

src/components/cylc/workflow/Toolbar.vue
```diff
@@ -77,9 +80,38 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+            <v-card-text>
+              <v-select
+                density="compact"
+                v-model="nWindow"
+                :items="[0,1,2,3]"
+              />
+            </v-card-text>
+          </v-card>
+        </v-menu>
+      </v-btn>
```
MetRonnie, line 113: The use of a button here leads to odd styling. Can just use a chip with the `link` prop instead.
```suggestion
      <v-chip
        link
        size="small"
      >
        N = {{ nWindow }}
        <v-menu activator="parent" :close-on-content-click="false">
          <v-card title="Graph Window Depth">
            <v-card-subtitle>
              This changes the number of tasks which are displayed.

              Higher values may impact performance.
            </v-card-subtitle>
            <v-card-text>
              <v-select
                density="compact"
                v-model="nWindow"
                :items="[0,1,2,3]"
              />
            </v-card-text>
          </v-card>
        </v-menu>
      </v-chip>
```

src/components/cylc/workflow/Toolbar.vue
```diff
@@ -273,6 +355,13 @@ export default {
           )
         )
       }
+    },
+    userInitials () {
+      return this.user.username[0].toUpperCase()
+    },
```
MetRonnie, line 361: Not used?

src/components/cylc/workflow/Toolbar.vue
```diff
@@ -285,6 +374,16 @@ export default {
...
     isStopped () {
       this.expecting.stop = null
+    },
+    nWindow (newVal) {
+      // the user has requested to change the window size
+      this.setGraphWindow(newVal)
+    },
+    nEdgeDistance (newVal) {
+      // the scheduler has reported that the window size has changed
+      if (newVal !== undefined) {
```
MetRonnie, line 384: I think it's safer to guard against the value being `null` too
```suggestion
      if (newVal != undefined) {
```

src/components/cylc/workflow/Toolbar.vue
```diff
@@ -77,9 +80,34 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           :disabled="!enabled.stopToggle"
           @click="onClickStop"
           size="small"
+          density="comfortable"
         />
       </div>
 
+      <!-- n-window selector -->
+      <v-chip
```
MetRonnie, line 88: Also how about
```suggestion
      <v-chip
        :disabled="isStopped"
```
or even

```suggestion
      <v-chip
        v-if="!isStopped"
```

src/components/cylc/workflow/Toolbar.vue
oliver-sanders, line 361:
```suggestion
```

src/components/cylc/workflow/Toolbar.vue
```diff
@@ -285,6 +372,16 @@ export default {
...
     isStopped () {
       this.expecting.stop = null
+    },
+    nWindow (newVal) {
+      // the user has requested to change the window size
+      this.setGraphWindow(newVal)
+    },
+    nEdgeDistance (newVal) {
+      // the scheduler has reported that the window size has changed
+      if (newVal != undefined) {
```
MetRonnie, line 382: Oh, for some reason ESLint only allows non-strict equality comparison with `null` not `undefined` (https://github.com/eslint/eslint/issues/13537)
```suggestion
      if (newVal != null) {
```

MetRonnie: (If using `!= null` you don't have to worry about whether it's `null` or `undefined`)

MetRonnie (approved):

hjoliver (approved):
Tested, love it.

Issue #1255, opened 2023-03-02
Table view sorting: sort empty entries last when in ascending mode


MetRonnie:
If you click e.g. the "Job ID" column to sort by the job ID, you get the tasks that haven't yet got a job ID showing before the tasks that do have a job ID. It should be the other way round because if you're sorting in such a way you're more likely to care about the values that are relevant, not empty values.

Note the more sensible sorting behaviour is already implemented for some of the columns e.g. "T-submit" (essentially treat empty value as `Infinity`).

PR #1762, opened 2024-04-18
Table view: sort empty entries last regardless of sort order

MetRonnie:
Closes https://github.com/cylc/cylc-ui/issues/1255

When sorting by e.g. Submitted Time, an absence of a value means a job has not yet been submitted by the task that this row corresponds to. If sorting ascending, it is therefore logical that empty values should come last. But if sorting descending, you are probably interested in the newest jobs so again the empty values should come last. Note this is somewhat different behaviour to the linked issue.

**Check List**

- [x] Tests are included 
- [x] No changelog entry included as minor

### DIFF
--- a/src/components/cylc/common/sort.js
+++ b/src/components/cylc/common/sort.js
@@ -28,21 +28,14 @@
 
 /**
  * The default comparator used to compare strings for cycle points, family proxies names,
- * task proxies names, and jobs. Note: nullish or empty strings are treated as infinity.
+ * task proxies names, and jobs.
  *
- * @param {string?} left
- * @param {string?} right
+ * @param {string} left
+ * @param {string} right
  * @returns {number}
  * @constructor
  */
 export const DEFAULT_COMPARATOR = (left, right) => {
-  if (!left && !right) {
-    return 0
-  } else if (!left) {
-    return 1
-  } else if (!right) {
-    return -1
-  }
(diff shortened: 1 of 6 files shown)


oliver-sanders:
I'm not so sure about putting undefined vales *after* defined ones, it might make more sense to put them *before*. E.g. it makes sense for a task which has not yet submitted to be listed before one which has (because it is yet to submit).

A notable use case for the table view is to view the most recent submitted/started/finished jobs (ala Cylc Review). Unfortunately, giving undefined items *after* defined ones makes this difficult as all the n>0 tasks come first:

Screenshot from 2024-05-22 11-24-02

MetRonnie:
> E.g. it makes sense for a task which has not yet submitted to be listed before one which has (because it is yet to submit).

No I don't think it does. If you sort by ascending submitted time, surely you want the oldest submissions at the top, but instead you would get tasks that have not been submitted yet.

So should we sort empty values after real values all the time, then (regardless of ascending/descending order)?

oliver-sanders:
> So should we sort empty values after real values all the time, then (regardless of ascending/descending order)?

That seems sensible (any other choice is arbitrary) but makes it a Vuetify problem?

MetRonnie (commented):
I found a way

oliver-sanders (approved):

Issue #1130, opened 2022-11-04
graph: family collapsing


oliver-sanders:
> Follow on from #1108

Implement family expand/collapse in the graph view.

In the Cylc 7 graph view families were represented with a special node shape & expansion was available via a right-click menu. When a family was expanded the special node was removed and the tasks it represented were added in its place.

With the Cylc 8 graph view we have more power over rendering so can do something more advanced. One thing I'd like to consider is expanding families into subgraphs so that the family name is preserved which should make graphs easier to understand.

**Pull requests welcome!**

oliver-sanders:
Attempt along with https://github.com/cylc/cylc-ui/issues/1280

markgrahamdawson:
For the time being - figure out a means of drawing a dotted box around families.

This can be done with graphviz [subgraphs ](https://graphviz.readthedocs.io/en/stable/manual.html#subgraphs-clusters)

The graphviz dot code is returned from the `getDotCode` function in `Graph.vue` view

Dot code can be rendered with 
`dot ffo.dot -Tpng -o fee.png`

markgrahamdawson:
When turning the dotcode into json objects - if two subgraphs are included with the exact same bounding box then the coordinate data is only contained in the first object 

In the image below subgraph cluster_1 and subgraph cluster_2 have the same coordinates so they are not included in the object returned for cluster_2

![image](https://github.com/cylc/cylc-ui/assets/7523921/4191ffb8-3729-4ee0-a3fa-28da311e56e2)

Not sure if this is going to be an issue or not but thought it was worth making a note of.

oliver-sanders:
Some ideas for a unified approach to grouping/collapsing cycles/families.

Note, for simplicity, this approach groups/collapses *all* instances of selected families rather than managing this at a per-cycle level. I think this is probably more aligned with expectations, but does represent a minor limitation, e.g. there's no ability to collapse all but one cycle.

![graph-grouping-and-collapsing](https://github.com/cylc/cylc-ui/assets/16705946/b1c2ba2e-2258-47e9-bce3-0a0a442d9919)

Issue #1635, opened 2024-01-17
table: task state unknown


oliver-sanders:
I spotted that sometimes the task status wasn't getting through to the task component in the  table view.

This results in a job in an active state associated with a task which appears to be in the waiting state:

![Screenshot from 2024-01-17 13-58-24](https://github.com/cylc/cylc-ui/assets/16705946/60650735-0e9c-461d-8d9a-82bf99eabe88)

But actually isn't:

![Screenshot from 2024-01-17 14-03-15](https://github.com/cylc/cylc-ui/assets/16705946/7504a8a6-88e1-4534-9434-4e57d074211c)

Running the example in https://github.com/cylc/cylc-ui/issues/1634, this appears to crop up fairly regularly.

I have only seen this in the table view so far.

Replicated with 2.3.0 and on master.

MetRonnie:
Also appears in the graph view. And the tree view shows an empty cycle point

![image](https://github.com/cylc/cylc-ui/assets/61982285/22ca4094-4bb1-44f0-b2b2-c168df94afa6)

MetRonnie:
Problem remains on browser refresh, go away on workflow reload. So this is presumably a data store issue

oliver-sanders:
Does this only occur in simulation mode?

PR #1763, opened 2024-04-19
added cyclepoint grouping with subgraphs

markgrahamdawson:
Partly addresses [#1130](https://github.com/cylc/cylc-ui/issues/1130)

- Added button to graph view to group nodes by cycle point.
- Works for timeseries and integer
- Option state saves when moving between workflows

---initial notes on work---
For the time being - figure out a means of drawing a dotted box around nodes that share a cyclepoint.
This can be done with graphviz [subgraphs](https://graphviz.readthedocs.io/en/stable/manual.html#subgraphs-clusters)
The graphviz dot code is returned from the getDotCode function in Graph.vue view
Dot code can be rendered with
dot foo.dot -Tpng -o foo.png
ultimately in the svg there will need to be a rect tag added...
<rect x="" y="" width="" height="">

**Check List**


markgrahamdawson:
The amount of marginbetween subgraphs is random unless made explicit

No margin
image

Margin set in code
image

Im guessing the second one is better?

hjoliver:
Def prefer the second one.

src/components/cylc/GraphSubgraph.vue
```diff
@@ -0,0 +1,51 @@
+<template>
+  <g class="c-graph-subgraph">
+    <rect
+      :width="subgraph.width"
+      :height="subgraph.height"
+      :x="subgraph.x"
+      :y="subgraph.y"
+      :visibility="subgraph.label=='margin' ?  'hidden' : 'visible' "
```
markgrahamdawson, line 8: In order to force a margin between the subgraphs Ive put an extra subgraph in with visibility set to hidden

Used the solution mentioned here:
https://stackoverflow.com/questions/9871062/padding-between-cluster-boundaries-and-nodes-when-using-graphviz-and-neato

src/views/Graph.vue
```diff
@@ -361,6 +379,13 @@ export default {
               icon: mdiArrowCollapse,
               action: 'callback',
               callback: this.decreaseSpacing
+            },
+            {
+              title: 'Organise by cycle point',
```
MetRonnie, line 384:
```suggestion
              title: 'Group by cycle point',
```

src/views/Graph.vue
```diff
@@ -689,26 +745,39 @@ export default {
...
-        // TODO convert (2) to maths OR fix it to avoid recomputation?
-        this.nodeTransformations[obj.name] = `
+        if (obj.bb) {
+        // // if the object is a subgraph
+          this.subgraphs[obj.name] = {
+            x: obj.bb.split(',')[0],
+            y: -obj.bb.split(',')[3],
+            width: obj.bb.split(',')[2] - obj.bb.split(',')[0],
+            height: obj.bb.split(',')[3] - obj.bb.split(',')[1],
+            label: obj.label
```
MetRonnie, line 766: Is this correct? If so, it's more readable IMO. Not sure why `y = -top` though
```suggestion
          // if the object is a subgraph
          const [left, bottom, right, top] = obj.bb.split(',')
          this.subgraphs[obj.name] = {
            x: left,
            y: -top,
            width: right - left,
            height: top - bottom,
            label: obj.label
```

src/views/Graph.vue
```diff
@@ -86,6 +86,12 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             />
           </g>
         </g>
+        <g v-if="groupCyclePoint" class="graph-subgraph-container">
+          <GraphSubgraph
+            v-for="(subgraph, index) in subgraphs"
+            :key="index"
```
MetRonnie, line 92: It's not an array, it's an object
```suggestion
            v-for="(subgraph, key) in subgraphs"
            :key="key"
```

src/views/Graph.vue
```diff
@@ -483,7 +508,15 @@ export default {
...
       return ret
     },
-    getDotCode (nodeDimensions, nodes, edges) {
+    getCycleSubgraph (nodes) {
+      if (!this.groupCyclePoint) return
+      const result = nodes.reduce((x, y) => {
+        (x[y.node.cyclePoint] = x[y.node.cyclePoint] || []).push(y)
+        return x
+      }, {})
+      return result
```
MetRonnie, line 517:
```suggestion
      return nodes.reduce((x, y) => {
        (x[y.node.cyclePoint] ||= []).push(y)
        return x
      }, {})
```

MetRonnie (commented):
Looking good, partial review so far

src/views/Graph.vue
markgrahamdawson: I used this phrasing as its what is used in Cylc 7 but can change it

src/views/Graph.vue
markgrahamdawson: This was a lot of trial and error tbh but without the - 
image

markgrahamdawson: Example of what comes back from graphviz:
` { "x": "7604", "y": -734.8, "width": 4025, "height": 676.8, "label": "20000104T1200Z" } `

src/views/Graph.vue
```diff
@@ -526,6 +559,27 @@ export default {
...
+      if (this.groupCyclePoint) {
+        // Loop over the subgraphs
+        Object.keys(cycleSubgraph).forEach((key, i) => {
+          // Loop over the nodes that are included in the subraph
+          const nodeFormattedArray = cycleSubgraph[key].map(a => `"${a.id}"`)
+          ret.push(`
+          subgraph cluster_margin
+          {
+            margin=50.0
+            label="margin"
```
oliver-sanders, line 572: Very neat margin trick! I did not know we could do this!

But I don't think it's doing quite what it says on the tin, because all of the `cluster_margin` subgraphs have been given the same name, so all of the `cluster_<i>` subgraphs are residing in the same margin. It does seem to work, if I increase the margin the clusters move apart, though not by the amount I would expect, but I really don't understand how this works.

If we give each `cluster_margin` subgraph a unique ID, then the margin applies to each contained subgraph individually not globally.

```suggestion
          subgraph cluster_margin_$(i)
          {
            margin=50.0
            label="margin"
```

Might be worth increasing this margin value, I had a play with the "complex" workflow which has a nasty graph and found ~100 worked well (I can get you a copy of the complex workflow, good for testing nasty use cases).

src/views/Graph.vue
```diff
@@ -526,6 +559,27 @@ export default {
...
+            subgraph cluster_${i} {${nodeFormattedArray};\n
+              label = "${key}";\n
```
oliver-sanders, line 574: Could do with upping the font-size of the label.

The label size can be defined in graphviz like this:

```
              label = "${key}";\n
              fontsize = 40
```

Will need to get the `GraphSubgraph` label font size to match this.

src/views/Graph.vue
```diff
@@ -526,6 +559,27 @@ export default {
...
+              color=blue
```
oliver-sanders, line 575: Nothing appears to be blue, intentional?

```suggestion
```

src/views/Graph.vue
oliver-sanders: Some systems disagree on the y axis, presumably GraphViz and SVG are an example. I think we do `-y` somewhere else in the graph view too.

src/components/cylc/GraphSubgraph.vue
oliver-sanders: Could we avoid creating a `<rect />` element for the hidden subgraph?

Not a big deal, but worth cutting down on the number of elements where possible.

src/views/Graph.vue
```diff
@@ -483,7 +508,15 @@ export default {
       }
       return ret
     },
-    getDotCode (nodeDimensions, nodes, edges) {
+    getCycleSubgraph (nodes) {
```
oliver-sanders, line 511: Possibly rename `getCycles`.

src/views/Graph.vue
```diff
@@ -689,26 +745,39 @@ export default {
      * @param {Object[]} edges
      * @param {{ [id: string]: SVGRect }} nodeDimensions
      */
-    async layout (nodes, edges, nodeDimensions) {
+    async layout (nodes, edges, nodeDimensions, cycleSubgraph) {
```
oliver-sanders, line 748: Possible rename `cycleSubgraph` to `cycles`?

oliver-sanders (commented):
Looks good, works nicely :+1:

Sadly, GraphViz doesn't stop edges from being routed through graph labels by the looks of it. Not much we can do about that, but we could try adding a white `stroke` (SVG lingo for border) around the text to keep it readable, e.g:

```css
element {
    fill: black;
    stroke-width: 2;
    stroke: white;
}
```

Could do with bumping the font-size of the graph labels a bit, not much, they don't need to be larger than task `GraphNode`s, but they do get lost in big graphs rather easily e.g:

Screenshot from 2024-05-08 17-14-24

src/components/cylc/GraphSubgraph.vue
```diff
@@ -0,0 +1,51 @@
...
+      required: true
+    }
+  },
+  methods: {
+    calcLabelXPosition (x, width) {
+      return (parseInt(x) + (parseInt(width) / 2))
+    },
+    calcLabelYPosition (y, height) {
+      return (parseInt(y) + 90)
+    }
```
MetRonnie, line 40: I think these could be computed properties instead?

src/views/Graph.vue
```diff
@@ -361,6 +379,13 @@ export default {
               icon: mdiArrowCollapse,
               action: 'callback',
               callback: this.decreaseSpacing
+            },
+            {
+              title: 'Organise by cycle point',
+              icon: mdiTimetable,
```
MetRonnie, line 385: How about [mdiVectorSelection](https://pictogrammers.com/library/mdi/icon/vector-selection/)?

src/components/cylc/GraphSubgraph.vue
```diff
@@ -0,0 +1,52 @@
...
+      :y="subgraph.y"
+      rx="50"
+      ry="50"
+      fill="none" />
+    <text
+      :x="calcLabelXPosition"
+      :y="calcLabelYPosition"
+      font-family="Roboto"
+      alignment-baseline="middle" text-anchor="middle"
+      font-size=70
```
oliver-sanders, line 16: This font size should match the one defined in the dot file to ensure that graphviz leaves the right amount of space for the label that we draw.

But note that the units may differ so you might need to specify `px` or `mm` as appropriate (I'm not sure what units GraphViz uses by default, I know dimensions are in inches so fonts might be odd).

src/views/Graph.vue
```diff
@@ -526,6 +564,27 @@ export default {
...
+          // Loop over the nodes that are included in the subraph
+          const nodeFormattedArray = cycles[key].map(a => `"${a.id}"`)
+          ret.push(`
+          subgraph cluster_margin_${i}
+          {
+            margin=100.0
+            label="margin"
+            subgraph cluster_${i} {${nodeFormattedArray};\n
+              label = "${key}";\n
+              fontsize = 40
```
oliver-sanders, line 580: This font size should match the one in the subgraph component (watch out for units, they may differ).

src/views/Graph.vue
```diff
@@ -689,26 +750,40 @@ export default {
...
+    async layout (nodes, edges, nodeDimensions, cycles) {
       // generate the GraphViz dot code
-      const dotCode = this.getDotCode(nodeDimensions, nodes, edges)
+      const dotCode = this.getDotCode(nodeDimensions, nodes, edges, cycles)
 
       // run the layout algorithm
       const jsonString = (await this.graphviz).layout(dotCode, 'json')
       const json = JSON.parse(jsonString)
 
+      this.subgraphs = {}
```
oliver-sanders, line 761: I used `map` to update the graph edges because I guessed that rebuilding the set in a single operation might play more nicely with Vue's reactivity than emptying and rebuilding in a series of steps.

This guess might be completely wrong, anyone know different?

src/components/cylc/GraphSubgraph.vue
```diff
@@ -0,0 +1,52 @@
...
+      :width="subgraph.width"
+      :height="subgraph.height"
+      :x="subgraph.x"
+      :y="subgraph.y"
+      rx="50"
+      ry="50"
+      fill="none" />
+    <text
+      :x="calcLabelXPosition"
+      :y="calcLabelYPosition"
```
oliver-sanders, line 13: It's a computed property now so it should get a "thing" name rather than a "do" name to make it clear what it is:

```suggestion
      :x="labelXPosition"
      :y="labelYPosition"
```

src/components/cylc/GraphSubgraph.vue
```diff
@@ -0,0 +1,52 @@
...
+    subgraph: {
+      type: Object,
+      required: true
+    }
+  },
+  computed: {
+    calcLabelXPosition () {
+      return (parseInt(this.subgraph.x) + (parseInt(this.subgraph.width) / 2))
+    },
+    calcLabelYPosition () {
```
oliver-sanders, line 39:
```suggestion
    labelXPosition () {
      return (parseInt(this.subgraph.x) + (parseInt(this.subgraph.width) / 2))
    },
    labelYPosition () {
```

src/components/cylc/GraphSubgraph.vue
```diff
@@ -0,0 +1,52 @@
...
+}
+</script>
+
+<style lang="scss">
+  rect {
+    stroke-width: 8px;
+    stroke: grey;
+    stroke-dasharray: 50 50;
+  }
+</style>
```
oliver-sanders, line 52: @MetRonnie, should this be `scoped`?

Or would it be easier to just inline this?

src/views/Graph.vue
```diff
@@ -86,6 +86,12 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             />
           </g>
         </g>
+        <g v-if="groupCycle" class="graph-subgraph-container">
+          <GraphSubgraph
+            v-for="(subgraph, key) in subgraphs"
```
oliver-sanders, line 91: We may be able to move that `-vif` out of the `GraphSubgraph` component to avoid creating component instances for the renderless "hidden" subgraphs.

```js
          <GraphSubgraph
            v-for="(subgraph, key) in subgraphs"
            v-if="subgraph.label!='margin'"
```

markgrahamdawson: Cant have v-for and v-if on the same element but think another solution is

```        
        <g v-if="groupCycle" class="graph-subgraph-container">
          <g v-for="(subgraph, key) in subgraphs"
            :key="key">
            <GraphSubgraph
              v-if="subgraph.label!='margin'"
              :subgraph="subgraph" />
          </g>
        </g>
```

src/views/Graph.vue
MetRonnie: I wouldn't have thought that would matter, as surely this code has to finish executing synchronously before any Vue reactivity can come into play.

Generally I do favour using `Array.prototype.map` for its elegance (and it's surely more efficient than creating an empty array and then pushing to it which means the JS engine will keep having to increase the size of the array). But in this case as we can iterate over `json.objects` once to extract the subgraphs and the node transformations I think it makes sense to do it the current way.

src/views/Graph.vue
MetRonnie: This would fail ESLint - https://eslint.vuejs.org/rules/no-use-v-if-with-v-for.html

src/components/cylc/GraphSubgraph.vue
MetRonnie: Yes, either one of these should be done

src/views/Graph.vue
MetRonnie: What is the purpose of the `v-if`? I saw in an earlier discussion https://github.com/cylc/cylc-ui/pull/1763#discussion_r1587800644

```
:visibility="subgraph.label=='margin' ?  'hidden' : 'visible' "
```
but `v-if` does not accomplish the same thing (hidden visibility means it still takes up space which I presume is the whole point)

src/components/cylc/GraphSubgraph.vue
markgrahamdawson: SVG default is pixels but accepts cm, mm, in, em, ex, pt, pc, and px

Graphviz uses points (pt)
https://graphviz.org/doc/info/attrs.html
but it seems happy enough when given pixels

src/views/Graph.vue
oliver-sanders: These "hidden" subgraphs are added into the GraphViz document in order to take up space in the virtual graph (relative coordinates) proving a bit of margin between subgraphs which is nice.

However, I don't think they need to be added into the SVG document (either as hidden elements or otherwise). You don't need them to take up space in the SVG (absolute coordinates), we only need the x,y,width,height of the subgraph outline.

Another way of doing it would be to filter out the "hidden" clusters from `this.subgraphs`. From an efficiency standpoint it's best to reduce the number of HTML elements and especially Vue components we use.

markgrahamdawson: No - the layout is done by graphviz and what we are rendering in the svg is a copy of that. The spacing between graphviz subgraphs is a bit random and so I added an extra subgraph to give a handle for controlling the 'margin' between the different subgraphs (or at least give a minimum spacing).
This means each cycle point has two subgraphs - one that it is meant to have and a second which is only there to control the spacing...
image

So just need to make that "margin" one not visible/render.

src/views/Graph.vue
```diff
@@ -603,6 +603,8 @@ export default {
         }
       }
       ret.push('}')
+      console.log("ret.join('\n')")
+      console.log(ret.join('\n'))
```
oliver-sanders, line 607: debug

src/views/Graph.vue
```diff
@@ -774,7 +776,9 @@ export default {
             label: obj.label
           }
         } else {
-        // else the object is a node
+          console.log("obj.name")
+          console.log(obj.name)
```
oliver-sanders, line 780: debug

src/views/Graph.vue
oliver-sanders: Fine by me.

> this code has to finish executing synchronously before any Vue reactivity can come into play.

Vue reacts to changes asynchronously, however I presume that Vue's detection of changes must somehow boil down to synchronous callbacks registered on reactive things. E.G. replacing an array in-situ might result in one "array modified" callback, but adding N items to the array might result in N "item added" callbacks. At a later time Vue then responds to the changes captured by these callbacks.

src/views/Graph.vue
MetRonnie: So the subgraph for the margin/spacing is needed in the dot code but not in the rendered DOM?

src/components/cylc/GraphSubgraph.vue
```diff
@@ -0,0 +1,48 @@
...
+      stroke-width="8px"
+      stroke="grey"
+      stroke-dasharray="50 50"
+      />
+    <text
+      :x="labelXPosition"
+      :y="labelYPosition"
+      font-family="Roboto"
+      alignment-baseline="middle" text-anchor="middle"
+      font-size="70px"
```
oliver-sanders, line 20: The fontsize here *should* match the "70px" defined in the graphviz document. But, unfortunately, the font family does not match, because different fonts may render different sizes (for the same `px` value) this value is a bit off. The result is that the label may overflow the box:

Screenshot from 2024-05-14 09-55-37

To correct this, we could define`fontfamily="roberto"` in the dot code. This works a charm locally, but sadly, it doesn't appear to work in the browser. I expect graphviz doesn't have access to this font so is falling back to something else.

So I think the best solution is to apply a fudge factor here to ensure the text fits, here's the number I came up with by trial and error:

```suggestion
      font-size="60px"
```

Which results in:

Screenshot from 2024-05-14 10-10-39

src/components/cylc/GraphSubgraph.vue
```diff
@@ -0,0 +1,48 @@
...
+      fill="black"
+      stroke-width=1.5
```
oliver-sanders, line 22: Currently the white stroke outline is being drawn over the top of the text which makes it look thin and limits how wide we can make the stroke (mb).

I found a nicer way of doing this using "paint-order" to make the browser draw the stroke first, then draw the text over it.

```suggestion
      stroke-width=5
      paint-order="stroke";
```

This gives a generous margin to arrows:

Screenshot from 2024-05-14 10-13-52

oliver-sanders (approved):

oliver-sanders:
Bumping this forward to the 8.3 meta-release as this should be ready to go in before cylc-flow blockers.

It would be good to get a screenshot of this into the "changes" section in cylc-doc: https://cylc.github.io/cylc-doc/nightly_8.3/html/reference/changes.html#cylc-8-3

src/views/Graph.vue
```diff
@@ -483,7 +510,20 @@ export default {
       }
       return ret
     },
-    getDotCode (nodeDimensions, nodes, edges) {
+    /**
+     * Get the nodes binned by cycle point
+     *
+     * @param {Object[]} nodes
+     * @returns {{ [dateTime: string]: Object[] nodes }} mapping of node to their cycle point.
```
MetRonnie, line 517: Not quite valid JSDoc syntax I don't think.
```suggestion
     * @returns {{ [dateTime: string]: Object[] }} mapping of cycle points to nodes
```

src/components/cylc/GraphSubgraph.vue
```diff
@@ -0,0 +1,49 @@
+<template>
```
MetRonnie, line 1: Missing the copyright notice. Also could do with a comment explaining what this component is for.

src/components/cylc/GraphSubgraph.vue
```diff
@@ -0,0 +1,49 @@
...
+      type: Object,
+      required: true
+    }
+  },
+  computed: {
+    labelXPosition () {
+      return (parseInt(this.subgraph.x) + (parseInt(this.subgraph.width) / 2))
+    },
+    labelYPosition () {
+      return (parseInt(this.subgraph.y) + 90)
```
MetRonnie, line 45: Worth a comment explaing why 90?

MetRonnie:
Got a PR against this at https://github.com/markgrahamdawson/cylc-ui/pull/5

MetRonnie:
Managed to get a workflow into a state where there's an grouping with the title "undefined". And transposing this graph gives a console error. Will see if I can figure out anything else tomorrow

image

markgrahamdawson:
> to a state where there's an grouping with the title "undefined". And transposing this graph gives a console error. Will see if I can figure out anything else tomorrow

Do you have the flow.cylc file?

markgrahamdawson:
I've also noticed that on the complex workflow switching on the 'group by cycle point' it can result in some graph edges (I think thats the correct term for arrows) go nowhere...

'group by cycle point off'
image


'group by cycle point on'
image

oliver-sanders:
Hmm, I didn't notice any issues with the complex workflow. Will need the dot code to debug this. Given that the edges (yes that's right word) lead outside of the subgraph, I expect there may be an issue with the formatting of the subgraph dot code.

When I read through the code I think I spotted a missing newline at the end of the subgraph. I thought it was harmless, but it might be worth checking that's not causing issues.

MetRonnie:
> Managed to get a workflow into a state where there's an grouping with the title "undefined"

Also the 3 tasks are showing as "state unknown" while the jobs are showing as running. But these are from an earlier cycle and have actually already succeeded. Might be something to do with simulation mode. Or actually #1635?

MetRonnie (approved):

src/services/mock/json/workflows/multi.json
```diff
@@ -141,9 +141,11 @@
           {
             "id": "~user/other/multi/run2//1/foo",
             "state": "waiting",
+            "cyclePoint": "1",
```
oliver-sanders, line 144:
```suggestion
```

src/services/mock/json/workflows/multi.json
```diff
@@ -157,9 +159,11 @@
           {
             "id": "~user/other/multi/run2//2/foo",
             "state": "waiting",
+            "cyclePoint": "2",
```
oliver-sanders, line 162:
```suggestion
```

src/views/Graph.vue
```diff
@@ -483,7 +509,20 @@ export default {
...
+    /**
+     * Get the nodes binned by cycle point
+     *
+     * @param {Object[]} nodes
+     * @returns {{ [dateTime: string]: Object[] }=} mapping of cycle points to nodes
+     */
+    getCycles (nodes) {
+      if (!this.groupCycle) return
+      return nodes.reduce((x, y) => {
+        (x[y.node.cyclePoint] ||= []).push(y)
```
oliver-sanders, line 521:
```suggestion
        (x[y.tokens.cycle] ||= []).push(y)
```

oliver-sanders (commented):
Regards that "undefined" error for tasks with the missing `cyclePoint` field.

There is a slightly nicer way to fetch the cycle point that is available for all tasks by default (no need to request the `cyclePoint` field explicitly) using the "tokens" which are provided for every object in the store. Due to the way tokens are provisioned, the cycle cannot be undefined by accident (it's derived from the ID).

I haven't tested with these changes so worth giving them a spin before committing to them, but hopefully that'll resolve the error.

oliver-sanders (approved):
(re-approving)

oliver-sanders:
:+1:

oliver-sanders:
(added a changes entry in cylc-doc https://github.com/cylc/cylc-doc/pull/733)

PR #1773, opened 2024-04-30
Switch to `lodash-es` for treeshaking

MetRonnie:
TIL the [`lodash`](https://www.npmjs.com/package/lodash) package is CommonJS and isn't tree-shaken!

This PR replaces `lodash` with [`lodash-es`](https://www.npmjs.com/package/lodash-es) which is an ES modules version of the package which allows tree-shaking.

I've confirmed this by comparing the non-minified built files before and after. There are a bunch of lodash functions included with the build on master that are not on this branch. The total reduction in the minified build is something like ~70kB

**Check List**

- [x] Changelog not needed

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -55,6 +55,7 @@ export default defineConfig(({ mode }) => {
       alias: {
         '@': path.resolve(__dirname, './src'),
         $tests: path.resolve(__dirname, './tests'),
+        lodash: 'lodash-es',
         react: 'preact/compat',
         'react-dom': 'preact/compat',
       }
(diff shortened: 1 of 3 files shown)


vite.config.js
```diff
@@ -55,6 +55,7 @@ export default defineConfig(({ mode }) => {
       alias: {
         '@': path.resolve(__dirname, './src'),
         $tests: path.resolve(__dirname, './tests'),
+        lodash: 'lodash-es',
```
MetRonnie, line 58: This means we don't have to update all the imports

markgrahamdawson (approved):

hjoliver (approved):

PR #927, opened 2022-02-15
Prevent mutation menu closing when clicking on another task or cycle

MetRonnie:
These changes close #916

demo6

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Appropriate tests are included (e2e).
- [ ] Appropriate change log entry included.
- [x] No documentation update required.

PR #1775, opened 2024-05-02
Simplify mutation menu

MetRonnie:
By figuring out how to use `VMenu` better I was able to drop a fair amount of cruft.

We were originally using `VMenu` but through various attempts to get it to behave the way we want (e.g. #927, #1269) I ended up re-implementing some of the behaviour manually (e.g. click-outside behaviour).

It turns out the key to using `VMenu` and getting it to behave the way we want is not too difficult:
- `:target` prop - this allows Vuetify to take care of positioning relative to the `c-data-interactive` element that was clicked on
- `:key` prop - this allows the menu to re-open when clicking between `c-data-interactive` elements  by using a unique identifier for each element (so Vue knows a different element has been clicked)

Update: also renamed the `v-cylc-object` directive to `v-command-menu` as that is more descriptive IMO

**Check List**

- [x] Tests are included 
- [x] `CHANGES.md` entry not needed as invisible to users


oliver-sanders (commented):
Looks good to me.

Spotted one issue in the graph view where opening the menu on different jobs didn't behave as elsewhere. Haven't checked this against master, however, it appears to be fixed with #1780

MetRonnie:
https://github.com/cylc/cylc-ui/compare/2309827454a7782fab1888be0f2c4b9c8d6370ea..e5363541968df64c5fa872544717cb5acbfbafae should make it immune to that kind of thing

oliver-sanders:
Fixed.

oliver-sanders (approved):

MetRonnie:
@markgrahamdawson Tagged you for review. If you've not come across the mutation menu code before, it might be best for me to explain how it works before you review

MetRonnie:
I've also renamed `v-cylc-object` to `v-command-menu` as that is more descriptive IMO

markgrahamdawson (approved):

oliver-sanders (approved):
I found one "quirk" when a task disappears from the n-window and is removed from the view, any open menu will jump to the top-left of the screen (because the target node has been removed).

Would be beneficial if we could disable this behaviour, but the menu remains functional so not a biggie.

PR #1778, opened 2024-05-03
Include XSRF token in userprofile request

minrk:
JupyterHub 4.1 applies XSRF checks consistently to authenticated GET requests, so apply the same `getCylcHeaders` logic in the graphQL POST request to all requests (userprofile was the only other one I found). As a result, the `getCylcHeaders` is moved to a common location in `utils/url`, rather than being confined to graphQL.

This solves the userprofile request, described in https://github.com/jupyterhub/jupyterhub/issues/4800

Together with https://github.com/cylc/cylc-uiserver/pull/592, cylc works with JupyterHub 4.1.5

**Check List**

- [x] Tests are not needed for this change

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,6 +60,7 @@ below.
  - Jamie Allen
  - Christopher Bennett
  - Mark Dawson
+ - Min RK
 <!-- end-shortlog -->
 
 (All contributors are identifiable with email addresses in the git version
--- a/changes.d/1778.fix
+++ b/changes.d/1778.fix
@@ -1,0 +1,1 @@
+Compatibility with JupyterHub 4.1 XSRF changes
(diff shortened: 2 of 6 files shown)


oliver-sanders:
Many thanks for your fixes @minrk, greatly appreciated!

minrk:
Worth linking [this comment](https://github.com/jupyterhub/jupyterhub/issues/4800#issuecomment-2092600254) where I suggested using API tokens to make API requests, which avoids all xsrf fiddliness. This PR is the smallest change to keep things working as they are.

I didn't make the API token change because I don't know the best way to get the token from the Python to javascript in your stack, but if someone wants to take that on, the token is available as `token = self.hub_auth.get_token(self)` when using JupyterHub authentication, and can then be injected into templates for authenticated pages, etc.

MetRonnie (approved):

oliver-sanders (approved):
Makes sense, tested with JupyterHub 4.1.5, couldn't force a page load error. Many thanks!

PR #1780, opened 2024-05-03
graph: fix graph rendering

oliver-sanders:
* There appears to have been a recent change in how multiple browsers measure the "bounding box" of SVG elements.
* This caused graph nodes to be measured much larger than the area they actually render to.
* This messed with the graph because we feed these dimensions into GraphViz when creating the virtual graph causing the nodes in the virtual graph to be given too much space. An error which is subsequently translated to the SVG graph.
* The changes in "bounding box" measurement appear to affect SVG Symbols and rotations.
* This commit removes the use of SVG symbols (and their related "use" elements) and moves the rotation from the circle element to a group.
* The use of SVG symbols made it easier for us to capture click events from behind the svgPanZoom layer.
* Removing SVG symbols required changes to the way click events are captured in the graph. Note in SVG the click events land on graphical elements (i.e. children), not their containing groups (i.e. parents).

I think this will do it, it's also a simplification that uses fewer elements in the graph which is good. We will need to try this out with a bunch of different browsers, especially Safari as we previously encountered an issue with Safari calculating the rotation centre differently.

Things to test:
* Opening / closing the command menu by clicking on icons in the graph view and elsewhere.
* The progress animations rendering correctly in the graph view and elsewhere.
* The graph layout fix itself.

**Check List**

- [ ] Please check the appearance of the generated component test screenshots by hand (these are not presently compared to known good output)

### DIFF
--- a/src/components/cylc/GraphNode.vue
+++ b/src/components/cylc/GraphNode.vue
@@ -18,43 +18,26 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 <template>
   <g class="c-graph-node">
     <!-- the task icon -->
-    <symbol :id="nodeID" viewBox="-40 -40 140 140">
-      <!--
-        Use a "symbol" for the task node in order to apply a viewBox to it.
-        This both contains it and makes it clickable.
-
-        NOTE: Due to the viewBox we use here the coordinate system ends up
-        offset by -20px. This doesn't impact most things, however, rotations
-        can be sensitive to this change causing the rotated elements to end up
-        in the wrong places. To counteract this we provide the coordinate offset
-        to the task component.
-      -->
-      <SVGTask
-        :task="task.node"
-        :modifierSize="0.5"
-        :startTime="startTime"
-        :coordinateOffset="-20"
-      />
-    </symbol>
-    <use
-      :href="`#${nodeID}`"
-      x="0" y="0"
-      width="150" height="150"
+    <SVGTask
+      :task="task.node"
+      :modifierSize="0.5"
+      :startTime="startTime"
+      viewBox="-40 -40 140 140"
       v-cylc-object="task"
+      x="0" y="0"
     />
 
+    <!-- the label -->
     <g :transform="labelTransform">
-      <!-- the task name -->
       <text
-        x="180" y="70"
+        x="130" y="25"
         font-size="45"
       >
         {{ task.name }}
       </text>
 
-      <!-- the cycle point -->
       <text
-        x="180" y="115"
+        x="130" y="65"
         font-size="30"
       >
         {{ task.tokens.cycle }}
@@ -64,7 +47,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
     <!-- the job(s) -->
     <g
       transform="
-        translate(180, 125)
+        translate(130, 75)
         scale(0.3, 0.3)
       "
     >
@@ -77,23 +60,10 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           scale(${ (index === 0) ? mostRecentJobScale : '1' })
         `"
       >
-        <symbol
-          :id="`${nodeID}-${index}`"
+        <job
+          :svg="true"
+          :status="job.node.state"
           viewBox="0 0 100 100"
-          :class="`job_theme--${jobTheme}`"
-        >
-          <!--
-            Use a "symbol" for job nodes in order to make them clickable.
-            The job theme must be set on the "symbol" for styling to work.
-          -->
-          <job
-            :svg="true"
-            :status="job.node.state"
-          />
-        </symbol>
-        <use
-          :href="`#${nodeID}-${index}`"
-          width="100" height="100"
           v-cylc-object="job"
         />
       </g>
@@ -174,7 +144,7 @@ export default {
       if (this.jobs.length) {
         return ''
       }
-      return 'translate(0, 14)'
+      return 'translate(0, 20)'
     },
     previousJobOffset () {
       // the most recent job is larger so all subsequent jobs need to be bumped
--- a/src/components/cylc/GraphNode.vue
+++ b/src/components/cylc/GraphNode.vue
@@ -60,7 +60,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           scale(${ (index === 0) ? mostRecentJobScale : '1' })
         `"
       >
-        <job
+        <Job
           :svg="true"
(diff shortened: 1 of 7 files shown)


src/components/cylc/GraphNode.vue
```diff
@@ -18,43 +18,26 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 <template>
   <g class="c-graph-node">
     <!-- the task icon -->
-    <symbol :id="nodeID" viewBox="-40 -40 140 140">
```
oliver-sanders, line 21: It would appear these symbol/use objects can no longer be reliably measured :shrug:.

Symbols are kinda like "sprites", usually used for repeated elements in the document for efficiency reasons which don't apply for our single-use use case.

We were using symbols because these were more easily clickable. Click events landed on the parent "use" element and they worked through the svgPanZoom layer with its default click event configuration.

src/components/cylc/GraphNode.vue
```diff
@@ -18,43 +18,26 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-        offset by -20px. This doesn't impact most things, however, rotations
-        can be sensitive to this change causing the rotated elements to end up
-        in the wrong places. To counteract this we provide the coordinate offset
-        to the task component.
-      -->
-      <SVGTask
-        :task="task.node"
-        :modifierSize="0.5"
-        :startTime="startTime"
-        :coordinateOffset="-20"
```
oliver-sanders, line 36: Coordinate offset is no longer needed now we are specifying the viewbox on the group rather than the symbol.

src/components/cylc/GraphNode.vue
```diff
@@ -18,43 +18,26 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-      />
-    </symbol>
-    <use
```
oliver-sanders, line 39: ("symbol"s define elements, "use" elements add them into the document)

src/components/cylc/GraphNode.vue
```diff
@@ -18,43 +18,26 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
       v-cylc-object="task"
+      x="0" y="0"
     />
 
+    <!-- the label -->
     <g :transform="labelTransform">
-      <!-- the task name -->
       <text
-        x="180" y="70"
+        x="130" y="25"
```
oliver-sanders, line 33: Coordinates have moved due to the way the task icon is added to the page.

src/components/cylc/GraphNode.vue
```diff
@@ -77,23 +60,10 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           scale(${ (index === 0) ? mostRecentJobScale : '1' })
         `"
       >
-        <symbol
-          :id="`${nodeID}-${index}`"
+        <job
+          :svg="true"
+          :status="job.node.state"
           viewBox="0 0 100 100"
-          :class="`job_theme--${jobTheme}`"
```
oliver-sanders, line 83: Job theme no longer needs to be specified because we are using the job icon in the document natively rather than embedding it as a symbol.

src/components/cylc/cylcObject/Menu.vue
```diff
@@ -305,9 +305,24 @@ export default {
...
+      // menu is open for
+      let target = e.target
+      while (target) {
+        if (target?.getAttribute('data-c-interactive')) {
+          // keep the menu open
+          // this.showMenu = true
+          shouldClose = false
+          break
+        }
+        target = target.parentElement
```
oliver-sanders, line 321: When we click on a task icon in the graph, the click event will land on render elements, e.g. the outline `<circle />` element within the task group rather than the task `<g />` element itself (the bit which has the `v-cylc-object` directive attached to it.

As a result we can't just look for the `c-interactive` data attribute, we have to walk up the tree as it will belong to a parent element not the thing we clicked on.

This will go away with Ronnie's PR.

src/styles/index.scss
```diff
@@ -58,11 +58,6 @@ html {
   &:hover {
     cursor: pointer;
   }
-
-  > * {
-    // Prevent click handler from thinking any child elements are the target
-    pointer-events: none;
-  }
```
oliver-sanders, line 65: This needs to go to make icons clickable in the graph view.

See https://github.com/cylc/cylc-ui/pull/1780/files#r1589015567

src/components/cylc/GraphNode.vue
oliver-sanders: Here's a screenshot showing the GraphNode component selected in the FF devtools showing the incorrect dimensions returned before this change:

graph-error

src/views/Graph.vue
```diff
@@ -404,7 +404,7 @@ export default {
           zoomEnabled: true,
           dblClickZoomEnabled: true,
           mouseWheelZoomEnabled: true,
-          preventMouseEventsDefault: true,
+          preventMouseEventsDefault: false,
```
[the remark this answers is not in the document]
markgrahamdawson: With this set to false things like double click dont zoom

MetRonnie (approved):
And tested in Firefox 115.10.0esr on linux

src/views/Graph.vue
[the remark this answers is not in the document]
markgrahamdawson: The widget seems to work the same if this is true, false or removed completely?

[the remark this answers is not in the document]
oliver-sanders: Ha, ok, thought I had to change this one, will look in...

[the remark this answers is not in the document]
oliver-sanders: Confirmed and reverted.

src/components/cylc/GraphNode.vue
```diff
@@ -77,23 +60,10 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           scale(${ (index === 0) ? mostRecentJobScale : '1' })
         `"
       >
-        <symbol
-          :id="`${nodeID}-${index}`"
+        <job
```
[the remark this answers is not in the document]
markgrahamdawson: Does this need a change or can I merge this one in?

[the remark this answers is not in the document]
oliver-sanders: I was just waiting to see if there were any other suggestions before applying them in a batch (I try to avoid cluttering things with "address feedback" commits).

markgrahamdawson (approved):

MetRonnie (approved):

PR #1786, opened 2024-05-07
Fix N-window selector sending mutation upon navigation to stopped workflows

MetRonnie:
Fixes an unreleased bug on master. Also adds some tests. Follow-up to #1755 

#### How to reproduce the bug on master

1. Start a workflow and use the UI to set the n-window to a value other than 1, e.g. 2
2. Stop the workflow
3. Navigate to the dashboard or a different workflow
4. Navigate back to the original workflow
5. You should see an error snackbar for a failed mutation

#### Check List

- [x] Tests are included
- [x] No changelog entry needed as fixes unreleased bug

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "preview": "node ./scripts/concurrently.cjs serve:jupyterhub preview",
     "serve": "node ./scripts/concurrently.cjs serve:jupyterhub serve:vue",
     "serve:vue": "vite --mode offline",
-    "serve:jupyterhub": "nodemon -e js,mjs,cjs,json src/services/mock/json-server.cjs",
+    "serve:jupyterhub": "nodemon src/services/mock/json-server.cjs --watch src/services/mock/",
     "test:component": "cypress open --component",
     "test:e2e": "yarn run serve e2e:open",
     "test:unit": "vitest",
--- a/src/components/cylc/workflow/Toolbar.vue
+++ b/src/components/cylc/workflow/Toolbar.vue
@@ -89,12 +89,14 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         :disabled="isStopped"
         link
         size="small"
+        data-cy="n-win-selector"
       >
         N={{ nWindow }}
         <v-menu
           activator="parent"
           :close-on-content-click="false"
           max-width="400"
+          data-cy="n-win-popup"
         >
           <v-card title="Graph Window Depth">
             <v-card-text>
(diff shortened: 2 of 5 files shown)


src/components/cylc/workflow/Toolbar.vue
```diff
@@ -357,9 +372,20 @@ export default {
...
+      },
+      async set (val) {
+        if (val == null || this.isStopped) return
+
+        this.changingNWindow = true
+        if (await this.setGraphWindow(val)) {
+          await until(() => this.currentWorkflow?.node?.nEdgeDistance === val)
+        }
+        this.changingNWindow = false
+      },
```
MetRonnie, line 388: By switching to a writeable computed property, the state relies on the subscription value and so cannot deviate from it, fixing the bug. But this means there is a delay in the update of the value displayed in the input after choosing a new value, so I added a loading spinner to give immediate feedback to the user

src/components/cylc/workflow/Toolbar.vue
```diff
@@ -357,9 +372,20 @@ export default {
         )
       }
     },
-    nEdgeDistance () {
-      // the graph window distance reported by the scheduler
-      return this.currentWorkflow?.node?.nEdgeDistance
+    nWindow: {
+      get () {
+        // the graph window distance reported by the scheduler
+        return this.currentWorkflow?.node?.nEdgeDistance ?? 1
```
oliver-sanders, line 378: For stopped workflows, the icon initially says `n=1`, but then pings to `n=0`.

Not a problem (and technically very accurate), but curious that this `n=1` default is overridden.

oliver-sanders (approved):

src/components/cylc/workflow/Toolbar.vue
MetRonnie: Not sure why, but stopped workflows have a reported `nEdgeDistance` of 0

image

package.json
```diff
@@ -17,7 +17,7 @@
     "preview": "node ./scripts/concurrently.cjs serve:jupyterhub preview",
     "serve": "node ./scripts/concurrently.cjs serve:jupyterhub serve:vue",
     "serve:vue": "vite --mode offline",
-    "serve:jupyterhub": "nodemon -e js,mjs,cjs,json src/services/mock/json-server.cjs",
+    "serve:jupyterhub": "nodemon src/services/mock/json-server.cjs --watch src/services/mock/",
```
markgrahamdawson, line 20: Whats this? is it related to the PR?

MetRonnie: Not related to the PR but worth getting in IMO.

The mock JSON server was unnecessarily restarting every time any JS file was edited, when it only needs to restart when files in `src/services/mock` are changed. See https://github.com/cylc/cylc-ui/pull/1786/commits/447657b1b22736945ab9879a0beaa6936b8398f3

wxtim (approved):
Works for me. Have not inspected the code. :)

markgrahamdawson (approved):

PR #1789, opened 2024-05-08
Disable Edit Runtime in mutation menu when workflow is stopped/stopping

MetRonnie:
Related to https://github.com/cylc/cylc-flow/pull/6095

Edit Runtime should be disabled in the mutation menu for stopped/stopping workflows as it is not possible/sensical to broadcast to them.

Includes some tidying

**Check List**

- [x] No tests included (this is what TypeScript is for)
- [x] No changelog entry needed, as minor

### DIFF
--- a/src/utils/aotf.js
+++ b/src/utils/aotf.js
@@ -50,7 +50,7 @@ import {
 import { Alert } from '@/model/Alert.model'
 import { store } from '@/store/index'
 import { Tokens } from '@/utils/uid'
-import { WorkflowState } from '@/model/WorkflowState.model'
+import { WorkflowState, WorkflowStateNames } from '@/model/WorkflowState.model'
 
 /** @typedef {import('@apollo/client').ApolloClient} ApolloClient */
 /** @typedef {import('graphql').IntrospectionInputType} IntrospectionInputType  */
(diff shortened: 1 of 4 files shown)


src/utils/aotf.js
```diff
@@ -118,27 +119,30 @@ import { WorkflowState } from '@/model/WorkflowState.model'
...
-  reload: mdiReload,
-  remove: mdiMinusCircleOutline,
-  resume: mdiPlay,
-  set: mdiVectorPolylineEdit,
-  stop: mdiStop,
-  trigger: mdiCursorPointer
+ * @param {string} name - mutation name
+ * @returns {string} - icon svg path
+ */
+export function getMutationIcon (name) {
```
wxtim, line 125: This reads more plesantly IMO: Are the computational advantages too?

wxtim (approved):
😄

src/utils/aotf.js
MetRonnie: The main advantage is it provides a proper way of having a default fallback icon.

In terms of performance, it shouldn't matter for this use, as it is only called a dozen times or so. In compiled languages, `switch` statements are usually equivalent to lookup tables. However I was curious and looked into JS, but can't find any definitive information. It looks like it may be entirely engine-dependent: https://jsperf.app/pekoba

MetRonnie:
I think 1 review should do

PR #1794, opened 2024-05-13
Fix router issues

MetRonnie:
- Set page title per-route rather than per-view
  - Multiple views may be included in a page, so the previous approach was flawed
  - This removes a blocker from #1785 (for some reason that Vuetify update has uncovered this flaw)
- Improve router error handling & tidy
- Address this warning that has been around for ages:

    ```
    [Vue Router warn]: <router-view> can no longer be used directly inside <transition>
    ```
- Removes `@unhead/vue` dependency as it is a bit overkill for just setting the page title, which can be done simply using `document.title = title` 

**Check List**

- [x] Tests are included 
- [x] Changelog not needed as not user-facing


oliver-sanders (approved):
(Not too fussed about loosing the fade transition)

wxtim (approved):

PR #1810, opened 2024-05-24
added family grouping plus family and cycle collapsing

markgrahamdawson:
Partly addresses issue https://github.com/cylc/cylc-ui/issues/1130
The grouping of nodes by cycle point is completed in this pr https://github.com/cylc/cylc-ui/pull/1763

----Notes on work----

Some ideas for a unified approach to grouping/collapsing cycles/families. I'm suggesting unifying the handling of cycles and families (note, cycles represent the "root" family so they are essentially the same thing).

**Grouping/Ungrouping** - Drawing dashed boxes around a cycle/family.

**Collapsing/Expanding** - Reducing a family down to a single node.

Limitations of the Cylc 7 approach:

- Once you expand a family it's gone, the tasks which belong to the expanded family are mixed in with other tasks in the graph, you can not tell what family they belong to. This is an issue if the user wants to examine a component within the workflow.
- No visibility of the inheritance hierarchy (i.e. what can we expand/collapse).
- No visibility of what you have expanded/collapsed (i.e. where are we in the hierarchy).

Note, for simplicity, this approach groups/collapses all instances of selected families rather than managing this at a per-cycle level. I think this is probably more aligned with expectations, but does represent a minor limitation, e.g. there's no ability to collapse all but one cycle. The ability to expand/collapse specific cycle instances would be a reasonable enhancement.

**Design Sketch**
image

Had a quick discussion on this (more to come):

- Can't really think of a valid use case for collapsing all cycles (users would do this in the tree view if they wanted to), so perhaps treat cycles differently (i.e. collapse per-cycle rather than all cycles) and remove from the menus.
- Better expand/collapse icon (obviously).
- The Cylc 7 default of only expanding the cycle point (i.e. show top-level families only) is a reasonable protection for viewing large workflows. We might want to continue with this, or perhaps do something smart (e.g. only collapse families if there are lots of tasks on load).

**Check List**


markgrahamdawson:
Still to do (in order of priority).

1. Work out how to implement nested families
2. Get graph status showing on graph nodes
3. When switching to another workflow and returning the graph fails to load
4. Expand collapse icons on each node (and subgraph) on the graph
5. If collapsed by family+ grouped by family - dont draw graph around collapsed node (see cycle point functionality)
6. The edges and nodes are now let variables (so they can be updated) - is this ok? Might want to change them to something else?

src/views/Graph.vue
```diff
@@ -394,6 +509,170 @@ export default {
...
+      if (nodeSearch === -1) {
+        return null
+      } else {
+        return nodeSearch
+      }
+    },
+    createEdge (sourceName, targetName, sourceCyclePoint, targetCyclePoint) {
+      // adds a new edge object to 'edges' array
+      const edgePath = this.workflowIDs[0]
+      const edgeId = `${edgePath}//$edge|${sourceCyclePoint}/${sourceName}|${targetCyclePoint}/${targetName}`
```
oliver-sanders, line 580: If you can pass this function the source and target tokens rather than their IDs, then this becomes a lot eaisier.

I think we can do it like this:

```js
// untested
const edge_tokens = src_tokens.clone(cycle=`\$edge|${src_tokens.relativeID}|${tgt_tokens.relativeID}`, task=null, job=null)
const edge_tokens.edge = [src_tokens, tgt_tokens]
const edge_id = edge_tokens.id
```

markgrahamdawson:
Im struggling with grouping by families due to the fact that they can be nested ...


### The context
Collapsing and Expanding nodes is easier as the .dot file just needs the node/s added/removed ....

```          
"~mdawson/runtime-tutorial-families-two/run1//20240724T0000Z/get_observations_aldergrove" [
            label=<
              <TABLE HEIGHT="132.447509765625">
                <TR>
                  <TD PORT="in" WIDTH="100"></TD>
                </TR>
                <TR>
                  <TD PORT="task" WIDTH="100" HEIGHT="132.447509765625">icon</TD>
                  <TD WIDTH="628.4224853515625">~mdawson/runtime-tutorial-families-two/run1//20240724T0000Z/get_observations_aldergrove</TD>
                </TR>
                <TR>
                  <TD PORT="out" WIDTH="100"></TD>
                </TR>
              </TABLE>
            >
          ]
```

and the edge (relationship between the node and other nodes) defined...
`  "~mdawson/runtime-tutorial-families-two/run1//20240724T0300Z/get_observations_aldergrove":out -> "~mdawson/runtime-tutorial-families-two/run1//20240724T0300Z/consolidate_observations":in`

This means I dont have to directly deal with a hierarchy (nested structure) as things are just being added/removed.

For grouping the syntax is a little different, using subgraphs ...

```
                  subgraph cluster_margin_family16
                  {
                    margin=100.0
                    label="margin"
                    subgraph cluster_family16 {"~mdawson/runtime-tutorial-families-two/run1//20240724T0000Z/get_observations_shetland","~mdawson/runtime-tutorial-families-two/run1//20240724T0000Z/get_observations_aldergrove";

                      label = "GET_OBSERVATIONS_NORTH20240724T0000Z";

                      fontsize = "70px"
                      style=dashed
                      margin=60.0
                  }
                }
```
For grouping by cycle point there are no nested cycle points (doesnt make sense) so its just a case of making subgraph for each cycle. The subgraphs do need to account for the fact that the nodes may have been expanded or collapsed but that can be managed by calculating what nodes need to be included from the `nodes` variable - which is up-to-date with what has been expanded/collapsed. Also understanding the hierarchical relationship is easier because its contained in the node id whether it has been expanded or collapsed - it will always have a cycle associated with it. 

### The problem
The problem is with nested grouping which is relevant for
1.  family groups inside cycle groups
2.  family groups inside family groups.

The way this is represented in the .dot code is by having subgraphs within subgraphs...

```
subgraph FAMILY {
  "~mdawson/run-name/run1/cycle/SUBFAMILY", "~mdawson/run-name/run1/cycle/task1", "~mdawson/run-name/run1/cycle/task2", ;
  label = Family
  subgraph SUBFAMILY {
    "~mdawson/run-name/run1/cycle/task1", "~mdawson/run-name/run1/cycle/task2", ;
    label = SubFamily
  }
}
```
image

The above is an simple example of some graphviz code for a simple nested family situation. Below is an example for a more complicated one...
image

```
subgraph FAMILY {
  "~mdawson/run-name/run1/cycle/SUBFAMILY_A", "~mdawson/run-name/run1/cycle/SUBFAMILY_B", "~mdawson/run-name/run1/cycle/SUBFAMILY_A1", "~mdawson/run-name/run1/cycle/SUBFAMILY_A2", "~mdawson/run-name/run1/cycle/SUBFAMILY_B1", "~mdawson/run-name/run1/cycle/SUBFAMILY_B2", "~mdawson/run-name/run1/cycle/Task_y", "~mdawson/run-name/run1/cycle/Task_x", "~mdawson/run-name/run1/cycle/Task_m", "~mdawson/run-name/run1/cycle/Task_n", "~mdawson/run-name/run1/cycle/Task_g", "~mdawson/run-name/run1/cycle/Task_h", "~mdawson/run-name/run1/cycle/Task_i", "~mdawson/run-name/run1/cycle/Task_j" ;
  label = Family
  subgraph SUBFAMILY_A {
 "~mdawson/run-name/run1/cycle/SUBFAMILY_A1", "~mdawson/run-name/run1/cycle/SUBFAMILY_A2", "~mdawson/run-name/run1/cycle/Task_y", "~mdawson/run-name/run1/cycle/Task_x", "~mdawson/run-name/run1/cycle/Task_m", "~mdawson/run-name/run1/cycle/Task_n" ;
    label = SubFamily_A
        subgraph SUBFAMILY_A1 { "~mdawson/run-name/run1/cycle/Task_y", "~mdawson/run-name/run1/cycle/Task_x" ;
          label = SubFamily_A1
      }
      subgraph SUBFAMILY_A2 { "~mdawson/run-name/run1/cycle/Task_m", "~mdawson/run-name/run1/cycle/Task_n" ;
          label = SubFamily_A1
      }
  }
  subgraph SUBFAMILY_B {
 "~mdawson/run-name/run1/cycle/SUBFAMILY_B1", "~mdawson/run-name/run1/cycle/SUBFAMILY_B2", "~mdawson/run-name/run1/cycle/Task_g", "~mdawson/run-name/run1/cycle/Task_h", "~mdawson/run-name/run1/cycle/Task_i", "~mdawson/run-name/run1/cycle/Task_j" ;
    label = SubFamily_B
        subgraph SUBFAMILY_B1 { "~mdawson/run-name/run1/cycle/Task_g", "~mdawson/run-name/run1/cycle/Task_h" ;
          label = SubFamily_B1
      }
      subgraph SUBFAMILY_B2 { "~mdawson/run-name/run1/cycle/Task_i", "~mdawson/run-name/run1/cycle/Task_j" ;
          label = SubFamily_B1
      }
  }
}
```

The subgraphs can be `n` layers deep so that needs to be handled programatically (cant be hard coded).
At the moment the graphviz .dot code is being written as an array of strings that all gets added to - pushing new values onto the end. And then using the join method on the array to make one big string.

oliver-sanders:
I think this is a problem that warrants recursion as it's tricky to unroll as an iterative loop.

Here's an idea of what that could look like (Python syntax):
* First, go through every task and create a dotcode entry for each (this is the bit that includes the `<TABLE />` label).
* Then go through every family in inheritance order and build the nested subgraphs for each, inserting the task entries we have just built into the relevant subgraph when we get to it.
* Then do the `'\n'.join(dotcode)` bit.

```python
from random import random

TASKS = {
    'foo': {
        'name': 'foo',
        'parent': 'FOO',
    },
    'FOO': {
        'name': 'FOO',
        'parent': 'root'
    },
    'bar': {
        'name': 'bar',
        'parent': 'BAR1',
    },
    'baz': {
        'name': 'baz',
        'parent': 'BAR2',
    },
    'BAR1': {
        'name': 'BAR1',
        'parent': 'BAR',
    },
    'BAR2': {
        'name': 'BAR2',
        'parent': 'BAR',
    },
    'root': {
        'name': 'root',
        'parent': None,
    },
}

TREE = {
    'root': {
        'FOO': None,
        'BAR': {
            'BAR1': None,
            'BAR2': None,
        },
    },
}

def add_subgraph(dotcode, pointer, graph_sections):
    for key, value in pointer.items():
        dotcode.append(
            f'subgraph cluster_{str(random())[2:]} {{'
            f'\nlabel = "{key}"'
        )

        if value:
            add_subgraph(dotcode, value, graph_sections)

        if key in graph_sections:
            dotcode.extend(graph_sections[key])

        dotcode.append('}')

    return dotcode

def get_dotcode(tasks):
    graph_sections = {}

    for task in tasks.values():
        parent = task['parent']
        if not parent:
            continue
        section = graph_sections.setdefault(parent, [])
        section.append(f'{task["name"]} [title="{task["name"]}"]')

    dotcode = ['digraph {']
    add_subgraph(dotcode, TREE['root'], graph_sections)
    return dotcode


for item in get_dotcode(TASKS):
    print(item)
```

```dot
digraph {

  subgraph cluster_23300787190407446 {
    label = "FOO"

    foo [title="foo"]
  }

  subgraph cluster_5025488657295563 {
    label = "BAR"

    subgraph cluster_2135762450670372 {
      label = "BAR1"

      bar [title="bar"]
    }

    subgraph cluster_4413670667138756 {
      label = "BAR2"

      baz [title="baz"]
    }

  BAR1 [title="BAR1"]
  BAR2 [title="BAR2"]
}
```


I haven't taken cycles into account in this solution, you'll need to add a `for cycle in cycles` loop at the top of this.

This solution will also add entries for families which have no tasks, so, you'll need some fancy logic for removing empty families, and any families that contain only empty families.

src/views/Graph.vue
```diff
@@ -387,6 +542,209 @@ export default {
...
+      edgeCheckTarget.forEach((edge) => {
+        removedEdges.push(edge)
+        const targetName = this.cylcTree.$index[edge.node.target].name
+        const targetCycle = this.cylcTree.$index[edge.node.target].tokens.cycle
+        const sourceName = this.cylcTree.$index[edge.node.source].name
+        const sourceCycle = this.cylcTree.$index[edge.node.source].tokens.cycle
+        edges = this.removeEdge(sourceName, sourceCycle, targetName, targetCycle, edges)
+      })
+      return [edges, removedEdges]
+    },
```
oliver-sanders, line 747: The source and target are available in the tokens (`edge[0] == source`, `edge[1] == target`) which avoids the need to go via the `cylcTree.$index`:

```suggestion
    removeEdgeBySource (edgeCheckSource, edges, removedEdges, config, cycle) {
      edgeCheckSource.forEach((edge) => {
        removedEdges.push(edge)
        edges = this.removeEdge(
          config.name,
          cycle,
          edge.tokens.edge[1].task,
          edge.tokens.edge[1].cycle,
          edges
        )
      })
      return [edges, removedEdges]
    },
    removeEdgeByTarget (edgeCheckTarget, edges, removedEdges) {
      edgeCheckTarget.forEach((edge) => {
        removedEdges.push(edge)
        edges = this.removeEdge(
          edge.tokens.edge[0].task,
          edge.tokens.edge[0].cycle,
          edge.tokens.edge[1].task,
          edge.tokens.edge[1].cycle, edges
        )
      })
      return [edges, removedEdges]
    },
```

src/views/Graph.vue
```diff
@@ -508,13 +866,101 @@ export default {
...
+     * Get the nodes binned by family
+     *
+     * @param {Object[]} nodes
+     * @returns {{ [dateTime: string]: Object[] }=} mapping of family to nodes
+     */
+    getFamilies (nodes) {
+      if (!this.groupFamily) return
+      return nodes.reduce((x, y) => {
+        if (y.node.firstParent) {
+          (x[y.node.firstParent.id.split('/')[y.node.firstParent.id.split('/').length - 1]] ||= []).push(y)
```
oliver-sanders, line 884: Probably something `Tokens` can do?

src/views/Graph.vue
```diff
@@ -508,13 +866,101 @@ export default {
...
+        }
+
+        if (value) {
+          this.addSubgraph(dotcode, value, graphSections)
+        }
+        if (Object.keys(graphSections).includes(key)) {
+          dotcode.push(graphSections[key.id])
+        }
+
+        if (key.node.name !== '01' &&
```
oliver-sanders, line 943: Why are we checking for the name `01`?

src/views/Graph.vue
```diff
@@ -594,6 +1083,8 @@ export default {
         }
       }
       ret.push('}')
+      // console.log('Dot Code')
+      // console.log(ret.join('\n'))
```
oliver-sanders, line 1087:
```suggestion
```

tests/unit/views/graph.vue.spec.js
```diff
@@ -116,215 +116,215 @@ describe('Graph view', () => {
...
-        },
-        namespaces () {
-          return namespaces()
-        },
-        cylcTree () {
-          return cylcTree
-        }
-      }
-    })
+  // it('it gets families', async () => {
```
oliver-sanders, line 119: Whoops!

Note, you can use `.only` during development to only run targeted tests, e.g:

```js
it.only('it works', ....
```

oliver-sanders:
@wxtim, could you join me on this review.

These things can be moved to follow up:

* Consider unifying the toolbar items (i.e. fewer buttons).
* Adding expand/collapse icons to families / cycles in the graph itself.
* Default collapse state (like the Cylc 7 GUI).
* Expand/collapse all buttons (or similar).

Note, the graph view is a pain to test.

wxtim:
[minor point] @oliver-sanders Observed that collapsed by Cycle point cycles probably don't need both of the Cycle point labels:

image

markgrahamdawson:
> ycle point cycles probably don't need both o

Would we rather keep the big one or the small one?

oliver-sanders:
> Would we rather keep the big one or the small one?

We should keep the big one.

Easy to do, but it'll need some `if` type logic in the `GraphNode` component. Can do as follow up.

changes.d/1810.feat.md
```diff
@@ -0,0 +1 @@
+Added graph view feature to group by family and collapse by family and cycle point
```
wxtim, line 1: Sig improvement in comprehendability IMO
```suggestion
Added graph view feature to group by family, and collapse by family and cycle point
```

wxtim:
Hi @markgrahamdawson 

I'm still getting missing dependency arrows, though it was harder to reproduce:

```cylc
[scheduler]
    allow implicit tasks = True
    cycle point format = %Y

[scheduling]
    initial cycle point = 1971
    [[graph]]
        P1Y = """
            X[-P1Y]:succeed-all => start_cycle
            start_cycle => X:succeed-all => _dummy_ => Y
        """

[runtime]
    [[root]]
        script = sleep 1000
    [[X, Y]]

    [[x]]
        inherit = X
    [[y]]
        inherit = Y
```

I had collapset 1972 and X

image

tests/e2e/specs/graph.cy.js
```diff
@@ -229,7 +229,7 @@ describe('Graph View', () => {
       })
   })
 
-  it('should collapse by family', () => {
+  it.only('should collapse by family', () => {
```
oliver-sanders: `.only`

src/components/cylc/ViewToolbar.vue
```diff
@@ -100,6 +129,22 @@ export default {
...
+  mounted () {
+    if (this.groups[0].title === 'Graph') {
+      const getControlValue = (groups, controlKey) => {
+        return groups[0].controls.find(control => control.key === controlKey).value
+      }
+      this.selectedItems.groupFamily = getControlValue(this.groups, 'groupFamily')
+      this.selectedItems.collapseCycle = getControlValue(this.groups, 'collapseCycle')
+      this.selectedItems.collapseFamily = getControlValue(this.groups, 'collapseFamily')
+    }
+  },
```
oliver-sanders, line 146: We should be able to avoid hardcoding graph view logic into the toolbar component.

This diff seems to play nicely with initial options:

```diff
diff --git a/src/components/cylc/ViewToolbar.vue b/src/components/cylc/ViewToolbar.vue
index cec9b130..fb26fd47 100644
--- a/src/components/cylc/ViewToolbar.vue
+++ b/src/components/cylc/ViewToolbar.vue
@@ -43,7 +43,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             </v-btn>
           </template >
           <v-treeview
-            v-model:selected="selectedItems[iControl.key]"
+            v-model:selected="iControl.value"
             v-on:update:selected="iControl.callback"
             :items="iControl.items"
             select-strategy='independent'
@@ -134,16 +134,6 @@ export default {
       selectedItems: {}
     }
   },
-  mounted () {
-    if (this.groups[0].title === 'Graph') {
-      const getControlValue = (groups, controlKey) => {
-        return groups[0].controls.find(control => control.key === controlKey).value
-      }
-      this.selectedItems.groupFamily = getControlValue(this.groups, 'groupFamily')
-      this.selectedItems.collapseCycle = getControlValue(this.groups, 'collapseCycle')
-      this.selectedItems.collapseFamily = getControlValue(this.groups, 'collapseFamily')
-    }
-  },
 
   computed: {
     iGroups () {
@@ -229,10 +219,9 @@ export default {
       control.callback()
       e.currentTarget.blur()
     },
-    select (control, e) {
+    select (control, value) {
       // call a control's callback
-      control.value = this.selectedItems[control.key]
-      this.$emit('setOption', control.key, control.value)
+      this.$emit('setOption', control.key, value)
     },
     getValues () {
       // an object with all defined values
```

wxtim:
I can't break it. 🚀

markgrahamdawson:
> [tangential to , but exacerbated by this PR]
> 
> We were talking about the funky lines GraphViz sometimes comes up with.
> 
> I suspect that these are the result of GraphViz trying to thread edges between tasks and subgraphs.
> 
> We currently apply spacing to subgraphs by nesting them inside another subgraph. There is a GraphViz [margin](https://graphviz.org/docs/attrs/margin/) attribute which might work for us?

On this one we, we do use that margin attribute - line 855 in the src/views/Graph.vue

src/views/Graph.vue
```diff
@@ -641,11 +1057,255 @@ export default {
...
+              // a family member up the tree is collapsed
+              const ancestor = this.allParentLookUp[nodeFirstFamily].some(element => {
+                return this.collapseFamily.includes(element)
+              })
+              if (firstParent && !ancestor) {
+                // the node is collapsed by its first parent
+                return nodeFirstFamily
+              } else if (ancestor) {
+                return this.allParentLookUp[nodeFirstFamily][0]
+                // this is almost certainly an over simplification -> better logic needed here
```
oliver-sanders, line 1117: TODO

I'm not 100% on what this function does, but I suspect it's missing a bit.

E.g, for this inheritance:

* A
  * A1
    * A11
      * A111

And this state:

| Fam | iscollapsed |
| --- |--- |
| A | :x: |
| A1 | :heavy_check_mark: |
| A11 | :heavy_check_mark: |
| A111 | :x: |

What family should be returned?

src/views/Graph.vue
```diff
@@ -641,11 +1057,255 @@ export default {
...
+                  [edges, removedEdges] = this.removeEdgeByTarget(edgeCheckTarget, edges, removedEdges)
+                }
+              }
+            })
+            // ...now we have removed any parts of child nodes that shouldnt be there we can add nodes and edges that should be...
+            // ---------------ADD NODES BASED ON FAMILY------------
+            if (!this.collapseCycle.includes(cycle)) { // cycle collapsing takes priority over family collapsing
+              nodes.push(indexSearch)
+            }
+            const isNodeCollapsedByFamily = (nodeFirstFamily) => {
```
oliver-sanders, line 1105: Can make this a method.

src/views/Graph.vue
```diff
@@ -508,12 +837,81 @@ export default {
...
+          children.length &&
+          this.groupFamily.includes(key.node.name) &&
+          !this.collapseFamily.includes(key.node.name) &&
+          !this.collapseFamily.includes(key.node.firstParent.name)) {
+          const nodeFormattedArray = children.filter((a) => {
+            // if its not in the list of families (unless its been collapsed)
+            const isFamily = !this.familyArrayStore.includes(a.name) || this.collapseFamily.includes(a.name)
+            // its the node has been removed/collapsed
+            const isRemoved = !removedNodes.includes(a.name)
+            return isFamily && isRemoved
```
oliver-sanders, line 907: This block looks identical to the one above (only without the isNumeric check).

So I think we can remove it completely like this:

```diff
diff --git a/src/views/Graph.vue b/src/views/Graph.vue
index 9973a43a..fa070fb2 100644
--- a/src/views/Graph.vue
+++ b/src/views/Graph.vue
@@ -855,7 +855,9 @@ export default {
             })
           }
         })
+
         // filter parent
+        let openedBrackets = false
         if (
           children.length &&
           this.groupFamily.includes(key.node.name) &&
@@ -872,6 +874,7 @@ export default {
             return isFamily && isRemoved && isNumeric
           }).map(a => `"${a.id}"`)
           if (nodeFormattedArray.length) {
+            openedBrackets = true
             dotcode.push(`
             subgraph cluster_margin_family_${key.name}${key.tokens.cycle}
               {
@@ -894,21 +897,8 @@ export default {
           dotcode.push(graphSections[key.id])
         }
 
-        if (
-          children.length &&
-          this.groupFamily.includes(key.node.name) &&
-          !this.collapseFamily.includes(key.node.name) &&
-          !this.collapseFamily.includes(key.node.firstParent.name)) {
-          const nodeFormattedArray = children.filter((a) => {
-            // if its not in the list of families (unless its been collapsed)
-            const isFamily = !this.familyArrayStore.includes(a.name) || this.collapseFamily.includes(a.name)
-            // its the node has been removed/collapsed
-            const isRemoved = !removedNodes.includes(a.name)
-            return isFamily && isRemoved
-          }).map(a => `"${a.id}"`)
-          if (nodeFormattedArray.length) {
-            dotcode.push('}}')
-          }
+        if (openedBrackets) {
+          dotcode.push('}}')
         }
       })
     },
```

src/views/Graph.vue
```diff
@@ -558,25 +956,66 @@ export default {
...
-            subgraph cluster_${i} {${nodeFormattedArray};\n
-              label = "${key}";\n
-              fontsize = "70px"
-              style=dashed
-              margin=60.0
-            }
-          }`)
+      const graphSections = {}
+      Object.keys(cycles).forEach((cycle, iCycle) => {
+        const indexSearch = Object.values(this.cylcTree.$index).filter((node) => {
```
oliver-sanders, line 961: This is searching over the full `cylcTree.$index`, this is the complete data store which also contains nodes for other workflows.

We only need to iterate over the cycles in this workflow, i.e. `Object.keys(cycles).forEach((key, i) => {`.

src/views/Graph.vue
```diff
@@ -558,25 +956,66 @@ export default {
...
+          return node.tokens.cycle === cycle
         })
-      }
+        if (indexSearch.length && !this.collapseCycle.includes(cycle)) {
+          indexSearch.forEach((task) => {
+            const parent = task.node.firstParent
+            if (!parent) { return }
+            const section = graphSections[parent.id] ??= []
```
oliver-sanders, line 968: Can avoid declaring an extra variable:

```suggestion
            if (!task.node.firstParent) { return }
            const section = graphSections[task.node.firstParent.id] ??= []
```

src/views/Graph.vue
```diff
@@ -329,6 +404,68 @@ export default {
...
+      const store = []
+      this.cycleArrayStore.forEach((name, id) => {
+        store.push({
+          id,
+          name
+        })
+      })
+      return store
+    },
+    allParentLookUp () {
```
wxtim, line 432: As someone who isn't super fluent in JS, could I request more docstring?

src/views/Graph.vue
```diff
@@ -329,6 +404,68 @@ export default {
...
+            parents = childNode.node.parents
+          })
+        }
+        lookup[namespace.name] = array
+      })
+      return lookup
+    },
+    allChildrenLookUp () {
+      const lookup = {}
+      // Calculate some values for familes that we need for the toolbar
```
wxtim, line 451: Which values?

src/views/Graph.vue
```diff
@@ -508,12 +836,78 @@ export default {
...
+          this.addSubgraph(dotcode, value, graphSections)
+        }
+        if (Object.keys(graphSections).includes(key)) {
+          dotcode.push(graphSections[key.id])
+        }
+        if (openedBrackets) {
+          dotcode.push('}}')
+        }
+      })
+    },
```
wxtim, line 910: This seemed fine:

```suggestion
      pointer.children.forEach((value, i) => {
        const children = this.allChildrenLookUp[value.id]
        if (!children) { return }
        const removedNodes = new Set()
        children.forEach((a) => {
          if (this.collapseFamily.includes(a.name)) {
            a.children.forEach((child) => {
              removedNodes.add(child.name)
            })
          }
        })
        // filter parent
        let openedBrackets = false
        if (
          children.length &&
          this.groupFamily.includes(value.node.name) &&
          !this.collapseFamily.includes(value.node.name) &&
          !this.collapseFamily.includes(value.node.firstParent.name)) {
          // filter child
          const nodeFormattedArray = children.filter((a) => {
            const isNumeric = !parseFloat(a.name)
            let isAncestor = true
            if (isNumeric) {
              const nodeFirstParent = this.cylcTree.$index[a.id].node.firstParent.name
              isAncestor = !this.isNodeCollapsedByFamily(nodeFirstParent)
            }
            return (
              // the node is not a numeric value
              isNumeric &&
              // if its not in the list of families (unless its been collapsed)
              (!this.familyArrayStore.includes(a.name) || this.collapseFamily.includes(a.name)) &&
              // the node has been removed/collapsed
              !removedNodes.has(a.name) &&
              // the node doesnt have a collapsed ancestor
              isAncestor
            )
          }).map(a => `"${a.id}"`)
          if (nodeFormattedArray.length) {
            openedBrackets = true
            dotcode.push(`
            subgraph cluster_margin_family_${value.name}${value.tokens.cycle}
              {
              margin=100.0
              label="margin"
              subgraph cluster_${value.name}${value.tokens.cycle}
                {${nodeFormattedArray}${nodeFormattedArray.length ? ';' : ''}
                  label = "${value.name}"
                  fontsize = "70px"
                  style=dashed
                  margin=60.0
            `)
          }
        }

        if (value) {
          this.addSubgraph(dotcode, value, graphSections)
        }
        if (Object.keys(graphSections).includes(value)) {
          dotcode.push(graphSections[value.id])
        }
        if (openedBrackets) {
          dotcode.push('}}')
        }
      })
    },
```

src/views/Graph.vue
```diff
@@ -508,12 +837,81 @@ export default {
...
     getCycles (nodes) {
-      if (!this.groupCycle) return
       return nodes.reduce((x, y) => {
         (x[y.tokens.cycle] ||= []).push(y)
         return x
       }, {})
     },
+    addSubgraph (dotcode, pointer, graphSections) {
+      pointer.children.forEach((key, i) => {
+        const value = key
```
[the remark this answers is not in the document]
wxtim: I'm curious about this too - tried getting rid of this either way and both seem to work.

MetRonnie:
Got some conflicts that I suggest resolving by merging master into this branch

src/views/Graph.vue
```diff
@@ -508,12 +837,81 @@ export default {
...
+          !this.collapseFamily.includes(key.node.name) &&
+          !this.collapseFamily.includes(key.node.firstParent.name)) {
+          // filter child
+          const nodeFormattedArray = children.filter((a) => {
+            // if its not in the list of families (unless its been collapsed)
+            const isFamily = !this.familyArrayStore.includes(a.name) || this.collapseFamily.includes(a.name)
+            // its the node has been removed/collapsed
+            const isRemoved = !removedNodes.includes(a.name)
+            // is not numeric
+            const isNumeric = !parseFloat(a.name)
```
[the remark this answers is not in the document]
oliver-sanders: Ach right, these are jobs rather than tasks.

The data store has a hierarchy like this:
* user
  * workflow
    * cycle
      * task
        * job

We can avoid the need to filter out job nodes, by not looking at the children of task nodes.

src/views/Graph.vue
oliver-sanders: Cheers.

There are a couple of more instances of this, e.g, this forEach is looking through all of the objects in the store which includes other workflows and job nodes which we aren't interested in:

```
Object.keys(this.cylcTree.$index).forEach((itemName) => {
```

We could do something like this instead:

```
for (const workflow of this.workflows) {
  for (const cycle of workflow.children) {
    for (const task of cycle.children {
       // stop here (don't look at jobs)
    }
  }
}
```

markgrahamdawson: I think this should now be resolved in 51fb8df837f1fb7022e5c7333572623310bd9808
I cant see any other examples of `forEach` on `this.cylcTree.$index`

src/views/Graph.vue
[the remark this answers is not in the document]
markgrahamdawson: Ive made a change to the data being returned from `allChildrenLookUp` so there shouldnt be any jobs in it now... and this check isn't needed
7f6a89c6539cfeacdfa504c9fca01d55933b4103

src/views/Graph.vue
```diff
@@ -645,11 +1293,237 @@ export default {
...
         return nodes.length ? nodes : false
       })
-      const edges = this.getGraphEdges()
+      let edges = this.getGraphEdges()
+
+      // ----------------------------------------
+      let removedEdges = []
+
+      // For all cycles...
+      this.cycleArrayStore.forEach((cycle) => {
```
MetRonnie, line 1306: Style note: generally I think we prefer `for...of` rather than `Array.prototype.forEach`:
- You can't break out of `forEach` apart from throwing an error
- Debugging is more natural as "step over" goes into `for` loops, but skips `forEach` callbacks

src/views/Graph.vue
```diff
@@ -333,6 +408,162 @@ export default {
...
+     */
+    treeDropDownCycle () {
+      const store = []
+      this.cycleArrayStore.forEach((name, id) => {
+        store.push({
+          id,
+          name
+        })
+      })
+      return store
```
MetRonnie, line 442:
```suggestion
      return this.cycleArrayStore.map((name, id) => ({
        id,
        name,
      }))
```

src/views/Graph.vue
oliver-sanders: I have been leaning towards `for of` rather than `forEach`.

From a quick search, it looks like `forEach` predated `for of`. `for of` seams cleaner to me and appears to be more efficient.

src/views/Graph.vue
```diff
@@ -168,19 +171,49 @@ fragment TaskProxyData on TaskProxy {
...
+    name
+  }
   task {
     meanElapsedTime
   }
   flowNums
 }
 
+fragment FamilyProxyData on FamilyProxy {
+  __typename
```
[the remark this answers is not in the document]
oliver-sanders: We do need this, it's how the data store differentiates between families and tasks.

https://github.com/cylc/cylc-ui/tree/master/src/views#subscriptions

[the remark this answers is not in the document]
MetRonnie: Ah right

src/views/Graph.vue
markgrahamdawson: This doesnt work correctly when I try it just return an empty array
I want it to return an array..
```
[
  {id:0, name: '20250125T0000Z'},
  {id:1, name: '20250124T2100Z'},
  {id:2, name: '20250124T1800Z'}
]
```

src/views/Graph.vue
oliver-sanders: @MetRonnie, there seems to be a rule for this in eslint recommended: https://github.com/github/eslint-plugin-github/blob/main/docs/rules/array-foreach.md

MetRonnie: I think that's GitHub's own plugin for ESLint, presumably what they use for their coding standards. So not included in ESLint itself. But looks like the popular unicorn plugin also has the same rule: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-for-each.md

src/views/Graph.vue
MetRonnie: It works for me. It's doing exactly the same as creating an empty array and appending to it, but it's more efficient as it already knows what size array it needs to make.

Note the round brackets around the curly brackets are necessary to make the map callback return the object.

src/views/Graph.vue
```diff
@@ -333,6 +408,162 @@ export default {
...
+     * @returns {Family[]} array containing nested structure of families
+     */
+    treeDropDownFamily () {
+      if (this.familyArrayStore.length) {
+        return this.getTree()
+      } else {
+        return [
+          {
+            name: 'No families',
+            disabled: true
```
MetRonnie, line 425: This `disabled` isn't having an effect

image

src/views/Graph.vue
```diff
@@ -333,6 +408,162 @@ export default {
...
+          if (type === 'task') {
+            return flattened
+              .concat([{ id, name, type }])
+          } else {
+            return flattened
+              .concat([{ id, name, children, type }])
+              .concat(children ? childArray(children, id) : [])
+          }
+        }, [])
+      }
```
MetRonnie, line 540: This function can be extracted out of the computed property so it is not re-declared every time it runs. It would also allow unit testing of this function.

MetRonnie: Also it looks like `flatMap` would make this simpler:

```suggestion
      // Function for getting a flattened array of the nested children
      function childArray (a) {
        return a.flatMap(({ id, name, children, type }) => {
          if (type === 'task') {
            return [{ id, name, type }]
          } else {
            return [{ id, name, children, type }]
              .concat(children ? childArray(children) : [])
          }
        })
      }
```

src/views/Graph.vue
```diff
@@ -391,6 +646,268 @@ export default {
...
+      const edgePath = this.workflowIDs[0]
+      const edgeSearchTerm = `${edgePath}//$edge|${sourceCyclePoint}/${sourceName}|${targetCyclePoint}/${targetName}`
+      const edgesFiltered = edges.filter((edge) => {
+        if (edge.id.indexOf(edgeSearchTerm) === -1) {
+          return true
+        } else {
+          return false
+        }
+      })
+      return edgesFiltered
```
MetRonnie, line 805:
```suggestion
      return edges.filter(
        (edge) => edge.id.indexOf(edgeSearchTerm) === -1
      )
```

src/views/Graph.vue
```diff
@@ -391,6 +646,268 @@ export default {
...
+      }
+      if (edgeType === 'collapsedTarget') {
+        src = `${sourceCyclePoint}/${sourceName}|${targetName}`
+        tgt = `${sourceCyclePoint}/${sourceName}|${targetName}`
+      }
+      if (edgeType === 'collapsedSource') {
+        src = `${sourceCyclePoint}|${targetCyclePoint}/${targetName}`
+        tgt = `${sourceCyclePoint}|${targetCyclePoint}/${targetName}`
+      }
+      if (edgeType === 'collapsedSourceAndTarget') {
```
MetRonnie, line 758:
```suggestion
      } else if (edgeType === 'collapsedTarget') {
        src = `${sourceCyclePoint}/${sourceName}|${targetName}`
        tgt = `${sourceCyclePoint}/${sourceName}|${targetName}`
      } else if (edgeType === 'collapsedSource') {
        src = `${sourceCyclePoint}|${targetCyclePoint}/${targetName}`
        tgt = `${sourceCyclePoint}|${targetCyclePoint}/${targetName}`
      } else if (edgeType === 'collapsedSourceAndTarget') {
```

src/views/Graph.vue
```diff
@@ -662,6 +649,26 @@ export default {
...
+      return nodeArray.flatMap(({ id, name, children, type }) => {
+        // We dont want to include jobs so dont include children if type is "task"
+        if (type === 'task') {
+          return [{ id, name, type }]
+        } else {
+          return [{ id, name, children, type }]
+            .concat(children ? this.childArray(children) : [])
+        }
+      })
+    },
```
MetRonnie, line 671: Even better if this was moved out of the component methods as it doesn't make use of component state. Would simplify the testing too

markgrahamdawson: Done in 9b791ce672b168eb81fbb8837595bbab55cc6ecc

src/views/Graph.vue
markgrahamdawson: Done in 0908cd64b34900e51744daa7a7f080fd719846f4

src/views/Graph.vue
markgrahamdawson: Done in 0908cd64b34900e51744daa7a7f080fd719846f4

src/views/Graph.vue
markgrahamdawson: Done in 0908cd64b34900e51744daa7a7f080fd719846f4

src/views/Graph.vue
markgrahamdawson: Fixed in 0908cd64b34900e51744daa7a7f080fd719846f4

src/views/Graph.vue
```diff
@@ -333,6 +408,162 @@ export default {
...
+     * example return object
+     * {
+     *  FAMILY: ['PARENT_FAMILY', 'GRANDPARENT_FAMILY', 'GREAT_GRANDPARENT_FAMILY', 'root' ],
+     *  PARENT_FAMILY: [ 'GRANDPARENT_FAMILY', 'GREAT_GRANDPARENT_FAMILY', 'root' ],
+     *  GRANDPARENT_FAMILY: [ 'GREAT_GRANDPARENT_FAMILY', 'root' ],
+     *  GREAT_GRANDPARENT_FAMILY : ['root'],
+     *  root: []
+     * }
+     *
+     * note: object value arrays do not contain family names as strings only - not nodes
```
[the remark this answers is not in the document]
markgrahamdawson: Yes, fixed in 0908cd64b34900e51744daa7a7f080fd719846f4

src/views/Graph.vue
markgrahamdawson: Done in 0a1f3c510660cfb628d3744a0d80cd986b97687b

src/views/Graph.vue
markgrahamdawson: These have all been updated in 
e035bfb48ecaaaa769935f95429c63bd8afd897a

src/views/Graph.vue
```diff
@@ -659,7 +1533,8 @@ export default {
       }
 
       const cycles = this.getCycles(nodes)
-
+      this.latestCycle = Object.keys(cycles)[0]
```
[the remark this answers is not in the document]
markgrahamdawson: removed in 076545609ad0bc1803e8709e46d63072e55465d8

src/services/mock/json/workflows/one.json
```diff
@@ -46,6 +46,100 @@
           "ancestors": []
         }
       ],
+      "families": [
```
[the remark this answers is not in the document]
markgrahamdawson: I have updated the multi.json with families in this commit
ca89af79716a4e7ad5b169a5f7bb5fb79c919b59

Im not sure about adding other fields, the only fields that are added are `families` and `familyProxies` and the latter is already included in the mock json

src/components/cylc/ViewToolbar.vue
```diff
@@ -30,7 +30,30 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         class="control"
         :data-cy="`control-${iControl.key}`"
       >
-        <v-btn
+        <v-menu
+        v-if="iControl.action==='select-tree'"
+        :close-on-content-click="false">
+          <template v-slot:activator="{ props }" >
```
MetRonnie, line 36: (Style)
```suggestion
          v-if="iControl.action==='select-tree'"
          :close-on-content-click="false"
        >
          <template v-slot:activator="{ props }">
```

src/components/cylc/ViewToolbar.vue
```diff
@@ -30,7 +30,30 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+            <v-btn
+            v-bind="{...$attrs, ...props, ...btnProps}"
+            :disabled="iControl.disabled"
+            :color="iControl.color">
```
MetRonnie, line 40:
```suggestion
              v-bind="{...$attrs, ...props, ...btnProps}"
              :disabled="iControl.disabled"
              :color="iControl.color"
            >
```

src/components/cylc/ViewToolbar.vue
```diff
@@ -30,7 +30,30 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+              <v-icon>{{ iControl.icon }}</v-icon>
+              <v-tooltip>{{ iControl.title }}</v-tooltip>
+            </v-btn>
+          </template >
```
MetRonnie, line 44:
```suggestion
          </template>
```

src/services/mock/json/workflows/one.json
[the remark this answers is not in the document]
MetRonnie: Also I think this is needed in `multi.json`

```diff
diff --git a/src/services/mock/json/workflows/multi.json b/src/services/mock/json/workflows/multi.json
index 14f7f37a..4232e136 100644
--- a/src/services/mock/json/workflows/multi.json
+++ b/src/services/mock/json/workflows/multi.json
@@ -168,4 +168,5 @@
             "firstParent": {
               "id": "~user/other/multi/run2//1/root",
+              "name": "root",
               "__typename": "FamilyProxy"
             },
@@ -186,4 +187,5 @@
             "firstParent": {
               "id": "~user/other/multi/run2//2/root",
+              "name": "root",
               "__typename": "FamilyProxy"
             },
```

otherwise I am getting 
> TypeError: Cannot read properties of undefined (reading 'some')
>    at Proxy.isNodeCollapsedByFamily (Graph.vue:888:62)

src/components/cylc/ViewToolbar.vue
markgrahamdawson: fixed in 7b225b317d42fad6d7d656ad01182ea41ff5f751

src/components/cylc/ViewToolbar.vue
markgrahamdawson: fixed in 7b225b317d42fad6d7d656ad01182ea41ff5f751

tests/unit/views/graph.vue.spec.js
```diff
@@ -0,0 +1,697 @@
...
+    )
+  })
+  it('it gets flattened array of the nested children', async () => {
+    const testData = [
+      {
+        id: 'user/one/run1//1',
+        name: 'cycle',
+        children: [
+          {
+            id: 'user/one/run1//1/child',
```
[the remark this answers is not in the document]
markgrahamdawson: fixed in 7b225b317d42fad6d7d656ad01182ea41ff5f751

src/views/Graph.vue
```diff
@@ -391,6 +647,268 @@ export default {
...
+     */
+    getTreeHelper (store, node, counter) {
+      let tempItem
+      const isParent = this.collapseFamily.includes(node.name)
+      const isAncestor = this.allParentLookUp[node.name].some(element => {
+        return this.collapseFamily.includes(element)
+      })
+      const disabled = isParent || isAncestor
+      for (const childFamily of node.node.descendants) {
+        if (this.namespaces.map((obj) => obj.name).includes(childFamily)) {
```
MetRonnie, line 712: Can short-circuit this
```suggestion
        if (this.namespaces.some((obj) => obj.name === childFamily)) {
```

MetRonnie:
I spotted an issue where no edges are being rendered at all (discussed in person)

src/views/Graph.vue
```diff
@@ -391,6 +647,268 @@ export default {
...
+      const counter = {
+        value: 0,
+        next () {
+          this.value = this.value + 1
+          return this.value
+        }
+      }
+
+      const root = {
+        id: counter.next(),
```
MetRonnie, line 685: What is the need for this id/counter?

src/views/Graph.vue
```diff
@@ -391,6 +647,268 @@ export default {
...
+          if (childNode.node.firstParent.id === node.id) {
+            tempItem = {
+              id: counter.next(),
+              name: childFamily,
+              children: [],
+              disabled
+            }
+            this.getTreeHelper(tempItem, childNode, counter)
+            store.children.push(tempItem)
+          }
```
MetRonnie, line 724:
```suggestion
        const childNamespace = this.namespaces.find((obj) => obj.name === childFamily)
        if (childNamespace?.node.firstParent.id === node.id) {
          tempItem = {
            id: counter.next(),
            name: childFamily,
            children: [],
            disabled
          }
          this.getTreeHelper(tempItem, childNamespace, counter)
          store.children.push(tempItem)
```

src/components/cylc/ViewToolbar.vue
```diff
@@ -120,7 +151,9 @@ export default {
           color = null
           callback = null
           disabled = false
-
+          if (control.items && control.items.length) {
+            disabled = control.items[0].disabled
```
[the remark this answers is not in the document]
markgrahamdawson: Ive spent a bit of time looking at this and Im not really sure what it was doing.
Its purpose would have been to set families as disabled in the drop down if a family further up the tree has already been selected. 
But looking at it now it looks  like it wouldn't do that - I can just remove this code and it still works as intended so I might go ahead and do that?

src/views/Graph.vue
markgrahamdawson: Its because it was written for the vuetify treeview component that has been added to the toolbar.
https://vuetifyjs.com/en/components/treeview/#open-all
In some of the examples it passes an `id` to the `item-value` prop - but looking now I can see in some examples it does not so I guess its not required.... 

In this pr the `v-treeview` the constant string `"name"` is passed to  `item-value`. I've just tried removing it and it makes the tree indents go away - it also stops the `open-all` prop working. So I think I can probably remove the `id`/`counter` and just leave it with this `name` string?

markgrahamdawson: fixed in 669fa07cc43b0f9ca283b9bcffe53c4afdffe96d

src/views/Graph.vue
markgrahamdawson: fixed in 669fa07cc43b0f9ca283b9bcffe53c4afdffe96d

src/services/mock/json/workflows/one.json
```diff
@@ -46,6 +46,107 @@
...
+          "id": "~user/one//$namespace|BAD",
+          "name": "BAD",
+          "firstParent": { "id": "~user/one//$namespace|root", "name": "root"},
+          "parents": [
+            {
+              "name": "root",
+              "_typename": "Family"
+            }
+          ],
+          "descendants": ["GOOD", "failed", "retrying"],
```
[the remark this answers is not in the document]
markgrahamdawson: fixed in 669fa07cc43b0f9ca283b9bcffe53c4afdffe96d

src/views/Graph.vue
```diff
@@ -333,6 +424,147 @@ export default {
...
+        }
+      }
+      return lookup
+    },
+    /**
+     * Get an array of family names
+     * @returns {String[]} array of family names
+     */
+    familyArrayStore () {
+      return this.namespaces.filter((family) => { return family.name !== 'root' }).map((family) => family.name)
```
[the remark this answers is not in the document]
markgrahamdawson: fixed in 669fa07cc43b0f9ca283b9bcffe53c4afdffe96d

src/views/Graph.vue
```diff
@@ -333,6 +424,147 @@ export default {
...
+        }
+        return ret
+      } else {
+        return [
+          {
+            name: 'No families',
+            disabled: true
+          }
+        ]
+      }
```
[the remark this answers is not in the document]
markgrahamdawson: fixed in 669fa07cc43b0f9ca283b9bcffe53c4afdffe96d

src/components/cylc/ViewToolbar.vue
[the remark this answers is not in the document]
markgrahamdawson: fixed in 669fa07cc43b0f9ca283b9bcffe53c4afdffe96d

MetRonnie:
@wxtim This is normally as a result of an uncaught error which will be visible in the console

src/views/Graph.vue
```diff
@@ -167,9 +167,6 @@ fragment EdgeData on Edge {
 fragment TaskProxyData on TaskProxy {
   id
   state
-  isHeld
-  isRunahead
-  isQueued
```
MetRonnie, line 169: Why have these been removed? I think they are needed for the task icons

src/views/Graph.vue
```diff
@@ -185,9 +182,6 @@ fragment FamilyProxyData on FamilyProxy {
   __typename
   id
   state
-  isHeld
-  isRunahead
-  isQueued
```
MetRonnie: Ditto

markgrahamdawson: fixed in 4d775f25446f833064f0e5f082595c4f0c28ea06

src/views/Graph.vue
```diff
@@ -645,11 +1275,239 @@ export default {
...
+        // ...loop through collapsed families...
+        for (const family of this.collapseFamily) {
+          // ...get the node from the index...
+          const indexSearch = Object.values(this.cylcTree.$index).find((node) => {
+            return node.name === family && node.tokens.cycle === cycle
+          })
+          // ...now we have a node - we have to understand all its relations in the graph...
+          if (indexSearch) {
+            // ---------------REMOVE NODES BASED ON FAMILY------------
+            // ṃust do this before removing nodes and edges based on cycle as
```
wxtim, line 1298:
```suggestion
            // must do this before removing nodes and edges based on cycle as
```

src/views/Graph.vue
```diff
@@ -645,11 +1275,239 @@ export default {
...
+            // cycle collapsing takes priority of families
+            // ...this node is collapsed so need to remove any of its children (nodes and edges) from the graph if it has any...
```
wxtim, line 1300: Consider splitting these very long lines.

src/views/Graph.vue
```diff
@@ -508,16 +1007,108 @@ export default {
...
+        }
+        let openedBrackets = false
+        if (
+          // If this pointer has grandchildren
+          grandChildren.length &&
+          // Is the child of the pointer a family that is grouped?
+          // If yes - we want to include in the grouping
+          this.groupFamily.includes(child.node.name) &&
+          // But if its collapsed then we dont want to group it
+          // We dont put boxes around collapsed nodes - design choice
```
wxtim, line 1065: I'd question that design choice, but I'm guessing that it's been thrashed out?

I think that collapsed nodes should retain the box to indicated that it's not just a single task.

tests/e2e/specs/graph.cy.js
```diff
@@ -150,17 +168,148 @@ describe('Graph View', () => {
...
+          }
+        }
+      })
+
+      // there should be 10 graph edges (all on-screen)
+      .get('.c-graph:first')
+      .find('.graph:first .edges:first')
+      .children()
+      .should('have.length', 10)
+      .should('be.visible')
```
wxtim, line 200: Consider checking that you have the failed task icon in here?

wxtim:
I've left a workflow running all day in the background with this display on and haven't noted any indication of slowdown.

wxtim (approved):
- [x] I have had a good stab at breaking this by throwing workflows at it.
- [x] I have read the code.

I've made a small number of suggestions, but I don't belive any of these should block the merging of this PR.

src/views/Graph.vue
```diff
@@ -562,24 +1156,63 @@ export default {
...
-              label = "${key}";\n
-              fontsize = "70px"
-              style=dashed
-              margin=60.0
+      const graphSections = {}
+      for (const cycle of Object.keys(cycles)) {
+        const indexSearch = cycles[cycle]
+        if (indexSearch.length && !this.collapseCycle.includes(cycle)) {
+          for (const task of indexSearch) {
+            if (!task.node.firstParent) { return }
```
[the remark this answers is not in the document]
markgrahamdawson: going back through the code Im pretty sure this is never actually called.
The only node that doesnt have a first parent is the `root` node - which has been removed from `nodes` at this point (dont show it in the graph). 
And `cycles` comes from `const cycles = this.getCycles(nodes)`

So I think it can be removed

src/views/Graph.vue
```diff
@@ -645,11 +1278,239 @@ export default {
...
+          }
+        }
+      }
+      for (const cycle of this.collapseCycle) {
+        const indexSearch = Object.values(this.cylcTree.$index).find((node) => {
+          return node.name === cycle
+        })
+        if (indexSearch) {
+          // ---------------REMOVE NODES BASED ON CYCLE POINT------------
+          if (!this.allChildrenLookUp[indexSearch.id]) { return }
```
MetRonnie, line 1430: Another `return` that looks erroneous?

MetRonnie (commented):
Unfortunately, edges are missing between collapsed families in different cycle points:

image

C.f.
```
cylc graph complex -g BRIEF -g LONG -g ENSEMBLE -c
```

image

markgrahamdawson:
> unately, edges are missing between collapsed famil

Hopefully fixed with 
b108b6f04935ab2d740de00cc3fad24f0c0400d3
image

src/views/Graph.vue
markgrahamdawson: removed in 8740871e50e7ee148200cc2b7e40669cf62de19a

markgrahamdawson:
> > unately, edges are missing between collapsed famil
> 
> Hopefully fixed with [b108b6f](https://github.com/cylc/cylc-ui/commit/b108b6f04935ab2d740de00cc3fad24f0c0400d3)

I have unfortunately found an similar issue with cycles

Collapsed families and grouped by cycle
image

Collapsed families and grouped by cycle... and then collapse by a cycle

image

markgrahamdawson:
> > > unately, edges are missing between collapsed famil
> > 
> > 
> > Hopefully fixed with [b108b6f](https://github.com/cylc/cylc-ui/commit/b108b6f04935ab2d740de00cc3fad24f0c0400d3)
> 
> I have unfortunately found an similar issue with cycles
> 

Hopefully fixed with 73d05f50086e96bff0b60b0b97721c4b78c3fccf

image

MetRonnie (approved):
I have created a short-lived feature branch for this and changed the base. I am happy for the PR to be merged into this branch and will open some follow-up PRs to address some technical debt. 

We may also hold off merging the feature branch into master until https://github.com/cylc/cylc-flow/issues/6689 is addressed.

PR #1842, opened 2024-06-24
Log view: display any errors in listing the log files

MetRonnie:
If there is a failure to list the log files, the user should be presented with an error message, not left in the dark.

**Check List**

- [x] No tests needed
- [x] No changelog needed

### DIFF
--- a/src/views/Log.vue
+++ b/src/views/Log.vue
@@ -198,6 +198,7 @@ import ViewToolbar from '@/components/cylc/ViewToolbar.vue'
 import DeltasCallback from '@/services/callbacks'
 import { debounce } from 'lodash-es'
 import CopyBtn from '@/components/core/CopyBtn.vue'
+import { Alert } from '@/model/Alert.model'
 import { getJobLogFileFromState } from '@/model/JobState.model'
 
 /**
@@ -558,7 +559,7 @@ export default {
         })
       } catch (err) {
         // the query failed
-        console.warn(err)
+        this.handleLogFileListingErr(err)
         this.handleNoLogFiles()
         return
       }
@@ -576,6 +577,9 @@ export default {
         this.fileDisabled = false
         this.logFiles = logFiles
       } else {
+        if (result.errors?.length) {
+          this.handleLogFileListingErr(result.errors[0].message)
+        }
         this.handleNoLogFiles()
       }
     },
@@ -602,6 +606,9 @@ export default {
       this.fileLabel = this.id ? `No log files for ${this.id}` : 'Enter a task/job ID'
       this.fileDisabled = true
       this.logFiles = []
+    },
+    handleLogFileListingErr (err) {
+      this.$store.dispatch('setAlert', new Alert(err, 'error'))
     },
   },
 


MetRonnie:
No longer applicable after change of tack

MetRonnie:
Actually this change still makes sense irrespective of https://github.com/cylc/cylc-uiserver/pull/607

MetRonnie:
Try
```py
    if mode == 'list-dir':
        sys.exit('foo bar baz')
```

MetRonnie:
Your first diff does the opposite? `!=` instead of `==`

MetRonnie:
Ah sorry, this only applies to network/graphql errors

oliver-sanders:
Removing the `cat_log_files` function will create an internal code error not a network error right? Note GraphQL errors are another form of internal error.

This PR does not appear to assist with diagnosing remote errors, e.g. try creating an SSH error with your config.

This diff introduces a common network error (host not found), but no error is reported in the log view:

```diff
diff --git a/cylc/flow/remote.py b/cylc/flow/remote.py
index e0ab7544a..e99903e0a 100644
--- a/cylc/flow/remote.py
+++ b/cylc/flow/remote.py
@@ -289,7 +289,8 @@ def construct_ssh_cmd(
     if stdin is None:
         command.append('-n')
 
-    command.append(host)
+    # command.append(host)
+    command.append('elephant')
 
     # Pass CYLC_VERSION and optionally, CYLC_CONF_PATH & CYLC_UTC through.
     command += ['env', quote(r'CYLC_VERSION=%s' % CYLC_VERSION)]
```

I think this PR may make sense, but at present I don't think it is helping as we need the UIS to capture and forward these errors which is clearly isn't doing at present. We should get the errors into the schema at the UIS end first, then make sure the log view is displaying them after.

MetRonnie:
We agreed it doesn't make sense to show an error snackbar for non-zero ret code of `cat-log` in https://github.com/cylc/cylc-uiserver/pull/607:

> I realised that there are a few ways `cat-log` will exit non-zero when navigating in the UI (e.g. navigating to `/log/non-exist`), so have settled for just logging the stderr and returning an empty list as before

Maybe we should ensure ret code 1 means no files found, and show an error snackbar for any ret code > 1?

oliver-sanders:
These changes may well make sense, but so far I have failed to come up with an example that triggers this code.

I even tried breaking the `cylc cat-log` command entirely with this diff:

```diff
diff --git a/cylc/flow/scripts/cat_log.py b/cylc/flow/scripts/cat_log.py
index 2b4f03ca6..0bd138f20 100755
--- a/cylc/flow/scripts/cat_log.py
+++ b/cylc/flow/scripts/cat_log.py
@@ -421,6 +421,7 @@ def _main(
       b) remote: re-invoke cylc cat-log as a) on the remote account
 
     """
+    sys.exit('foobar')
     if options.filename is not None:
         _check_fs_path(options.filename)

```

The error gets logged at the UIS:

```
[E 2025-05-13 13:53:03.405 CylcUIServer] Command failed (1): cylc cat-log -m l -o ~oliver.sanders/generic/run1
    ERROR: foobar
```

But no error message is produced in the UI (it just says "no log files for ...").

MetRonnie:
It's been a while but I think this only applies to errors in graphql execution, not errors in the cat-log command. That comes back to my question

> We agreed it doesn't make sense to show an error snackbar for non-zero ret code of `cat-log` in [cylc/cylc-uiserver#607](https://github.com/cylc/cylc-uiserver/pull/607): ...
>
> Maybe we should ensure ret code 1 means no files found, and show an error snackbar for any ret code > 1?

MetRonnie:
Here's an example
```diff
diff --git a/cylc/uiserver/resolvers.py b/cylc/uiserver/resolvers.py

     @classmethod
-    async def cat_log_files(cls, id_: Tokens, log: 'Logger') -> List[str]:
+    async def cat_log_files(cls, log: 'Logger') -> List[str]:
         """Calls cat log to get list of available log files.
```

On cylc-ui master, there is no indication of anything going wrong other than it saying "no log files for <workflow id>"

On this branch you should get:

image

PR #1886, opened 2024-08-08
metadata view: introduce a minimal task metadata view

### LINKED ISSUE #1071: task metadata view
**Cylc 7**

Cylc supports task metadata e.g:

```ini
[runtime]
    [[mytask]]
        [[[meta]]]
            # the standard metadata fields used by some cylc commands
            title = My Task
            url = https://...
            description = """
                Lorel ipsum ...
            """
            # you can create custom fields too
            system = xyz
            group = abc
```

Cylc 7 made minimal use of task metadata, namely in commands like `cylc show` but some metadata was also used in the GUI.

**Cylc 8**

Cylc 8 has greater potential here. Earlier design sketches had an info "pane", "tounge" or "panel" which was intended to show the `[meta]` info for the selected task (or the workflow itself as workflows also support `[meta]`). The GUI has moved on since then, the  Lumino tab system is now the natural place for displays, I think dividing the screen beyond this would be unhelpful.

So here's my suggestion, we implement a "metadata view" (or "info" view). This view gives you a text box to type in the name of the task (we should be able to make this a combo box with all task names pre-loaded). Once you've selected a task it displays the `[meta]`. For convenience we can create a shortcut from the task menu which opens a new "info" view pre-initalised at the specified task. We should also consider adding the "title" field to the task menu.

**Sketch**

Here we have an info view showing some basic metadata. The task menu has been opened for the task "baz" on the right. The "title" has been added to the top of the menu along with an info button which is the shortcut for opening the info tab on the left.

task-metadata

**Metadata & Formats**

Some sites are already making use of the "URL" field. In [at least] one case, using HTML pages generated from markdown. We can make this much easier (rendering markdown in a web app is trivial).

* Suggest rendering `[meta]` items as markdown by default.

> This part is more cylc-flow / cylc-rose focused but bringing it up here for context

Many tasks already have extensive documentation in the form of Rose application metadata. There is an existing Rose config to ReST generator which is used by the Rose documentation.

* Consider adding a `[meta]_markup_format` config to allow this to be configurable in the future?
* Or should this be a marker in the content itself e.g. `#markdown`, `#rest`, ...?
* Consider a cylc-rose plugin to drag in Rose application metadata where available.

**Workflows & Documentation**

Workflows can have `[meta]` too. We might want to add an option for viewing the workflow metadata here too (and possibly a shortcut for it in the workflow toolbar).

Cylc has documentation too! We might want to add an option view viewing it here!

**Additional context**

Once discussed / agreed this should be quite quick to implement.

* List of task names can be obtained by listing taskdefs (called `Tasks` in the schema).
* Metadata can be obtained from the taskdef.

**Pull requests welcome!**

### LINKED ISSUE #1896: info view: add completion expression info
Add completion expression information into the "Info View".

See the "completion" section of the `cylc show` command to see what this should include.

oliver-sanders:
Closes #1071
Closes #1896

* Bare bones task metadata view.
* Display task metadata, prerequisites and outputs.
* Display / functionality improvements and support for other information to follow in future PRs.
 
Screenshot from 2024-08-08 15-44-54

**Check List**

- [ ] Cylc-Doc - bumped changes entry to a [PR](https://github.com/cylc/cylc-doc/issues/756) (view is likely to change further before release)

### DIFF
--- a/cypress/component/info.cy.js
+++ b/cypress/component/info.cy.js
@@ -1,0 +1,220 @@
+/**
+ * Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import InfoComponent from '@/components/cylc/Info.vue'
+import { Tokens } from '@/utils/uid'
+
+const DESCRIPTION = `Lorem ipsum dolor sit amet, consectetur adipiscing elit.
+Sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi.
+`
+const TOKENS = new Tokens('~user/workflow//1234/foo')
+const TASK = {
+  id: TOKENS.id,
+  name: TOKENS.task,
+  tokens: TOKENS,
+  node: {
+    state: 'running',
+    task: {
+      meta: {
+        title: 'My Foo',
+        description: DESCRIPTION,
+        URL: 'https://cylc.org',
+        customMeta: {
+          answer: '42',
+          question: 'mutually exclusive',
+        }
+      }
+    },
+    prerequisites: [
+      {
+        satisfied: false,
+        expression: '(c0 & c1) | c2',
+        conditions: [
+          {
+            taskId: 'a',
+            message: 'succeeded',
+            reqState: 'succeeded',
+            exprAlias: 'c0',
+            satisfied: true,
+          },
+          {
+            taskId: 'b',
+            message: 'custom message',
+            reqState: 'custom_output',
+            exprAlias: 'c1',
+            satisfied: false,
+          },
+          {
+            taskId: 'a',
+            message: 'expired',
+            reqState: 'expired',
+            exprAlias: 'c2',
+            satisfied: false,
+          },
+        ],
+      },
+      {
+        satisfied: true,
+        expression: 'c0',
+        conditions: [
+          {
+            taskId: 'x',
+            message: 'succeeded',
+            reqState: 'succeeded',
+            exprAlias: 'c0',
+            satisfied: true,
+          },
+        ],
+      },
+    ],
+    outputs: [
+      {
+        label: 'started',
+        message: 'started',
+        satisfied: true,
+      },
+      {
+        label: 'succeeded',
+        message: 'succeeded',
+        satisfied: false,
+      },
+      {
+        label: 'failed',
+        message: 'failed',
+        satisfied: false,
+      },
+    ]
+  },
+  children: [
+    {
+      id: TOKENS.clone({ job: '01' }).id,
+      tokens: TOKENS.clone({ job: '01' }),
+      name: '01',
+      node: {
+        state: 'failed'
+      }
+    },
+    {
+      id: TOKENS.clone({ job: '02' }).id,
+      tokens: TOKENS.clone({ job: '02' }),
+      name: '02',
+      node: {
+        state: 'succeeded'
+      }
+    },
+  ],
+}
+
+describe('Info component', () => {
+  it('displays task information', () => {
+    cy.vmount(InfoComponent, {
+      props: {
+        task: TASK,
+        class: 'job_theme--default',
+        // NOTE: expand all sections by default
+        panelExpansion: [0, 1, 2],
+      }
+    })
+
+    // there should be a task icon (running)
+    cy.get('.c-graph-node .c8-task.running').should('be.visible')
+
+    // and two job icons (succeeded & failed)
+    cy.get('.c-graph-node .c-job').should('have.length', 2)
+      .get('.c-graph-node .c-job .failed').should('be.visible')
+      .get('.c-graph-node .c-job .succeeded').should('be.visible')
+
+    // the metadata panel
+    cy.get('.metadata-panel.v-expansion-panel--active').should('be.visible')
+      .contains('My Foo')
+      .get('.metadata-panel') // the description should be displayed
+      .contains(/Lorem ipsum dolor sit amet.*/)
+      .get('.metadata-panel a:first') // the URL should be an anchor
+      .should('have.attr', 'href', 'https://cylc.org')
+      .contains(/^https:\/\/cylc.org$/)
+
+    // the prerequisites panel
+    cy.get('.prerequisites-panel.v-expansion-panel--active').should('be.visible')
+      .find('.prerequisite-alias.condition')
+      .should('have.length', 6)
+      .then((selector) => {
+        expect(selector[0]).to.contain('(c0 & c1) | c2')
+        expect(selector[0].classList.toString()).to.equal('prerequisite-alias condition')
+
+        expect(selector[0]).to.contain('c0')
+        expect(selector[1].classList.toString()).to.equal('prerequisite-alias condition satisfied')
+
+        expect(selector[0]).to.contain('c1')
+        expect(selector[2].classList.toString()).to.equal('prerequisite-alias condition')
+
+        expect(selector[0]).to.contain('c2')
+        expect(selector[3].classList.toString()).to.equal('prerequisite-alias condition')
+
+        expect(selector[0]).to.contain('c0')
+        expect(selector[4].classList.toString()).to.equal('prerequisite-alias condition satisfied')
+
+        expect(selector[0]).to.contain('c0')
+        expect(selector[5].classList.toString()).to.equal('prerequisite-alias condition satisfied')
+      })
+
+    // the outputs panel
+    cy.get('.outputs-panel.v-expansion-panel--active').should('be.visible')
+      .find('.condition')
+      .should('have.length', 3)
+      .then((selector) => {
+        expect(selector[0]).to.contain('started')
+        expect(selector[0].classList.toString()).to.equal('condition satisfied')
+
+        expect(selector[1]).to.contain('succeeded')
+        expect(selector[1].classList.toString()).to.equal('condition')
+
+        expect(selector[2]).to.contain('failed')
+        expect(selector[2].classList.toString()).to.equal('condition')
+      })
+  })
+
+  it('should expand sections as intended', () => {
+    const spy = cy.spy()
+    cy.vmount(InfoComponent, {
+      props: {
+        task: TASK,
+        class: 'job_theme--default'
+      },
+      listeners: {
+        'update:panelExpansion': spy,
+      }
+    }).as('wrapper')
+
+    // ONLY the metadata panel should be expanded by default
+    cy.get('.v-expansion-panel--active')
+      .should('have.length', 1)
+      .should('have.class', 'metadata-panel')
+
+    // the update:panelExpansion event should be emitted when a panel is
+    // expanded/collapsed
+    cy.get('.prerequisites-panel')
+      .find('button')
+      .should('be.visible')
+      .click({ force: true })
+      .get('@wrapper').then(({ wrapper }) => {
+        expect(
+          wrapper.emitted('update:panelExpansion')[0][0]
+        ).to.deep.equal([0, 1])
+      })
+  })
+})
--- a/cypress/component/info.cy.js
+++ b/cypress/component/info.cy.js
@@ -217,4 +217,34 @@ describe('Info component', () => {
         ).to.deep.equal([0, 1])
       })
   })
+
+  it('should work for a task with no data', () => {
+    // ensure the component can be mounted without errors for empty states
+    // i.e. no metadata, prerequisites, outputs or jobs
+    const tokens = new Tokens('~user/workflow//1234/foo')
+    const task = {
+      id: tokens.id,
+      name: tokens.task,
+      tokens,
+      node: {
+        task: {
+          meta: {
+            customMeta: {}
+          }
+        },
+        prerequisites: [],
+        outputs: [],
+      },
+      children: [],
+    }
+
+    cy.vmount(InfoComponent, {
+      props: {
+        task,
+        class: 'job_theme--default',
+        // NOTE: expand all sections by default
+        panelExpansion: [0, 1, 2],
+      }
+    })
+  })
 })
--- a/cypress/component/info.cy.js
+++ b/cypress/component/info.cy.js
@@ -153,22 +153,22 @@ describe('Info component', () => {
       .find('.prerequisite-alias.condition')
       .should('have.length', 6)
       .then((selector) => {
-        expect(selector[0]).to.contain('(c0 & c1) | c2')
+        expect(selector[0]).to.contain('(0 & 1) | 2')
         expect(selector[0].classList.toString()).to.equal('prerequisite-alias condition')
 
-        expect(selector[0]).to.contain('c0')
+        expect(selector[0]).to.contain('0')
         expect(selector[1].classList.toString()).to.equal('prerequisite-alias condition satisfied')
 
-        expect(selector[0]).to.contain('c1')
+        expect(selector[0]).to.contain('1')
         expect(selector[2].classList.toString()).to.equal('prerequisite-alias condition')
 
-        expect(selector[0]).to.contain('c2')
+        expect(selector[0]).to.contain('2')
         expect(selector[3].classList.toString()).to.equal('prerequisite-alias condition')
 
-        expect(selector[0]).to.contain('c0')
+        expect(selector[0]).to.contain('0')
         expect(selector[4].classList.toString()).to.equal('prerequisite-alias condition satisfied')
 
-        expect(selector[0]).to.contain('c0')
+        expect(selector[0]).to.contain('0')
         expect(selector[5].classList.toString()).to.equal('prerequisite-alias condition satisfied')
       })
 
--- a/cypress/component/info.cy.js
+++ b/cypress/component/info.cy.js
@@ -98,7 +98,15 @@ const TASK = {
         message: 'failed',
         satisfied: false,
       },
-    ]
+      {
+        label: 'x',
+        message: 'xxx',
+        satisfied: true,
+      }
+    ],
+    runtime: {
+      completion: '(succeeded and x) or failed'
+    }
   },
   children: [
     {
@@ -127,7 +135,7 @@ describe('Info component', () => {
         task: TASK,
         class: 'job_theme--default',
         // NOTE: expand all sections by default
-        panelExpansion: [0, 1, 2],
+        panelExpansion: [0, 1, 2, 3],
       }
     })
 
@@ -175,7 +183,7 @@ describe('Info component', () => {
     // the outputs panel
     cy.get('.outputs-panel.v-expansion-panel--active').should('be.visible')
       .find('.condition')
-      .should('have.length', 3)
+      .should('have.length', 4)
       .then((selector) => {
         expect(selector[0]).to.contain('started')
         expect(selector[0].classList.toString()).to.equal('condition satisfied')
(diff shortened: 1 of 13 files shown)


hjoliver:
Ping @ScottWales @ColemanTom (you were wanting this for viewing prerequisites etc.)

ScottWales:
Great! thanks for this

oliver-sanders:
Added offline-data and tests.

Many, many enhancements and features left to add to this view, but will leave it at basic metadata, prereqs and outputs for this PR.

Further enhancements will slot into this template fairly easily provided the required data is present (and accurate) in the schema.

oliver-sanders:
Documented desired future enhancements in issues:

* https://github.com/cylc/cylc-ui/issues/1893
* https://github.com/cylc/cylc-ui/issues/1894
* https://github.com/cylc/cylc-ui/issues/1895
* https://github.com/cylc/cylc-ui/issues/1896
* https://github.com/cylc/cylc-ui/issues/1897
* https://github.com/cylc/cylc-ui/issues/1898

src/components/cylc/Info.vue
```diff
@@ -0,0 +1,257 @@
...
+-->
+
+<template>
+  <div class="c-info">
+    <!-- The task summary
+
+    * The "-40" comes from the GraphNode offset, see the GraphNode compoent.
+    * The height is "200" + "40".
+    * The "99999" is the maximum component width (the xMinYMin value prevents
+      this from taking up more horizontal space than it requires).
```
MetRonnie, line 25: It doesn't on Chrome

image

oliver-sanders: Thanks for testing, will take a look.

oliver-sanders: Resolved by setting `overflow-x: hidden`.

src/components/cylc/Info.vue
```diff
@@ -0,0 +1,261 @@
...
+    * The "99999" is the maximum component width (because we don't know
+      how wide it will be until we render it).
+    * The "overflow-x: hidden" prevents this 99999 width from causing
+      horizontal scrolling.
+    * The "8em" is the height that this SVG will be scaled to fill.
+    -->
+    <div style="overflow-x: hidden;">
+      <svg
+        preserveAspectRatio="xMinYMin"
+        viewBox="-40 -40 99999 200"
```
MetRonnie, line 33: This is not so good for long task names on a narrow window.

Happy to bump to a follow-up issue however

wxtim: Follow up pls.

wxtim: I think this should be a follow up issue.

oliver-sanders: Besides making the icon smaller, I don't think there is actually very much we can do about this. Word wrapping is not an option here.

However, https://github.com/cylc/cylc-ui/issues/1898 will add inputs to allow the user to manually enter the task & cycle similar to the log view. These will likely use smaller font so may display more of the task name, although the log view task input will also get cut off after a point.

src/components/cylc/Info.vue
```diff
@@ -0,0 +1,261 @@
...
+    padding: 0.5em;
+
+    // for user-defined text where whitespace should be preserved
+    .markup {
+      white-space: pre;
+    }
+
+    // prefixes a tick or cross before the entry
+    .condition:before {
+      content: '✕';
```
wxtim, line 214: I think that this makes more sense: Each output can be produced or not, where as to my mind the tick/cross suggests a more succeed/failed outcome.
```suggestion
      content: '☐';
```

src/components/cylc/Info.vue
```diff
@@ -0,0 +1,261 @@
...
+      padding-right: 0.5em;
+      color: rgb(255, 100, 100);
+    }
```
wxtim, line 217: I think that this makes more sense: Each output can be produced or not, where as to my mind the tick/cross suggests a more succeed/failed outcome.
```suggestion
      content: '☐';
```

```suggestion
    .condition:before {
      content: '☐';
      padding-right: 0.5em;
      color: rgb(0, 0, 0);
    }
```

src/components/cylc/Info.vue
```diff
@@ -0,0 +1,261 @@
...
+    </div>
+
+    <!-- The task information -->
+    <v-expansion-panels
+      multiple
+      variant="accordion"
+      v-model="panelExpansionModel"
+    >
+      <!-- The metadata -->
+      <v-expansion-panel class="metadata-panel">
```
wxtim, line 51: IMO most users will be coming here for the pre-reqs and outputs and not the metadata - these should go first and probably not be collapsed. Could they go in neighbouring Columns?

Happy to spin this out into an issue and not block this PR.

wxtim (approved):
- [x] Manually tested
- [x] Read PR content

I've made some comments, none of which should block the PR going in if they cost any effort to implement.

src/components/cylc/Info.vue
oliver-sanders: I don't think an empty box is clearer than a cross?

This view copies the format used in the scheduler logs and the `cylc view` command.

src/components/cylc/Info.vue
oliver-sanders: Columns aren't an option (or at least aren't always an option) because they won't work in narrow widths.

We can't expand all of the sections by default as there will be a lot of them, from prerequisites, to completion expression satisfaction, to runtime configuration and broadcasts. Prereqs are important, but so are all the other sections.

We could start with nothing expanded, but I thought the title, description and URL (often used to link to documentation) are notably important and deserve to be displayed at all times. Important messages like "do not re-run this task" or "if it fails, try re-running in shortstep mode" will be prominent as a result.

src/components/cylc/Info.vue
MetRonnie: I don't think we should necessarily use the `GraphNode` component to display the task name and jobs, we can do this more bespoke for this view. But happy to accept the quick win here and punt this to either #1898 or a new issue

oliver-sanders: Aw, I quite like the `GraphNode`, it gives easy access to the task and jobs in a familiar format.

oliver-sanders: Haven't really thought through #1898 yet, we could lift-and-shift the task/workflow selectors from the log view as we need much the same thing here, e.g. this view should display both workflow and task metadata.

src/components/cylc/Info.vue
MetRonnie: I agree with Tim's point about green/red being interpreted as success/error.

In the screenshot it looks like the `failed` output is a good thing, while 

image

Also would good to differentiate expected outputs, e.g. by greying out `expired`, `submit-failed` etc. if they are not completed

src/components/cylc/Info.vue
```diff
@@ -0,0 +1,261 @@
...
+      <svg
+        preserveAspectRatio="xMinYMin"
+        viewBox="-40 -40 99999 200"
+        height="6em"
+      >
+        <GraphNode
+          :task="task"
+          :jobs="task.children"
+          :jobTheme="jobTheme"
+        />
```
MetRonnie, line 40: Something is going wrong with the command menu. It does not know what type of Cylc object it is representing (task or cycle or workflow etc.)

image

src/components/cylc/Info.vue
```diff
@@ -0,0 +1,261 @@
...
+        <v-expansion-panel-title color="blue-grey-lighten-2">
+          Prerequisites
+        </v-expansion-panel-title>
+        <v-expansion-panel-text>
+          <ul>
+            <li v-for="prereq in prerequisites" :key="prereq.expression">
+              <span
+                class="prerequisite-alias condition"
+                :class="{satisfied: prereq.satisfied}"
+              >{{ prereq.expression }}</span>
```
MetRonnie, line 98: The prereq expression is a little confusing - why is `c` in it? 

image

src/components/cylc/Info.vue
oliver-sanders: We don't have required / optional information available via GraphQL at the moment, so all we can do is report if an output has been completed or not. We cannot determine whether the completion or non-completion of an output is a good or a bad thing without that information

src/components/cylc/Info.vue
oliver-sanders: Curious, will take a look.

I'm guessing the graph view doesn't show this issue?

src/components/cylc/Info.vue
oliver-sanders: This is (currently) how Cylc represents conditional expressions. Please try using the `cylc show` command which this view is replicating.

At the moment, this is the only way the GUI can represent a conditional expression. I have opened an issue to look at improving this, see #1893 for more deets.

MetRonnie: `cylc show` does not have `c` in it, just numbers

oliver-sanders: Interesting. The `c` prefix is present in the GraphQL response:

```graphql
query {
  workflows {
    taskProxies {
      id
      prerequisites {
        satisfied
        expression
        conditions {
          taskId
          message
          reqState
          exprAlias
          satisfied
        }
      }
    }
  }
}
```

E.G:

```
{
  "data": {
    "workflows": [
      {
        "taskProxies": [
          {
            "id": "~me/show/run1//1/b",
            "prerequisites": []
          },
          {
            "id": "~me/show/run1//1/x",
            "prerequisites": [
              {
                "satisfied": false,
                "expression": "c0 & (c1 | c2) & (c3 | c4)",
                "conditions": [
                  {
                    "taskId": "1/a",
                    "message": "satisfied naturally",
                    "reqState": "succeeded",
                    "exprAlias": "c0",
                    "satisfied": true
                  },
```

I guess we could strip the first character? Presumably `cylc show` does that itself?

oliver-sanders: Yep, `cylc show` strips the `c`:

https://github.com/cylc/cylc-flow/blob/b8be2e3ef5b37288df5de4d62642b0d84e8eb02d/cylc/flow/scripts/show.py#L315

I'll update this to do the same.

oliver-sanders: Done.

MetRonnie: Here is where the `c` comes from (suddenly remembered seeing this on my `cylc-remove` work): https://github.com/cylc/cylc-flow/blob/63a6a377982eb91875a260e6b734435b3ae00740/cylc/flow/prerequisite.py#L239

Don't know why we can't just remove it?

src/components/cylc/Info.vue
oliver-sanders: Sorted.

The data-store node standin thinggy was missing the `type` field which the menu apparently uses.

src/components/cylc/Info.vue
oliver-sanders: `cylc show` uses `string.strip('c')` and I've used `string.replace(/c/g, '')` here, so we should be safe to remove it.

oliver-sanders: For #1893, I think we should remove the "alias" as a user-facing thing and find another way to display it (we're not bound by plain text here).

src/components/cylc/Info.vue
```diff
@@ -0,0 +1,261 @@
...
+              <dd><span class="markup">{{ value }}</span></dd>
+              <v-divider />
+            </template>
+          </dl>
+        </v-expansion-panel-text>
+      </v-expansion-panel>
+
+      <!-- The prereqs -->
+      <v-expansion-panel class="prerequisites-panel">
+        <v-expansion-panel-title color="blue-grey-lighten-2">
```
MetRonnie, line 89: This one is inconsistent with the others
```suggestion
        <v-expansion-panel-title color="blue-grey-lighten-1">
```

oliver-sanders: On purpose.

There will be other sections in the near future. I tried adding a few extras and it ended up as a giant block of "blue-grey" which didn't make it especially obvious that these are individual sections that can be expanded.

I thought it was clearer if the sections were striped ("lighten-1", "lighten-2", "lighten-1", ...).

src/services/mock/json/infoView.json
```diff
@@ -0,0 +1,81 @@
...
+                {
+                  "taskId": "20000102T0000Z/succeeded",
+                  "reqState": "succeeded",
+                  "expreAlias": "c0",
+                  "satisfied": true
+                },
+                {
+                  "taskId": "20000102T0000Z/eventually_succeeded",
+                  "reqState": "succeeded",
+                  "expreAlias": "c1",
```
MetRonnie, line 48: Typo
```suggestion
                  "exprAlias": "c0",
                  "satisfied": true
                },
                {
                  "taskId": "20000102T0000Z/eventually_succeeded",
                  "reqState": "succeeded",
                  "exprAlias": "c1",
```

src/services/mock/json/infoView.json
```diff
@@ -0,0 +1,81 @@
...
+                "my custom field": "My custom value!"
+              }
+            }
+          },
+
+          "jobs": [
+            {
+              "id": "~user/one//20000102T0000Z/failed/01",
+              "jobId": "1234",
+              "startedTime": 0,
```
MetRonnie, line 29: Getting a console warning
```
[Vue warn]: Invalid prop: type check failed for prop "startTime". Expected String with value "0", got Number with value 0. 
```

```suggestion
              "startedTime": "0",
```

src/components/cylc/Info.vue
MetRonnie: To me it doesn't really indicate clickability, it looks like it's selected or has keyboard focus or something. Not too important

src/components/cylc/Info.vue
```diff
@@ -0,0 +1,261 @@
...
+  },
+
+  props: {
+    task: {
+      required: true,
+    },
+    panelExpansion: {
+      required: false,
+      default: [0],
+    },
```
MetRonnie, line 159: Any need to have this as a prop?

src/components/cylc/Info.vue
```diff
@@ -0,0 +1,261 @@
...
+  },
+
+  setup (props, { emit }) {
+    return {
+      jobTheme: useJobTheme(),
+    }
+  },
+
+  computed: {
```
MetRonnie, line 168: (Funny, now that I am used to using `setup` composition API, having separate sections for `computed`, `methods` etc seems rather arbitrary)

src/components/cylc/Info.vue
oliver-sanders: This enables clean separation of the component and view.

The view handles data provision, configuration, session restoration, etc.

The component just renders the content.

oliver-sanders:
Unfortunately, this work may be blocked by https://github.com/cylc/cylc-flow/issues/6307 which also imposes some back-compatibility constraints :(

I may be able to hack around it with special logic for handling the `prerequisites` and `outputs` fields as a workaround though.

src/components/cylc/Info.vue
```diff
@@ -0,0 +1,261 @@
...
+    padding: 0.5em;
+
+    // for user-defined text where whitespace should be preserved
+    .markup {
+      white-space: pre;
+    }
+
+    // prefixes a tick or cross before the entry
+    .condition:before {
+      content: '☐';
```
MetRonnie, line 214: https://github.com/oliver-sanders/cylc-ui/pull/113

> Had a little play about - I like `mdiProgressHelper` as an icon to indicate outputs that haven't been completed yet
>
> image

oliver-sanders:
We could overcome the block by moving from a delta subscription to a regular subscription.

However, this would request the complete data for the task on every change.

src/components/cylc/Info.vue
```diff
@@ -0,0 +1,261 @@
...
+}
+</script>
+
+<style scoped lang="scss">
+  .c-info {
+    padding: 0.5em;
+
+    // for user-defined text where whitespace should be preserved
+    .markup {
+      white-space: pre;
```
MetRonnie, line 209: To allow to text to wrap instead of overflowing
```suggestion
      white-space: pre-wrap;
```

src/components/cylc/Info.vue
oliver-sanders: Used filled and unfilled circles in native CSS/unicode (avoids need for surplus Vue components) to match the Cylc task icon convention.

oliver-sanders:
The cylc-flow blocker to this has not been resolved, however, I have come up with a workaround: https://github.com/cylc/cylc-flow/pull/6403

I have also added an output completion expression so this should now be more or less a direct analogue of `cylc show` with a similar display interface.

src/components/cylc/Info.vue
oliver-sanders: And using one line of CSS avoids the need for a component?!

MetRonnie: It's 10 lines of CSS + 4 lines of template, vs 11 lines of template + 2 imports: https://github.com/oliver-sanders/cylc-ui/pull/113/commits/61f46c0c441be5589f2b59995af7f98210a7c05f#diff-11a1594bf34a47d021d9d6a0cfe88f5b235480cba1d7cb743a4ad7134238a594

oliver-sanders: eh? +25/-20!

MetRonnie:
Have you/are you considering a client-side workaround (i.e. in the `onUpdated` method of the `InfoCallback`) rather than https://github.com/cylc/cylc-flow/pull/6403?

Other feedback: to me, these look like bullet points rather than yes/no indicators. (Tested on Windows, FF & Chrome)

image

oliver-sanders:
> Have you/are you considering a client-side workaround 

It's possible, but it would require iterating over the received delta and the fix would only apply to the info view and not to other views which might want this information (e.g. the tree view currently uses it).

I think the long-term fix should be cylc-flow end so would rather stick any bodges / workarounds there if possible. Happy to undraft the hack PR, was hoping we could come up with a more global fix.

hjoliver:
Unblocked by merge of https://github.com/cylc/cylc-flow/pull/6403 ... I think

cypress/component/info.cy.js
```diff
@@ -0,0 +1,282 @@
...
+        expect(selector[2].classList.toString()).to.equal('prerequisite-alias condition')
+
+        expect(selector[0]).to.contain('2')
+        expect(selector[3].classList.toString()).to.equal('prerequisite-alias condition')
+
+        expect(selector[0]).to.contain('0')
+        expect(selector[4].classList.toString()).to.equal('prerequisite-alias condition satisfied')
+
+        expect(selector[0]).to.contain('0')
+        expect(selector[5].classList.toString()).to.equal('prerequisite-alias condition satisfied')
```
MetRonnie, line 180: - These `expect(selector[0]).to.contain('N')` were all testing the topmost item in the list
- Less brittle to check for/against the existence of the particular class rather than the whole class list
```suggestion
        expect(selector[0].innerText).to.equal('(0 & 1) | 2')
        expect(selector[0]).to.not.have.class('satisfied')

        expect(selector[1].innerText).to.equal('0 a:succeeded')
        expect(selector[1]).to.have.class('satisfied')

        expect(selector[2].innerText).to.equal('1 b:custom_output')
        expect(selector[2]).to.not.have.class('satisfied')

        expect(selector[3].innerText).to.equal('2 a:expired')
        expect(selector[3]).to.not.have.class('satisfied')

        expect(selector[4].innerText).to.equal('0')
        expect(selector[4]).to.have.class('satisfied')

        expect(selector[5].innerText).to.equal('0 x:succeeded')
        expect(selector[5]).to.have.class('satisfied')
```

src/components/cylc/commandMenu/Menu.vue
```diff
@@ -222,7 +222,7 @@ export default {
 
   methods: {
     isEditable (mutation, authorised) {
-      return mutation.name !== 'log' && !this.isDisabled(mutation, authorised)
+      return mutation.name !== 'log' && mutation.name !== 'info' && !this.isDisabled(mutation, authorised)
```
MetRonnie, line 225:
```suggestion
    isEditable (mutation = {}, authorised = false) {
      return !['log', 'info'].includes(mutation.name) && !this.isDisabled(mutation, authorised)
```

tests/e2e/specs/info.cy.js
```diff
@@ -0,0 +1,47 @@
...
+  it('works', () => {
+    // test opening the "Info View" from a task in the "Tree View"
+    cy.visit('/#/workspace/one')
+      // click on task 20000102T0000Z/failed
+      .get('.c-treeitem .c-treeitem .c-treeitem:first')
+      .find('.c-task')
+      .click({ force: true })
+
+      // from the menu select the "Info" psudo-mutation
+      .get('.v-list > :nth-child(6)')
```
MetRonnie, line 28: Can make these less brittle
```suggestion
      // click on task
      .get('.node-data-task [data-c-interactive]:first')
      .click({ force: true })

      // from the menu select the "Info" psudo-mutation
      .get('.v-list-item')
```

src/views/views.js
```diff
@@ -56,16 +72,14 @@ export const TREE = 'Tree'
...
 export const allViews = new Map([
-  [TREE, { component: TreeView, icon: mdiFileTree }],
-  ['Table', { component: TableView, icon: mdiTable }],
-  ['Graph', { component: GraphView, icon: mdiGraph }],
-  ['Log', { component: LogView, icon: mdiFileDocumentMultipleOutline }],
-  ['Analysis', { component: AnalysisView, icon: mdiChartLine }],
-  ['Gantt', { component: GanttView, icon: mdiChartGantt }]
+  ...workspaceViews,
+  ['Info', { component: InfoView, icon: mdiInformationOutline }],
 ])
```
MetRonnie, line 77: Note uses of `allViews` elsewhere. The info view shouldn't be selectable as the default view in settings: 
image

MetRonnie (approved):
Approved although ideally pending the fix for default view selection

wxtim:
Re-reviewed and happy. 👍🏼

PR #1904, opened 2024-08-14
yarn up

oliver-sanders:
Upgrade deps.


oliver-sanders:
Had a run through various functionalities of the UI, appears to be in good order.

MetRonnie (approved):
tested

Issue #1917, opened 2024-08-29
Gantt chart view - choose timezone


ColemanTom:
### Problem

I was looking at the Gantt view to see timings and was getting confused because the time along the x-axis was local time. I was expecting UTC because that is what I was wanting - workflow is triggered/runs using UTC, so analysis of it I tend to do in UTC.

### Proposed Solution

Provide the ability to choose timezone, or at least a switch to allow changing between Local and UTC. Or perhaps, have both visible along the x-axis.

Also, the x-axis should not just say `Time`, it should be `Time (tz)` where tz is the display timezone.

ChrisPaulBennett:
Ah, yes. Slight oversight on my part. I'm currently swamped at the moment. If someone else wants to deal this, you are very welcome to. Other wise I'll try and fit this in

ChrisPaulBennett:
I've had a look at this and none of my code is altering the time. I get the time values from a GraphQL query, which gives me the time in Zulu time (I'm assuming that's what the z means). Zulu time is the same as UTC.
The x-axis is auto-generated by chart library (I will guess the scale from the max and min times in the data). So the x-axis will always match the data.

Can you try the same GraphQL commands as shown below.

Are you sure that you are getting local time?
If you definitely are, I think it might be graphQL giving you time based on your time zone

![image](https://github.com/user-attachments/assets/4a1297b2-f139-4a08-8979-55cc3eba633e)

ColemanTom:
> Are you sure that you are getting local time?

Definitely. I just ran a suite, Gantt chart is sowint 905am instead of 2305UTC

>  If you definitely are, I think it might be graphQL giving you time based on your time zone

When I run your query, it returns the results in UTC time, not local time. 

![image](https://github.com/user-attachments/assets/c132b0c7-9472-428d-857e-08e8bb1ec2fe)

JS frontend perhaps?

ChrisPaulBennett:
Curious. I'll have another look

ChrisPaulBennett:
The JavaScript functions that get time all convert to local time by default (Which should have been obvious to me).
x-axis strings are now UTC 
https://github.com/cylc/cylc-ui/pull/1946

MetRonnie:
To summarise what is left to do:
- (cylc-flow) expose UTC mode via the GraphQL schema so it can automatically be chosen by the Gantt view
- Dropdown to choose between local time or UTC?

PR #1946, opened 2024-09-24
Display x-axis in UTC

ChrisPaulBennett:
This pull request fixes this issue
https://github.com/cylc/cylc-ui/issues/1917

**Check List**


hjoliver:
Sorry I'm late to the party on this issue.

This PR clearly works as advertised, but we probably need to use the appropriate time zone for the workflow, not just assume UTC.

If you're not running NWP or climate science workflows the local timezone is probably preferred, and I don't think you'll want to see analysis view time axes in UTC.

```cylc
[scheduler]
    UTC mode = False  # scheduler log timestamps, defaults to local time via global config
    cycle point time zone = +1300  # cycle point time zone, defaults to UTC
[scheduling]
    initial cycle point = now
    [[graph]]
        P1D = "foo"
[runtime]
    [[foo]]
        script = "sleep $(( RANDOM % 10 ))"
```

For the gantt view time axis, it should be determined by `[scheduler]UTC mode`.

MetRonnie:
I imagine it would be useful to be able to choose the timezone from a dropdown (either workflow cycle point time zone, UTC, or local).

The cycle point time zone for the workflow can be got by adding this to the query:
```diff
 query ganttQuery ($workflows: [ID]) {
+  workflows(ids: $workflows) {
+    timeZoneInfo {
+      hours
+      minutes
+      stringExtended // if needed
+    }
+  }
   jobs(live: false, workflows: $workflows) {
     ${jobFields.join('\n')}
   }
 }
```

N.B. the UTC mode setting is not exposed in the GraphQL schema at the moment

oliver-sanders (commented):
Tested as working.

It would be good to change the x-axis label to "Time (UTC)" to make this clear.

hjoliver:
> N.B. the UTC mode setting is not exposed in the GraphQL schema at the moment

Hmmm yeah you're right. That's a problem, because cycle point time zone is not what we need for the gantt time axis - which applies even for integer cycling. If you set scheduler `UTC mode` to local time, that's because you want scheduler events logged in local time, which implies you'll also want to see said events in local time in the gantt view.

I guess for now we can just hardwire UTC - because most of our users do NWP and climate workflows - and post an issue to fix it in the future.

hjoliver (approved):

PR #1966, opened 2024-10-16
Add copy button for ID in command menu

MetRonnie:
Brushed off this branch which had been languishing for a while.

image

image

**Check List**

- [x] No docs PR needed


tests/e2e/specs/menu.cy.js
```diff
@@ -120,4 +120,29 @@ describe('Command Menu component', () => {
       .get('.c-mutation-menu')
       .should('not.exist')
   })
+
+  it("copies the object's name to the clipboard", { browser: 'electron' }, () => {
+    // (Access to the clipboard in Cypress only reliably works in Electron)
```
MetRonnie, line 125: We could add Electron to the GH Actions testing matrix if desired

hjoliver (approved):
(quick functional review)

MetRonnie:
Will also add to the log view path by popular request...

oliver-sanders:
> Also removed username from ID displayed in the menu, to make it shorter

The user name was intended to be displayed as the UI was written in a way that allowed it to display multiple user's workflows (see BoM requirement). This is also the reason that the `cylcTree` has a user-name level at the top of the tree.

Given lack of Cylc resourcing and the BoM development of a top-level multi-user interface, I think it is ok to drop this, but we should leave the door open to multi-user functionality in the future (we didn't decide to drop this, we just didn't get around to it).

> We could add Electron to the GH Actions testing matrix if desired

This is what we're testing with locally, makes sense to test in CI.

src/components/cylc/commandMenu/Menu.vue
```diff
@@ -192,6 +198,10 @@ export default {
       return this.mutations
     },
 
+    title () {
+      return this.node.tokens.clone({ user: undefined }).id
```
oliver-sanders, line 202: Sneaky, make sure this is nailed down with a test.

oliver-sanders (approved):

src/components/cylc/commandMenu/Menu.vue
MetRonnie: Added a unit test

PR #2017, opened 2024-12-06
added log autoscroll

markgrahamdawson:
**Check List**

### DIFF
--- a/src/views/Log.vue
+++ b/src/views/Log.vue
@@ -171,6 +171,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             :timestamps="timestamps"
             :word-wrap="wordWrap"
           />
+          <!-- a div to use for autoscrolling -->
+          <div class="auto-scroll-end"></div>
         </template>
       </v-col>
     </v-row>
@@ -188,6 +190,7 @@ import {
   mdiPowerPlugOff,
   mdiPowerPlug,
   mdiWrap,
+  mdiMouseScrollWheel,
 } from '@mdi/js'
 import { btnProps } from '@/utils/viewToolbar'
 import graphqlMixin from '@/mixins/graphql'
@@ -411,6 +414,8 @@ export default {
 
   data () {
     return {
+      timer: null,
+      autoScroll: false,
       controlGroups: [
         {
           title: 'Log',
@@ -428,6 +433,13 @@ export default {
               action: 'toggle',
               value: this.wordWrap,
               key: 'wordWrap',
+            },
+            {
+              title: 'Auto scroll',
+              icon: mdiMouseScrollWheel,
+              action: 'toggle',
+              value: this.autoScroll,
+              key: 'autoScroll',
             },
           ]
         }
@@ -478,6 +490,14 @@ export default {
   },
 
   methods: {
+    scrollToElement (elementCLassName, options) {
+      const el = this.$el.getElementsByClassName(elementCLassName)[0]
+
+      if (el) {
+        // Use el.scrollIntoView() to instantly scroll to the element
+        el.scrollIntoView(options)
+      }
+    },
     setOption (option, value) {
       // used by the ViewToolbar to update settings
       this[option] = value
@@ -562,6 +582,10 @@ export default {
     },
   },
 
+  beforeUnmount () {
+    clearInterval(this.timer)
+  },
+
   watch: {
     jobLog (val, old) {
       // reset the filename when the log mode changes
@@ -569,6 +593,15 @@ export default {
       // go back to last chosen job if we are switching back to job logs
       this.relativeID = val ? this.previousRelativeID : null
     },
+    autoScroll (val, old) {
+      if (old) {
+        clearInterval(this.timer)
+      } else {
+        this.timer = setInterval(() => {
+          this.scrollToElement('auto-scroll-end', { behavior: 'smooth' })
+        }, 1000)
+      }
+    }
   },
 
   // Misc options


markgrahamdawson:
The log e2e tests include tests for toggling timestamps and toggling word wrap
`cylc-ui/tests/e2e/specs/log.cy.js`
I wanted to add one for toggling the auto-scroll but because of the way the mock data is loaded for the tests - and the way this pr is implemented (scroll to the bottom of the log file only happens when new lines are written) its difficult.

markgrahamdawson:
There was discussion regarding the UX. 
It was suggested having the autoscroll automatically turn off if the user scrolls on a log (whilst autoscroll is on). This hasnt been included yet. It would require something like this...

```
export default {
  created () {
    window.addEventListener('scroll', this.handleScroll);
  },
  unmounted () {
    window.removeEventListener('scroll', this.handleScroll);
  },
  methods: {
    handleScroll (event) {
      if (this.autoScroll) {
         this.autoScroll == false
      }
    }
  }
}
```

src/components/cylc/log/Log.vue
```diff
@@ -47,14 +66,29 @@ export default {
...
-      match: ''
+      match: '',
+      ro: null
     }
   },
 
+  mounted () {
+    this.ro = new ResizeObserver(this.onResize)
+    this.ro.observe(this.$el)
+  },
```
MetRonnie, line 90: This does not need to be reactive
```suggestion
    }
  },

  mounted () {
    const ro = new ResizeObserver(this.onResize)
    ro.observe(this.$el)
  },
```

src/views/Log.vue
```diff
@@ -146,32 +146,23 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+      />
+      <template v-else>
+        <log-component
+          data-cy="log-viewer"
+          :logs="results.lines"
+          :timestamps="timestamps"
+          :word-wrap="wordWrap"
+          :error="results.error"
+          :autoScroll="autoScroll"
+          @updating="updating"
```
MetRonnie, line 163: Unused?
```suggestion
```

markgrahamdawson: fixed in 49544afb8128260fe05f5020e87ca8767e666521

src/components/cylc/log/Log.vue
markgrahamdawson: fixed in 49544afb8128260fe05f5020e87ca8767e666521

MetRonnie (commented):
Works well :+1: However I added a check that it scrolls to the bottom when you turn on autoscroll, but it didn't. See https://github.com/markgrahamdawson/cylc-ui/pull/6

MetRonnie (approved):
Works well

src/components/cylc/log/Log.vue
```diff
@@ -16,14 +16,35 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-  >{{ log }}</span></pre>
+  <div ref="wrapper">
+    <v-alert
+      v-if="error"
+      type="error"
+      variant="tonal"
+      density="comfortable"
+      class="mb-4"
+      :icon="$options.icons.mdiFileAlertOutline"
+    >
```
oliver-sanders, line 27: Why did this move from the view into the component?

src/components/cylc/log/Log.vue
```diff
@@ -47,12 +68,40 @@ export default {
...
+  setup (props) {
+    const wrapperRef = useTemplateRef('wrapper')
+    const autoScrollEndRef = useTemplateRef('autoScrollEnd')
+
+    function scrollToEnd () {
+      autoScrollEndRef.value?.scrollIntoView({ behavior: 'smooth' })
     }
+
+    when(wrapperRef, () => {
+      const ro = new ResizeObserver(scrollToEnd)
```
oliver-sanders, line 91: Do we need an `unmount` function to ensure this observer is destroyed correctly when the view is closed?

src/views/Log.vue
```diff
@@ -146,32 +146,22 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
       no-gutters
       class="overflow-auto px-4 pb-2"
     >
-      <v-col>
-        <v-skeleton-loader
-          v-if="id && file && results.connected == null"
-          type="text@5"
-          class="mx-n4 align-content-start"
+    <v-col>
```
oliver-sanders, line 149: I think these indentation changes are erroneous?

src/views/Log.vue
```diff
@@ -358,6 +348,9 @@ export default {
     /** Wrap lines? */
     const wordWrap = useInitialOptions('wordWrap', { props, emit }, false)
 
+    /** AutoScroll? */
+    const autoScroll = useInitialOptions('autoScroll', { props, emit }, false)
```
oliver-sanders, line 352: We might want to consider making the default `true`, thoughts?

MetRonnie: Not until we have a solution for turning off autoscroll when the user scrolls up from the bottom, IMO. We could also automatically turn it on when the user scrolls down to the bottom.

oliver-sanders: > Not until we have a solution for turning off autoscroll when the user scrolls up

I'm hoping this is easy to do?

I would quite like to deliver this functionality before we release this feature.

src/components/cylc/log/Log.vue
oliver-sanders: Can the auto-scrolling be covered in a component test cutting out websockets, data provision and all that?

src/views/Log.vue
```diff
@@ -183,11 +173,11 @@ import { usePrevious, whenever } from '@vueuse/core'
 import { useStore } from 'vuex'
 import {
   mdiClockOutline,
-  mdiFileAlertOutline,
   mdiFolderRefresh,
   mdiPowerPlugOff,
   mdiPowerPlug,
   mdiWrap,
+  mdiMouseScrollWheel,
```
oliver-sanders, line 180: It isn't that obvious that the mouse icon means scroll to the bottom of the page.

Maybe suggest something with a down arrow in it.

src/components/cylc/log/Log.vue
MetRonnie: Yes that's what Mark has done

src/views/Log.vue
MetRonnie: Maybe:
- https://pictogrammers.com/library/mdi/icon/chevron-triple-down/
- https://pictogrammers.com/library/mdi/icon/transfer-down/
- https://pictogrammers.com/library/mdi/icon/mouse-move-down/

src/components/cylc/log/Log.vue
MetRonnie: The observer object should be garbage collectable after the component is unmounted, but yes it is probably a good idea to guarantee it is freed up by doing
```js
onBeforeUnmount(ro.disconnect)
```

src/components/cylc/log/Log.vue
oliver-sanders: @markgrahamdawson, could you move the `<v-alert>` part of this back into the log view (but leave the wrapper stuff that's needed for testing here).

This keeps the error reporting with the bit of the code that generates the errors.

src/components/cylc/log/Log.vue
```diff
@@ -86,12 +87,11 @@ export default {
   },
 
   mounted () {
-    // this.ro.observe(this.$el)
     window.addEventListener('wheel', this.handleScroll)
```
MetRonnie, line 90: I don't think this event includes if the user drags the scrollbar to the bottom.

It looks like `vueuse` has a composable that you can use to detect when you are scrolled to the bottom: https://vueuse.org/core/useScroll/#usescroll: `arrivedState.bottom`

markgrahamdawson: Yes I tried `window.addEventListener('scroll', this.handleScroll) ` but couldnt get it working. Ill take a look at the composable

markgrahamdawson: The vueuse composable doesnt work - Im guessing for the same reason `window.addEventListener('scroll', this.handleScroll)` doesnt work as its doing the same thing under the hood

markgrahamdawson: Is there anything other 'scroll' relating that could be effecting this?
The only other event listener in the app is in `Drawer.vue` and is listening to `mousemove`

MetRonnie: I had a go at it here, main difficulty was sorting out a few niggles with the behaviour: https://github.com/markgrahamdawson/cylc-ui/pull/7

src/views/Log.vue
markgrahamdawson: I could only try the last one of these as the others are not in 
cylc-ui/node_modules/@mdi/js/mdi.d.ts

src/components/cylc/log/Log.vue
```diff
@@ -68,8 +68,9 @@ export default {
       autoScrollEndRef.value?.scrollIntoView({ behavior: 'smooth' })
     }
 
+    const ro = new ResizeObserver(scrollToEnd)
+
     when(wrapperRef, () => {
-      const ro = new ResizeObserver(scrollToEnd)
       watch(
```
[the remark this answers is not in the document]
markgrahamdawson: Fixed in 09bd29f4658c3d2d50682db58c94dd720e824f37

src/views/Log.vue
markgrahamdawson: Set the default to true in 
95554aca3233d61a711a33f378838facc009b625

src/views/Log.vue
markgrahamdawson: icon changed to https://pictogrammers.com/library/mdi/icon/mouse-move-down/
in this commit...
689627cd0d6481bde6c1cec88fb0d2f35b542fdd

src/components/cylc/log/Log.vue
markgrahamdawson: merged here
130858204f18b0df507805455397b88786784aa7

src/views/Log.vue
markgrahamdawson: ...of course this breaks the tests. Before I fix them, do we definitely want autoscroll on by default @oliver-sanders ?

src/views/Log.vue
```diff
@@ -382,7 +382,7 @@ export default {
     }, 500)
 
     /** AutoScroll? */
-    const autoScroll = useInitialOptions('autoScroll', { props, emit }, false)
+    const autoScroll = useInitialOptions('autoScroll', { props, emit }, true)
```
MetRonnie, line 385: Did we discuss this being the default?

src/components/cylc/log/Log.vue
markgrahamdawson: resolved in this commit
09bd29f4658c3d2d50682db58c94dd720e824f37

src/components/cylc/log/Log.vue
markgrahamdawson: resolved in 689627cd0d6481bde6c1cec88fb0d2f35b542fdd

src/views/Log.vue
MetRonnie: Ah I forgot Oliver was contemplating setting the default to autoscroll. Not sure I agree, on the basis that there is no easy way to scroll back to the top if you didn't want it set on. Whereas if you turn it on it will scroll to the bottom for you with one click.

src/views/Log.vue
oliver-sanders: With the scroll-detection stuff working, I think this should be the default.

MetRonnie: > Not sure I agree, on the basis that there is no easy way to scroll back to the top if you didn't want it set on. Whereas if you turn it on it will scroll to the bottom for you with one click.

oliver-sanders: Page Up?

MetRonnie: I don't think a lot of users know they can use home/end keys, or 2-finger scrolling etc!

There are also files where it doesn't make sense to scroll to the bottom, such as the config log and job script

oliver-sanders: Meh, I think autoscroll will still come out favourably on balance.

If you are concerned about people struggling to scroll to the top of the page, it would be easy to add a "scroll to top" button an stick it under the log file.

oliver-sanders: It would be good for us to get https://github.com/cylc/cylc-ui/issues/1146 sorted at some point to allow easier configuration of user preferences.

markgrahamdawson:
> LGTM
> 
> One small issue, changing the log file appears to turn autoscroll off.

This happens when switching between the `workflow` and `job` logs due to this...
```
    // Turn off autoscroll when user scrolls up:
    whenever(() => directions.top, () => {
      // autoScroll.value = false
    })
```

As `directions.top` is true when you switch between the two

markgrahamdawson:
> It seems to happen when switching between log files within the workflow/job modes too?

I think this may be a race condition thing?

1) the file loads 
2) its at the top

What we **want** if autoscroll is on...
3) it scrolls to the bottom 

What we **get** if autoscroll is on...
3) autoScroll is turned off because its at the top


```
    // Turn off autoscroll when user scrolls up:
    whenever(() => directions.top, () => {
      // autoScroll.value = false
    })
```
...commenting this out fixes it.

Is this really the functionality that is wanted? when the user scroll to the **top** autoscroll is turned off?
In the comment it states `// Turn off autoscroll when user scrolls up:`
The useScroll composable has an option for scrolling up

https://vueuse.org/core/useScroll/#usage

Setting autoScroll to off when user scrolls up makes more sense than setting it off when user scroll to top right?

MetRonnie:
`directions.top` means "is scrolling up", not to be confused with `arrivedState.top`

src/views/Log.vue
MetRonnie: Personally, I'd estimate 75% of the time I open the log view I want to go to the bottom, but it automatically happening the other 25% of the time would outweigh the convenience. I just don't see a strong need to turn it on automatically when it can so easily be turned on with the button or by scrolling to the bottom.

markgrahamdawson:
Ive done a fix. 
Only apply the - turn off autoscroll when user scrolls up - behavior when log file data has loaded after the file change.

src/views/Log.vue
```diff
@@ -474,10 +478,44 @@ export default {
...
+        }
+      ]
     }
   },
 
   methods: {
+    scrollToTop () {
+      this.autoScroll = false
+      this.logScrollEl.scrollTo(0, 0)
+    },
```
MetRonnie, line 518: Would be good to keep this alongside the rest of the autoscroll code inside `setup`

src/views/Log.vue
```diff
@@ -152,26 +152,33 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+            :timestamps="timestamps"
+            :word-wrap="wordWrap"
+            :autoScroll="autoScroll"
+          />
+          <v-btn
+            v-if="results.lines.length"
+            class="my-3 mx-auto"
+            @click="scrollToTop">scroll to top
+          </v-btn>
+        </div>
```
MetRonnie, line 181: It is quite common in UIs to see a floating button at the bottom right that takes you up to the top. How about something like
```vue
<v-btn
  v-if="results.lines.length"
  position="fixed"
  location="bottom right"
  class="ma-3"
  @click="scrollToTop"
  :icon="mdiFormatVerticalAlignTop"
/>
```

oliver-sanders:
I've experienced an odd issue where the log view scrolls back to the top of the page, before scrolling back to the bottom of the page every time a log line is added.

This doesn't make any logical sense (there's nothing that should cause this) and it only affects me, not anyone else!

Hopefully just something strange with my browser? Dunno!

Suggest ignoring this for now and investigating if it is ever reported.

oliver-sanders (approved):
Slick.

(nicer than the GitHub actions auto-scrolling feature!)

tests/component/specs/log.cy.js
```diff
@@ -0,0 +1,108 @@
...
+
+  it('scroll to the top', () => {
+    // see: https://on.cypress.io/mounting-vue
+    cy.mount(LogComponent, merge(mountOpts, {
+      props: {
+        logs: logLines(30),
+        autoScroll: true,
+      },
+    })).as('component')
+    cy.get('.v-btn').click({ force: true }).click({ force: true })
```
MetRonnie, line 104: There is an issue where this button needs clicking twice. I have a fix at https://github.com/cylc/cylc-ui/pull/2106

PR #2025, opened 2024-12-10
set default log file by job outcome

### LINKED ISSUE #1241: log view: set default log file by job outcome
The default log file selected should depend on the outcome of a job:

| Outcome | Default File |
| --- | --- |
| submitted, running & succeeded | `job.out` |
| failed | `job.err` |
| submit-failed | `job-activity.log` |

**Pull requests welcome!**

markgrahamdawson:
This pr now implements setting default log file by job outcome in two ways:

1.  **If the user navigates to the log view through the node menu** the job state is passed to the view as in this pr...
https://github.com/cylc/cylc-ui/compare/master...MetRonnie:cylc-ui:menu-log-file

2.  **If the user manually types in the path of the job via the input in the log view** a new query is made as in this pr...
https://github.com/cylc/cylc-uiserver/pull/657

Closes #1241

**Check List**


src/views/Log.vue
```diff
@@ -474,6 +460,18 @@ export default {
...
+      if (this.workflows[0].node.stateTotals.failed) {
+        return 'failed'
+      }
+      if (this.workflows[0].node.stateTotals.submitted | this.workflows[0].node.stateTotals.running | this.workflows[0].node.stateTotals.succeeded) {
+        return 'succeeded'
+      }
+      if (this.workflows[0].node.stateTotals['submit-failed']) {
+        return 'submit-failed'
+      }
+      return 0
```
oliver-sanders, line 474: The workflow's `stateTotals` tells us how many tasks are in each state (e.g. if there are *any* failed tasks in the workflow), not the state of the particular job that we are requesting the log file for.


To get the job status, I think you want something along these lines:

```js
if (this.jobLog) {
  // we are looking at a job log => pick the appropriate log file based on its state

  // get the job from the store
  this.$workflowService.$index[this.id]

  return ...
} else {
   // we are viewing the workflow log => always default to the latest log file

  return ...
}
```

oliver-sanders: Dammit, good point, we could use an offline query, but I'm not sure if we can retrieve the status field without making changes to the backend.

MetRonnie: @markgrahamdawson From discussing today, we'll settle on a live query, like this:
```gql
query JobState($id: ID!) {
  job (id: $id) {
    state
  }
}
```
(As I mentioned, we don't need the query if the `file` in `intialOptions` is pre-populated as in https://github.com/cylc/cylc-ui/compare/master...MetRonnie:cylc-ui:menu-log-file)

src/views/Log.vue
```diff
@@ -509,21 +511,43 @@ export default {
...
+          return
+        }
+        if (result.data.jobs[0]) {
+          if (result.data.jobs[0].state === 'failed') {
+            return 'job.err'
+          }
+          if (result.data.jobs[0].state === 'submit-failed') {
+            return 'job-activity.log'
+          }
+          return 'job.out' // rather than undefined
```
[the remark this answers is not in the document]
markgrahamdawson: added in a52e45089ea747a5dbebf547c124bd29afdc20b5

src/views/Log.vue
```diff
@@ -455,10 +449,17 @@ export default {
...
     workflowTokens () {
       // tokens for the workflow this view was opened for
       return new Tokens(this.workflowId)
     },
+    workflowIDs () {
+      return [this.workflowId]
+    },
+    workflows () {
+      return this.getNodes('workflow', this.workflowIDs)
+    },
```
oliver-sanders, line 462: I'm not sure we're actually using this computed property?

src/views/Log.vue
```diff
@@ -455,10 +449,17 @@ export default {
...
 
   computed: {
+    ...mapGetters('workflows', ['getNodes']),
     workflowTokens () {
       // tokens for the workflow this view was opened for
       return new Tokens(this.workflowId)
     },
+    workflowIDs () {
+      return [this.workflowId]
+    },
```
oliver-sanders, line 459: I'm not sure we're actually using this computed property?

src/views/Log.vue
```diff
@@ -503,6 +504,52 @@ export default {
...
+          return JobStateLogFileMap.get(job.state) || 'job.out'
+        } else {
+          return 'job.out'
+        }
+      } else {
+        // we are viewing the workflow log => always default to the latest log file
+        if (logFiles.length) {
+          // loop through all the options in LOG_FILE_DEFAULTS
+          for (const filePattern of LOG_FILE_DEFAULTS) {
+            // Loop through all the filenames e.g.[job.out, job.status]
```
oliver-sanders, line 542:
```suggestion
            // Loop through all the filenames e.g. [config/..., scheduler/...]
```

markgrahamdawson: Done in 622c4f938da8f03a3adde278b34a24f58dc6fea5

src/views/Log.vue
markgrahamdawson: done in 36274f6f186134586cabb8a2a6688bbfb7709b6e

src/views/Log.vue
markgrahamdawson: Done in 36274f6f186134586cabb8a2a6688bbfb7709b6e

src/views/Log.vue
```diff
@@ -503,6 +497,52 @@ export default {
...
+     * @returns {?string}
+     */
+    async getDefaultFile (logFiles) {
+      if (this.jobLog === 1) {
+        // we are looking at a job log => pick the appropriate log file based on its state
+        // get the job from the store
+        let result
+        try {
+          // get the list of available log files
+          result = await this.$workflowService.apolloClient.query({
```
oliver-sanders, line 515: We've got an interface for performing queries like this. It's got a silly name and isn't documented or explained (sorry). The workflow service code is in need of refactor but rather low priority.

```suggestion
          result = await this.$workflowService.query2({
```

src/views/Log.vue
```diff
@@ -235,6 +236,20 @@ query LogFiles($id: ID!) {
...
+ *
+ * @type {DocumentNode}
+*/
+const JOB_QUERY = gql`
+query JobState($id: [ID!]) {
+  jobs (live: false, workflows: $id) {
+    id
+    state
+  }
+}
```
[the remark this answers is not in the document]
oliver-sanders: Here's the source of the problem:

https://github.com/cylc/cylc-uiserver/blob/aeddbcb01ca68de01258ccdd1c129614361fdde9/cylc/uiserver/schema.py#L306-L307

It only makes use of the `task` argument (which does work but can return the same task from multiple cycles and multiple job submissions for the same cycle)  and completely ignores the `id` argument (because this hasn't been implemented yet).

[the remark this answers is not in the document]
markgrahamdawson: Will the `cylc/uiserver/schema.py` need something like this?? @oliver-sanders 

```

def run_ids_query(conn, workflow, ids):

    # TODO: support all arguments including states
    # https://github.com/cylc/cylc-uiserver/issues/440
    jobs = []

    # Create sql snippet used to limit which tasks are returned by query
    if ids:
        where_clauses = "' OR name = '".join(ids)
        where_clauses = f" AND (name = '{where_clauses}')"
    else:
        where_clauses = ''
    for row in conn.execute(f'''
SELECT
    name,
    cycle,
    submit_num,
    submit_status,
    time_run,
    time_run_exit,
    job_id,
    platform_name,
    time_submit,
    STRFTIME('%s', time_run_exit) - STRFTIME('%s', time_submit) AS total_time,
    STRFTIME('%s', time_run_exit) - STRFTIME('%s', time_run) AS run_time,
    STRFTIME('%s', time_run) - STRFTIME('%s', time_submit) AS queue_time
FROM
    task_jobs
WHERE
    run_status = 0
    {where_clauses};
'''):
        jobs.append({
            'id': workflow.duplicate(
                cycle=row[1],
                task=row[0],
                job=row[2]
            ),
            'name': row[0],
            'cycle_point': row[1],
            'submit_num': row[2],
            'state': row[3],
            'started_time': row[4],
            'finished_time': row[5],
            'job_ID': row[6],
            'platform': row[7],
            'submitted_time': row[8],
            'total_time': row[9],
            'run_time': row[10],
            'queue_time': row[11]
        })
    return jobs

```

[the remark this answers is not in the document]
markgrahamdawson: graphql query

```
query JobState {
  jobs (
    live:false
    workflows: ["runtime-tutorial-families/run1"], 
    # tasks: ["get_observations_aldergrove"], 
    ids:["*/get_observations_aldergrove/*"]
  ) {
    id
    state
  }
}

```

src/views/Log.vue
markgrahamdawson: fixed in ad47dc208af76bb2da0142a835d6dac5de5645ef

src/views/Log.vue
```diff
@@ -503,6 +497,52 @@ export default {
...
+          result = await this.$workflowService.apolloClient.query({
+            query: JOB_QUERY,
+            variables: { id: [this.id] }
+          })
+        } catch (err) {
+          // the query failed
+          console.warn(err)
+          return
+        }
+        const job = result.data.jobs.find(x => x.id === this.id)
```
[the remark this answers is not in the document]
markgrahamdawson: fixed in ad47dc208af76bb2da0142a835d6dac5de5645ef

src/views/Log.vue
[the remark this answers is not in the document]
markgrahamdawson: fixed in ad47dc208af76bb2da0142a835d6dac5de5645ef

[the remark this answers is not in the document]
markgrahamdawson: But now requires https://github.com/cylc/cylc-uiserver/pull/657 this update to UI server to work

oliver-sanders:
LGTM. Works well.

Looking at the "network" tab in the web inspector, it looks like we get two `jobState` queries, even when we are opening a log from the Tree view where we already know what the job state is?

We should be able to disable this check if the file is specified in the initial inputs (the query is only needed if the file is not specified).

src/views/Log.vue
```diff
@@ -503,6 +498,56 @@ export default {
...
+                workflowId: this.workflowTokens.workflow
+              }
+            )
+          }
+        } catch (err) {
+          // the query failed
+          console.warn(err)
+          return
+        }
+        if (result && result.data.jobs.length) {
```
MetRonnie, line 530:
```suggestion
        if (result?.data?.jobs.length) {
```

src/views/Log.vue
```diff
@@ -503,6 +498,56 @@ export default {
...
+      } else {
+        // we are viewing the workflow log => always default to the latest log file
+        if (logFiles.length) {
+          // loop through all the options in LOG_FILE_DEFAULTS
+          for (const filePattern of LOG_FILE_DEFAULTS) {
+            // Loop through all the filenames e.g. [config/..., scheduler/...]
+            for (const fileName of logFiles) {
+              if (filePattern.exec(fileName)) {
+                return fileName
+              }
```
MetRonnie, line 544:
```suggestion
          // Loop through all the filenames e.g. [config/..., scheduler/...]
          // (lexographical sorting ensures the latest log)
          for (const fileName of logFiles) {
            if (fileName.startsWith('scheduler/')) {
              return fileName
```

src/views/Log.vue
```diff
@@ -248,27 +263,6 @@ const LOG_FILE_DEFAULTS = [
   /^scheduler\/*/
 ]
```
MetRonnie, line 264: `LOG_FILE_DEFAULTS` is not really needed anymore (see previous comment)

src/views/Log.vue
```diff
@@ -503,6 +498,56 @@ export default {
...
+            // Loop through all the filenames e.g. [config/..., scheduler/...]
+            for (const fileName of logFiles) {
+              if (filePattern.exec(fileName)) {
+                return fileName
+              }
+            }
+          }
+        }
+      }
+      return null // rather than undefined
```
MetRonnie, line 549: (Why null rather than undefined?)

markgrahamdawson: This is left over from the original code 

```

/**
 * Return the default log file from the given log filenames, if there is a
 * matching filename. Relies on the filenames having been sorted in descending
 * order.
 *
 * @param {string[]} logFiles - list of available log filenames
 * @returns {?string}
 */
export const getDefaultFile = (logFiles) => {
  if (logFiles.length) {
    for (const filePattern of LOG_FILE_DEFAULTS) {
      for (const fileName of logFiles) {
        if (filePattern.exec(fileName)) {
          return fileName
        }
      }
    }
  }
  return null // rather than undefined
}

```

oliver-sanders (approved):
:+1:

Please squash merge.

oliver-sanders:
Note, can't be merged until https://github.com/cylc/cylc-uiserver/pull/657 is in.

src/views/Log.vue
MetRonnie: These 2 test cases are no longer valid as they are testing the code that gets the workflow log because `this.joblog = 0`: https://github.com/cylc/cylc-ui/blob/6e3f8d1eddb5062767868e65a23766c6471dd78e/tests/unit/views/log.vue.spec.js#L79-L99

IMO it would make sense to split this into 2 functions, one for getting the default job log and one for getting the default workflow log

markgrahamdawson: Does this make any sense or not at all?

```
it.each([
    {
      files: [
        'job.err',
        'job.out',
        'job',
        'job-activity.log',
        'job.status',
        'zjob.out',
      ].sort().reverse(),
      expected: {workflow: null, job: 'job.out'},
    },
    {
      files: [
        'job',
        'job-activity.log',
        'job.status',
        'scheduler/pluto',
      ].sort().reverse(),
      expected: {workflow: 'scheduler/pluto', job: 'job.out'},
    },
    {
      files: [
        'scheduler/02-restart-02.log',
        'scheduler/01-start-01.log',
        'install/02-reinstall.log',
        'config/flow-processed.cylc',
        'config/20240212T155825+0000-rose-suite.conf',
        'config/02-restart-02.cylc',
      ].sort().reverse(),
      expected: {workflow: 'scheduler/02-restart-02.log', job: 'job.out'},
    },
    {
      files: [],
      expected: {workflow: null, job: 'job.out'},
    },
    {
      files: ['ceres', 'vesta', 'aphosis'].sort().reverse(),
      expected: {workflow: null, job: 'job.out'},
    },
  ])('workflow log getDefaultFile($files) == $expected', async ({ files, expected }) => {
    const wrapper = mountFunction()
    wrapper.vm.jobLog = 0
    expect(await wrapper.vm.getDefaultFile(files)).toBe(expected.workflow)
    wrapper.vm.jobLog = 1
    expect(await wrapper.vm.getDefaultFile(files)).toBe(expected.job)
  })

```

MetRonnie: That looks sensible enough, but I do think it would be worth splitting the function in two

markgrahamdawson: Like this...

```
it.each([
    {
      files: [
        'job.err',
        'job.out',
        'job',
        'job-activity.log',
        'job.status',
        'zjob.out',
      ].sort().reverse(),
      expected: null,
    },
    {
      files: [
        'job',
        'job-activity.log',
        'job.status',
        'scheduler/pluto',
      ].sort().reverse(),
      expected: 'scheduler/pluto',
    },
    {
      files: [
        'scheduler/02-restart-02.log',
        'scheduler/01-start-01.log',
        'install/02-reinstall.log',
        'config/flow-processed.cylc',
        'config/20240212T155825+0000-rose-suite.conf',
        'config/02-restart-02.cylc',
      ].sort().reverse(),
      expected: 'scheduler/02-restart-02.log',
    },
    {
      files: [],
      expected: null,
    },
    {
      files: ['ceres', 'vesta', 'aphosis'].sort().reverse(),
      expected: null,
    },
  ])('workflow log getDefaultFile($files) == $expected', async ({ files, expected }) => {
    const wrapper = mountFunction()
    wrapper.vm.jobLog = 0
    expect(await wrapper.vm.getDefaultFile(files)).toBe(expected)
  })

  it.each([
    {
      files: [
        'job.err',
        'job.out',
        'job',
        'job-activity.log',
        'job.status',
        'zjob.out',
      ].sort().reverse(),
      expected: 'job.out',
    },
    {
      files: [
        'job',
        'job-activity.log',
        'job.status',
        'scheduler/pluto',
      ].sort().reverse(),
      expected: 'job.out',
    },
    {
      files: [
        'scheduler/02-restart-02.log',
        'scheduler/01-start-01.log',
        'install/02-reinstall.log',
        'config/flow-processed.cylc',
        'config/20240212T155825+0000-rose-suite.conf',
        'config/02-restart-02.cylc',
      ].sort().reverse(),
      expected: 'job.out',
    },
    {
      files: [],
      expected: 'job.out',
    },
    {
      files: ['ceres', 'vesta', 'aphosis'].sort().reverse(),
      expected: 'job.out',
    },
  ])('job log getDefaultFile($files) == $expected', async ({ files, expected }) => {
    const wrapper = mountFunction()
    wrapper.vm.jobLog = 1
    expect(await wrapper.vm.getDefaultFile(files)).toBe(expected)
  })

```

MetRonnie: I am thinking along the lines of
```js
function getDefaultFile() {
  return jobLog ? getDefaultJobLog() : getDefaultWorkflowLog()
}
```

In terms of the testing we wouldn't expect job log files to be in the workflow log dir and vice versa. However, it is still worth a simple case testing an unexpected file in the dir doesn't affect the result

src/views/Log.vue
markgrahamdawson: fixed in 4b2d273d733fab1472e62574592473636174177d

tests/unit/setup.js
```diff
@@ -14,3 +14,4 @@ window.ResizeObserver ??= ResizeObserverStub
 // Mock element scroll API as not currently included in jsdom:
 // https://github.com/jsdom/jsdom/issues/1422
 Element.prototype.scrollBy ??= function () { }
+Element.prototype.scrollIntoView ??= function () { }
```
MetRonnie, line 17: Ok, might as well put the whole lot in to avoid potential future pain for devs:
```suggestion
Element.prototype.scrollIntoView ??= function () { }
Element.prototype.scroll ??= function () { }
Element.prototype.scrollTo ??= function () { }
```

MetRonnie: Although I'm curious as to why this is needed on this branch?

wxtim (approved):

src/model/JobState.model.js
```diff
@@ -42,4 +42,23 @@ class JobState extends Enumify {
...
+ */
+export function getJobLogFileFromState (state) {
+  switch (state) {
+    case JobState.FAILED.name:
+      return 'job.err'
+    case JobState.SUBMITTED.name:
+    case JobState.SUBMIT_FAILED.name:
+      return 'job-activity.log'
+    case JobState.RUNNING.name:
+    case JobState.SUCCEEDED.name:
```
oliver-sanders, line 59: [Minor] A little neater, but possibly a little safer (we have had issues with missing states):

```suggestion
    default:
```

oliver-sanders (commented):
:+1:

oliver-sanders (approved):

Issue #1999, opened 2024-11-17
Old cycles still showing on the WUI


esteban-abellan:
### Description

<img width="213" alt="Screenshot 2024-11-18 092320" src="https://github.com/user-attachments/assets/44c3195e-0fc3-415b-a0bf-14811d72bcb2">

I noticed that WUI is showing old cycles in one of the suites despite these cycles have been completed. I checked the `cylc workflow-state` and all tasks have been succeeded. I played with different `Graph Window Depth` but it doesn't have any impact on the old cycles. We are using the Cylc 8.3.5 version, and Cylc UI 2.5.0. 

Note that this behaviour is not preventing the suite from cycling and it's only present in one of the suites. The rest of the suites, which use the same settings, are ok and do not display any previous cycle if `N=0`, only one if `N=1` etc.

oliver-sanders:
Hi :wave:.

I can't replicate this with my own workflows, so I suspect that there is something about the workflow you are trying to run that is exposing this bug.

If you are able to share the `[scheduling]` section of the workflow, I can have a go at replicating the issue. Also, it would be handy to know if the workflow is defined in a `suite.rc` (back-compat mode) or `flow.cylc` (new style) file.

MetRonnie:
Possible duplicate of https://github.com/cylc/cylc-ui/issues/1813?

ColemanTom:
![Image](https://github.com/user-attachments/assets/c2b475a8-a1ab-417d-b585-da8043b17b25)

This is a fresh load of the browser, so refreshing the tab does nothing to help. This workflow has not been showing the old cycles for a long time (although I notice it is the same one reported by @esteban-abellan). When we did a `cylc stop; sleep; cylc play`, they started building up again. But, other workflows which have near identical graph (changing some numbers in task names only), which stopped and started at the same time, do not have a buildup of old cycles.

Workflows are all defined in `flow.cylc`. There are no deprecation warnings. It may or may not matter, but, the workflows do use the `cylc-rose` plugin.

ColemanTom:
> Possible duplicate of [#1813](https://github.com/cylc/cylc-ui/issues/1813)?

It might be, but in this case, I do not believe the n-window was adjusted since it started again. Maybe the n-window was different, then stop and play'd again, which changed the n-window, so that could be the same root cause. I would need to go through logs to check.

MetRonnie:
> This is a fresh load of the browser, so refreshing the tab does nothing to help.

[...]

PR #2055, opened 2025-01-31
Fix display of cycle points that don't have children in tree view

MetRonnie:
Partially addresses #1999 by fixing the display, but does not fix the cause of the issue.

**Check List**

- [x] Tests are not needed
- [x] No docs needed


oliver-sanders:
This isn't a display issue, it's a problem with the data in the UI store. We should fix the issue rather than trying to cover it over in this way.

MetRonnie:
This is not covering anything up, it's just aiming to fix the display aspect of the issue. However I've realised it still leaves the expand/collapse button visible and it would be tricky to include that, so I'll just pull out the tidying into a new PR

PR #2069, opened 2025-02-12
Downgrade vite to v5.4.14

MetRonnie:
We are having a problem with the rebuild during `yarn build:watch` with vite v6:
- https://github.com/vitejs/vite/issues/19410


MetRonnie:
Ugh, tests failing now, reason unclear.

PR #2100, opened 2025-03-12
CPU and Max RSS Analysis tools

ChrisPaulBennett:
This apart of 3 pull requests for adding CPU time and Max RSS analysis to the Cylc UI.

This adds the Max RSS and CPU time (as measured by cgroups) to the table view, box plot and time series views.

Linked to;
https://github.com/cylc/cylc-flow/pull/6663
https://github.com/cylc/cylc-uiserver/pull/675

**Check List**


MetRonnie:
Btw it looks like you have accidentally pushed this branch both to your fork and this upstream repo. I'm deleting the branch on here but it will remain on your fork.

MetRonnie:
We'll get round to this when we can, but a little bit swamped at the moment!

wxtim:
I've had a first look - 

The "chips" look good: 

image

* Is there a reason I don't get maxrss for the hpc background job?
* Might these be more useful if the data was converted to human readable?

I'm currently getting errors when trying to load the analysis view, causing the view not to load at all:

[console-export-2025-4-16_13-8-9.txt](https://github.com/user-attachments/files/19776758/console-export-2025-4-16_13-8-9.txt)

ChrisPaulBennett:
Oliver recommended the best way to move data around cylc was to use a debug message. Those chips only only appear in dev mode and won't be visible to normal users.
What's a HPC background job?
The analysis view being stuck in that state was introduced when I merged in @oliver-sanders changes to SQL. I haven't figured out what is causing this yet. Clicking refresh will solve the issue.

wxtim:
> What's a HPC background job?

Using the supercomputer login node as just another computer. We use the term background to mean execution a job in bash without using a job runner like PBS or Slurm.

E.g. `> ssh exab bash program.sh` rather than `ssh exab qsub program.sh`

oliver-sanders:
> Is there a reason I don't get maxrss for the hpc background job?

1) You would need to configure this platform to run the profiler.
2) There are no cgroups to poll for background jobs so it wouldn't work if you tried.

src/components/cylc/analysis/AnalysisTable.vue
```diff
@@ -128,50 +132,82 @@ export default {
...
+      const timingHeaders = []
+      // Check if there are any stats to show
+      let stats = false
+      for (let i = 0; i < this.tasks.length; i++) {
+        if (this.tasks[i].count > 1) {
+          stats = true
+          break
         }
-      ]
+      }
```
MetRonnie, line 143:
```suggestion
      const stats = this.tasks.some((task) => task.count > 1)
```

src/utils/tasks.js
```diff
@@ -96,28 +98,83 @@ export function jobMessageOutputs (jobNode) {
...
+    if (value < 5000) {
+      const kilobytes = value
+      return kilobytes.toPrecision(3) + ' KB'
+    } else if (value / 1024 < 1000) {
+      const megabytes = value / 1024
+      return megabytes.toPrecision(3) + ' MB'
+    } else {
+      const gigabytes = value / 1048576
+      return gigabytes.toPrecision(3) + ' GB'
+    }
```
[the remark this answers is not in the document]
ChrisPaulBennett: Actually, I think the name of the function should be change. Originally, when we only had timing on the analysis view, this function displayed the timings in a sensible format. 
It's ensuring the appropriate scale, for the appropriate unit is displayed on the Y-axis. Wouldn't make sense to have two functions doing that.
`formatUnits()` ?

src/components/cylc/analysis/AnalysisTable.vue
ChrisPaulBennett: `some` is a terrible name for a method. I like my old school loops!

src/components/cylc/analysis/BoxPlot.vue
```diff
@@ -241,10 +245,10 @@ export default {
         { title: 'Task name', value: 'name' },
         { title: 'Platform', value: 'platform' },
         { title: 'Count', value: 'count' },
-        { title: `Mean ${this.timingOption} time`, value: `mean${upperFirst(this.timingOption)}Time` },
-        { title: `Median ${this.timingOption} time`, value: `median${upperFirst(this.timingOption)}Time` },
-        { title: `Min ${this.timingOption} time`, value: `min${upperFirst(this.timingOption)}Time` },
-        { title: `Max ${this.timingOption} time`, value: `max${upperFirst(this.timingOption)}Time` },
+        { title: `Mean ${formatChartLabels(this.timingOption)}`, value: `mean${upperFirst(getTimingOption(this.timingOption))}` },
+        { title: `Median ${formatChartLabels(this.timingOption)}`, value: `median${upperFirst(getTimingOption(this.timingOption))}` },
```
MetRonnie, line 249: Not sure what is happening here but in the offline mode, sorting by median time is not working for any metric

Image

Possibly missing from offline data?

src/utils/tasks.js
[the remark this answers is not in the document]
ChrisPaulBennett: Not sure I get you? 
The the function won't change. I'm just changing the name to better reflect it's purpose

src/components/cylc/analysis/BoxPlot.vue
oliver-sanders: Ping @ChrisPaulBennett

src/utils/tasks.js
```diff
@@ -132,7 +132,9 @@ export function formatDuration (value, allowZeros = false, timingOption = true)
 
 function formatRSS (value) {
   // Format the max RSS value in a human-readable format
-  if (value / 1024 < 1000) {
+  if (value === 0 || value === undefined || value === null) {
+    return undefined
```
MetRonnie: https://github.com/cylc/cylc-uiserver/pull/675#issuecomment-3480728965

> I don't think that fix is valid - the UI should treat 0 as 0 seconds, and null/undefined should be treated as empty. So the uiserver should send null when there is no data rather than 0. Some tasks may run in ~0 seconds so 0 is a valid normal value.

ChrisPaulBennett:
Please squash merge

src/utils/tasks.js
```diff
@@ -135,17 +135,17 @@ export function formatDuration (value, allowZeros = false, timingOption = true)
 }
 
 function formatRSS (value) {
-  // Format the peak RSS value in a human-readable format
+  // Format the peak RSS value in a human-readable format.
+  // The value is provided in megabytes (MB).
```
MetRonnie: Megabytes or mebibytes? (see https://github.com/cylc/cylc-flow/pull/7451/changes#r3977949450)

ChrisPaulBennett: You are technically correct. Which, I have been reliably informed, is the best type of correct.

src/utils/tasks.js
```diff
@@ -97,26 +99,114 @@ export function jobMessageOutputs (jobNode) {
...
+  } else if (value < 1) {
+    const kilobytes = value * 1024
+    return kilobytes.toPrecision(3) + ' KB'
+  } else if (value < 1000) {
+    return value.toPrecision(3) + ' MB'
+  } else {
+    const gigabytes = value / 1024
+    return gigabytes.toPrecision(3) + ' GB'
+  }
+}
```
MetRonnie, line 151: The value from GraphQL is remaining in bytes now, so this needs changing. Simply revert `dbd36d3795d1003124c27fed90d85e783ae16313` and make sure the units are stated as binary units (KiB, MiB etc)?

ChrisPaulBennett: Done

PR #2102, opened 2025-03-12
Xtriggers and task run mode in info view

wxtim:
Requires https://github.com/cylc/cylc-flow/pull/6671 [MERGED]

**Check List**


wxtim:
Broken for this workflow::

```flow.cylc
#!/bin/bash
TMP=$(mktemp -d)
echo Files extracted to ${TMP}
mkdir -p "${TMP}/."

cat > "${TMP}/./flow.cylc" <<__ICI__
[scheduler]
    allow implicit tasks = True
    cycle point format = %Y

[scheduling]
    initial cycle point = 2026
    [[graph]]
        R1 = """
            @myxt => foo
            @echo => foo
        """
    [[xtriggers]]
        echo = echo('Hello, world!', succeed=False)
        myxt = myxt():PT1S

[runtime]
    [[foo]]
        [[[meta]]]
            description = """
                Xtriggers dissapear from the gui

                to make myxt return true

                touch ~/foo.flag
            """__ICI__

mkdir "${TMP}/./lib"

mkdir "${TMP}/./lib/python"

cat > "${TMP}/./lib/python/myxt.py" <<__ICI__
from pathlib import Path

def myxt():
    """Switched by the existance of a file foo.flag in the users home
    directory.
    """
    if (Path.home() / 'foo.flag').exists():
        return True, {}
    return False, {}
__ICI__
```

When one touches the `foo.flag` file the state of `myxt` correctly displays as satisfied, but the still unsatisfied `echo` trigger dissapears.

wxtim:
Note to self - remove the note from cylc doc added in https://github.com/cylc/cylc-doc/pull/816 about info view not showing xtriggers.

wxtim:
PR is ready for review, but only works in combination with https://github.com/cylc/cylc-flow/pull/6671

wxtim:
Checks cancelled because diff is https://github.com/cylc/cylc-ui/compare/077f8cd1ae3363314ac038512e84b079f8bfee10..73a93e22501b5e39032ad35029c8db0f81184004 no change

MetRonnie:
I get a blank Info view on this branch, I'm afraid (cylc-flow PR checked out, cylc-uiserver at master)

wxtim:
> I get a blank Info view on this branch, I'm afraid (cylc-flow PR checked out, cylc-uiserver at master)

Can confirm - I've made a mistake in the graphql - I'll have a poke at some point, but i think some of the reviewing I have is more urgent

wxtim:
> Unfortunately, if there is a problem/mistake in a subscription query, we don't seem to be getting any error reporting

My evidence is that I checked out the upstream/master `src/views/Info.vue` and it all worked fine. This also demonstates that I haven't used the extra data, so it's pointless, and that I hadn't finished something I was doing on branch!

tests/e2e/specs/info.cy.js
```diff
@@ -42,5 +42,37 @@ describe('Info View', () => {
       .click({ force: true })
       .get('.v-expansion-panel--active')
       .should('have.length', 2)
+      .and('contain', 'Live')
+
+      // Xtriggers should be ticked or unticked
+      // depending on fulfillment:
+      .get('.c-info .v-expansion-panel:nth-child(3)')
```
wxtim, line 49: This feels unduly fragile? It certainly doesn't tally with https://docs.cypress.io/app/core-concepts/best-practices#Selecting-Elements.

But I left it because I thought that we might want to test the order of info view sections.

hjoliver:
(I've started looking at this one)

oliver-sanders:
For context, xtrigger state filtering, and task icon badges will follow in future PRs, this one is just about the info view.

tests/e2e/specs/info.cy.js
wxtim: The best practices didn't seem that keen on just re-using classes - or is this ok because it's a well defined class and we're happy to take on any fragility?

MetRonnie: Using class names is more fragile than data-cy attributes, but not too bad. I'm happy with either.

oliver-sanders: > Using class names is more fragile than data-cy attributes, but not too bad. I'm happy with either.

Class names are no more fragile than data attributes, classes are flags, attributes are key=value pairs, use the on that makes sense for the application.

oliver-sanders:
Something is going wrong with the trigger time field:

Screenshot from 2025-06-10 13-00-40

And what's up with the sort order? Why is the latest xtrigger in the middle?

src/components/cylc/Info.vue
```diff
@@ -84,6 +84,42 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+        <v-expansion-panel-title color="blue-grey-lighten-2">
+          Run Mode
+        </v-expansion-panel-title>
+        <v-expansion-panel-text>
+          <v-icon>{{ runModeIcon }}</v-icon>  {{ runMode }}
+        </v-expansion-panel-text>
+      </v-expansion-panel>
+
+      <v-expansion-panel class="xtriggers-panel">
+        <v-expansion-panel-title color="blue-grey-lighten-2">
```
oliver-sanders, line 97: I was going for striping with these panels because the default style doesn't make it overly obvious that this is a collection of panels when there are lots of them and they are all collapsed. I'm sure we'll come up with something nicer in the future.

```suggestion
        <v-expansion-panel-title color="blue-grey-lighten-1">
```

src/components/cylc/Info.vue
```diff
@@ -84,6 +84,42 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+              <td>{{ xt.label }}</td>
+              <td>{{ xt.id }}</td>
+              <td class="d-flex justify-center"><v-icon>{{ xt.satisfactionIcon }}</v-icon></td>
+              <td>{{ xt.trigger_time }}</td>
+            </tr>
+          </tbody></table>
+          <p class="footnote">
+            * For wall_clock triggers.
+          </p>
+          </div>
```
MetRonnie, line 119:
```suggestion
```

tests/e2e/specs/info.cy.js
```diff
@@ -42,5 +42,37 @@ describe('Info View', () => {
       .click({ force: true })
       .get('.v-expansion-panel--active')
       .should('have.length', 2)
+      .and('contain', 'Live')
+
+      // Xtriggers should be ticked or unticked
+      // depending on fulfillment:
```
MetRonnie, line 48: This comment does not match the test, which is only checking the xtrigger names are in the table.
```suggestion
```

src/components/cylc/Info.vue
```diff
@@ -295,15 +294,17 @@ export default {
         }
 
         // Extract the trigger time from the ID
-        const re = /trigger_time=(?<unixTime>[0-9]+)/
+        const re = /trigger_time=(?<unixTime>[0-9.]+)/
```
oliver-sanders, line 297: `[0-9.]+` is equivalent to `[.]+`?

oliver-sanders:
The re-formatted trigger-time field looks great.

The xtriggers are still being displayed in the order the backend provides them in which is unsorted. Not a biggie, but I think it would be good for the info view to sort by label, then signature.

Screenshot from 2025-06-20 10-34-40

src/components/cylc/Info.vue
oliver-sanders: I think this got wiped out in a force-push or something?

tests/e2e/specs/info.cy.js
```diff
@@ -42,5 +42,35 @@ describe('Info View', () => {
...
+      .find('button')
+      .click({ force: true })
+      .get('.v-expansion-panel--active')
+      .should('have.length', 5)
+
+      .get('.c-info .v-expansion-panel:nth-child(6)')
+      .find('button')
+      .click({ force: true })
+      .get('.v-expansion-panel--active')
+      .should('have.length', 6)
```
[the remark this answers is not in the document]
wxtim: > Edit: https://github.com/cylc/cylc-ui/pull/2102#discussion_r2094993022 was marked as resolved but is not?

It's not resolved, but it's orthoganal to the main purpose of the PR.

MetRonnie (approved):
Noticed wallclock xtriggers that are past the point where they should be satisfied are not showing as satisfied so long as there are other unsatisfied prerequisites. But this is surely a cylc-flow problem and not this PR

Issue #1536, opened 2023-10-30
Use `eslint-plugin-vue` strongly-recommended rules


MetRonnie:
We should turn these ESLint rules to improve consistency. https://eslint.vuejs.org/rules/

oliver-sanders:
Looks good.

Note, will result in a lot of diff (maybe wait for a quiet period) and a bit of work (not all rules have fixers).

PR #2125, opened 2025-04-04
ESLint: enforce trailing commas and html indentation

MetRonnie:
Enough is enough 😄 

Partially addresses https://github.com/cylc/cylc-ui/issues/1536 (`vue/html-indent` is one of the strongly-recommended rules)

The 2 rules are added in separate commits, and then then lint is fixed in the last commit where I simply ran `yarn lint --fix` - **you don't need to review every line of that commit**

> [!NOTE]
> Merging this will most certainly cause other PRs to become conflicting, but that is not a problem as they can just run `yarn lint --fix`

> [!TIP]
> If not already done so, set your IDE to automatically fix fixable ESLint problems on save, you'll be doing yourself a massive favour

**Check List**

- [x] Tests, changelog and docs not needed

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -15,6 +15,14 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+const commaDangle = {
+  arrays: 'always-multiline',
+  objects: 'always-multiline',
+  imports: 'always-multiline',
+  exports: 'always-multiline',
+  functions: 'only-multiline',
+}
+
 module.exports = {
   root: true,
   parserOptions: {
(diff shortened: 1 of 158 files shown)


MetRonnie:
Ah no I was wrong, before we were merely allowing dangling commas when on multiline arrays and objects (but if a trailing comma was left on a non-multiline array/object, the formatter would strip it). Now I've set it to enforce for multiline arrays/objects so that if a trailing comma is missing from a multiline array/object the formatter will add it (non-multiline dangling commas will still be stripped). This should make life easier as you can add entries to multiline arrays/objects without having to fiddle with missing dangling commas on the line above.

samuel-denton (approved):
LGTM.

ChrisPaulBennett (approved):
Trailing commas feel wrong to me. Like, they feel like a mistake. 
But apparently they help with editing and diffs. Reluctant approval.....ughhh....

MetRonnie:
Shh, just turn on auto lint fixes in your IDE and forget all your troubles...

PR #2129, opened 2025-04-07
Graph grouping/collapsing code tidy

MetRonnie:
Follow-up to #1810
Partly addresses #1130

**Check List**

- [x] Tests are updated
- [x] No docs needed

### DIFF
--- a/src/views/Graph.vue
+++ b/src/views/Graph.vue
@@ -678,6 +678,7 @@ export default {
         }
       }
     },
+
     /**
      * Get a nested object of families
      * @property {Family} store - nested object of families
@@ -706,6 +707,7 @@ export default {
       }
       return store
     },
+
     /**
      * Removes a node from an array of nodes
      * @property {String} nodeName - name of node to be removed
@@ -720,6 +722,7 @@ export default {
       })
       return nodesFiltered
     },
+
     /**
      * Creates an edge from the name and cycle of source and target nodes
      * @property {String} edgeType - config option 'noCollapsed'|'collapsedTarget'|'collapsedSource'|'collapsedSourceAndTarget'
@@ -768,6 +771,7 @@ export default {
         },
       }
     },
+
     /**
      * Removes an edge from an array of edges
      * @property {String} sourceName - name of source node of edge to be removed
@@ -785,6 +789,7 @@ export default {
         (edge) => edge.id.indexOf(edgeSearchTerm) === -1
       )
     },
+
     /**
      * Removes an edges from an array of edges based on source node
      * @property {Node[]} edgeCheckSource - array of edges that have been identified as needing removal from the graph
@@ -806,6 +811,7 @@ export default {
       }
       return [edges, removedEdges]
     },
+
     /**
      * Removes an edges from an array of edges based on target node
      * @property {Node[]} edgeCheckTarget - array of edges that have been identified as needing removal from the graph
@@ -827,6 +833,7 @@ export default {
       }
       return [edges, removedEdges]
     },
+
     /**
      * Filters edges array based on source node
      * @property {String} sourceName - name of source node of edge to be removed
@@ -843,6 +850,7 @@ export default {
       })
       return edgeSearch
     },
+
     /**
      * Filters edges array based on target node
      * @property {String} targetName - name of target node of edge to be removed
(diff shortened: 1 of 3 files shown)


src/views/Graph.vue
```diff
@@ -1335,92 +1326,89 @@ export default {
         // ...loop through collapsed families...
         for (const family of this.collapseFamily) {
           // ...get the node from the index...
-          const indexSearch = Object.values(this.cylcTree.$index).find((node) => {
-            return node.name === family && node.tokens.cycle === cycle
-          })
+          const indexSearch = Object.values(this.cylcTree.$index).find(
+            (node) => node.name === family && node.tokens.cycle === cycle
+          )
```
oliver-sanders, line 1263: (unrelated) Walking over the full index is really slow, it can contain nodes from other workflows.

src/views/Graph.vue
```diff
@@ -1335,92 +1326,89 @@ export default {
...
+          if (!indexSearch) continue
```
oliver-sanders, line 1264: Does linting really allow omitting the braces?

src/views/Graph.vue
```diff
@@ -968,26 +904,22 @@ export default {
...
 
     getGraphEdges () {
-      // list graph edges from the store (non reactive list)
+      // map graph edges from the store (non reactive)
       const ret = []
-      for (const workflow of this.workflows) {
-        for (const edge of workflow.$edges || []) {
-          ret.push(edge)
+      for (const { $edges } of this.workflows) {
+        for (const { tokens } of $edges || []) {
```
oliver-sanders, line 918: Maybe a bit to clever, the previous syntax is much more obvious.

src/views/Graph.vue
```diff
@@ -968,26 +904,22 @@ export default {
...
+          ret.push([tokens.id, new Edge(tokens)])
         }
       }
-      return ret
+      return new Map(ret)
```
oliver-sanders, line 922: Why build an array then convert to a map?

src/views/Graph.vue
```diff
@@ -1243,8 +1175,8 @@ export default {
     hashGraph (nodes, edges) {
       // generate a hash for this list of nodes and edges
       return nonCryptoHash(
-        nodes.map(n => n.id).reduce((x, y) => { return x + y }) +
-        (edges || []).map(n => n.id).reduce((x, y) => { return x + y }, 1)
+        nodes.map(n => n.id).reduce((x, y) => x + y) +
+        Array.from(edges.keys()).reduce((x, y) => x + y, 1)
```
oliver-sanders, line 1179: Turning the map into an array might be a bit expensive.

oliver-sanders:
LGTM, but doesn't work right.

I tested with the complex workflow, the graph didn't produce any errors, but it also didn't layout and left all of the tasks in a long line.

When I tried to group by cycle I got this error in the console:

```
TypeError: this.allParentLookUp[l] is undefined
```

src/views/Graph.vue
MetRonnie: Seeing as originally it was doing an `Array.prototype.map()` followed by an `Array.prototype.reduce()`, it shouldn't be any more expensive as there were 2 loops before. However, newer browsers support `edges.keys().reduce()` directly, so we could do something like
```js
const edgesReducer = (x, y) => x + y
// If browser supports Iterator.prototype.reduce(), use it:
edges.keys().reduce?.(edgesReducer) ?? Array.from(edges.keys()).reduce(edgesReducer)
```
to take advantage of it on the browsers that support it?

src/views/Graph.vue
MetRonnie: Yes for control statements on the same line as the condition

MetRonnie:
As discussed in person, I can't reproduce the problem with the same workflow

wxtim:
> I tested with the complex workflow, the graph didn't produce any errors, but it also didn't layout and left all of the tasks in a long line.

I wasn't able to replicate this.

oliver-sanders:
Sorry for the delay.

The error was very persistent, I finally managed to get past it by:
* Nuking `node_modules`.
* Updating `nodejs` from 22.12.0 to 22.13.0.

:sweat: 

Don't understand this, but it's working for me now...

oliver-sanders (approved):
* Code minfied (haven't done too deep a dive into the diff at this stage but makes sense).
* Tested with the "complex workflow", everything seems to be in good order.

Screenshot from 2025-05-13 13-29-16

(1 review is enough for this feature branch)

PR #2142, opened 2025-04-15
Migrate to ESLint flat config & neostandard

MetRonnie:
Supersedes #1747 

The flat config format in ESLint 9 in some ways makes more sense than the old format, but the documentation is rather lacking. Essentially, you've got an array of configuration objects, that ESLint combines.

I have moved from [standard](https://github.com/standard/eslint-config-standard) to [neostandard](https://github.com/neostandard/neostandard), as standard is [apparently not maintained anymore](https://github.com/standard/standard/issues/1948). Unfortunately, neostandard's documentation is also rather lacking. It aims to replace the need for a normal ESLint config, except that only seems to work for very simple use cases. But it can stil be combined with a normal flat ESLint config. Several of the ESLint plugins we use are bundled with neostandard, so I've removed them from our dependencies.

Needs:
- https://github.com/cylc/cylc-ui/pull/2630

**Check List**

- [x] No new tests are needed
- [x] No changelog or docs needed


eslint.config.mjs
```diff
@@ -0,0 +1,123 @@
...
+      'no-console': ['off'],
+    },
+  },
+
+  {
+    files: cypressTestFiles,
+    plugins: {
+      cypress: pluginCypress,
+    },
+    extends: ['cypress/recommended'],
```
MetRonnie, line 94: I found https://github.com/eslint/eslint/issues/19116 more illuminating than anything in the ESLint docs!

MetRonnie:
Got a genuine lint failure which must be being newly picked up by `eslint-plugin-vue` since the update from 9.33.0 to 10.7.0
> error  Unexpected mutation of "data" prop  vue/no-mutating-props

Fixed by https://github.com/cylc/cylc-ui/pull/2630

PR #2154, opened 2025-04-29
Enhance filtering in Analysis tab

ChrisPaulBennett:
In some feedback from users. They would like a more powerful filtering mechanism.
This change allows for multiple terms to be filtered for. The Analysis view will display all jobs that match any of the terms. Terms are separated by a space. A tool tip has been added to the filtering box to help inform users this can be done
Unit tests do not need to be updated

**Check List**

- [x ] Contains logically grouped changes (else tidy your branch by rebase).

### DIFF
--- a/src/components/cylc/analysis/filter.js
+++ b/src/components/cylc/analysis/filter.js
@@ -26,9 +26,20 @@
  * @return {boolean} Boolean determining if task should be displayed
  */
 export function matchTask (task, tasksFilter) {
-  let ret = true
-  if (tasksFilter.name?.trim()) {
-    ret &&= task.name.includes(tasksFilter.name)
+  // Split the filter into individual words and conduct a search for each one
+  const filter = tasksFilter.name.split(/(\s+)/).filter(
+    function (e) { return e.trim().length > 0 }
+  )
+  let ret = false
+  // If the filter is empty, return true
+  if (!filter.length) {
+    ret = true
+  } else {
+    for (const element of filter) {
+      if (task.name.includes(element)) {
+        ret = true
+      }
+    }
   }
   if (tasksFilter.platformOption.trim?.()) {
     ret &&= task.platform === tasksFilter.platformOption
--- a/src/views/Analysis.vue
+++ b/src/views/Analysis.vue
@@ -28,14 +28,20 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           md="4"
           class="pr-md-2 mb-2 mb-md-0"
         >
-          <v-text-field
-            id="c-analysis-filter-task-name"
-            clearable
-            placeholder="Filter by task name"
-            v-model.trim="tasksFilter.name"
-            ref="filterNameInput"
-            :disabled="chartType === 'timeSeries'"
-          />
+          <v-tooltip>
+            <template #activator="{ props }">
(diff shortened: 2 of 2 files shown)


src/views/Analysis.vue
```diff
@@ -28,14 +28,20 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+            <template #activator="{ props }">
+              <v-text-field
+                id="c-analysis-filter-task-name"
+                clearable
+                placeholder="Filter by task name"
+                v-model.trim="tasksFilter.name"
+                ref="filterNameInput"
+                :disabled="chartType === 'timeSeries'"
+                v-bind="props"
+              />
```
oliver-sanders, line 41: We should be able to make the space separated bit a little more intuitive.

Check out the Vuetify Combobox component, it has a [multiple option](
https://vuetifyjs.com/en/components/combobox/#multiple-combobox) that allows multiple values to be specified (note you don't *have* to pre-populate suggestions to a combox). It can place each item in a "chip" and allow each item [to be individually clearable](https://vuetifyjs.com/en/components/combobox/#advanced-custom-options). It should also take care of the `/\s/` stuff.

ChrisPaulBennett: Done

MetRonnie:
Want me to review this?

src/components/cylc/analysis/filter.js
```diff
@@ -26,9 +26,16 @@
...
-  let ret = true
-  if (tasksFilter.name?.trim()) {
-    ret &&= task.name.includes(tasksFilter.name)
+  let ret = false
+  if (tasksFilter.name.length > 0) {
+    for (let i = 0; i < tasksFilter.name.length; i++) {
+      if (tasksFilter.name[i] === task.name) {
+        ret = true
+      }
+    }
```
MetRonnie, line 35: This ought to be a short-cutting loop.
```suggestion
  if (tasksFilter.name?.length) {
    ret = tasksFilter.name.some((name) => task.name === name)
```
`Array.prototype.some()` is like Python's `any()`

tests/e2e/specs/analysis.cy.js
```diff
@@ -61,13 +61,16 @@ describe('Analysis view', () => {
...
         cy
           .get('td')
           .contains('waiting')
           .should('be.visible')
+        cy
+          .get('#c-analysis-filter-task-name')
+          .click()
+          .get('.v-list-item')
+          .contains('waiting')
+          .click({ force: true })
```
MetRonnie, line 73: This has come underneath the check that `waiting` is visible in the table. It ought to be before it

src/components/cylc/analysis/filter.js
ChrisPaulBennett: I'm an old man that doesn't understand these new fangled loop styles

tests/e2e/specs/analysis.cy.js
ChrisPaulBennett: Done

src/components/cylc/analysis/filter.js
MetRonnie: In English think of it like `if there is some name in the tasksFilter.name array, that is equal to task.name`

PR #2163, opened 2025-05-08
Fix teleport to wrong instance when multiple analysis views open

MetRonnie:
Fixes a bug not recorded in an issue.

#### Steps to reproduce

Open 2 analysis view tabs in the same workspace, then switch to the box plot or time series mode in the second one. You would find the sorting control does not appear here, but if you go back to the first analysis view tab it will be there instead.

image

#### Check List


src/views/Analysis.vue
```diff
@@ -268,8 +267,7 @@ export default {
      */
     const chartType = useInitialOptions('chartType', { props, emit }, 'table')
 
-    /** @type {import('vue').Ref<HTMLElement>} template ref */
-    const toolbar = ref(null)
+    const toolbarRef = useTemplateRef('toolbar')
```
MetRonnie, line 271: This is somewhat tangential, but is a newer (clearer IMO) way of defining template refs in Vue 3.5

oliver-sanders (approved):
LGTM, but haven't tested.

wxtim (approved):

PR #2171, opened 2025-05-13
workflows table: add cylc version column

oliver-sanders:
* Add a version column to the workflow table (makes it easier for users to find workflows running under older versions as a stopgap measure).
* Rename the "workflows table" to "workflow table" (note the URL was already "workflow-table" so this is a non-change).
* Fix a problem where the "workflow-table" was not registering its subscription properly so was relying on the "workflows view" subscription for the data it needed.
* Removed fields from the "workflows view" subscription that are not needed for the workflows view.

**Check List**

- [x] Changelog entry included if this is a change that can affect users - v-minor

### DIFF
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -65,7 +65,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               Workflows Table
             </v-list-item-title>
             <v-list-item-subtitle>
-              View name, host, port, etc. of your workflows
+              View name, host, version, etc. of your workflows
             </v-list-item-subtitle>
           </v-list-item>
           <v-list-item to="/user-profile" data-cy="user-settings-link">
--- a/src/views/Workflows.vue
+++ b/src/views/Workflows.vue
@@ -60,9 +60,6 @@ fragment WorkflowData on Workflow {
   id
   status
   statusMsg
-  owner
-  host
-  port
   stateTotals
   latestStateTasks(states: [
     "failed",
--- a/src/views/WorkflowsTable.vue
+++ b/src/views/WorkflowsTable.vue
@@ -114,7 +114,7 @@ fragment WorkflowData on Workflow {
 `
 
 export default {
-  name: 'WorkflowTable',
+  name: 'WorkflowsTable',
 
   mixins: [
     subscriptionComponentMixin,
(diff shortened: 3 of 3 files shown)


src/views/WorkflowTable.vue
```diff
@@ -103,17 +106,18 @@ subscription Workflow {
...
   port
 }
 `
 
 export default {
-  name: 'WorkflowsTable',
+  name: 'WorkflowTable',
 
   mixins: [
-    subscriptionMixin
```
oliver-sanders, line 116: This seems to have been the wrong mixin. The workflows view was only working because the data it needed happened to have been requested (in error) by the workflows view.

src/views/Workflows.vue
```diff
@@ -59,10 +59,6 @@ fragment WorkflowData on Workflow {
   # (it would cause a race condition with the workflow subscription)
   id
   status
-  statusMsg
-  owner
-  host
-  port
```
oliver-sanders, line 65: These fields do not appear to be needed by the workflows view.

oliver-sanders: Aah, good spot, will add it back.

src/lang/en-GB/Workflows.json
```diff
@@ -1,5 +1,5 @@
 {
-  "tableHeader": "Workflows Table",
```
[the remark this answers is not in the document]
oliver-sanders: It doesn't to me, but I can always revert.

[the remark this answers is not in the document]
hjoliver: Actually, having "Table" in the heading is pointless, and "Workflows" (plural) makes sense by itself.  Can we just drop "Table"?

src/views/Dashboard.vue
```diff
@@ -62,10 +62,10 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               <v-icon size="1.6em">{{ $options.icons.table }}</v-icon>
             </template>
             <v-list-item-title class="text-h6 font-weight-light">
-              Workflows Table
+              Workflow Table
```
hjoliver, line 65: The dashboard could use a better link name here, to indicate what distinguishes this "Workflow(s) Table" from the workflows listed in the side bar.  It's basically workflow contact information: owner, host, post (it should have PID too) plus status (running, stopped, paused). What can we call this?  *Workflow Status*, or *Workflow Details*, or *Workflow Locations*, ...?

oliver-sanders: > to indicate what distinguishes this "Workflow(s) Table" from the workflows listed in the side bar

The thing that distinguishes it, is being a table!

oliver-sanders: I'll kill this diff.

MetRonnie (approved):
Good

PR #2172, opened 2025-05-13
Graph grouping/collapsing: remove need for descendants field in query

MetRonnie:
Follow-up to https://github.com/cylc/cylc-ui/pull/1810
Partly addresses https://github.com/cylc/cylc-ui/issues/1130

It turns out we don't need the `descendants` field added in https://github.com/cylc/cylc-flow/pull/6610, and can actually simplify the `getTree` function

**Check List**

- [x] Tests are not needed
- [x] No docs needed

### DIFF
--- a/src/views/Graph.vue
+++ b/src/views/Graph.vue
@@ -214,7 +214,7 @@ fragment FamilyProxyData on FamilyProxy {
   isQueued
   name
   ancestors {
-     name
+    name
   }
   childTasks {
     id
@@ -335,6 +335,19 @@ export const childArray = (nodeArray) => {
         .concat(children ? childArray(children) : [])
     }
   })
+}
+
+/**
+ * Convert a mapping to a format that VTreeView can handle.
+ * @param {Map<string, { children: Map, diabled: Boolean }>} map - The mapping to convert
+ * @returns {Object[]} - The converted mapping
+ */
+export function convertTree (map) {
+  return Array.from(map, ([name, { children, disabled }]) => ({
+    name,
+    children: convertTree(children),
+    disabled,
+  }))
 }
 
 export default {
@@ -484,7 +497,7 @@ export default {
      * @returns {Family[]} array containing nested structure of families
      */
     treeDropDownFamily () {
-      return this.familyArrayStore.length ? this.getTree() : [{ name: 'No families', disabled: true }]
+      return this.allParentLookUp.size ? this.getTree() : [{ name: 'No families', disabled: true }]
     },
     /**
      * Gets the array of cycles for use in vuetify toolbar drop down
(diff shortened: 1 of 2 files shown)


oliver-sanders:
Found one issue, I haven't checked to see if this is a new issue on this PR though.

Workflow:

```
#!Jinja2
#

[scheduler]
    UTC mode = True
    allow implicit tasks = True

[scheduling]
    initial cycle point = 2019-12-09T09:00Z
    [[graph]]
        R1 = foo_cold => foo_start
        R1/T00 = foo_start[^] => FOO
        T00, T12 = """
            cycle_end[-PT12H] => FOO
            FOO:succeed-all => BAR
            BAR:succeed-any => wipe_bar
            BAR:succeed-all & wipe_bar => cycle_end
        """

[runtime]
    [[root]]
        script = sleep {{ SLEEP | default('1') }}
    [[FOO]]
    [[BAR]]
    [[foo, foot, fool]]
        inherit = FOO
    [[bar, bool]]
        inherit = BAR
    [[PUB]]
        inherit = BAR
    [[pub]]
        inherit = PUB
```

To reproduce, collapse PUB, then collapse BAR. The edges disappear with the following traceback:

```
TypeError: f is undefined
    layout Graph.vue:1493
    refresh Graph.vue:1415
    updateTimer Graph.vue:854
    mounted Graph.vue:465
    promise callback*it runtime-core.esm-bundler.js:289
    mounted Graph.vue:463
```

src/views/Graph.vue
```diff
@@ -796,33 +775,18 @@ export default {
...
-            return this.allParentLookUp[nodeFirstParent][i]
-          }
-        }
-      } else {
-        // the node is not collapsed
-        return null
+    getCollapsedAncestor (id) {
+      const { name } = this.cylcTree.$index[id].node.firstParent
+      if (name !== 'root') {
+        // TODO: could use Set.intersection if browser support is good enough?
```
oliver-sanders, line 786: IT looks reasonable: https://caniuse.com/?search=set.intersection

src/views/Graph.vue
```diff
@@ -796,33 +775,18 @@ export default {
...
+        return this.collapseFamily.find((family) => (
+          name === family || this.allParentLookUp.get(name).includes(family)
+        ))
```
oliver-sanders, line 789: A non-iterative approach would be very worthwhile here as this gets called a lot as part of graphviz document building.

src/views/Graph.vue
MetRonnie: Unfortunately I think the ESLint plugin we use for checking compat has issues: https://github.com/amilajack/eslint-plugin-compat/issues/636

oliver-sanders:
Confirmed the previous issue has been resolved.

I have found another case where a graph edge gets missed based as the result of family collapsing (sorry). It's a complex example, but I've saved the DB for reproduction.

Screenshot from 2025-06-11 16-12-01
Screenshot from 2025-06-11 16-12-19

The ENSEMBLE family is the one that triggers the behaviour. The workflow config and DB are in ~/cylc-ran/complex.

I'm happy if you want to merge this one and tackle later.

MetRonnie:
OK, I'll merge but leave as saved in my inbox so I come back to this

PR #2176, opened 2025-05-16
Save layout

### LINKED ISSUE #1837: lumino: preserve layouts in LocalStorage
Currently the UI will remember tab layouts within a session but not between them.

Using LocalStorage would allow tab layouts to be preserved between sessions. This would be especially relevant for saving the default layout (#1834) and restarts of the UIS.

Preserving layouts presents the additional complexity of working out how to behave when multiple sessions are present.

See #662

MetRonnie:
Closes #1837

**Check List**

### DIFF
--- a/src/components/cylc/workspace/Lumino.vue
+++ b/src/components/cylc/workspace/Lumino.vue
@@ -39,7 +39,7 @@ import {
   onMounted,
   ref,
 } from 'vue'
-import { useStore } from 'vuex'
+import { useLocalStorage } from '@vueuse/core'
 import { startCase, uniqueId } from 'lodash'
 import WidgetComponent from '@/components/cylc/workspace/Widget.vue'
 import LuminoWidget from '@/components/cylc/workspace/lumino-widget'
@@ -66,8 +66,6 @@ import '@lumino/default-theme/style'
  * @property {string} name - the view to add
  * @property {Record<string,*>} initialOptions - prop passed to the view component
  */
-
-const $store = useStore()
 
 const props = defineProps({
   workflowName: {
@@ -103,6 +101,29 @@ const mainDiv = ref(null)
 const views = ref(new Map())
 
 const defaultView = useDefaultView()
+
+const layoutStorage = useLocalStorage(
+  'workspaceLayout',
+  new Map(),
+  {
+    serializer: {
+      read: (str) => JSON.parse(str, (key, value) => {
+        if (value?.__type === 'Map') {
+          return new Map(value.arr)
+        }
+        if (key === 'widgets') {
+          return value.map((w) => new LuminoWidget(w.id, w.name, w.closable))
+        }
+        return value
+      }),
+      write: (obj) => JSON.stringify(obj, (key, value) => (
+        value instanceof Map
+          ? { __type: 'Map', arr: Array.from(value) }
+          : value
+      )),
+    },
+  }
+)
 
 // create a box panel, which holds the dock panel, and controls its layout
 const boxPanel = new BoxPanel({ direction: 'left-to-right', spacing: 0 })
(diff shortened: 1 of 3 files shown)


src/components/cylc/workspace/Lumino.vue
```diff
@@ -138,85 +151,105 @@ onBeforeUnmount(() => {
...
+    views: views.value,
+  }, replacer)
+  const cache = await layoutsCache
+  // Overrides on FIFO basis:
+  await cache.put(props.workflowName, new Response(
+    serializedLayout,
+    { headers: { 'Content-Type': 'application/json' } }
+  ))
+  const keys = await cache.keys()
+  if (keys.length > 100) {
```
wxtim, line 206: Does this create an in-practice limit of 100 workflows stored.

tests/e2e/specs/workspace.cy.js
wxtim, line 1: General tangential thought on this test: The `BeforeEach` is somewhat slow - do we really need it, or could we modify the same instance for all the tests?

tests/e2e/specs/workspace.cy.js
```diff
@@ -162,7 +166,34 @@ describe('Workspace view and component/widget', () => {
...
+        // Test page refresh
+        cy.reload()
+        expectRememberedLayout()
+      } else {
+        cy.log('Cache API not supported; skipping page refresh test.')
+      }
+    })
+  })
+
+  it('Resets layout', () => {
```
wxtim, line 182: Could you add this to another test to save the setup time?

wxtim (approved):
Nice.

Something to watch for: I did briefly manage to produce a horrid effect where content ended up in the wrong tab, but I've not been able to reproduce without a debug statement in the code.

I've made some comments about the e2e tests, but I don't think they should block this PR.

- [x] Read the code
- [x] Tried the code
  - [x] With lots of workflows
  - [x] With lots of windows
  - [x] With different options in different tabs
  - [x] With `debugger` statements in the code so I could see what the stored JSON looked like.
  - [x] Ran the e2e test locally and confirmed that the behaviour looks sensible.

tests/e2e/specs/workspace.cy.js
MetRonnie: > The `BeforeEach` is somewhat slow

Is it?

tests/e2e/specs/workspace.cy.js
MetRonnie: There are downsides to overloading tests, as if something changes in the middle it might lead to invalidating the rest of the test, and failures can be more difficult to isolate

tests/e2e/specs/workspace.cy.js
oliver-sanders: I expect the test overheads are high rather than BeforeEach being slow.

I don't think the test runtime is at the point where we need to start gaming things yet.

src/components/cylc/workspace/Lumino.vue
```diff
@@ -114,21 +113,35 @@ const resizeObserver = new ResizeObserver(() => {
   boxPanel.update()
 })
 
-onMounted(() => {
+const layoutsCache = window.caches.open('workspace-layouts')
```
oliver-sanders: I think caches are tied to the domain. Because of Jupyter Server and Jupyter Hub services, there can be multiple web apps served from the same domain.

We should probably namespace our caches to avoid collisions, e.g:

```suggestion
const layoutsCache = window.caches.open('cylc/workspace-layouts')
```

The MDN docs for caches use a version identifier (i.e. `v1`) which might be a good idea (e.g. `cylc/workspace-layouts/v1`), but we could always add the version number later (i.e. for `v2`) if we need it.

src/components/cylc/workspace/Lumino.vue
```diff
@@ -140,71 +153,104 @@ onBeforeUnmount(() => {
...
 
+/**
+ * Reset the workspace layout to a single tab with the default view.
+ */
+async function resetToDefault () {
+  layoutWatcher.pause()
+  await closeAllViews()
+  await addView({ name: defaultView.value })
+  // (addView resumes the layout watcher)
+}
```
oliver-sanders: :+1:

For future work, it would be great to allow the user to replace the "default view" with a "default layout" so we can allow users to save the current layout as their default view.

src/components/cylc/workspace/luminoWidget.js
```diff
@@ -87,4 +87,25 @@ export default class LuminoWidget extends Widget {
...
+  toJSON () {
+    // Allow the widget to be serialized.
+    // We only need to store this limited info when saving the layout,
+    // allowing us to entirely recreate the widget when restoring the layout.
+    return {
+      name: this.name,
+      id: this.id,
+      closable: this.closable,
+    }
+  }
```
oliver-sanders, line 100: Wow, that's all there was to it!

src/utils/json.js
```diff
@@ -0,0 +1,46 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// JSON utilities.
+
+/**
+ * Custom replacer function for JSON.stringify to handle JS objects that
+ * cannot otherwise be serialized.
+ */
```
oliver-sanders, line 23: Consider adding an example of this for noobs (me):

```suggestion
/**
 * Custom replacer function for JSON.stringify to handle JS objects that
 * cannot otherwise be serialized.
 *
 * > const json = JSON.stringify(obj, replacer)
 * > const obj = JSON.parse(json, reviver)
 */
```

src/utils/json.js
```diff
@@ -0,0 +1,46 @@
...
+export function replacer (key, value) {
+  if (value instanceof Map) {
+    return { _jsonType: 'Map', _val: Array.from(value) }
```
oliver-sanders, line 26: Out of interest, is there any convention / library for doing this?

Seems like something that would be useful, e.g. for Python <-> JS stuff.

src/components/cylc/workspace/Lumino.vue
oliver-sanders: Yes (as suggested) and yes (understood).

oliver-sanders: @MetRonnie, I still think it's a good idea to namespace our caches right?

oliver-sanders (approved):

wxtim (approved):

PR #2204, opened 2025-06-16
Workspace layouts: avoid duplicate layout saving when resetting to default

MetRonnie:
Follow up to #2176 

I spotted that clicking the "reset layout" button caused two layout saves to occur instead of just one when it has finished resetting the layout.

**Check List**

- [x] Changelog entry not needed as invisible to users
- [x] Docs not needed


oliver-sanders (approved):
:1+:

Issue #1225, opened 2023-02-06
aotf: load the schema for each workflow


oliver-sanders:
See also https://github.com/cylc/cylc-uiserver/issues/419

Presently the UI loads the UIS schema from which it gets the list of mutations.

This means that if the UIS is talking to two schedulers with different mutations (e.g. different Cylc versions), then mutations issued to one of the two will be drawn up with the wrong schema and will fail if the interfaces differ.

Rather than loading the schema once per-app, we should load it once-per-workflow.

Note: Since the AOTF work was put in, we have come to rely on introspection for the edit-runtime forms too, the same problem applies here also.

**Pull requests welcome!**

dwsutherland:
Since the UI only interacts with the UIS, how does it make sense to load a scheduler schema that might be incompatible with the UIS?

oliver-sanders:
Providing that the UIS passes the mutations through to the scheduler without meddling with them this should work fine.

* The scheduler declares its capabilities.
* The UI requests actions based on those capabilities.
* The UIS acts as a forwarding layer passing these requests on to the scheduler.

The design and motive are documented here - https://github.com/cylc/cylc-flow/issues/2123

tldr; We keep fiddling the workflow interfaces for valid reasons, adding mutations, options, etc. But our users have valid reasons for working with different versions of Cylc simultaneously (e.g. research & production workloads). This has caused a number of problems with GUIs over the years and has made user-facing documentation tricky.

PR #2230, opened 2025-06-30
AOTF: Do not send non-required mutation args when they have the default value

MetRonnie:
As mentioned in https://github.com/cylc/cylc-flow/pull/6820, the reload command in the GUI currently will be broken for workflows running in Cylc < 8.5.0.

This is because https://github.com/cylc/cylc-flow/pull/6509 added a new argument in the schema. However we can work around this by not sending that arg when it is the default value.

This workaround can be removed when #1225 is addressed.

**Check List**

- [x] Changelog entry not needed as the problem will not be released until 8.5.0
- [x] No docs needed

### DIFF
--- a/src/utils/aotf.js
+++ b/src/utils/aotf.js
@@ -762,10 +762,11 @@ export function argumentSignature (arg) {
 /** Construct a mutation string from a mutation introspection.
  *
  * @param {Mutation} mutation - A mutation as returned by an introspection query.
+ * @param {Record<string,any>} variables
  *
  * @returns {string} A mutation string for a client to send to the server.
  */
-export function constructMutation (mutation) {
+export function constructMutation (mutation, variables) {
   // the scan mutation has no arguments
   if (!mutation.args.length) {
     return dedent`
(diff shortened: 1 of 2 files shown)


src/utils/aotf.js
```diff
@@ -517,20 +517,12 @@ export function getMutationExtendedDesc (text) {
...
-    cylcObject = null
-    cylcType = null
-    if (pointer?.kind === 'NON_NULL') {
-      required = true
-    }
+    let pointer = arg.type
+    let multiple = false
+    let cylcObject = null
+    let cylcType = null
+    const required = arg.type?.kind === 'NON_NULL'
```
oliver-sanders, line 525: Jslint used to tell us not to declare variables within loops like this. More performant to declare once and reuse?

MetRonnie: I can't find much info on that and perhaps that dated back to when `var` was used to declare variables?

I prefer this way because a) it reduces duplication and makes the code more readable, b) it is less likely to lead to the value from one iteration accidentally leaking into the next iteration - the scope is only within a single iteration.

oliver-sanders: > dated back to when var was used to declare variables?

That's probably it. Using `var` within a loop would re-declare the variable over and over.

MetRonnie: Actually the opposite I think `var` declarations were hoisted 😬

Issue #2180, opened 2025-05-23
task icon: add xtrigger derived modifiers


oliver-sanders:
Task modifiers tell us that a task is waiting on some external (non dependency) based condition.

### Modifiers

Currently this covers:
* Queued
* Held
* Runahead

We [can now extend this to cover](https://github.com/cylc/cylc-flow/pull/6671) to reflect:

* Clock triggers
* Retry triggers
* Generic xtriggers

To complete this, we would also need to add:

* Ext triggers

Ext-triggers are a push-based alternative to pull-based xtriggers. We may be able to roll these in with xtriggers, at least for the purpose of display. These are seldom used and a topic for another day.

### Proposal

The presence of the modifier is the most important piece of information, however, we use a different modifier for each type of external dependency to convey extra information.

We need three more modifier icons to cover the three new xtrigger derived modifier states.

Icons should be:

* Graphically simple - they are small and need to be recognisable.
* Vectorally simple - the fewer SVG elements the better (all modifiers in the below image use three elements or fewer).
* Intuitive - obvs!

Here is my initial proposal.

![Image](https://github.com/user-attachments/assets/26dac31c-9a44-46d9-b4b1-e78971e40e65)

### Related

We may go beyond the task icon modifier on a view-by-view basis in the future, however, that is orthogonal to this issue:

* [Graph view](https://github.com/cylc/cylc-ui/issues/131)
* [Tree view](https://github.com/cylc/cylc-ui/issues/331)

Additionally, we may add the ability to filter tasks by modifier (dependent on this issue):

* https://github.com/cylc/cylc-ui/issues/1666

wxtim:
> I think that looks OK [@wxtim](https://github.com/wxtim)

But I couldn't put the arrowhead in the retry and make it look good

oliver-sanders:
I'll have a go at redrawing the retry and clock icons to make them clearer in smaller sizes.

wxtim:
Thanks @oliver-sanders  - That looks much more sane.

PR #2254, opened 2025-07-23
Display xtrigger task state modifiers in command menu

MetRonnie:
Follow-up to #2180 

**Check List**

- [x] Changelog entry not needed as minor
- [x] Docs not needed

### DIFF
--- a/src/views/Tree.vue
+++ b/src/views/Tree.vue
@@ -187,9 +187,9 @@ fragment TaskProxyData on TaskProxy {
   isHeld
   isQueued
   isRunahead
-  isXtriggered
   isRetry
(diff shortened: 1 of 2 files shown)


src/views/Tree.vue
```diff
@@ -187,9 +187,9 @@ fragment TaskProxyData on TaskProxy {
   isHeld
   isQueued
   isRunahead
-  isXtriggered
   isRetry
   isWallclock
+  isXtriggered
```
MetRonnie, line 192: (I just did this so that the order was consistent with the other views)

wxtim (approved):
Thank you!

PR #2339, opened 2025-10-14
workflows: omit stopped workflows from cumulative state totals

### LINKED ISSUE #1930: gscan: omit stopped workflows from cumulative state totals
The cumulative state totals include the states of all workflows in the hierarchy.

The Cylc UI doesn't list tasks for stopped workflows, unless the workflow stopped whilst the UI Server was connected to it, in which case it preserves the last known state of the workflow before it shut down.

This can lead to the slightly confusing situation where a stopped workflow is reported as having active tasks. These tasks contribute the gscan state totals (even if stopped workflows are filtered out) which isn't great.

oliver-sanders:
* Closes #1930
* Have purposefully left the "latest state tasks" logic unchanged as we are planning to remove this imminently.

It doesn't look like there are any tests for this logic at present. I haven't added any in this PR as yet.

There's ongoing refactor work here, so might be something to pick up on the other side?

**Check List**

### DIFF
--- a/changes.d/2339.feat.md
+++ b/changes.d/2339.feat.md
@@ -1,0 +1,1 @@
+In the sidebar, don't include task states for stopped workflows in any parent state totals.


wxtim (approved):

src/components/cylc/tree/GScanTreeItem.vue
```diff
@@ -109,20 +109,26 @@ import { useWorkflowWarnings } from '@/composables/localStorage'
  * @param {Object} node
  * @param {Record<string, number>} stateTotals
  * @param {Record<string, string[]>} latestTasks
+ * @param {Boolean} topLevel - true if the traversal depth is 0, else false.
  */
-function traverseChildren (node, stateTotals = {}, latestTasks = {}) {
+function traverseChildren (node, stateTotals = {}, latestTasks = {}, topLevel = true) {
```
MetRonnie: I can't see how `topLevel` is needed

MetRonnie:
Will add tests in #2331 as it will be easier with the numbers included

Issue #1666, opened 2024-02-08
filtering: task modifier filtering


oliver-sanders:
We can filter by task states, but not task state modifiers.

It might be useful in some cases to filter for modifiers, main use cases:

* Filtering out runahead tasks (these are superfluous for most use cases and were filtered out by default at Cylc 7).
* Filtering for queued tasks (e.g. I want to see all "active" tasks in a workflow i.e. waiting(queued), preparing, submitted, running where this definition of "active" differs subtly from the n=0 definition which also contains tasks with partially satisfied prereqs).

See also the [queue view](https://github.com/cylc/cylc-ui/issues/462).

Issue #1264, opened 2023-03-29
views: filter tasks by pattern


oliver-sanders:
We currently allow filtering tasks by name e.g. in the tree view.

### Problem

There is a good use case for filtering by pattern too.

### Proposed Solution

Two options:

* Regex
* [ext]glob

Note that extglobs are currently used for task filtering by Cylc Review and are also used by the CLI and for mutations. So using extglobs for task filtering would assist in issuing multi-workflow/multi-task mutations.

### Question

1. Regex?
2. [ext]Glob?
3. Both?

hjoliver:
Globbing is probably sufficient for this, and is a lot more intuitive to most users.

ColemanTom:
Does extglob support things like `a|b` in case you want to be just watching tasks `a` and `b`?

oliver-sanders:
Yes, as `(a|b)`

oliver-sanders:
Space separated patterns were suggested.

This might work nicely, we could use "chips" to separate the terms for clarity.

Issue #991, opened 2022-04-08
Make filter controls "sticky"


MetRonnie:
**Describe exactly what you would like to see in an upcoming release**
Pin the filter controls to the top of the lumino tree & table views so they don't scroll out of view.

From Hilary on Element:
>Anyone noticed this cylc-ui regression?
>
>We merged this PR late January: https://github.com/cylc/cylc-ui/pull/861 
>
>As I recall it pinned the filter view toolbars to the top of the lumino tabs, for table and tree view - i.e. independent of view scrolling.
>
>However, since then (and possibly for rc1 which was released not long after) the tool bars have become unpinned again.

>Looks like the regression occurred on the PR itself.  I just built cylc-ui on the first commit of that branch, and the filter toolbar was pinned to the top of the table view at that point.

**Pull requests welcome!**

oliver-sanders:
See also #471

PR #2370, opened 2025-11-26
view toolbar: add search & filter functionality

oliver-sanders:
This adds task modifier filtering to the Tree & Table views as well as a bunch of related things:

* centralise and improve task search and filter controls:
    * Add functionality to the ViewToolbar component:
      * Add `action=menu` for dropdowns.
      * Add `action=input` for text input.
      * Allow `icon=Object` to configure state-dependent icons.
      * Reduce icon spacing and improve group divider.
    * Move the task search & filter controls into the ViewToolbar component.
      * Switch the Tree and Table views over to the ViewToolbar.
      * Addresses https://github.com/cylc/cylc-ui/issues/471
    * Improve the task state filter control:
      * Add support for task modifiers - closes https://github.com/cylc/cylc-ui/issues/1666
      * Collapse the control into a single button/menu (take up less space).
      * Reduce the width of the task search input, automatically increase
        this width when focused or when text is present in it.
* tree: make the toolbar sticky
    * Ensure the ViewToolbar doesn't scroll out of view.
    * Closes https://github.com/cylc/cylc-ui/issues/991
* view toolbar: support globs in task ID search
    * Closes https://github.com/cylc/cylc-ui/issues/1264
* vite: ignore nfs files
    * These can cause vite to crash.
* update caniuse-lite

**Check List**

### DIFF
--- a/tests/e2e/specs/tree.cy.js
+++ b/tests/e2e/specs/tree.cy.js
@@ -184,12 +184,12 @@ describe('Tree view', () => {
           .should('not.be.visible')
       }
       // It should stop filtering when input is cleared
-      cy.get('.c-view-toolbar input')
+      cy.get('[data-cy=control-taskIDFilter] input')
         .clear()
         .get('.node-data-task:visible')
         .should('have.length', initialNumTasks)
       // It should filter by cycle point
-      cy.get('.c-view-toolbar input')
+      cy.get('[data-cy=control-taskIDFilter] input')
         .type('2000') // (matches all tasks)
         .get('.node-data-task:visible')
         .should('have.length', initialNumTasks)


src/components/cylc/common/filter.js
```diff
@@ -36,8 +43,25 @@ export function matchID (node, id) {
  * @param {?string[]} states
  * @returns {boolean}
  */
-export function matchState (node, states) {
-  return !states?.length || states.includes(node.node.state)
+export function matchState (
```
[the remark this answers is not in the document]
wxtim: Fixed. :)

src/components/cylc/ViewToolbar.vue
MetRonnie: (Discussion to possibly take into another issue, not hold up this PR)

IMO this is becoming increasingly difficult to use. This single component controls the presentation and business logic of several input types, making it difficult to reason about how they are working.

In future, I think it would be worth extracting out the input types that are used here into their own SFCs, leaving this as a component that controls presentation and vuetify defaults, with a slot for allowing each view to set its controls (the more conventional way of using Vue components).

While I understand the idea behind this existing implementation (https://github.com/cylc/cylc-ui/issues/1809#issuecomment-2126935245), the main advantage being having an object that contains the configuration that can be re-used in the settings page, this can still be achieved under what I'm proposing.

tests/e2e/specs/table.cy.js
```diff
@@ -182,18 +180,17 @@ describe('State saving', () => {
...
     cy
-      .get('[data-cy="filter task state"]:last')
+      .get('[data-cy=control-taskStateFilter]:last')
       .click()
     cy
-      .get('.v-list-item')
-      .contains(TaskState.SUCCEEDED.name)
-      .click({ force: true })
+      .get('.v-list-item[state=succeeded]')
+      .click({ force: true, multiple: true })
```
MetRonnie: This suggests the menu for the tree view tab is also being rendered (although probably has `display: none`). Normally vuetify does not render a menu until the first time it is opened (which was the case before this PR).

src/components/cylc/ViewToolbar.vue
```diff
@@ -30,24 +30,75 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-        <v-btn
-          @click="iControl.callback"
-          v-bind="btnProps"
-          :disabled="iControl.disabled"
-          :aria-checked="iControl.value"
-          :color="iControl.value ? 'blue' : undefined"
-          role="switch"
+        <!-- menu component (to support dropdowns) -->
+        <v-menu
+          eager
```
wxtim: Is this a speed thing (moving delay from onClick to page load)? The API docs suggest SEO use cases?

wxtim (approved):
Legit - tested as working. Code looks good. :)

src/components/cylc/common/filter.js
```diff
@@ -17,15 +17,55 @@
...
+    RegExp.escape(' ' + glob.trim())
+      .substr(4)
+      // `*` -> `.*`
+      .replace(/\\\*/, '.*')
+      // `?` -> `.`
+      .replace(/\\\?/, '.')
+      // `[X]` -> `[X]`
+      .replace(/\\\[\\x21([^]*)\\\]/, '[^$1]')
+      // `[^X]` -> `[^X]`
+      .replace(/\\\[([^]*)\\\]/, '[$1]')
```
MetRonnie: Backwards?
```suggestion
      .replace(/\\\[([^]*)\\\]/, '[$1]')
      // `[!X]` -> `[^X]`
      .replace(/\\\[\\x21([^]*)\\\]/, '[^$1]')
```

oliver-sanders:
> Suggestion for simplifying some stuff at https://github.com/oliver-sanders/cylc-ui/pull/122

Suggestion breaks some things, took the table tidy part and rolled it in with above suggestions:

https://github.com/cylc/cylc-ui/compare/ddd6bc3d4038b25babd19d9bc42ddf611a067f1e..90841e339423bfcc36c3d09942bf49a54b8808ac

<br />

> How are users going to know that you can use globs?

For now, Discourse.

I was looking to extend this to cover Regex support and add a little glob/regex selector chip on the right of the search bar.

I didn't get onto this, largely as it would have required adding a new field (ID filter type) into the view defaults (and upgrading the undefined value).

I'll try and look into the chip for this PR if able to (thinking something like the "Markdown Supported" thinggy).


<br />

> Glob matching is not working in a weird way for the table view:

:eyes:

src/components/cylc/common/filter.js
```diff
@@ -17,15 +17,55 @@
...
+  if (!glob || !glob.trim()) {
+    // no glob provided
+    return null
+  }
+  return new RegExp(
+    // escape any regex characters in the glob
+    // NOTE: the first character is escaped using the `\x` syntax, so we
+    // prefix a space then subtract the first four characters (`\x20`)
+    // from the result
+    RegExp.escape(' ' + glob.trim())
```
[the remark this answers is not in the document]
MetRonnie: (We have not been using Babel since the move to Vite; it uses esbuild instead)

Note Vite does transform some newer ECMAScript _syntax_ (stuff like operators, e.g. [`&&=`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND_assignment) was introduced relatively recently and might be transformed into a more compatible syntax if you set an old enough build target). But it can't transform _APIs_ (e.g. [`ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) or indeed `RegExp.escape`)

Ah, after typing that all out I realise it's stated here lol: https://esbuild.github.io/api/#target

N.B. Also the next major version of Vite will be moving away from esbuild to their own (feature-equivalent) tool: https://vite.dev/blog/announcing-vite8-beta

wxtim (approved):
Reviewed changes since last review & happy

tests/component/specs/viewToolbar.cy.js
```diff
@@ -224,4 +224,93 @@ describe('View Toolbar Component', () => {
...
+
+      // implementation detail: the setOption event is fired again
+      // but the waiting state is not removed
+      .get('@wrapper').then(({ wrapper }) => {
+        expect(
+          wrapper.emitted().setOption.at(-1)[1]
+        ).to.deep.equal(['isRetry', 'waiting'])
+      })
+
+      // the waiting state should still be selected
```
MetRonnie: This doesn't seem right - surely the expected UX is that trying to unselect waiting should unselect it and its children?

oliver-sanders: That's another way of doing it.

Happy either way.

wxtim: I think that's better if we have time, but this is better than nothing.

wxtim: In that case I think that this is the best we can get right now, and I'm happy to run with it.

oliver-sanders:
Addressed comments (bar one).

Fixed a small pre-existing issue where the search feature was matching family names in the tree view even when families were turned off - 9028ce44997134b59b8763d29cd5c7cb5969f77a

Added a filter "reset" button - https://github.com/cylc/cylc-ui/compare/14fc4ecea00adac5cb851b760a776b29d6af403e..9028ce44997134b59b8763d29cd5c7cb5969f77a

tests/component/specs/viewToolbar.cy.js
```diff
@@ -224,4 +224,93 @@ describe('View Toolbar Component', () => {
...
+        controls: [
+          {
+            title: 'Filter',
+            action: 'taskStateFilter',
+            value: [],
+            key: 'filter'
+          }
+        ]
+      },
+    ])
```
oliver-sanders: In recent changes I have fixed a reactivity issue (which means that the `eager` is no longer needed).

Unfortunately, that has broken this test. In order for this to work, we need to implement a `setOption` callback like we do in each view, i.e:

```vue
<ViewToolbar @setOption="setOption" />
```

I think the solution should be something along the lines of (see `listeners` in the Info View tests):

```js
import { reactive } from 'vue'

const model = reactive([])
mountToolbar({
    props: { groups: [...] },
    listeners: { setOption: (key, value) => { model = value } }
})
```

However, I've so far been unable to get the `setOption` callback to fire (even though we know the event has been emitted by inspecting the wrapper).

tests/e2e/specs/tree.cy.js
```diff
@@ -184,12 +184,12 @@ describe('Tree view', () => {
           .should('not.be.visible')
       }
       // It should stop filtering when input is cleared
-      cy.get('[data-cy=filter-id] input')
+      cy.get('.c-view-toolbar input')
```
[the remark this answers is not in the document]
oliver-sanders: Will change on account of the selector being more precise (the toolbar may contain multiple inputs), but the CSS selector is otherwise fine.

MetRonnie (approved):

PR #2372, opened 2025-12-01
Info view: improve styling

MetRonnie:
image

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Tests are not needed
- [x] Changelog entry included if this is a change that can affect users
- [x] Docs not needed

PR #2373, opened 2025-12-01
Info view: show full family inheritance

### LINKED ISSUE #2257: Info view: display full list of families a task inherits from
This can be gotten from the subscription

MetRonnie:
Closes #2257 

Built on:
- #2372

image

**Check List**

- [x] Docs not needed

### DIFF
--- a/src/components/cylc/Info.vue
+++ b/src/components/cylc/Info.vue
@@ -123,7 +123,18 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           data-cy="inheritance-panel"
         >
           <v-expansion-panel-title>
-            Inheritance
+            <template #default="{ expanded }">
+              Inheritance
+                <v-icon
+                  v-if="expanded"
+                  :icon="icons.mdiHelpCircleOutline"
+                  class="ml-2"
+                  v-tooltip="{
+                    text: 'Shows the linearised family inheritance hierarchy for this task. The order of precedence is determined by the C3 algorithm used in Python.',
+                    location: 'top',
+                  }"
+                />
+            </template>
           </v-expansion-panel-title>
           <v-expansion-panel-text>
             <v-breadcrumbs :items="inheritance">
@@ -265,6 +276,7 @@ import {
   mdiDramaMasks,
   mdiCheckboxOutline,
   mdiCheckboxBlankOutline,
+  mdiHelpCircleOutline,
 } from '@mdi/js'
 import { cloneDeep } from 'lodash-es'
 import { formatDatetime } from '@/utils/datetime'
@@ -294,6 +306,9 @@ export default {
     return {
       inheritance,
       jobTheme: useJobTheme(),
+      icons: {
+        mdiHelpCircleOutline,
+      },
     }
   },
 


src/components/cylc/Info.vue
```diff
@@ -111,8 +117,30 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+            <v-breadcrumbs :items="inheritance">
+              <template #divider>
+                <v-icon
+                  :icon="mdiChevronRight"
+                  size="small"
+                  class="mx-n1"
+                />
+              </template>
+            </v-breadcrumbs>
+          </v-expansion-panel-text>
```
oliver-sanders: I think this will go wrong in the event of multiple inheritance.

E.g, given this hierarchy for the task "foo":

foo

The returned namespace will be:

```
"namespace": [
              "root",
              "A",
              "B",
              "BAR",
              "F",
              "FOO",
              "foo"
            ]
```

Which would be displayed as `root > A > B > BAR > F > FOO`, however, `B` does not inherit `F` as they are at the same level, but on different branches.

<details>
<summary>flow.cylc</summary>

```cylc
[scheduling]
    [[graph]]
        R1 = foo

[runtime]
    [[foo]]
        inherit = FOO, BAR
    [[FOO]]
        inherit = F
    [[BAR]]
        inherit = B
    [[F, B]]
        inherit = A
    [[A]]
```

</details>

<details>
<summary>GraphQL Example</summary>

Query:

```graphql
query {
  workflows {
    id
    tasks {
      name
      namespace
    }
  }
}
```

Response:

```json
{
  "data": {
    "workflows": [
      {
        "id": "~oliver.sanders/tmp.QdOZMqsUFA/run1",
        "tasks": [
          {
            "name": "foo",
            "namespace": [
              "root",
              "A",
              "B",
              "BAR",
              "F",
              "FOO",
              "foo"
            ]
          }
        ]
      }
    ]
  }
}
```

</details>

oliver-sanders (commented):
It's a good idea to display inheritance in the info view.

But we might need to stop for a min to work out what exactly we are trying to display:

1. Inheritance.
2. Linearised inheritance.
3. Display inheritance.

As all three may differ.

Perhaps it might be best to display an inheritance (1) graph?

src/components/cylc/Info.vue
MetRonnie: The main motivation for this as I see it is to be able to tell what families a task belongs to. The graphical order when it comes to multiple inheritance would be nice but is not particularly important. If the `>` separator would cause confusion in the case of multiple inheritance then I can change it.

Edit: Or if we can come up with a straightforward way to do a sort of tree view like
```
A
|- B
|  `- BAR
`- F
   `- FOO
```
but that doesn't seem possible using just the `namespace` GraphQL field

oliver-sanders (commented):
LGTM.

Haven't tested expansion and tab restoration yet (which has changed, e.g, what happens if it tries to restore the view from previous state?)

It might be worth adding an info icon or something to explain what this inheritance means as inheritance in Cylc is a bit confusing (linearised vs display vs full inheritance), might help to add an explanation.

oliver-sanders (approved):

oliver-sanders:
> Assuming you mean if an info tab was open before the UI was updated, then all the panels will be collapsed when you revisit the workflow. Not ideal but not much of a problem either

Fair, only important thing is to make sure it doesn't traceback and that the previous saved layout will still get updated with the new collapsed value.

PR #2380, opened 2025-12-12
workflows: make tree more compact

oliver-sanders:
Add new user configuration which reduces the vertical height of the workflows in this example by ~22%:

<table>
<thead>
<tr>
<td>before</td>
<td>after</td>
</tr>
</thead>
<tbody>
<tr>
<td>
Screenshot from 2025-12-12 10-09-25
</td><td>
Screenshot from 2025-12-12 10-08-11
</td>
</tr>
</tbody>
</table>

**Check List**

### DIFF
--- a/src/views/UserProfile.vue
+++ b/src/views/UserProfile.vue
@@ -131,40 +131,42 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
                   v-model="jobTheme"
                 >
                   <table class="c-job-state-table">
-                    <tr>
-                      <th>State</th>
-                      <th
-                        v-for="theme in $options.jobThemes"
-                        :key="theme"
+                    <tbody>
+                      <tr>
+                        <th>State</th>
+                        <th
+                          v-for="theme in $options.jobThemes"
+                          :key="theme"
+                        >
+                          {{ upperFirst(theme.replace('_', ' ')) }}
+                        </th>
+                      </tr>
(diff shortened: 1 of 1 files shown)


src/components/cylc/tree/GScanTreeItem.vue
```diff
@@ -33,6 +34,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
       :to="workflowLink"
       :class="nodeClass"
       class="flex-grow-1 flex-shrink-1 px-2 ml-1"
+      :density="compactMode ? null : 'compact'"
```
MetRonnie: Wrong way round, also `null` doesn't work:
```suggestion
      :density="compactMode ? 'compact' : 'default'"
```
Or if we want to reduce the default padding slightly:

```suggestion
      :density="compactMode ? 'compact' : 'comfortable'"
```

oliver-sanders: And yet it works?!

Tried out comfortable, that looks like a sensible default to me.

wxtim (approved):
Re-checked

src/components/cylc/tree/GScanTreeItem.vue
oliver-sanders: (the original commit seems to have worked)

I took your suggestion.

PR #2386, opened 2025-12-16
Fix empty cycle point bug

### LINKED ISSUE #1922: Empty cycle point - `getFamilyTree()` TypeError
### Description

I had a workflow open with the tree view. I uninstalled, reinstalled, and started it. 

```
Uncaught TypeError: Cannot read properties of undefined (reading 'slice')
```

Location: https://github.com/cylc/cylc-ui/blob/b0204f2d7d2d4848bd5f6d7ec9972f4ef23fc691/src/store/workflows.module.js#L322

`node` has no property `ancestors`.

It caused a cycle point to show up as empty, similar to in #1813

> image

MetRonnie:
Closes #1922

Guard against `node.ancestors` being `undefined` rather than an empty array.

**Check List**

- [x] Tests not needed?
- [x] Changelog entry
- [x] Docs not needed

### DIFF
--- a/src/store/workflows.module.js
+++ b/src/store/workflows.module.js
@@ -386,14 +386,13 @@ function getFamilyTree (tokens, node) {
   }
 
   // add family levels below the cycle point
-  if (node.ancestors?.length) {
-    for (const ancestor of node.ancestors.slice().reverse()) {
-      ret.push([
-        'family',
-        ancestor.name,
-        lastTokens.clone({ task: ancestor.name })
-      ])
-    }
+  for (let i = node.ancestors?.length ?? 0; i > 0; i--) {
+    const { name } = node.ancestors[i - 1]
+    ret.push([
+      'family',
+      name,
+      lastTokens.clone({ task: name })
+    ])
   }
 
   // add the family itself
--- a/src/store/workflows.module.js
+++ b/src/store/workflows.module.js
@@ -386,12 +386,14 @@ function getFamilyTree (tokens, node) {
   }
 
   // add family levels below the cycle point
-  for (const ancestor of node.ancestors.slice().reverse()) {
-    ret.push([
-      'family',
-      ancestor.name,
-      lastTokens.clone({ task: ancestor.name })
-    ])
+  if (node.ancestors?.length) {
+    for (const ancestor of node.ancestors.slice().reverse()) {
+      ret.push([
+        'family',
+        ancestor.name,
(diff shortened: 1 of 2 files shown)


src/store/workflows.module.js
```diff
@@ -386,11 +386,12 @@ function getFamilyTree (tokens, node) {
...
+  for (let i = node.ancestors?.length ?? 0; i > 0; i--) {
+    const { name } = node.ancestors[i - 1]
     ret.push([
       'family',
-      ancestor.name,
-      lastTokens.clone({ task: ancestor.name })
+      name,
+      lastTokens.clone({ task: name })
     ])
   }
```
oliver-sanders: Hmmm, this suggests that we are receiving an `added` delta for the family, in which `ancestors` is `undefined`? If so, we have a data store bug :(

Note, this is also relevant to the graph view which needs to request the family hierarchy.

By this logic:

* If `ancestors` is `undefined` on the first delta we receive, then we won't create the families under the cycle point.
* If `ancestors` is defined on a subsequent delta, that won't trigger this to subsequently run (because the node already exists in the store).

This will certainly give us a partial tree at first, but I'm not sure if something else in the store will subsequently force these nodes to be created on future deltas, so the issue might get fixed in time.

MetRonnie: (Discussed in person after some digging, summarising here for the record)

> Does this mean: The added delta just never got sent, something was lost in translation (family should be added to the store).

Yes, that. Possibly fixed by https://github.com/cylc/cylc-uiserver/pull/761

N.B. `updated` deltas for `root` families will never have `ancestors` as we have `updated(stripNull = true)` in the tree view subscription.

Btw, this PR will also defend against the bug if `stripNull` is ever turned on for `added` deltas

src/store/workflows.module.js
```diff
@@ -378,11 +386,12 @@ function getFamilyTree (tokens, node) {
...
 
   // add family levels below the cycle point
-  for (const ancestor of node.ancestors.slice().reverse()) {
+  for (let i = node.ancestors?.length ?? 0; i > 0; i--) {
+    const { name } = node.ancestors[i - 1]
```
oliver-sanders, line 392: I think we are working around a situation that shouldn't logically exist (i.e, a UIS bug).

Should add a comment here to make sure this isn't undone by a subsequent "tidy" commit.

oliver-sanders (approved):

PR #2387, opened 2025-12-16
Show last update time in gscan workflow title tooltip 

MetRonnie:
Plus homogenise tooltips.

Follow-up to https://github.com/cylc/cylc-ui/pull/2381

**Check List**

- [x] Tests are not needed
- [x] Changelog entry included
- [x] Docs not needed


src/components/cylc/tree/GScanTreeItem.vue
```diff
@@ -45,6 +45,12 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               style="overflow-wrap: anywhere;"
             >
               {{ node.id }}
+              <template v-if="node.type === 'workflow' && node.node.status !== WorkflowState.STOPPED.name">
+                <br/>
+                <span class="text-grey-lighten-1">
+                  Last update: {{ useTimeAgo(node.node.lastUpdated * 1e3, { showSecond: true, updateInterval: 5e3 }) }}
```
wxtim: Is using exponential notation in this context a standard JS script thing? Or are you playing code-golf? ;p

oliver-sanders (changes requested):
The conclusion of the meeting was not to do this?

MetRonnie:
This was requested by one of the PSAs. Pasted from Teams chat:

> In Cylc 7 we can easily assess the health of a suite by looking at the time it was last updated _[and also by visually checking it has a blue hollow square...]_
> In Cylc 8, this is much more difficult to determine, suites can look inert / dead for many hours and we have to go in and visually check each one. 
> image
> This could be potentially be addressed with:
> a) Including a timestamp in the Recently Submitted task info hover box
> _[b) Including a Task Waiting square indicator.]_

This PSA was not in the meeting you refer to, though I know the one that was in the meeting didn't think he would find it useful to have in a sidebar tooltip. By including it in the sidebar we at least give them options and they aren't forced to go to the workflow table for this info.

src/components/cylc/tree/GScanTreeItem.vue
MetRonnie: Perfectly standard: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#numeric_literals

Easier to eyeball than counting zeros IMO

oliver-sanders:
One says it would be useful, the other says it will get in the way.

Ask the PSA's find out what the consensus is?

PR #2404, opened 2026-01-08
fix command editor

oliver-sanders:
Fix two issues with the command editor spotted at release time:

1. Edits were ignored when the form was submitted.
2. Edits to the form were preserved when subsequently opening the form again.


wxtim (approved):
- Task status changes to fail when I set output = fail
- Get a yellow triangle if I try to set output = woteva
- Form doesn't remember content

Should setting output to fail kill the task?

PR #2405, opened 2026-01-08
bump version

oliver-sanders:


MetRonnie:
Why has this been done? We are going to do another release PR which will bump the version

wxtim:
> Why has this been done? We are going to do another release PR which will bump the version

I think that the relase PR will bump minor rather than maintaince version?

MetRonnie:
I've force pushed master back to 4a8e26339a44131982e1857221326c53fc7cfbca

PR #2413, opened 2026-01-09
tree: fix ID filtering in flat mode and complex globs

oliver-sanders:
1)
  * ID filtering in flat mode was broken by a bugfix which attempted to
    prevent family names from impacting searching in this mode.
  * Reversion of https://github.com/cylc/cylc-ui/commit/8ede99b543628047c73afa6798518e36d5d53d0b
  * Improve tests.
2)
  * Glob support only allowed a pattern to occur once, e.g, `foo*bar` but not `foo*bar*baz`.
  * Removed limitation and extended tests.

This goes back to the "status quo" where family names do impact searching in flat mode. I'm not sure this makes sense, but that's a topic for another day...

**Check List**

### DIFF
--- a/changes.d/2413.fix.md
+++ b/changes.d/2413.fix.md
@@ -1,0 +1,3 @@
+Fixed tree view search bugs:
+- Didn't work when families are toggled off.
+- Globs containing multiple `*` didn't work.


MetRonnie (approved):
I've pushed a changelog entry

PR #2415, opened 2026-01-09
Limit range of jobs rendered to improve performance

### LINKED ISSUE #2393: gantt: performance / load issues
The Gantt view seems to struggle with larger / longer-running workflows where there is much history to load/display.

ChrisPaulBennett:
This PR closes #2393. The gantt chart can become very unperformant when using a large dataset and a machine that has no graphics acceleration.
I have implement a feature that adds a default amount of jobs to render. The default is an arbitrary choice of 100. I'm open to ideas. 
I have also implemented a text field for users to alter the number of jobs rendered. When the field is focused, a counter on the bottom of the field will tell the user how many jobs in total there are.

**Check List**

### DIFF
--- a/tests/e2e/specs/gantt.cy.js
+++ b/tests/e2e/specs/gantt.cy.js
@@ -82,3 +82,31 @@ describe('Filter save state', () => {
     checkOption('#c-gantt-tasks-per-page', '25')
   })
 })
+
+describe('Render the correct amount of jobs', () => {
+  it('allows changing the number of jobs displayed', () => {
+    cy.visit('/#/gantt/one')
+    let jobsToShow = 5
+
+    cy.get('#c-gantt-job-range')
+      .clear()
+      .type(jobsToShow)
+      .blur() // Trigger the change event
+      .should('have.value', jobsToShow)
+
+    cy.get('.vue-apexcharts')
+      .should('be.visible')
+      .contains('d1')
+
+    jobsToShow = 4
+
+    cy.get('#c-gantt-job-range')
+      .clear()
+      .type(jobsToShow)
+      .blur() // Trigger the change event
+      .should('have.value', jobsToShow)
+
+    cy.get('.vue-apexcharts')
+      .contains('d1').should('not.exist')
+  })
+})


tests/e2e/specs/gantt.cy.js
```diff
@@ -82,3 +82,31 @@ describe('Filter save state', () => {
...
+    jobsToShow = 4
+
+    cy.get('#c-gantt-job-range')
+      .clear()
+      .type(jobsToShow)
+      .blur() // Trigger the change event
+      .should('have.value', jobsToShow)
+
+    cy.get('.vue-apexcharts')
+      .contains('d1').should('not.exist')
```
ChrisPaulBennett: Unit testing the gantt chart is a pain. I cannot find a way to see how many tasks the chart has rendered. The closest I can come is to test that a certain job is no longer rendered when the job range is lowered. It's not robust, but it's best I could come up with. If someone could figure out how, I'd appreciate it.

oliver-sanders: This might be fragile as we are essentially testing the implementation details of a library, but, to answer the question, we can take a peek inside the SVG.

It looks like the "task" labels down the left of the Gantt chart each have the class `apexcharts-yaxis-label`.

So you should be able to do something like this:

```
cy
  .get('.c-gantt')  // find the gantt view on the page
  .find('svg:first')  // find the plot
  .find('.apexcharts-yaxis-label')  // find the task labels
  .should('have.length', x)
```

There might be an additional step needed between "find the plot" and "find the task labels" to switch from searching the HTML document to searching the SVG document. I'm not sure how Cypress behaves in this regard, from (flaky) memory, I think you have to do something like this in jQuery: `$($('svg:first').document).find('.apexcharts-yaxis-label')`.

oliver-sanders:
I think this is limiting the display to show the most recent N jobs (but there's no pagination for viewing older jobs).

As we increase the number of jobs, we also increase the number of tasks which is a tad confusing.

It might make more sense to restrict the view to a given time window and allow the user to scroll through this range.

hjoliver:
Did you intend to merge this @wxtim? There's only one approval and comments by @oliver-sanders don't seem to have been addressed.

Issue #2412, opened 2026-01-08
command editor: consider allowing the Edit Runtime form to be opened in a new tab


oliver-sanders:
There were issues with the restoration of the Edit Runtime tab from previous state.

As a result we don't allow it to open in a new tab: https://github.com/cylc/cylc-ui/pull/2407

On reflection it might not be a good idea to allow the Edit Runtime form to be left open like this as it increases the chance of broadcast conflicts (we don't track broadcasts made whilst the form is open). So perhaps this is a reasonable restriction.

However, if that's the way we decide to go, we may want to revert some of the edits from: https://github.com/cylc/cylc-ui/pull/2394

See also https://github.com/cylc/cylc-ui/pull/2407

PR #2416, opened 2026-01-09
reduce the size of the introspection query

oliver-sanders:
Remove unused fields from the introspection query.

This is just a further cut-down of the default introspection query we are currently using that removes a couple of field types that we are not currently using in the hope of reducing result size.

This is a partial response to https://github.com/cylc/cylc-ui/pull/2394#pullrequestreview-3626892904. Though that matter may not be relevant pending the outcome of https://github.com/cylc/cylc-ui/issues/2412.

**Check List**

### DIFF
--- a/src/utils/aotf.js
+++ b/src/utils/aotf.js
@@ -530,6 +530,38 @@ export function getIntrospectionQuery () {
         types {
           ...FullType
         }
+        # omits:
+        # * subscriptionType
+        # * directives
+        # * locations and args
+      }
+    }
+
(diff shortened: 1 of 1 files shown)


MetRonnie (commented):
I'm not sure how this is supposed to work. And I don't get any difference in the response to the introspection query before and after

oliver-sanders:
This should override part of the introspection query, the query code here is extracted from the introspection query, but cut-down.

Try comparing results against the full query.

PR #2423, opened 2026-01-16
Add empty state notices

MetRonnie:
Users can sometimes be confused upon navigating back to a workflow and not seeing anything - either because:
- They closed all views at some point
- They applied filters at some point and the tasks that were matched have gone now

This PR adds empty state notices:

image

image

**Check List**

- [x] Tests are included 
- [x] Docs not needed
- [x] On a 2.x milestone as no backend changes needed

### DIFF
--- a/src/components/cylc/workspace/Lumino.vue
+++ b/src/components/cylc/workspace/Lumino.vue
@@ -20,20 +20,17 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
     class="main pa-2 fill-height position-relative"
   >
     <!-- Lumino box panel gets inserted here -->
-    <Transition appear>
-      <!-- Transition needed to avoid flashing this notice upon navigation -->
-      <v-empty-state
-        v-if="!views.size"
-        title="Workspace is empty"
-        text="To get started, add a view"
-        class="position-absolute top-0 left-0 h-100 w-100 text-medium-emphasis"
-        id="empty-workspace-notice"
-      >
-        <template #text>
-          To get started, add a view using the toolbar above
-        </template>
-      </v-empty-state>
-    </Transition>
+    <v-empty-state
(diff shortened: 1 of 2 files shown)


src/components/cylc/workspace/Lumino.vue
```diff
@@ -15,9 +15,27 @@ You should have received a copy of the GNU General Public License
...
 -->
 <template>
-  <div ref="mainDiv" class="main pa-2 fill-height">
+  <div
+    ref="mainDiv"
+    class="main pa-2 fill-height position-relative"
+  >
     <!-- Lumino box panel gets inserted here -->
+    <Transition appear>
+      <!-- Transition needed to avoid flashing this notice upon navigation -->
```
oliver-sanders: Hasn't quite worked unfortunately (FF Linux).

Image

It's not too bad, but a bit annoying.

src/components/cylc/workspace/Lumino.vue
```diff
@@ -15,9 +15,27 @@ You should have received a copy of the GNU General Public License
...
+      <v-empty-state
+        v-if="!views.size"
+        title="Workspace is empty"
+        text="To get started, add a view"
+        class="position-absolute top-0 left-0 h-100 w-100 text-medium-emphasis"
+        id="empty-workspace-notice"
+      >
+        <template #text>
+          To get started, add a view using the toolbar above
```
wxtim: Top right?

MetRonnie: > Actually, from a UX point of view, would it make more sense to spin the "Add View" button into its own component and insert it here?

Started doing that, but the button in the toolbar is designed to collapse when the workflow title and status message are long enough to reach the end, so it wasn't immediately straightforward

PR #2481, opened 2026-02-19
Remove unnecessary code

MetRonnie:
Partially addresses https://github.com/cylc/cylc-ui/issues/1130

**Check List**

- [x] Tests not needed
- [x] Correct base branch

### DIFF
--- a/src/views/Graph.vue
+++ b/src/views/Graph.vue
@@ -802,7 +802,7 @@ export default {
     getCollapsedAncestor (id) {
       const { name } = this.cylcTree.$index[id].node.firstParent
       if (name !== 'root') {
-        // TODO: could use Set.intersection if browser support is good enough?
+        // TODO: use Set.intersection when browser support reaches baseline widely available
         return this.collapseFamily.find((family) => (
           name === family || this.allParentLookUp.get(name).includes(family)
         ))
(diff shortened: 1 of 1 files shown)


src/views/Graph.vue
```diff
@@ -1058,21 +1044,9 @@ export default {
...
-                for (const child of this.allChildrenLookUp[node.id]) {
-                  removedNodes.add(child.name)
-                }
-              }
-            }
             const subgraphNodes = nodesInCycle.filter((a) => (
               // if its not in the list of families (unless its been collapsed)
               (!this.allParentLookUp.has(a.name) || this.collapseFamily.includes(a.name)) &&
-              // the node has been removed/collapsed
-              (!removedNodes.has(a.name) || this.collapseFamily.includes(a.name)) &&
```
MetRonnie: In this case, `removedNodes` was a set of all nodes below collapsed nodes, so it could contain the node itself. But this is redundant as we are checking if the node has a collapsed ancestor a few lines below.

src/views/Graph.vue
```diff
@@ -1058,21 +1044,9 @@ export default {
...
-              // its not a node representing this cycle
-              a.name !== cycle &&
```
MetRonnie: This filter is over `nodesInCycle` which cannot contain the cycle itself. So this check was pointless.

oliver-sanders (approved):
I'm happy with these explanations, but haven't manually tested.

wxtim (approved):

PR #2507, opened 2026-03-31
Added modifier properties to FamilyProxyData

### LINKED ISSUE #2421: implement modifiers on family icons
Task modifiers (e.g, "held") apply to tasks, they may bubble up to the workflow level and be displayed in the workflow sidebar (namely "held" and "retry"), however, they are never displayed on families (or cycles which represent the "root" family).

This has been highlighted as a regression from Cylc 7 where the "held" and "retry" modifiers were implemented as task states causing them to be reflected in the aggregate family state.

I don't think this omission was ever intentional, can't remember a decision not to do this, I think it just slipped the net.

### LINKED ISSUE #2514: tree: modifier cut off for the first row
Image

The task modifier on the first row gets cut off by the toolbar.

Can probs be solved just by fiddling the z-index.

See also: https://github.com/cylc/cylc-ui/pull/2507#issuecomment-4212954235

samuel-denton:
Closes #2421 and closes #2514

Added modifier properties from childTasks to propagate up to FamilyProxyData so task modifiers are displayed at the family level.  So if the current task within a family is held, the family shows held too.

**Check List**

- [x] Tests are included (or explain why tests are not needed).  -  Not sure if there are automated tests for this
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.  -  This might be something that should be documented?  I can add this if needed.

### DIFF
--- a/src/styles/cylc/_tree.scss
+++ b/src/styles/cylc/_tree.scss
@@ -50,6 +50,9 @@ $icon-width: 1.5rem;
 }
 
 .c-tree {
+  .c-view-toolbar {
+    margin-bottom: 5px;
+  }
   .c-task, .c-job {
     display: flex;
     align-items: center;
--- a/src/styles/cylc/_tree.scss
+++ b/src/styles/cylc/_tree.scss
@@ -51,6 +51,7 @@ $icon-width: 1.5rem;
 
 .c-tree {
   .c-view-toolbar {
+    // Give space to any task icon modifiers at the top of the tree:
     margin-bottom: 0.3em;
   }
   .c-task, .c-job {
--- a/src/styles/cylc/_tree.scss
+++ b/src/styles/cylc/_tree.scss
@@ -51,7 +51,7 @@ $icon-width: 1.5rem;
 
 .c-tree {
   .c-view-toolbar {
-    margin-bottom: 5px;
+    margin-bottom: 0.3em;
   }
   .c-task, .c-job {
     display: flex;
--- a/src/views/Tree.vue
+++ b/src/views/Tree.vue
@@ -122,7 +122,19 @@ fragment FamilyProxyData on FamilyProxy {
   }
   childTasks {
     id
+    isHeld
+    isQueued
+    isRunahead
+    isRetry
+    isWallclock
+    isXtriggered
   }
+  isHeld
+  isQueued
+  isRunahead
+  isRetry
+  isWallclock
+  isXtriggered
 }
 
 fragment TaskProxyData on TaskProxy {
--- a/src/views/Tree.vue
+++ b/src/views/Tree.vue
@@ -122,12 +122,6 @@ fragment FamilyProxyData on FamilyProxy {
   }
   childTasks {
     id
-    isHeld
-    isQueued
-    isRunahead
-    isRetry
-    isWallclock
-    isXtriggered
   }
   isHeld
   isQueued
--- a/.mailmap
+++ b/.mailmap
@@ -26,4 +26,4 @@ Jamie Allen           <69790602+JAllen42@users.noreply.github.com>
 Christopher Bennett   <christopher.bennett@metoffice.gov.uk>   ChrisPaulBennett
 Mark Dawson           <mark.dawson@metoffice.gov.uk>       Mark Dawson         <mark.dawson@metoffice.gov.ukpython>
 Harleen Kaur          <harleen9355@gmail.com>                  harleenkaur2003         <160705023+harleenkaur2003@users.noreply.github.com>
-
+Samuel Denton         <samuel.denton@metoffice.gov.uk>         samuel-denton           <samuel.denton@metoffice.gov.uk>


samuel-denton:
Opened as draft as I need to test different scenarios and confirm behaviour is as intended. 
With my changes, it shows task modifiers at the family level for the current active task.  It will not propagate modifiers for task further down the graph.  
So, where task [a] is held and is next in the graph to trigger, the held status is propagated up:
image
Equally, if [b] is held and it's dependencies are met, the held status is propagated:
image
However, if [b] is held but [a] is queued to trigger next, or is already running:
image
I don't think it makes sense to have a held task after it's run, but it does not propagate in this case:
image

Essentially, only the next queued or running tasks modifiers are propagated, which I think is correct.  I have only tested 'held' so far and am still testing edge cases.

samuel-denton:
There are 2 things in this example I do not quite understand:
image
1). I am not sure why [c] is not grouped under [fam].  Is this a restriction in what the graph can display due to dependencies?
2). In cycle 8, Im not sure why task [a] hasn't triggered as it is not dependant on [c], the held task

@oliver-sanders I am assuming these are both just me not understanding exactly how the graph displays families or how holding a task works?

samuel-denton:
image
Using more realistic family groupings make it a bit easier for the graph to visually group them properly.  However I am still unsure if holding a task should prevent another task which has no dependency on the first from triggering.  See cycle 7; [a] has not started.

oliver-sanders:
> I am not sure why [c] is not grouped under [fam]. Is this a restriction in what the graph can display due to dependencies?

I've had a go with this example, but couldn't reproduce the issue.

If you've got the workflow kicking around we can have a poke at to try to find out what happened.

oliver-sanders:
>  In cycle 8, Im not sure why task [a] hasn't triggered as it is not dependant on [c], the held task

Task "badges" indicate that the task is not being run for some reason. That bullet icon represents the runahead limit.

See the task state key in the "Cylc UI Quickstart" found in the GUI for more info on badges.

Because Cylc workflows can be infinite (e.g, `inital cycle point` configred, but no `final cycle point` defined), Cylc needs a mechanism to prevent the workflow from running too many cycles ahead. Otherwise, with your example, as soon as you started the workflow, Cylc would enter an infinte submission job loop trying to run every instance of task "a" from cycle 1 to cycle infinity.

src/views/Tree.vue
```diff
@@ -122,7 +122,19 @@ fragment FamilyProxyData on FamilyProxy {
   }
   childTasks {
     id
+    isHeld
+    isQueued
+    isRunahead
+    isRetry
+    isWallclock
+    isXtriggered
```
oliver-sanders: I suspect that you don't need to request these fields for `childTasks`.

We request fields for each "type" separately (lines 78-82), we are already requesting these fields for tasks, we just need to extend this to families.

samuel-denton: Yeah your right, these are not needed.

oliver-sanders:
LGTM.

Please add "Closes #2421" to the PR description (this links the issue).

oliver-sanders:
Looks good, makes sense, tested as working.

Some before:after screenshots for comparison:

Screenshot from 2026-04-09 10-04-28
Screenshot from 2026-04-09 10-03-44
Screenshot from 2026-04-09 10-02-18
Screenshot from 2026-04-09 10-02-00

Note, the task modifier on the first cycle point gets cut off by the toolbar, technically unrelated to this change, but we aught to improve this (probs just fiddle the z-index). Opened - https://github.com/cylc/cylc-ui/issues/2514

oliver-sanders (approved):

src/styles/cylc/_tree.scss
```diff
@@ -50,6 +50,9 @@ $icon-width: 1.5rem;
 }
 
 .c-tree {
+  .c-view-toolbar {
+    margin-bottom: 5px;
+  }
   .c-task, .c-job {
```
samuel-denton, line 57: Closes #2514
Adding a margin-bottom targeting just the toolbar when within the tree view to prevent the top of the tree being cut off:
Image
After:
Image
Used 5px for simplicity and efficiency, I've tested at various scales so I don't think using anything reactive is necessary, but it could be argued using 0.3rem is more correct.

It's also possible to alter the toolbar's z-index from 1 to -1 but doing so may effect other things, I've found if a z-index is set, it's probably for a reason.

src/styles/cylc/_tree.scss
```diff
@@ -50,6 +50,9 @@ $icon-width: 1.5rem;
 }
 
 .c-tree {
+  .c-view-toolbar {
+    margin-bottom: 5px;
```
oliver-sanders: This should be fine.

But please translate this from px to em to make sure it scales with the `font-size`.

samuel-denton: I've switched to 0.3em and tested at different scales and font sizes.

src/styles/cylc/_tree.scss
MetRonnie: Remember to add "closes" keywords to the original PR body so that GitHub links them 👍 Edit: I see you have but each issue number needs to come after its own "closes" in order to work. I.e. 
```diff
-Closes #2421 and #2514
+Closes #2421, closes #2514
```

samuel-denton: Good to know thanks

MetRonnie (approved):
I've just pushed a code comment

Issue #2503, opened 2026-03-26
README instructions not working


samuel-denton:
In the README it instructs to launch a gui for development using:
`cylc gui --ui-build-dir=<cylc-ui-path>/dist/`
Which I was unable to get working.

Using --help-all I found:
`cylc gui --CylcUIServer.ui_build_dir=<cylc-ui-path>/dist/`
Which does work for me.  

Not sure if this is due to my setup, where I was running from, or if it needs updating on the README.

oliver-sanders:
Not a "you" thing, just a bug, ,makes sense, please change.

samuel-denton:
Closed by #2513

PR #2513, opened 2026-04-07
Updated the example cylc gui command to launch a dev build

samuel-denton:
#2503
The argument for the cylc gui command telling it to use a specific build using a path needed updating.

**Check List**

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ yarn run build
 # build for development (rebuilds on change)
 yarn run build:watch
 # and launch using
-cylc gui --ui-build-dir=<cylc-ui-path>/dist/
+cylc gui --CylcUIServer.ui_build_dir=<cylc-ui-path>/dist/
 
 # start dev server in offline mode, using the build instead of source files
 yarn run preview


ChrisPaulBennett (commented):
Both commands seem to work for me.
What is the reason for this change?

samuel-denton:
> Both commands seem to work for me. What is the reason for this change?

For me at least, the current command does not work.  See #2503 
Oliver seemed to think this was not due to my setup but it's possible I have things slightly different to you @ChrisPaulBennett?

oliver-sanders (approved):

oliver-sanders:
> Both commands seem to work for me. What is the reason for this change?

The original command doesn't actually work.

To test, try changing the build dir to `$HOME`. This *should* mess the cylc-ui up because there is no cylc-ui build in the `$HOME` directory:

```
cylc gui --ui_build_dir="$HOME" --new
```

But it doesn't because the unknown argument is quietly ignored (Jupyter Server behaviour).

Then try replacing `--ui_build_dir` with the full version, it should fail.

PR #2531, opened 2026-04-23
Renovate: block apexcharts upgrades due to licence change

MetRonnie:
See
- #1978
- #2360

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -143,8 +143,21 @@
         "vue3-apexcharts"
       ],
       "addLabels": [
-        "data workflows team"
-      ]
+        "analysis views"
+      ],
+      "prPriority": -1
+    },
+    {
+      "matchPackageNames": [
+        "apexcharts"
+      ],
+      "allowedVersions": "<5"
+    },
+    {
+      "matchPackageNames": [
+        "vue3-apexcharts"
+      ],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -143,8 +143,21 @@
         "vue3-apexcharts"
       ],
       "addLabels": [
-        "data workflows team"
-      ]
+        "analysis views"
+      ],
+      "prPriority": -1
```
oliver-sanders, line 148: Could you explain this bit?

oliver-sanders (approved):
LGTM.

Do we need to do this for Dependabot too?

renovate.json
MetRonnie: Trying to drop apexcharts to the bottom of the priority order

> If you have dependencies that are more or less important than others then you can use the prPriority field for PR sorting. The default value is 0, so setting a negative value will make dependencies sort last, while higher values sort first.

https://docs.renovatebot.com/configuration-options/#packagerulesprpriority

MetRonnie:
> Do we need to do this for Dependabot too?

Not sure we can do anything about Dependabot security PRs, but we do not have npm updates enabled in the Dependabot config

---

Hitting merge, let's see how this goes

Issue #1455, opened 2023-08-30
Combine the two toolbar components


MetRonnie:
### Problem

We have two toolbar components with some duplicated code:
- `src/components/cylc/Toolbar.vue` - Used to display the type of view  on narrow viewports, and to display a burger button to hide the drawer component.
- `src/components/cylc/workflow/Toolbar.vue` - Toolbar for the workspace view. Also fulfills the above toolbar's purpose. 

### Proposed Solution

Combine the toolbars into one component that is always shown. This is more relevant after #1444 has been merged, as the workspace view toolbar will have a user avatar with a dropdown menu, which is not shown on other views.

MetRonnie:
Bumped up to 2.14.0 as a result of https://github.com/cylc/cylc-ui/pull/2540#issuecomment-4351797853

PR #2540, opened 2026-04-29
Display drawer expand/collapse button at all times

### LINKED ISSUE #1137: Display drawer expand/collapse button at all times
When you drag the drawer to zero width, a button appears to re-expand the drawer.

This button should probably be available at all times to allow collapsing of the drawer without having to drag it.

**Pull requests welcome!**

ChrisPaulBennett:
Closes #1137

The toolbar is now always displayed when regardless of whether the drawer is displayed or not.
Also changed the burger icon to a left arrow icon as to better reflect what it does when the drawer is expanded.

Also, altered test that checks that the toolbar is not displayed when the drawer is expanded as that is no longer desired behavior.

**Check List**

### DIFF
--- a/src/components/cylc/Toolbar.vue
+++ b/src/components/cylc/Toolbar.vue
@@ -23,7 +23,6 @@ component. Note: this is not used for the workflow view, see
 <template>
   <v-toolbar
     id="core-app-bar"
-    absolute
     :height="toolbarHeight"
     flat
     class="c-toolbar"
@@ -35,7 +34,7 @@ component. Note: this is not used for the workflow view, see
       @click.stop="toggleDrawer"
       id="toggle-drawer"
     >
-      <v-icon>{{ $options.icons.mdiViewList }}</v-icon>
+      <v-icon>{{ drawer ? $options.icons.mdiArrowLeft : $options.icons.mdiViewList }}</v-icon>
     </v-btn>
     <v-toolbar-title>
       {{ title }}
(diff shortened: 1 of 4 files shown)


Scott-Owen-James (approved):
Saw it working and seems fine to me. Does what it says on the tin and nothing else.

samuel-denton (approved):
Looks good to me.  I cant find any window size or arrangement of views that overlap or cause issues.

MetRonnie:
Couple of issues with this, I'm afraid...
- The menu collapse button looks like a backwards navigation button which will confuse some users. Perhaps the "backburger" icon would be more appropriate: https://pictogrammers.com/library/mdi/icon/backburger/
- On some views we now have duplicate toolbar and banners telling you what page you are on
  - See also https://github.com/cylc/cylc-ui/issues/1455

@ChrisPaulBennett Can you follow-up and tackle #1455 in the process?

image

samuel-denton:
> The menu collapse button looks like a backwards navigation button which will confuse some users. Perhaps the "backburger" icon would be more appropriate

I didn't think that was an issue when I reviewed but playing with the different screens again I actually did just press it intending to press the back button!  Not seen the back-burger before but I agree it works here.

> On some views we now have duplicate toolbar and banners telling you what page you are on

My bad, I never use the workflows table or any of those pages which have their own banner so didn't think to check them...  I agree with the proposed solution to merge them, having a consistent banner on all pages is nice.

samuel-denton:
@MetRonnie @ChrisPaulBennett should I revert the merge or is it easier to make a new PR for this?

ChrisPaulBennett:
I didn't know the back burger was a thing. That seems perfect for this situation. Will change.
I did notice the clash with the workflow toolbar, but addressed that.
I'll create another PR to address the icon and combine the toolbars

MetRonnie:
> should I revert the merge or is it easier to make a new PR for this?

We should be able to get the follow-up PR in soon, would avoid the need for any revert as nothing is broken

MetRonnie:
Reverted in #2543, superseded by #2542

PR #2542, opened 2026-04-30
Combine the two toolbars

### LINKED ISSUE #1137: Display drawer expand/collapse button at all times

### LINKED ISSUE #1455: Combine the two toolbar components

ChrisPaulBennett:
This addresses the comments in (https://github.com/cylc/cylc-ui/pull/2540) and closes #1455.
It combines both toolbars into one and implements the backburger icon.

**Check List**


samuel-denton:
The merge of the two toolbar logics looks good to me.  It looks to me like one of the 2 was never used but might be wrong about that.

The backburger works I think.  I'm not used to it but I think it's obvious what it means.  Only thing is that the icon looks like a different font or font weight to the menu burger:
image
Giving the icon a strokeWidth of 1.1 helps them look more similar, although I think I would prefer if both where thiner to be honest:
image
That can be achieved with:
```html
    <v-btn
      icon
      @click.stop="toggleDrawer"
      id="toggle-drawer"
    >
      <v-icon
        :style="drawer
          ? { stroke: 'currentColor', strokeWidth: 1.1 }
          : {}"
      >
        {{ drawer ? icons.backBurger : icons.list }}
      </v-icon>
    </v-btn>
```
The alternative would be to unify to this set of icons for both:
image



Doing some digging, it looks like there is a further "toolbar" that needs merging (or just removing)!
image
The example @MetRonnie gave where the workflows table now has 2 headings in 2 tool bars is caused by the workflows table view hacking the alert system to add an additional heading.  I would guess this was a workaround due to the main toolbar previously being hidden on this page.  So the problem code is in: src/views/WorkflowsTable.vue lines 27 to 33 (I'm sure there is a way with GH I could link to that bit of code but I'm not sure how to review code thats not been touched by this PR...)
```html
        <v-alert
          :icon="icons.mdiTable"
          prominent
          color="grey-lighten-3"
        >
          <h3 class="text-h5">{{ $t('Workflows.tableHeader') }}</h3>
        </v-alert>
```
Just removing that block fixes the duplicate headings / toolbar but does leave the formatting of the workflows table a little off.

Also, some of the page headings on the new toolbar might need some formatting for users E.G.:
image
Should probably be "Workflows Table" rather than "WorkflowsTable".  

Similarly the settings page uses an alert for its header, however the text is different so isn't as noticeable.

2 ways I can think to solve this are just to remove the alerts from the pages that use them, or to have the new toolbar display something generic while on these pages, maybe staying as 'dashboard' or maybe just 'cylc' or 'config' or similar.  Not sure.

Style wise, for some reason the blue headings doesn't feel quite right to me.  It works for workflow headings:
image

But im less of a fan on the workflows table and settings pages, especially after removing the additional heading:
image

ChrisPaulBennett:
The mdi library does not have a ForwardBurger icon (which does seem like an oversight on their part) and importing extra icons seems a bit much.
I'm not touching the finer points of the design. Feel free to create your own PR for that. I'm a just trying to combine the toolbars.

samuel-denton (approved):
Functionally I'm happy it works.
The settings page still has 2 toolbars/headings but they have different text so arguable if thats ok.
The icons aren't perfectly matched style wise but they work fine.
There are a few small style improvements that could be made but aren't critical, such as text colour on the toolbar on the workflows table page, or a margin being added back to the workflows table page so it matches other pages.

oliver-sanders:
FYI, we're not actually restricted to MDI icons.

If you can draw it (e.g, on paper) I can SVG'ize it for use in `<v-icon />` blocks.

samuel-denton:
I can't identify what changed to cause this, but the toolbar is now not displaying in a few places it should, such as the workflows table, the dashboard and the settings page.
image

Oliver also suggested testing 'standalone views' which I think means replacing 'workspace' in the URL with 'tree' or 'graph'.
These pages do not have a toolbar or menu, but I don't know if that's intended.

MetRonnie (changes requested):
> Also, some of the page headings on the new toolbar might need some formatting for users E.G.:
> image
> Should probably be "Workflows Table" rather than "WorkflowsTable".

Addressed along with some other suggested improvements at https://github.com/ChrisPaulBennett/cylc-ui/pull/5/commits

tests/unit/components/cylc/toolbar.vue.spec.js
```diff
@@ -50,9 +55,83 @@ describe('Toolbar component', () => {
...
+    })
+    await wrapper.vm.$nextTick()
+    // Burger button and title should still be present
+    expect(wrapper.find('#toggle-drawer').exists()).to.equal(true)
+    expect(wrapper.find('.c-toolbar-title').exists()).to.equal(true)
+    // Workflow-specific controls should not be present
+    expect(wrapper.find('#workflow-play-button').exists()).to.equal(false)
+    expect(wrapper.find('#workflow-stop-button').exists()).to.equal(false)
+    expect(wrapper.find('#workflow-mutate-button').exists()).to.equal(false)
+    expect(wrapper.find('.c-workflow-controls').exists()).to.equal(false)
```
MetRonnie: This test isn't really valid. Try adding the `workflowName` prop and it still passes, because the controls are also only shown if the workflow exists in the store:
https://github.com/cylc/cylc-ui/blob/77e9905654e5776e0975b0162f91f9ce186c8bfa/src/components/cylc/Toolbar.vue#L409-L411

It would be easier to test this in Cypress where it's running with the full mocked offline data, and avoid testing implementation details. E.g. modify this section:

https://github.com/cylc/cylc-ui/blob/a8dbe1804ee6b7b97dec5c7546b96028433e4007/tests/e2e/specs/toolbar.cy.js#L21-L42

It should test:
- It displays the toolbar but not the workflow controls when looking at the dashboard (`/#/`)
- It displays the toolbar and the workflow controls after navigating to a workflow (`/#/workspace/one`)
- It displays the toolbar but not the workflow controls after navigating to a non-existent workflow (`/#/workspace/non-exist`)

N.B. this series of navigations and assertions can be done in a single `it()`, which would speed up the tests a bit

MetRonnie (changes requested):

tests/unit/components/cylc/toolbar.vue.spec.js
ChrisPaulBennett: Done

tests/e2e/specs/toolbar.cy.js
```diff
@@ -25,12 +25,29 @@ describe('Toolbar component', () => {
       .get('#core-app-bar')
       .should('exist')
   })
-  it('Is NOT displayed when looking at the dashboard', () => {
+  it('Is NOT displayed when using a standalone view', () => {
     cy.visit('/#/')
     cy
       .get('#core-app-bar')
       .should('not.exist')
   })
```
MetRonnie, line 35: Testing it is not displayed when looking at the dashboard... immediately followed by testing it _is_ displayed when looking at the dashboard (below).

(The reason this one is passing is probably the page has not fully loaded by the time this assertion is made - which is a gotcha with Cypress.)

Your new test below can replace both of these two above.

tests/unit/components/cylc/toolbar.vue.spec.js
```diff
@@ -50,9 +55,78 @@ describe('Toolbar component', () => {
...
+      },
+      props: {
+        views: new Map(),
+      },
+    })
+    await wrapper.vm.$nextTick()
+    // Burger button and title should still be present
+    expect(wrapper.find('#toggle-drawer').exists()).to.equal(true)
+    expect(wrapper.find('.c-toolbar-title').exists()).to.equal(true)
+  })
```
MetRonnie: Also replaced by your new E2E test?
```suggestion
```

MetRonnie (changes requested):

MetRonnie:
This composable is no longer applicable:
https://github.com/cylc/cylc-ui/blob/e8922ea934d71b4cb38105fd6546d44b29b9fce9/src/utils/toolbar.js#L39-L43

tests/e2e/specs/toolbar.cy.js
ChrisPaulBennett: I think we should still keep it for completeness sake.
But I have amended the test to actually test the thing it's supposed to test.
(Also, what is the proper way to get cypress to wait? The linter doesn't like`cy.wait()`)

MetRonnie: > (Also, what is the proper way to get cypress to wait? The linter doesn't like`cy.wait()`)

Like what you've done, I think the best you can do is check for the existence of a nearby element

MetRonnie (approved):
Nice 👍

samuel-denton (approved):
All LGTM

Issue #2398, opened 2026-01-06
analysis: duplicate tasks and duplicate filters


oliver-sanders:
A couple of quick things noticed when trying the Analysis view out on some real workflows.

1. If a task has been run on multiple platforms, it appears multiple times in the list.
2. The per-cycle graph seems to have a separate task filter to the main view?
   * And it works differently, when I select a completion, it leaves behind the previously entered text?

PR #2544, opened 2026-05-06
Replace Apex charts with Apache eCharts

### LINKED ISSUE #2256: gantt: view goes blank with excessive scrolling
If you zoom in/out using the scroll wheel, the view will eventually go blank.

I think it might happen when all the data has scrolled outside of the bounds of the view?

### LINKED ISSUE #2403: gantt: scrambled x-axis
When zooming in/out on the Gantt chart, the axis often becomes scrambled.

Image

### LINKED ISSUE #2406: apex charts: licence issues
We are using Apex Charts V3 which is MIT licence (all good).

However, since then, the project has switched to a freeware model which prohibits corporate use (which will apply to Cylc users) and re-licencing (which causes issues with our GPL licence).

This means we can't upgrade :(

Ok for now, but not sustainable going forward. We will either have to:

* Switch to an open source fork of Apex (do any exist).
* Find something else :(

ChrisPaulBennett:
Apex Charts has changed it's licensing so we need to pivot to a different charting library.
Apache eCharts seems like a good fit for us. It uses the Apache 2.0 licence (https://www.apache.org/licenses/LICENSE-2.0.html).
This started of a play branch to see if it meets our requirements which has gone very well. 
This PR closes #2256 and closes #2406

**Check List**


src/components/cylc/analysis/BoxPlot.vue
```diff
@@ -40,14 +40,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
   </Teleport>
-  <VueApexCharts
-    type="boxPlot"
-    :options="chartOptions"
-    :series="series"
-    :height="105 + series[0].data.length * 60"
-    width="95%"
-    class="d-flex justify-center"
-  />
+  <div ref="chart" :style="{ height: `${100 + series[0].data.length * 50}px`, width: '100%' }" class="flex-grow-1" />
```
samuel-denton:
```suggestion
  <div
    ref="chart"
    :style="{
      height: `${100 + series[0].data.length * 50}px`,
      width: '100%'
    }"
    class="flex-grow-1"
  />
```

src/components/cylc/analysis/TimeSeries.vue
```diff
@@ -65,30 +65,21 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
-    <VueApexCharts
-      type="line"
-      :options="miniChartOptions"
-      :series="series"
-      height="120"
-      width="95%"
-      @selection="zoomMainChart"
-      class="d-flex justify-center"
-    />
+    <div ref="mainChart" style="height: 550px; width: 100%;" class="flex-grow-1" />
```
samuel-denton:
```suggestion
    <div
      ref="mainChart"
      style="{
        height: 550px;
        width: 100%;
      }"
      class="flex-grow-1"
    />
```

src/components/cylc/gantt/GanttChart.vue
```diff
@@ -16,14 +16,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
 <template>
-  <VueApexCharts
-    type="rangeBar"
-    :options="chartOptions"
-    :series="series"
-    width="100%"
-    height="auto"
-    class="d-flex justify-center"
-  />
+  <div ref="chartContainer" class="gantt-container" :style="{ height: `${Math.max(300, 60 + displayedTaskCount * 55)}px` }" />
```
samuel-denton:
```suggestion
  <div
    ref="chartContainer"
    class="gantt-container"
    :style="{
      height: `${Math.max(300, 60 + displayedTaskCount * 55)}px` }
    }"
  />
```

tests/unit/components/cylc/analysis/analysistimeseries.vue.spec.js
```diff
@@ -26,6 +26,7 @@ import { vuetifyOptions } from '@/plugins/vuetify'
 const vuetify = createVuetify(vuetifyOptions)
 const analysisJobs = analysisJobQuery.data.jobs
 const $workflowService = sinon.createStubInstance(WorkflowService)
+$workflowService.query2.resolves({ data: { tasks: [{ name: 'succeeded' }, { name: 'waiting' }, { name: 'eventually_succeeded' }] } })
```
samuel-denton:
```suggestion
$workflowService.query2.resolves({
  data: {
    tasks: [
      { name: 'succeeded' },
      { name: 'waiting' },
      { name: 'eventually_succeeded' },
    ]
  }
})
```

samuel-denton:
I've spent a while playing with this; mostly really good and proves the package can do the job.

- Sometimes charts do not load and we get a blank page, it might be because the workflow in question has no data to display in that format, but in that case it needs to show 'No Data' or something like that.  
- ~~I found it a bit confusing that with the analysis timeseries you get no data until you have selected a task to view. Obvious once I spotted it, but different from every other chart.  I guess this is because trying to display all tasks by default would be too much data?  If thats the case, when no tasks are selected, I think it needs to say 'No tasks selected' somewhere in the chart, or it should pick the first task by default or something.~~ While this is the case, it was the same on the Apex chart, not a change in this PR.
- ~~It's also a bit confusing that it has it's own independent task filter under the general analysis view one.~~ Again looks like this was the same before, not changed in this PR
image
image

- I think a bit more needs doing on dynamic resizing as the charts often ended up squished and unreadable:
image

There are a few other bits like some of the hover effect for overlapping colours on Gantt, the timeline on the timeseries etc. that might want a bit of polish but they work.

Looks good overall, just needs some playing with styles and defaults.

samuel-denton (approved):
I'm happy to approve this as is, the previous functionality has been replicated and the charts look good.

There are a few small outstanding issues that could be improved but they were present before this PR.  Mostly related to letting a user know why there is nothing showing (No data, workflow not run, no task selected etc).

There are some style improvements that can be addressed in a follow up PR at some point, like the timeline on the time-series looked a bit nicer before, but functionality wise I think its good.

samuel-denton:
> * I think a bit more needs doing on dynamic resizing as the charts often ended up squished and unreadable:

I think this was caused by a slightly obscure bug, if you open the Gantt view while the window is very narrow, then maximise it again, the chart remains squished.  I think this is unlikely to come up in practice, but should be fixable if we want with an additional width re-calculation after a resize event.

ChrisPaulBennett:
I'm going to action some of those points, as this is an easy place to do it.

package.json
```diff
@@ -30,8 +30,8 @@
     "@lumino/widgets": "2.9.0",
     "@mdi/js": "7.4.47",
     "@vueuse/core": "14.1.0",
-    "apexcharts": "3.54.1",
     "dedent": "1.7.2",
+    "echarts": "^6.0.0",
```
MetRonnie, line 34: Dependencies should be pinned to exact versions as this is an application rather than a library. (This way, every dev who has the same commit checked out will get the exact same versions from `yarn install`)

`yarn add echarts@latest -E`

```suggestion
    "echarts": "6.1.0",
```

renovate.json
```diff
@@ -145,25 +145,12 @@
...
       "groupName": "analysis view packages",
       "matchPackageNames": [
-        "apexcharts",
-        "vue3-apexcharts"
+        "echarts"
       ],
       "addLabels": [
         "analysis views"
       ],
       "prPriority": -1
```
MetRonnie, line 161: Thanks for updating this.

Btw, we don't need to de-prioritise echarts
```suggestion
      ]
```

src/views/views.js
```diff
@@ -125,8 +125,10 @@ export const allViews = new Map([
...
   const defaultView = useLocalStorage('defaultView', TREE)
   // Check if the view is implemented (in case we remove/rename a view in future)
-  if (!allViews.has(defaultView.value)) {
-    defaultView.value = TREE
-  }
+  watchEffect(() => {
+    if (!allViews.has(defaultView.value)) {
+      defaultView.value = TREE
+    }
+  })
```
MetRonnie, line 132: Why use `watchEffect` here?

src/views/Analysis.vue
```diff
@@ -24,6 +24,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
       <!-- Filters -->
       <v-row no-gutters>
         <v-col
+          v-if="chartType !== 'timeSeries'"
```
MetRonnie, line 27: This change leaves an awkward gap. Best to revert until addressing https://github.com/cylc/cylc-ui/issues/2398 IMO?

image

src/components/cylc/analysis/BoxPlot.vue
```diff
@@ -57,10 +57,15 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 </template>
 
 <script>
-import { computed } from 'vue'
-import VueApexCharts from 'vue3-apexcharts'
+import * as echarts from 'echarts/core'
```
MetRonnie, line 60: (This is a bit 🤮 but apparently what they say to do in their docs if you want tree-shaking 😕 )

src/components/cylc/analysis/BoxPlot.vue
```diff
@@ -128,89 +139,112 @@ export default {
...
-      chartOptions,
+      reducedAnimation,
+    }
+  },
+
+  data () {
+    return {
+      chart: null
+    }
+  },
```
MetRonnie, line 154: (FYI the Vue options API is [increasingly becoming a legacy feature](https://github.com/vuejs/core/releases/tag/v3.6.0-rc.1), please do not make new developments in it going forward, use the composition API instead)

src/components/cylc/analysis/BoxPlot.vue
```diff
@@ -269,14 +307,14 @@ export default {
...
-    sortDesc: mdiSortVariant,
+    sortAsc: mdiSortVariant,
+    sortDesc: mdiSortReverseVariant,
   },
 }
 </script>
 
 <style>
-.apexcharts-text {
+.echarts-text {
```
MetRonnie, line 317: This does not match any elements as far as I can tell - can remove the style section?

src/components/cylc/analysis/BoxPlot.vue
```diff
@@ -128,89 +139,112 @@ export default {
...
+          nameLocation: 'middle',
+          nameGap: 30,
+          scale: true,
+          axisLabel: {
+            formatter: (value) => formatDuration(value, { allowZeros: true })
+          }
+        },
+        yAxis: {
+          type: 'category',
+          data: this.series[0].data.map(d => d.x),
```
MetRonnie, line 229: While this works, it looks like `series` was originally structured for apexcharts and now we're having to go out of our way to make use of it in a way that echarts expects? Potential for some refactor here?

src/components/cylc/analysis/BoxPlot.vue
```diff
@@ -73,12 +78,18 @@ import {
...
+  CanvasRenderer
+])
+
 export default {
   name: 'BoxPlot',
 
-  components: {
-    VueApexCharts,
-  },
+  components: {},
```
MetRonnie, line 93:
```suggestion

```

src/components/cylc/analysis/BoxPlot.vue
```diff
@@ -249,14 +283,18 @@ export default {
...
-  },
-
   methods: {
+    initChart () {
+      this.chart = echarts.init(this.$refs.chart)
+      this.updateChart()
+    },
+    updateChart () {
+      if (!this.chart) return
+      this.chart.setOption(this.chartOptions, true)
```
MetRonnie, line 293:
```suggestion
      this.chart?.setOption(this.chartOptions, true)
```

src/components/cylc/analysis/BoxPlot.vue
```diff
@@ -128,89 +139,112 @@ export default {
...
+  mounted () {
+    this.$nextTick(() => {
+      this.initChart()
+    })
+    this.resizeObserver = new ResizeObserver(() => {
+      this.chart?.resize()
+      this.updateChart()
+    })
+    this.resizeObserver.observe(this.$refs.chart)
+    window.addEventListener('resize', this.handleResize)
```
MetRonnie, line 165: I don't think the window event listener is needed as resize observer will respond to window resizes anyway?

src/components/cylc/analysis/BoxPlot.vue
```diff
@@ -128,89 +139,112 @@ export default {
...
+    }
+  },
+
+  mounted () {
+    this.$nextTick(() => {
+      this.initChart()
+    })
+    this.resizeObserver = new ResizeObserver(() => {
+      this.chart?.resize()
+      this.updateChart()
```
MetRonnie, line 162: This seems to lead to a more responsive resize
```suggestion
      this.chart?.setOption({ animation: false })
```

MetRonnie (commented):
Mainly reviewed the box plot so far.

Getting some console errors that ought to be investigated.
```
[ECharts] Specified `grid.containLabel` but no `use(LegacyGridContainLabel)`;use `grid.outerBounds` instead.

[ECharts] cartesian2d cannot be found for series.boxplot (index: 0).
```

tests/unit/setup.js
```diff
@@ -18,3 +23,49 @@ Element.prototype.scrollBy ??= function () { }
...
+      store[key] = value?.toString() ?? null
+    },
+    removeItem: function (key) {
+      delete store[key]
+    },
+    clear: function () {
+      store = {}
+    }
+  }
+})()
```
MetRonnie, line 45: Why is this needed now?

PR #2392, opened 2025-12-29
guide: add more content and make available as view

oliver-sanders:
Built on https://github.com/cylc/cylc-ui/pull/2391

* Loop the running task animation (makes the clockface clearer).
* Allow the guide to be opened as a view (requested by users for acclimatation purposes).
* Provide basic documentation for the n-window - link to the full docs (requested by operators to clarify the "grey" icons).
* List and provide basic documentation for the views (nice to have, good for acclimatisation).
* Provide basic documentation for tab layouts (requested by operators, some of them had missed this feature).

Lots more to do here for the future, e.g:
* Pull in the interventions section from the docs.
* Allow the docs to be open in a tab (or within the guide).
* Better layout.
* Graph docs?
* Warnings, etc.

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [x] Tests are included (or explain why tests are not needed).
- [x] Changelog entry included if this is a change that can affect users
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

Issue #374, opened 2020-01-20
Investigate a Cylc UI within JupyterLab


kinow:
**Describe exactly what you would like to see in an upcoming release**

From comments in this PR: https://github.com/cylc/cylc-ui/issues/90

The pull request was closed, but there was a different approach, suggested by @bollwyvl, where we could actually have the Cylc UI within JupyterLab.

JupyterLab is extensible, and allows widgets/plugins to customize and manipulate the interface. As we already use JupyterHub for authentication and starting and managing the UI Server, perhaps that would be a good idea too.

**Additional context**

**Pull requests welcome!**

kinow:
Did some experimentation with JupyterLab yesterday and posted to Element. Almost forgot to write here too, if it were not for @hjoliver reminding me (thanks! :+1: )

---

Went as far as I could with GScan in JupyterLab. It almost worked: https://github.com/kinow/cylc-gscan

issues:

*    it's really hard to load Vue in React due to outdated libraries.
*    The current code starts a new Vue app, which is more expensive (we don't want to start all over for each component/widget, sharing data would be problematic too)
*    JupyterLab 3 is coming with build improvements. Developing with JupyterLab is quite annoying, having to re-run commands. There migth be a --watch command somewhere, but I haven't found it yet
*    JupyterLab does not expose webpack and babel objects/configuration; with this we could modify the build of the application
*    In the current code, I tried to load the template by compiling it then rendering, but it has complications with the runtime as we cannot modify webpack/babel. It would be really hard to get our Vue components to work in JupyterLab
*    The simplest way might be to use React (i.e. rewrite components in React)
*    Vuetify is for Vue, so our templates would need udpating for a Material Design with React
*    Oh, there's no easy way to get our data store (Vuex) to work with Redux (I'm still not sure what options we have for data sharing between components & widgets in JupyterLab, there might be something useful/cool)
*    I only did this work today because I had some spare time between tasks, and because this weekend I found a good repository with an example app (https://github.com/mkery/JupyterLab-CodeAnalysisDemo) via Twitter. This demo loads a sidebar, which is exactly what I had in mind for GScan (but it could be another kind of widget) - the interesting thing here is that you need to learn what classes you will extend, in order to have access to the sidepanel, or to have a listener to changes in the app, or to create widgets. The docs are not great, so the curve for learning is a little steep.
[...]

PR #2557, opened 2026-06-09
Simplify app toolbar

MetRonnie:
In #2542 we were not 100% agreed on the new icon for the nav sidebar. I am proposing this icon for when the sidebar is open:

image

and this for when it closed:

image

---

This PR also includes a few changes to eliminate technical debt that I had gradually accumulated to the point where I think we should get them in.
- Removing the app module from the vuex store as all it is used for is setting the toolbar title and there are simpler ways to do this
- Replacing the GraphQL mixin with a [composable](https://vuejs.org/guide/reusability/composables) and simplifying data flow across the app by eliminating some props

> Users coming from Vue 2 may be familiar with the [mixins](https://vuejs.org/api/options-composition#mixins) option, which also allows us to extract component logic into reusable units. There are three primary drawbacks to mixins:
>
> 1. **Unclear source of properties**: when using many mixins, it becomes unclear which instance property is injected by which mixin, making it difficult to trace the implementation and understand the component's behavior.
> 1. **Namespace collisions**: multiple mixins from different authors can potentially register the same property keys, causing namespace collisions.
> 1. **Implicit cross-mixin communication**: multiple mixins that need to interact with one another have to rely on shared property keys, making them implicitly coupled.
>
> For the above reasons, we no longer recommend using mixins in Vue 3. The feature is kept only for migration and familiarity reasons.

**Check List**

- [x] No changelog entry needed as invisible to users


ChrisPaulBennett (approved):
There is still an inconsistency between `workflowID` vs `workflowId`.
I think something is changing the case (GraphQL maybe?)
This may not actually be an issue, just thought I'd bring it up since you have altered a few of them.
Much better icons!

MetRonnie:
I was hesitant to change the GraphQL variable `$workflowId` but actually the variable name doesn't matter - it won't break anything. Changed to `$workflowID` now

MetRonnie:
Ping @samuel-denton are you happy with these icons for the sidebar toggle button?

samuel-denton:
> Ping @samuel-denton are you happy with these icons for the sidebar toggle button?

Yep I like those, consistent with everything else.

src/components/cylc/Toolbar.vue
```diff
@@ -285,8 +285,8 @@ import WarningIcon from '@/components/cylc/WarningIcon.vue'
 import { workflowViews } from '@/views/views'
 
 const QUERY = gql(`
-subscription Workflow ($workflowId: ID) {
-  deltas(workflows: [$workflowId]) {
+subscription Workflow ($workflowID: ID) {
+  deltas(workflows: [$workflowID]) {
```
oliver-sanders, line 289: > Consistent camel case for `workflowID`

I generally prefer it the way you've changed it, but for context, some styles recommend treating acronyms as words which is considered "correct" CamelCase in some circles.

I think [one of?] the reason(s) given is when acronyms appear back-to-back in the same word e.g, (if Cylc was an acronym) `CylcId` would likely be preferable to `CYLCID` (which mangles the acronyms together and is indistinguishable from upper-case generally reserved for constants). Combine that with, certain, variable conventions and the problem multiplies up `StrHtmlTmpl` / `STRHTMLTMPL`...

The "acronyms as words" approach is common in Java, the "acronyms in caps" approach is common in Python. The person who laid the groundwork of this project had a Java background (as did I but I try hard to forget it).

src/layouts/Default.vue
```diff
@@ -18,11 +18,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 <template>
   <div>
     <ConnectionStatus :is-offline="offline" />
-    <Drawer v-if="showSidebar" />
+    <Drawer v-if="drawerEnabled" />
     <CommandMenu/>
 
     <v-main>
-      <Toolbar v-if="showToolbar" />
```
oliver-sanders, line 25: Sooo, this change is unexpected, I'm guessing it just made more sense to you include the toolbar here?

The toolbar was very purposefully omitted from the standalone views, but admittedly, the standalone views never got much attention and their purpose is unclear / lost in the midst of time.

The ideas behind the standalone views were:
* Embeddable (i.e, `<iframe />`able) components for incorporation in other tools/apps/dashboards.
  * The idea to assimilate the Cylc GUI into other apps.
  * Jupyter Lab (#374) was one idea.
  * But operators also expressed a desire for "display panels" they could put on wall monitors or whatever.
  * Potential for other high-level integrations - site dashboards or whatever.
  * Situations where you probably don't want the toolbar on the view, but might want it available as a standalone component so it can be embedded in the appropriate location.
* Bare-bones testing:
  * To allow the views to be tested standalone with only their subscription and none other active.
  * We didn't have Cypress component testing as an option back then so the ability to test the bare minimum of the view was somewhat important.
  * Helps avoid the views developing dependencies on the toolbar, etc.

MetRonnie: I always found the lack of toolbar on standalone views a little odd, as it is unclear what workflow you are looking at? And it means you can't play/pause the workflow etc. But I always chalked it up to lack of development of standalone views as a feature.

Also there was never a spec that tested it was not displayed for a standalone view until recently ([added in #2542](https://github.com/cylc/cylc-ui/pull/2542/changes#diff-4886fa4ca8fbaca4f8220dc77465400693f186257d8632d269c7e99214b1e6d3L28-R31), which looked to me like more of a translation of the previous spec that was just checking only 1 of the old 2x toolbars was displayed when not in the workspace view)

src/components/cylc/Toolbar.vue
MetRonnie: As ID isn't an acronym the common usage in JS seems to be as in `getElementById` (excluding cases when it is part of an acronym like UUID).

However I think the most important thing is consistency in our usage. Happy to make it consistently `workflowId` instead but note it is `WorkflowID` in the schema which can't be changed

src/layouts/Default.vue
MetRonnie: Actually I've just remembered one of the reasons I changed this was because the `Guide` is now in the `workflowsView` map but should still show the toolbar, which was broken recently (in #2392 I think).

We could go back to not showing the toolbar in standalone views if really desired, but instead of the previous approach would have to use route params like we do for `showSidebar`

oliver-sanders: Yes, I'm pretty sure the toolbar was *not* intended to go here as per the embeddable components thing, but that's not gone anywhere to date. I don't get the guide should show the toolbar bit, but whatevs, was mostly trying to work out why it had changed - the answer being it just made more sense to you - which is fair enough.

Beyond the tests, we don't actually have any uses of the standalone views at present - so while they did not need to change - there's little harm in them changing.

Happy for you to steer this in a different direction. If this actually matters, we could probs stick it on a toggle controlled by a URL parameter or whatever.

src/components/cylc/Toolbar.vue
oliver-sanders: There are inconsistencies with it the way it was and the way it's changed - and double inconsistencies between the Python backend and JS frontent. I really don't mind much either way, just keen not to get into a a large-scale variable renaming drive!

oliver-sanders (approved):
LGTM, tested as working.

Issue #470, opened 2020-07-06
visual filtering


oliver-sanders:
*awaiting sketch*

A system for changing the appearance of task nodes based on properties of the task, e.g:

* Family
* Parameterisation
* (Re)Flow

This "visual filtering" is intended to make it easier for users to gain an at-a-glance overview of the higher-level-structure of a flow without needing to delve into deeper inspection.

Implementation:

* Visual filters are orthogonal:
   * You can only have one filter at a time.
   * Visual filtering local to a tab.
   * Turned off by default though can be automatically activated by events (e.g. user creates a new [re]flow).
* Use colour as the primary means to distinguish between nodes:
   * Since task-job separation nodes are now all grey.
   * Define a colourmap which doesn't conflict too much with the job status colours (preferably duller).
   * *Accessibility* Allow a user-definable colourmap for colour blind users.
* Use opacity as the secondary means to distinguish between nodes:
   * Users toggle the opacity of a group of tasks by (de)selecting them or by hovering over the options from the filter menu.
   * *Accessibility* This does not depend on colour at all.

oliver-sanders:
This has risen up the priority ladder partly as a result of planned changes to `cylc remove` which would see it strip flow-numbers from tasks. This needs to be communicated through the UI to make sense.

MetRonnie:
Proof of concept added in https://github.com/MetRonnie/cylc-ui/tree/flow-nums

![image](https://github.com/cylc/cylc-ui/assets/61982285/08d6c51e-89bb-4c28-97ac-6d6174631c7c)

hjoliver:
From https://github.com/cylc/cylc-ui/pull/2585#issuecomment-5001213053  - for N-window level: suggest at a minimum allow the user to gray-out all n != 0 nodes (and edges) as a way to visually emphasize the important n=0 tasks.

Issue #133, opened 2019-07-17
graph: edge style


oliver-sanders:
In the present GUIs there are two different edge styles:

* Solid arrows for `and` type dependencies (e.g. `foo => baz` or `foo & bar => pub`)
* Empty arrows for `or` type dependencies (e.g. `foo | bar => pub`)

Some ideas to consider for edge style in the new GUI:

* Dependency type (e.g. `and` or `or`)
* Qualifier (e.g. `:succeed` or `:fail`)
* Satisfaction (see bellow)

In this GUI we would also like the edge style to make it clear if a task has been manually triggered before the completion of its upstream dependencies (see also https://github.com/cylc/cylc-flow/issues/3223#issuecomment-512039292). Options:

1. Make edge style change depending on whether a dependency is satisfied or not
   * `+` simple, can be done client side with little to no change to Cylc Flow
   * `-` doesn't leave a permanent record that the dependency was ignored
2. If a task is manually triggered and some of its upstream dependencies are not met change their edge style to reflect this
   * `+` leaves a permanent record that the dependency was ignored
   * `-` more involved change, this state will need to be held at the workflow
3. Some hybrid of the above?

oliver-sanders:
Design ideas please...

matthewrmshin:
What are the choices? I suppose we have:
* Solid/empty arrow heads?
* Arrow head style?
* Solid/dotted lines?
* Colours?

On qualifier. Can we use colours for the edge to corresponding task state that provides the qualifier? (Obviously this will not work for custom qualifiers/outputs.) If there are multiple downstream paths, I guess we should use dotted lines?

Do we need to consider clock and xtriggers? (Or in another issue?)

oliver-sanders:
> What are the choices?

We could also use text or symbols.

oliver-sanders:
Some ideas:

1. Edges use a grey colour and turn darker when the dependency is satisfied.
2. If a task is manually triggered before it's dependencies are satisfied the arrow heads of the "ignored" dependencies turn to crosses.
3. Edges representing dependencies other than succeeded should have the qualifier (e.g. `:failed`) written next to the arrow.

hjoliver:
> At present xtriggers triggers are special graph nodes so the edge style isn't effected. 

My thinking was: xtriggers (including clock triggers) are just other entities that tasks can depend on, so they should be represented by other nodes in the graph, with dependence represented by edges (just as for dependence on other tasks).

[...]

PR #2585, opened 2026-07-08
ghost tasks: switch from flow=None to graphDepth=0

### LINKED ISSUE #2420: ghost tasks: switch from flow=None to graphDepth=0
We currently display `flow=None` tasks in grey (ghost tasks in Cylc 7 slang).

### Status Quo

The motivation for this was:
a. Indicate `cylc remove`d tasks.
b. Visually distinguish between `n=0` and `n>0` waiting tasks.

As a side effect, any task actually triggered in `flow=None` (extremely rare) would also show in grey.

### Issues

There are some issues with this approach:
1. `n>0` waiting tasks are grey, but `n<0` finished tasks are not 
   * This give us an asymmetry in the time dimension which is perhaps confusing.
   * Sometimes it's useful to differentiate between `n=0` and `n<0` finished tasks, e.g, failed tasks where success is optional.
2. Tasks get greyed out, but families don't:
   * Families don't have flow numbers because they are groupings.
   * If the tree view is collapsed down, there is no visual indication of a removed family.
   * This also gets confusing with group-trigger.
3. Manually triggered `--flow=None` tasks get greyed out:
   * Arguably either confusing or a feature depending on standpoint.
   * Never used at our site so not a concern for us.

### Alternative

A, perhaps, more logical approach would be to grey out `n!=0` tasks.

This would still answer to the original use cases:
a. `cylc remove`d tasks are all `n!=0` by definition.
b. `n>0` tasks are obviously `n!=0`!

This would wipe out the issues above.

ChrisPaulBennett:
Closes #2420.
Alters the logic so that all tasks that are not N=0 (N!=0) are dimmed to make it clearer what is currently active

**Check List**


src/styles/index.scss
```diff
@@ -98,7 +98,8 @@ html {
 }
 
 .c-tree, .c-table, .c-graph {
-  .flow-none {
-    opacity: 0.6;
+  // Dim nodes that are outside the n=0 (active) window, i.e. n != 0.
+  .dimmed {
+    opacity: 0.5;
```
ChrisPaulBennett, line 103: The dimming didn't seem to be that visible on my display. So I've cranked it up a notch

oliver-sanders:
(only assign two reviewers - ovbs copilot don't count)

src/utils/tasks.js
```diff
@@ -137,13 +137,20 @@ export function formatFlowNums (flowNums) {
...
+ * they are not dimmed.
+ *
+ * @param {number=} graphDepth
  * @returns {boolean}
  */
-export function isFlowNone (flowNums) {
-  return Boolean(flowNums && !JSON.parse(flowNums).length)
+export function isN0 (graphDepth) {
+  return graphDepth == null || graphDepth === 0
 }
```
oliver-sanders: I think we can get away without this function:

```js
> Boolean(0)
false
> Boolean(null)
false
> Boolean(1)
true
```

src/views/Graph.vue
```diff
@@ -175,6 +175,7 @@ fragment TaskProxyData on TaskProxy {
     meanElapsedTime
   }
   flowNums
```
oliver-sanders, line 176: Is `flowNums` used by this view any more?

Run through the modified queries and check usage. Note the field might be used by a component used by the view rather than in the view code itself.

If happy the field is not being used, remove it from the query.

ChrisPaulBennett: Look up at the co pilot suggestion

src/utils/tasks.js
ChrisPaulBennett: Oh, how embarrassing :/
Fixed

samuel-denton (approved):
LGTM

hjoliver:
I like this. A couple of comments though.  Here's it is for the graph view in my environment:

image

We don't want to dim ghost tasks too much, but on the other hand n=0 tasks aren't 100% obvious without staring at the screen for bit.  Could we try dimming the ghosts a bit more, and/or making the dimming level a user-configurable setting? 

Also, it doesn't help that the arrows are all dark black. Ideally only arrows between n=0 tasks (which is somewhat rare) should be black. If that's difficult to achieve, it might be better to dim all the arrows. It's the n=0 tasks that we want to emphasize here after all, not the arrows.

oliver-sanders:
> Could we try dimming the ghosts a bit more, and/or making the dimming level a user-configurable setting?

Oof, trying to avoid too much dimming (accessibility issue), yes, we could feasibly make this configurable (though would rather just cut the compromise right).

I think the graph view edge styling would require more data interfaces. We've currently got the family collapsing branch (slow progress) in play, so don't want to go messing with the Graph view too much until we're able to get that in.

Have linked this comment into https://github.com/cylc/cylc-ui/issues/133#issuecomment-4980667180

hjoliver:
OK, agreed ideally edge styling would reflect graph depth too, but that might be harder to implement.

In the meantime, how about these ideas:
- for the edges, it looks kinda weird to have dark black edges connecting grayed-out ghost tasks. How about we just gray out all the edges. It's the n=0 nodes that most need to be emphasized after all, not the edges.
- for the nodes, what about a toggle button in all views that heavily grays out n > 0 nodes? When toggled on, the n=0 distinction will be very clear, but you can toggle off again if you need to see the details of the other nodes.

oliver-sanders:
> How about we just gray out all the edges

We could certainly make the edges grey-er without little work and without creating difficult conflicts. Might not go quite as light as the n!=0 task grey for visibility reasons, but can fiddle to taste.

<br />

> what about a toggle button in all views that heavily grays out n > 0 nodes

See this issue https://github.com/cylc/cylc-ui/issues/470 (suggested groupings - point 4).

Do-able, and a worthwhile feature, but something to consider for the future IMO.

ChrisPaulBennett:
Agreed. Good ideas. But better in a separate PR?

hjoliver (approved):
Approving, can tweak further with follow-ups as suggested.  The code LGTM, with the usual disclaimer that my Vue JS chops are pretty limited.

ChrisPaulBennett:
Can someone merge it please?

samuel-denton:
2 Approvals and I think Oliver's comments are resolved so Ill merge.

MetRonnie:
Missing a changelog entry?

oliver-sanders:
UI-changelog of low importance.

But a cylc-doc changes entry would be greatly helpful.

MetRonnie:
Changelog entries are still helpful to fellow developers :)

oliver-sanders:
@ChrisPaulBennett, we need a cylc-doc "Major Changes" entry to communicate this. Here's the entry from when we first introduced this feature - https://cylc.github.io/cylc-doc/stable/html/reference/changes.html#ghost-tasks

---

Have just built the UI to review a PR and noticed that families/tasks *are* dimming, but cycles and jobs aren't. I think not dimming jobs is probably ok (or at least debatable), but not dimming cycles is more obviously inconsistent (this would useful for determining the runahead limit, etc).

Screenshot from 2026-08-11 15-02-15

PR #2618, opened 2026-08-06
Renovate: 4 day min age for security

MetRonnie:
Provides some buffer for any supply chain compromises to be flushed out before we pull anything into our CI.

Also reduce updates for simple-icons to every 6 months, as we don't need regular updates for it.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -146,7 +146,9 @@
       "matchPackageNames": [
         "simple-icons"
       ],
-      "prPriority": -1
+      "schedule": [
+        "* * * 1,6 *"
+      ]
     },
     {
       "groupName": "analysis view packages",
--- a/renovate.json
+++ b/renovate.json
@@ -8,6 +8,12 @@
   ],
   "platformAutomerge": true,
   "automergeStrategy": "squash",
+   "github-actions": {
+    "enabled": false
+  },
+  "minimumReleaseAge": "4 days",
+  "internalChecksFilter": "strict",
+  "prCreation": "not-pending",
   "packageRules": [
     {
       "matchDepNames": [
@@ -170,9 +176,6 @@
     "addLabels": [
       "security"
     ]
-  },
-  "github-actions": {
-    "enabled": false
   },
   "postUpdateOptions": [
     "yarnDedupeHighest"


renovate.json
```diff
@@ -140,7 +146,7 @@
       "matchPackageNames": [
         "simple-icons"
       ],
-      "prPriority": -1
+      "dependencyDashboardApproval": true
```
oliver-sanders: Please could you explain this one to me?

oliver-sanders (commented):
P4D seems reasonable.

Obvs, this isn't a "live" product, the pre-release incubation period is likely longer than P4D, but is still good to give us a buffer for CI and local development purposes.

renovate.json
MetRonnie: Instead of automatically getting PRs for simple-icons, we would have to explicitly tick a box in https://github.com/cylc/cylc-ui/issues/1014 in order to get a PR. Seeing as all these updates do is add more icons (or occasionally retire them), which we have no particular desire for, this would get rid of the noise

oliver-sanders: So we would have to remember to do this manually?

MetRonnie: I don't think we would ever do it unless there was an new icon we wanted

oliver-sanders: Can we just configure this to update infrequently instead - like P6M or something?

MetRonnie: Done. Though to be honest I expect the PRs to end up kicking around for months before anyone actually looks at them 😕 (e.g. https://github.com/cylc/cylc-ui/pull/2534)

oliver-sanders: True, but irrelevant as that can be said for all dependency updates?

renovate.json
```diff
@@ -140,7 +146,9 @@
       "matchPackageNames": [
         "simple-icons"
       ],
-      "prPriority": -1
+      "schedule": [
+        "* * * 1,6 *"
```
oliver-sanders, line 150: I think that's every Tuesday and every Saturday. Is that what you wanted?

I think monthly would be `* * 1 * *` and every 3 months `* * 1 */3 *`

MetRonnie: Every January and June: https://crontab.guru/#*_*_*_1,6_*

(Reason I have done it as a cron expression is [the docs](https://docs.renovatebot.com/configuration-options/#schedule) say they are deprecating the textual form)

renovate.json
MetRonnie: Hopefully not? We have plenty of dependencies that have actual useful updates to keep us going

renovate.json
oliver-sanders: Aah, right, I had missed a column.

oliver-sanders (approved):

PR #2620, opened 2026-08-06
Log view tail-end mode

samuel-denton:
Trio of PRs to add a new tail mode to cat-log and make use of it in the GUI
 - cylc-flow: https://github.com/cylc/cylc-flow/pull/7414
 - cylc-uiserver: https://github.com/cylc/cylc-uiserver/pull/878
 - cylc-ui: https://github.com/cylc/cylc-ui/pull/2620

The three PR's will close 
 - https://github.com/cylc/cylc-uiserver/issues/802

Changes will be documented in cylc-doc: 
 - https://github.com/cylc/cylc-doc/pull/965

-----------------------------------------------------------------------------------

Added a toggle to switch between tail and head (tail_from_start) in the log view.
Added a toggle between popping old lines as new ones are added vs keeping them all. (removed the UI for this so it's just a dev toggle now)
Added a MAX_LINES selector to the user settings.

**Check List**


src/views/Log.vue
```diff
@@ -442,6 +454,9 @@ export default {
     /** AutoScroll? */
     const autoScroll = useInitialOptions('autoScroll', { props, emit }, true)
 
+    /** Tail mode? (true = TAIL, follow the end; false = HEAD, follow the start) */
+    const tailMode = useInitialOptions('tailMode', { props, emit }, true)
```
samuel-denton: Currently set to use the new tail mode by default.  False here would switch that.

src/views/Log.vue
```diff
@@ -504,6 +520,15 @@ export default {
...
     )
+
+    // re-subscribe when the log view mode (TAIL/HEAD) is toggled
+    this.$watch(() => this.tailMode, (tailMode) => {
+      if (tailMode) {
+        // in TAIL mode jump to the end of the file and follow new lines
+        this.autoScroll = true
+      }
+      this.updateQuery()
+    })
```
samuel-denton: This switches on auto-scroll when toggling to tail mode.  I like this as scrolling down is a pain, but not sure about overriding what might have been a user choice?

samuel-denton:
Improved the truncation message
image

samuel-denton:
Discussion moved to Issue:
https://github.com/cylc/cylc-uiserver/issues/802#issuecomment-5267981093

samuel-denton:
Think I got that conflict right.  I believe it was just the trailing commas PR from @MetRonnie so I have kept my changes and added in the commas.

samuel-denton:
I think we agreed that the inline truncation warnings are sufficient and the banner can be removed.  I haven't done that yet as I am not sure if I think truncation is obvious using just the inline warning. 
Looking for feedback for/against each approach.

samuel-denton:
I think the MAX_LINES user setting is particularly in need of testing.  It seems to work for me, but I expect there are scenarios I haven't thought of like views already being open etc.

src/views/Log.vue
```diff
@@ -238,21 +253,59 @@ import CopyBtn from '@/components/core/CopyBtn.vue'
...
+/**
+ * Dividers shown where the log file has been truncated.
+ *
+ * The uiserver sends a structured `truncated` value ("start" or "end"); these
+ * are the human-readable blocks shown in its place.
+ */
+const LOG_TRUNCATION_MARKERS = {
+  start: truncationMarker('earlier lines omitted (file truncated)'),
+  end: truncationMarker('later lines omitted (file truncated)'),
+}
```
samuel-denton, line 270: Various things we could do to improve this message, including printing the MAX_LINES value. Left simple until we decide on implementation between this and a banner.

MetRonnie: I think it would be good if the banner was moved to take the place of these messages. Does that make sense?

src/views/Log.vue
```diff
@@ -548,6 +724,17 @@ export default {
               value: this.autoScroll,
               key: 'autoScroll',
             },
+            {
+              title: this.logMode
+                ? 'HEAD: showing the start of the file'
+                : 'TAIL: showing the end of the file',
+              icon: this.logMode
+                ? mdiFormatVerticalAlignTop
+                : mdiFormatVerticalAlignBottom,
```
MetRonnie: I think people will confuse this icon for scroll jumping to the start/end. Difficult to think of a better one, but how about [mdiInvoiceTextOutline](https://pictogrammers.com/library/mdi/icon/invoice-text-outline/) where the jagged end is indicative of truncation? Unfortunately there is no upside-down variant of this icon, nor is it easy to flip the icon at the moment until #2472 is merged, but should be possible to use CSS to flip it in the `<style>` section.

src/views/UserProfile.vue
```diff
@@ -225,6 +225,22 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+                  <v-text-field
+                    v-model.number="logMaxLines"
+                    type="number"
+                    :min="1"
+                    :max="$options.logMaxLinesMax"
+                    @blur="logMaxLines = normalizeLogMaxLines(logMaxLines)"
+                    data-cy="input-log-max-lines"
+                  />
+                </v-col>
+              </v-row>
```
MetRonnie, line 243: (Ideally this would be in the log view toolbar itself, perhaps under a settings dropdown, but blocked by https://github.com/cylc/cylc-ui/pull/2472)

changes.d/2620.feat.md
```diff
@@ -0,0 +1 @@
+The log viewer in the GUI/web UI now supports two viewing modes toggled via a HEAD/TAIL button: HEAD mode (default) shows the start of the file and follows new lines, while TAIL mode shows the most recent lines. Users can configure a global "Maximum log lines" setting in their User Profile (default 5000, max 50000) to control memory usage and UI performance. A truncation banner indicates which end of the file has been omitted when the file exceeds the configured limit.
```
MetRonnie:
```suggestion
The log view now supports two viewing modes toggled via a HEAD/TAIL button: HEAD mode (default) shows the start of the file and follows new lines, while TAIL mode shows the most recent lines. Users can configure a global "Maximum log lines" setting in their User Profile to control memory usage and UI performance.
```

src/views/Log.vue
```diff
@@ -443,6 +547,33 @@ export default {
     /** AutoScroll? */
     const autoScroll = useInitialOptions('autoScroll', { props, emit }, true)
 
+    /**
+     * The log view mode
+     * true - HEAD (cat-log "tail" mode) shows the start of the file and follows it;
+     * false - TAIL (cat-log "tail-end" mode) shows the end.
```
MetRonnie:
```suggestion
     * The cat-log mode
     * @type {import('vue').Ref<'tail'|'tail-end'>}
```

src/views/Log.vue
```diff
@@ -443,6 +547,33 @@ export default {
...
+     */
+    const logMode = useInitialOptions('logMode', { props, emit }, false)
```
MetRonnie:
```suggestion
    const logMode = useInitialOptions('logMode', { props, emit }, 'tail-end')
```

MetRonnie: Instead of a boolean, could you do something like this
```suggestion
     * The cat-log mode
     * @type {import('vue').Ref<'tail'|'tail-end'>}
     */
    const logMode = useInitialOptions('logMode', { props, emit }, 'tail-end')
```

MetRonnie (commented):
Works fairly slickly :+1: I just think we need to iron out a few things

src/views/Log.vue
samuel-denton: > I think it would be good if the banner was moved to take the place of these messages. Does that make sense?

Yep that should be doable.

> I think people will confuse this icon for scroll jumping to the start/end.

Agreed, I dont love those Icons so happy for alternative suggestions.  The jagged edge one you suggest is quite good. Possibly worth getting a few opinions as Icons can be interpreted differently by different people.  
The other option I thought about is just a stylised string that switches between Top / Bottom, maybe combined with an Icon?
@ChrisPaulBennett @oliver-sanders - Thoughts?
Example 1:
image
image
Example 2:
image

src/views/UserProfile.vue
samuel-denton: Thats where I had it previously but Oliver thought it better as a user setting to keep the toolbar from getting too busy.  I think either is fine so happy to let you argue that one with @oliver-sanders

changes.d/2620.feat.md
samuel-denton: Good changes and I have merged them.  I was expecting to have to re-do the changelog after review tbh, I suspect there are a few changes still needed.

src/views/Log.vue
samuel-denton: I can implement that if thats preferable?

src/views/Log.vue
oliver-sanders: As icons go, Ronnie's suggestion is probably about as good as we'll get.

This feature might not be all too obvious to users (we don't usually have to decide whether we view the top or bottom of a file), so it might be worth considering a text based toggle instead (e.g. head/tail or top/bottom).

Or a combination of text and icon?

src/views/Log.vue
MetRonnie: Or rename it to something like `headMode`? 
```suggestion
     * Whether the log view is in HEAD mode:
     * HEAD (cat-log "tail" mode) shows the start of the file and follows it;
     * TAIL (cat-log "tail-end" mode) shows the end.
     */
    const headMode = useInitialOptions('headMode', { props, emit }, false)
```

src/views/Log.vue
samuel-denton: Ill have a play with combining an Icon with some text, if I can make it look decent I think thats the best option.

src/views/Log.vue
samuel-denton: Yeah that's sensible, logMode was a hangover from the 3 way toggle.

samuel-denton:
> I think it would be good if the banner was moved to take the place of these messages.

Is this about what you where hoping for @MetRonnie?  It's a bit rough at the moment, I plan to copy the styles of the vue banners but this was just to get it working:
image
image

EDIT since I am away next week I will commit this change so it can be reviewed while away.

src/components/cylc/log/Log.vue
```diff
@@ -24,10 +39,16 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
       v-for="(log, index) in computedLogs"
       :key="index"
-    >{{ log }}</span></pre>
+    ><span
+      v-if="log?.truncation"
+      class="log-truncation-banner"
+      :data-cy="`log-truncation-${log.truncation}`"
+    >{{ log.message }}</span><span
+      v-else
+    >{{ log }}</span></template></pre>
```
MetRonnie, line 51: This is pretty nifty - however could it be moved outside of the `pre` element, seeing as it isn't actual log line content?

MetRonnie (commented):
I've got a PR to allow easily testing this feature in the offline mode: https://github.com/samuel-denton/cylc-ui/pull/1

src/components/cylc/log/Log.vue
samuel-denton: I have had a go at this, it's not as neat but it works.  Moving it outside the pre also means it can be a proper div rather than a span so can use some of vue's built in styles:

```html
    <v-alert
      v-if="truncatedStart"
      type="warning"
      variant="tonal"
      density="compact"
      class="my-2"
      data-cy="log-truncation-start"
    >{{ $options.truncationMessages.start }}</v-alert>
    <pre
      ref="logText"
      :class="wordWrap ? 'text-pre-wrap text-break' : 'text-pre'"
      data-cy="log-text"
    ><span
      v-for="(log, index) in computedLogs"
      :key="index"
    >{{ log }}</span></pre>
    <v-alert
      v-if="truncatedEnd"
      type="warning"
      variant="tonal"
      density="compact"
      class="my-2"
      data-cy="log-truncation-end"
    >{{ $options.truncationMessages.end }}</v-alert>
```

image
image

MetRonnie: Note you can use a `<v-defaults-provider>` to avoid specifying all the props on the 2 alerts twice

samuel-denton:
```html
    <v-defaults-provider
      :defaults="{
        VAlert: {
          type: 'warning',
          variant: 'tonal',
          density: 'compact',
          class: 'my-2',
        },
      }"
    >
      <v-alert
        v-if="truncatedStart"
        data-cy="log-truncation-start"
      >{{ $options.truncationMessages.start }}</v-alert>
      <pre
        ref="logText"
        :class="wordWrap ? 'text-pre-wrap text-break' : 'text-pre'"
        data-cy="log-text"
      ><span
        v-for="(log, index) in computedLogs"
        :key="index"
      >{{ log }}</span></pre>
      <v-alert
        v-if="truncatedEnd"
        data-cy="log-truncation-end"
      >{{ $options.truncationMessages.end }}</v-alert>
    </v-defaults-provider>
```

Much neater thanks!

samuel-denton:
I am playing with different icons / button styles for the toggle.  These are the current options.  They are a bit rough currently so don't worry too much about sizing, positioning etc, whichever we like I will spend some time making nice.  I'm also not sure about the colours, i'm not sure if it makes sense to have it coloured like it's on or off when it's actually switching between 2 states:

1). Icon above / below text
-----------------------------
End Truncated:
image
Start Truncated:
image

2). Icon next to text
-----------------------------
End Truncated:
image
Start Truncated:
image

3). Just text in a button
-----------------------------
End Truncated:
image
Start Truncated:
image

4). Text + Icon in a button
-----------------------------
End Truncated:
image
Start Truncated:
image

5). Just stylised text.  The colours aren't great but something like this could be made to look nice with some work
-----------------------------
End Truncated:
image
Start Truncated:
image

6). Just simple text
-----------------------------
End Truncated:
image
Start Truncated:
image

7). Simple text in button
-----------------------------
End Truncated:
image
Start Truncated:
image

8). Text in button with icon
-----------------------------
End Truncted:
image
Start Truncated:
image
8.5). Larger icon
End Truncated:
image
Start Truncated:
image

Whichever of these we like, if any, I will spend some time making it a bit cleaner.

MetRonnie:
What do you think of putting the button in the yellow banner instead? That way users wouldn't have to worry about which mode they're using unless they find the log file truncated and then click to see the other end of the file

samuel-denton:
> What do you think of putting the button in the yellow banner instead? That way users wouldn't have to worry about which mode they're using unless they find the log file truncated and then click to see the other end of the file

I like that idea, it would make it more intuitive, i.e. a user wants to see the start of the file so they scroll to the top, see the truncation warning, and can then click to load the hidden content.  The downside is that if a user wants to switch between the start and end, they need to first scroll / click auto-scroll to find the truncation banner, then click to change mode, whereas currently a single click takes them to the start/end.  
Regardless if we keep the button on the toolbar, I think having it on the banner makes sense.

samuel-denton:
Additional note that the button should not say 'End Truncated' or should maybe not be visible at all if the file is less than MAX_LINES.  Easy fix but there are other things to decide on first.  See the following post.

samuel-denton:
So currently need to decide on:
- Where the button appears, either in the toolbar, in the truncation warning banner or both
- The appearance and text of the button.  See above post for examples
- The appearance and text of the truncation banner. Currently using the yellow vue warning banner styles. 
    - "later lines omitted (file truncated)" / "earlier lines omitted (file truncated)"
- What the default should be.  I think it should be start truncation with auto-scroll on
- Whether switching the truncation mode should also automatically switch auto-scroll on/off

samuel-denton:
Currently looks like:
image

oliver-sanders:
> I am playing with different icons / button styles for the toggle.

How about a switch-like toggle (e.g, look at the workflow/task toggle).

> What do you think of putting the button in the yellow banner instead?

:+1:

samuel-denton:
> How about a switch-like toggle (e.g, look at the workflow/task toggle).

I like that idea, it might take up quite a lot of space though.  If we decide to move the button to the banner and remove it from the toolbar then it doesn't matter how much space it takes up so ill have a play.

samuel-denton:
image

I like this.  It needs some work if its going on the banner, but its an improvement overall. Ill have a go on Monday at making it fit a little better.  It needs some sort of title or label show what Start/End refers to.

EDIT: Something like this:
image

MetRonnie:
What are your thoughts on this very simple layout: If the start is truncated, show a button there to load the start, otherwise if the end is truncated, show the button down there. And don't show a toggle for the tail mode as it's redundant.

(FYI, might want to rebase onto master so you pick up the new Vuetify 4 styling)



localhost-cylc

samuel-denton:
> What are your thoughts on this very simple layout: If the start is truncated, show a button there to load the start, otherwise if the end is truncated, show the button down there. And don't show a toggle for the tail mode as it's redundant.
> 
> (FYI, might want to rebase onto master so you pick up the new Vuetify 4 styling)

Thats a good idea, it will look much simpler and be more obvious which is the current state and which you are switching too.  I guess it might want a tooltip saying that if you load the end, the start will unload and vice versa.  Ill have a play.

changes.d/2620.feat.md
```diff
@@ -0,0 +1 @@
+The log view now supports two viewing modes toggled via a HEAD/TAIL button: HEAD mode (default) shows the start of the file and follows new lines, while TAIL mode shows the most recent lines. Users can configure a global "Maximum log lines" setting in their User Profile to control memory usage and UI performance.
```
samuel-denton, line 1: Need to remember to change this to whatever we decide the actual default will be.  I suspect TAIL (truncating the start) will be default.

samuel-denton:
Bit rough at the moment but I much prefer this layout.  Ill clean it up a bit:
image

samuel-denton:
Remembered that the current default is TAIL_END (start truncated), which does not have a limit set on follow length.
Do we think this could cause memory issues if lots of uses leave logs open in the GUI while workflows are running?

It would be possible to add a follow limit, I previously implemented a system that popped off the first line as new ones were added to maintain max_lines.  I removed it based on discussion that just holding lines that are already fetched has minimal impact.  Just checking that we agree that is the case.  I can spend some time trying to benchmark memory use if we think thats helpful?

PR #2632, opened 2026-08-19
Refactor user service

MetRonnie:
The "user service" reads permissions and other information from an endpoint provided by cylc-uiserver and provision this information to the rest of the UI application.

This info is crucial for the application to function. For example:
- We need to know whether a user is authorised to view the UI, and redirect to the no-auth page if they are not.
- We need to know what permissions the user has for running commands like playing & stopping workflows, triggering tasks etc.
- We need to know whether the uiserver was started by JupyterHub ("multi-user mode") or by `cylc gui` ("single-user mode").

The way this is currently provisioned to the rest of the application is via the vuex store. But this does not really make sense because the vuex store is for reactive data, whereas the user object should be immutable once the info is loaded from the uiserver (for example, it is not possible for permissions to change without restarting the uiserver (and permissions should not be alterable within the UI source code by any vulnerability!)).

This PR removes the user module from the vuex store, and instead makes the user object immutable and provisions it via a composable.

Although this didn't urgently need addressing, this is work I have done in the sidelines over the last few months and other stuff I am working on builds on top of this commit, so I've split it off into its own PR.

**Check List**

- [x] Tests are included 
- [x] Changelog entry not needed as not user-facing.
- [x] Docs not needed


src/router/index.js
```diff
@@ -72,19 +70,6 @@ const router = createRouter({
...
-    store.commit('user/SET_USER', user)
-  }
-  if (!store.state.user.user.permissions?.includes('read')) {
-    if (to.name !== 'NoAuth') { // Avoid infinite redirect?
-      return { name: 'NoAuth' }
-    }
-  } else if (to.name === 'NoAuth') {
-    // If authorized, redirect no-auth page to home page
-    return { path: '/' }
-  }
```
MetRonnie, line 87: This part of the navigation guard has been moved to the services plugin: https://github.com/cylc/cylc-ui/blob/e18b2a1545200f4d429c1934492b7d08ccd1e956/src/services/plugin.js#L50-L61

src/services/plugin.js
```diff
@@ -54,5 +79,24 @@ export default {
...
+    )
+  }
+
+  /**
+   * Return the user profile, fetching it from the backend server if it hasn't already been loaded.
+   * @returns {Promise<User>}
+   */
+  getUser () {
+    return (this.#user ??= this.#fetchUserProfile())
+  }
```
samuel-denton, line 96: Having a play, I think there is a bug related to this block.  I think it is caching the userprofile promise to avoid duplicate requests, however, if the first request fails, it will have already cashed the promise and prevent retries, since ??= only assigns if a variable is currently null or undefined.  This test fails currently (inserted at line 45 of tests/unit/services/user.service.spec.js):

```js
  it('allows the user profile to be re-fetched after a failed attempt', async () => {
    // create failure fetching the user profile
    vi.stubGlobal('fetch', () => ({ ok: false, status: 500, statusText: 'Test Status' }))
    await expect(servicesPlugin.getUser()).rejects.toThrow()

    // subsequent attempt should now succeed
    const expected = {
      username: 'cylc-user-01',
      mode: 'single user',
      permissions: ['read'],
      owner: 'cylc-user-02',
    }
    vi.stubGlobal('fetch', () => ({ ok: true, json: () => Promise.resolve({ ...expected }) }))
    const user = await servicesPlugin.getUser()
    expect(user).toMatchObject(expected)
  })
```

There are 2 easy fixes:

```js
  async getUser () {
    return (this.#user ??= await this.#fetchUserProfile())
  }
```

This only caches a successful response, but seems to then cause duplicate requests for reasons I am not 100% on.  Or this more explicit version which avoids async:

```js
  getUser () {
    this.#user ??= this.#fetchUserProfile().catch((err) => {
      this.#user = null // don't cache failures
      throw err
    })
    return this.#user
  }
```

Either of those fixes cause the unit test to pass.
Tested using "cylc gui --CylcUIServer.ui_build_dir=/local_build" which I think starts a proper ui-server instance in single-user mode and exercises this part of the auth system properly.

However, while trying to reproduce the bug, I have realised there is no way to actually retry an authentication request currently.  On a failure, the only option is to refresh which would bypass this bug.  So this fix would allow some future feature to attempt a retry but doesn't currently have any impact.  If, for example this TODO is intending to include either a repeating retry, or have a retry button "// TODO: catch error getting user profile and redirect to static error page", this fix would allow that to work.

MetRonnie: No change in behaviour in this PR with regards to that. The TODO here is pre-existing (moved from `src/router/index.js`)

https://github.com/cylc/cylc-ui/blob/4a0a2291c8e2906c5b2f076174b24a2e716fcd48/src/services/plugin.js#L51-L52

samuel-denton: Not sure I follow, this block looks like it is new, and introduces the bug:
```js
  getUser () {
    return (this.#user ??= this.#fetchUserProfile())
  }
```
Previous behaviour didn't cache the result, so after a failure a retry could succeed (again, not sure there is actually a mechanism to retry at the moment) while new behaviour is to cache the promise, which cannot be overridden on a retry.  I think the advantage of caching the promise is that it doesn't need to make 2 requests here:
```js
    this.getUser().then((user) => {
      app.provide('user', user)
    })
    router.beforeEach(async (to, from) => {
      const user = await this.getUser()
...
```
 
Currently the bug has 0 impact since there is no retry mechanism, so I am not worried about it.  However, if the intent is to add a retry button to the static error page, this bug would become relevant.

MetRonnie: Both on master and this branch, we await fetching the `/userprofile` endpoint. If this rejects, the failure is unhandled and the GUI fails to load.

So it doesn't matter if the rejected promise is stored as a property on the `ServicesPlugin` object or not, as we are awaiting the promise in the navigation guard and it will thus throw. This is the same as master.

And you're right there is no retry mechanism on master or this branch.

samuel-denton (approved):
LGTM

src/services/plugin.js
MetRonnie: (Also, if we did add a retry mechanism it would be redirecting them to an error page which would instruct them to reload the page in the browser, which would mean caching the rejected promise wouldn't be preserved, so I don't think there is a future problem with this)

samuel-denton: Yeah it's not really a problem.
I got a bit invested after going down a long rabbit hole looking into ??= which is new syntax to me and only realised the issue cannot surface after spending ages researching it and writing the post.  I should have checked the bug could actually surface before going too deep into it!  
It has solidified the new syntax in my memory pretty well at least :D

Issue #2590, opened 2026-07-15
multiple selection: table view


oliver-sanders:
Implement multiple selection in the table view.

See the umbrella issue https://github.com/cylc/cylc-ui/issues/434

The table view is likely to be the easiest place to start as it is uneffected by issue (1) outlined [in this comment](https://github.com/cylc/cylc-ui/issues/434#issuecomment-4980541645).

Challenges (numbers match [these](https://github.com/cylc/cylc-ui/issues/434#issuecomment-4980541645)):

2) [Handling "out of window" selection](https://github.com/cylc/cylc-ui/issues/434#issuecomment-4980541645):
   * E.g, user selects a task which subsequently drops out of the n-window.
   * UI needs to handle this some how, either by updating the selection (i.e, to remove the out-of-context item), or by communicating somehow that something invisible is still filtered.

<br />

4) Handling "off page" selection:
    * Because the table view is paginated and items may change page while being viewed an item selected may move away from the current page while being viewed.
    * This issue would go away if we adopted a virtual-table approach which doesn't use pagination.
    * Note, there may be the potential to make filtered items float to the top of the table which may help - or hinder.

MetRonnie:
One potential but somewhat restrictive solution to both might be to "freeze" the table view when you enter selection mode, i.e. the tasks that are displayed on that page remain displayed even if they drop out of the N-window, and disable changing pages while in selection mode.

(Edit: I do not mean freeze the Vue reactivity somehow, just freeze the tasks that are displayed, so they will still update task state etc. Currently the tasks that are displayed is a computed property built from the data store, so it might be possible to adapt this to stop adding/removing tasks when they are added/pruned from the data store, perhaps using a watcher instead, or https://vueuse.org/shared/computedWithControl/)

hjoliver:
(comment moved to the umbrella issue)

MetRonnie:
This is my current proof of concept:

<img width="832" height="664" alt="Image" src="https://github.com/user-attachments/assets/e8a6b711-a395-4729-b47d-3cdebd040c26" />

And a fairly quick go at freezing the tasks list:

https://github.com/user-attachments/assets/4f4dcd04-6b93-4c26-8905-fbfccebcf328

(Still issues to resolve with info disappearing as the tasks drop out of N=1, but should be possible)

MetRonnie:
[...]

PR #2649, opened 2026-09-03
Move `initialOptions` from component to view

MetRonnie:
`useInitialOptions()` should only be used in views, not the child component as well, as this leads to confusing data flow (my bad).

Other work built on this, including #2590 

**Check List**

- [x] Tests are included 
- [x] Changelog entry not needed as invisible to users
- [x] Docs not needed

### DIFF
--- a/src/components/cylc/table/Table.vue
+++ b/src/components/cylc/table/Table.vue
@@ -149,45 +149,23 @@ import {
   formatDuration,
   isTruthyOrZero,
 } from '@/utils/tasks'
-import { useCyclePointsOrderDesc } from '@/composables/localStorage'
-import {
-  initialOptions as initialOptionsProp,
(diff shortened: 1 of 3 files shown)


ChrisPaulBennett (approved):
Looks good.
If this is the pattern that is supposed to be used, should we also be giving the same treatment to all the other views?

MetRonnie:
The other components that doubly-nest `initialOptions` are the analysis view components. Ideally we would update those too, but not now because:
- Unlike this PR, it doesn't affect https://github.com/cylc/cylc-ui/issues/2590
- It may conflict with #2544 
- It's easier when the child component is using `<script setup>` as it allows use of the [`defineModel()` macro](https://vuejs.org/api/sfc-script-setup.html#definemodel)

PR #2651, opened 2026-09-07
Turn off graphql-tag non-unique fragment name warnings

### LINKED ISSUE #1757: Warning: fragment with name X already exists
We get a lot of these warnings from `graphql-tag` in the console:
```
index.ts:46 Warning: fragment with name AddedDelta already exists.
graphql-tag enforces all fragment names across your application to be unique; read more about
this in the docs: http://dev.apollodata.com/core/fragments.html#unique-names
```
(Original link in the warning was dead, new link is https://www.apollographql.com/docs/react/data/fragments#unique-names)

From https://github.com/apollographql/graphql-tag/issues/269:
> we have a global module cache... of the fragments we parse — that means that somewhere else in your application, there's another fragment defined with the same name. we protect against this because if you were to interpolate both of them into the same GraphQL document, it would fail validation. you can disable this warning by invoking disableFragmentWarnings ( https://github.com/apollographql/graphql-tag#warnings )

However we also get this warning for fragments defined only once in the whole app, because at 
https://github.com/cylc/cylc-ui/blob/88753e233a99c5d1643b210b2cd98c232ba2e6d8/src/services/workflow.service.js#L546-L550

it is re-evaluating the fragment definitions (which have already been cached the first time) each time the query changes because they are included in the query document.

MetRonnie:
Closes https://github.com/cylc/cylc-ui/issues/1757

Turns off a noisy devtools console warning

**Check List**

- [x] Tests, changelog, docs not needed

### DIFF
--- a/src/graphql/index.js
+++ b/src/graphql/index.js
@@ -28,6 +28,12 @@ import { WebSocketLink } from '@apollo/client/link/ws'
 import { setContext } from '@apollo/client/link/context'
 import { store } from '@/store/index'
 import { createUrl, getXSRFHeaders } from '@/utils/urls'
+import { disableFragmentWarnings } from 'graphql-tag'
+
+// graphql-tag warns about fragments with the same name, as if they end up in the same GraphQL document, it will cause an error.
+// However, we have logic to merge subscription documents that means we shouldn't have two fragments with the same name in the same document.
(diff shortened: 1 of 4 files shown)


samuel-denton (approved):
LGTM, nice to clean up the console.

